PR #1, opened 2016-04-07
Adding general project structure, navigation drawer and list, misc

damyanpetev:


kdinev:
The list does not quite follow the guidelines where templates and styles are in a separate file. Please work on improving it!

PR #10, opened 2016-04-14
lite server config, readme, samples list and landing page

damyanpetev:
redirecting live server runs directly to a new samples landing page, adding one for the list
updating readme


kdinev:
Which issue is this associated with, or is it just related to the local repo samples?

damyanpetev:
Both now that https://github.com/Infragistics/zero-blocks/issues/9 is logged

PR #26, opened 2016-05-25
Add styling to tab control

petko5kov:


kdinev:
@petko5kov something went wrong with the merge I guess.

PR #31, opened 2016-06-07
RC1 update

### LINKED ISSUE #23: Update to Angular 2 RC
The current dependency is on beta 15. Update to RC.

damyanpetev:
**DO NOT merge yet** - there should be an RC2 this week that [might drop platform browser dynamic](https://docs.google.com/document/d/150lerb1LmNLuau_a_EznPV1I1UHMTbEl61t4hZ7ZpS0/edit).

Updated main Angular2 and other modules. There's no single "angular.dev" bundle anymore, polyfills module changed to [core-js](https://github.com/zloirock/core-js). Bumping version to "0.1.x" to avoid minor-level range versions from picking this up automatically.

### Samples
 - Due to changes above all html files have been updated to the new resources
 - Share single updated SystemJS config file.
 - Removed list from Nav Drawer sample, updated templates due to "#" inside of expressions being deprecated.

### General
- Updated test configuration
- Bundle package.json and min file updated
- [List / Tab/ related directives]: Replaced DomAdapter use in List/tabs with `getDom()` that should deliver the correct adapter, **BUT** it's not a public API and importing it from '@angular/platform-browser/src/dom/dom_adapter' will only work if each module is registered (read won't work with the new UMD bundles) so usage should be removed (see angular#7561) - I believe that's already in progress.

  List and tab samples temporarily request modules separately with `System.config({ packageWithIndex: true });`.
- Commented out Animation in NavigationDrawer as that feature is still in development. To be updated once available.
- For IE-copatability the extra `shims_for_IE.js` we used to pick from the internal testing source is no longer required.

**P.S.** Just FYI - decoration inheritance completely stopped as the AppComponentPin got errors for not being declared as directive after the update so now it includes a copy of the `@Component` decorator setup. I'm thinking whatever little support there was for direct inherit (not even merge) is gone as suggested in angular#7968.
(closes #23)


kdinev:
@damyanpetev so I should keep it open until RC2? Would the changes be added to this pull request after RC2?

damyanpetev:
@kdinev I was initially hoping to skip changing modules again and just update to RC2 in one go. Saw Angular build is failing though, so we might as well use RC1 and the next update shouldn't require as many changes.
Will merge changes from last PR in a bit to clear out conflict.

PR #338, opened 2017-03-06
Grid component - PoC 

rkaraivanov:
Initial PoC of the grid component.

Things yet to be done:
- [ ] Basic styling/responsive design.
- [ ] Template structure/markup changes and finalization.
- [ ] Review/discuss the API.
- [ ] More unit tests is always good.


src/grid/column.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    @Input() public width: string;
+    @Input() public index: number;
+    @Input() public filteringCondition: Function = FilteringCondition.string.contains;
+    @Input() public filteringIgnoreCase: boolean = true;
+    @Input() public dataType: DataType = DataType.String;
+    @ContentChild(IgxCellTemplateDirective) protected cellTemplate: IgxCellTemplateDirective;
+    @ContentChild(IgxCellHeaderTemplateDirective) protected headTemplate: IgxCellHeaderTemplateDirective;
+
+    public bodyTemplate: TemplateRef<any>;
+    public headerTemplate: TemplateRef<any>;
```
kdinev: Should a `footerTemplate` also be added? Features like column summaries, for example, should be achieved with a footerTemplate IMO. What do you think?

src/grid/grid.common.ts
```diff
@@ -0,0 +1,179 @@
...
+            "$implicit": this.column,
+            colIndex: this.colIndex
+        });
+    }
+}
+
+@Component({
+    moduleId: module.id,
+    selector: "igx-col-filter",
+    styles: [
```
kdinev: Can this go into a separate stylesheet, or is there a specific reason why it's defined here?

src/grid/grid.common.ts
```diff
@@ -0,0 +1,179 @@
...
+    @Output() protected onFilter = new EventEmitter<IFilteredEvent>();
+
+    protected filterData(event): void {
+        this.onFilter.emit(<IFilteredEvent> {
+            column: this.column,
+            value: event.target.value,
+        });
+    }
+
+    protected toggle(): void {
```
kdinev: Shouldn't there be specific methods to show and hide as well?

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+    row: Object;
+}
+
+export interface IgxGridCellSelectionEvent {
+    index: number;
+    value: any;
+    row: Object;
+}
+
+export interface IgxGridFilterEvent {
```
kdinev: Should be the column, and the filtering settings IMO.

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+    // tBD
+}
+
+export interface IgxGridEditEvent {
+    row: Object;
+}
+
+export interface IgxGridSortEvent {
```
kdinev: I think the direction is contained in the sorting expression, and is an enum?

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+ * @class IgxGridComponent
+ * @implements {OnInit}
+ * @implements {AfterContentInit}
+ * @implements {DoCheck}
+ * @implements {OnDestroy}
+ */
+@Component({
+    moduleId: module.id,
+    selector: "igx-grid",
+    styles: [
```
kdinev: Same question about the styles.

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+        } else if (this.dataContainer) {
+            this.dataContainer.state.paging = {
+                index: 0,
+                recordsPerPage: this.rowsPerPage,
+            };
+        }
+        this.shouldDataBind = true;
+    }
+
+    @Input() public rowsPerPage: number = 25;
```
kdinev: Stick to the same naming. If in the event the `rowsPerPage` is called `perPage`, then either change it here, or change it in the event.

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+            this.dataContainer = new DataContainer(this.data);
+            if (this.paging) {
+                this.dataContainer.state.paging = {
+                    index: 0,
+                    recordsPerPage: this.rowsPerPage,
+                };
+            }
+            // should we attach to all browsers and only when we have sorting??
+            if (this.hasSorting) {
+                this.dataContainer.state.sorting.strategy = new StableSortingStrategy();
```
kdinev: Do a feature check. Perform a small sort to see if the browser already performs a stable sort. If it doesn't then attach our stable sorting strategy.

kdinev: Btw, what is the default strategy?

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+
+    /**
+     * @hidden
+     */
+    public ngDoCheck(): void {
+        let diff: IterableDiffer = this._differ.diff(this.data);
+        if (this.shouldDataBind || diff) {
+            this.onBeforeProcess.emit(this.dataContainer.state);
+            this.dataContainer.process();
+            this.shouldDataBind = false;
```
kdinev: Is `shouldDataBind` used only for initial databinding?

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+    /**
+     * @hidden
+     */
+    public ngOnDestroy(): void {
+        this.colDiffer.unsubscribe();
+    }
+
+    public getColumnByIndex(index: number): any {
+        if (this.columns[index] !== undefined) {
+            return this.columns[index];
```
kdinev: In the case where `this.columns[index] === undefined` you are returning undefined. So just don't do the check, just `return this.columns[index];`

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+    /**
+     * Returns the cell at rowIndex/columnIndex.
+     *
+     * @param {number} rowIndex
+     * @param {number} columnIndex
+     * @returns
+     *
+     * @memberOf IgxGrid
+     */
+    public getCell(rowIndex: number, columnIndex: number): any {
```
kdinev: Shouldn't `getCell` work with `columnKey` instead of `columnIndex`?

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+
+    /**
+     * Returns
+     *
+     * @param {number} rowIndex
+     * @returns
+     *
+     * @memberOf IgxGrid
+     */
+    public getRow(rowIndex: number): any {
```
kdinev: Make a return type here. It should be a `GridRow` type, or something of the sort. The type should consist of record and element associated with the record, as well as collection of `GridCell` for this row.

src/grid/grid.component.ts
kdinev: Also there should be a `GridCell` type, same as with `getRow`, and the cell should consist of element and dataItem.

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+    /**
+     * Focuses the grid cell at position row x column.
+     * Fires the onCellSelection event.
+     *
+     * @param {(number | string)} rowIndex      : the index of the row
+     * @param {(number | string)} columnIndex   : the index of the column
+     *
+     * @memberOf IgxGrid
+     */
+    public cellAt(rowIndex: number | string, columnIndex: number | string): void {
```
kdinev: This method focuses a cell, so it shouldn't be called `cellAt` because the naming infers that it would return a cell.

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+
+    /**
+     * Focuses the grid row at index `index`.
+     * Fires the onRowSelection event.
+     *
+     * @param {(number | string)} index
+     *
+     * @memberOf IgxGrid
+     */
+    public rowAt(index: number | string): void {
```
kdinev: Same comment as for `cellAt`

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+
+    /**
+     * @hidden
+     */
+
+    protected onRowFocus(event: any, index: number): void {
+        let el: HTMLElement = event.target;
+        this._renderer.setElementAttribute(el, "aria-selected", "true");
+        this._renderer.setElementClass(el, "selected-row", true);
+        this.onRowSelection.emit({
```
kdinev: This should carry a `GridRow` object for the row.

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    protected onCellFocus(event: any, index: number, item: any): void {
+        let el: HTMLElement = event.target ? event.target : event;
+        this._renderer.setElementAttribute(el, "aria-selected", "true");
+        this._renderer.setElementClass(el, "selected-cell", true);
+        this._renderer.setElementClass(el.parentElement, "selected-row", true);
+        this.onCellSelection.emit({
+            index,
+            row: this.dataContainer.transformedData[index],
```
kdinev: `GridRow`

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+        if (this.paging) {
+            this.dataContainer.state.paging.index = 0;
+        }
+        this.shouldDataBind = true;
+        this.onFilterDone.emit({filtered: this.dataContainer.transformedData});
+    }
+
+    public addRow(rowData: Object, at?: number): void {
+        this.dataContainer.addRecord(rowData, at);
+        this.shouldDataBind = true;
```
kdinev: That answered my question about `shouldDataBind`.

src/grid/grid.component.ts
```diff
@@ -0,0 +1,630 @@
...
+            this.dataContainer.deleteRecordByIndex(row);
+        } else if (typeof row === "string" && !isNaN(parseInt(row, 10))) {
+            this.dataContainer.deleteRecordByIndex(parseInt(row, 10));
+        }
+        this.shouldDataBind = true;
+    }
+
+    public updateRow(index: number, rowData: Object): void {
+        this.dataContainer.updateRecordByIndex(index, rowData);
+        this.onEditDone.emit({row: rowData});
```
kdinev: `GridRow`

src/grid/grid.component.ts
kdinev: Ignore my second question. I looked at the default :)

rkaraivanov:
The styling is still in the .ts files just to have some "form" in the grid sample.
Once @simeonoff is complete with the styling they will be removed.

The grid samples are also WIP.

kdinev:
@rkaraivanov Is there a way to get this coverage up a little bit more?

rkaraivanov:
@kdinev it seems Coveralls now calculates the branches covered in the final score which is the reason for the sudden coverage drop. See PR #341 which brings the coverage down while the PR itself is just styling and no code.

kdinev:
@rkaraivanov Oh year, I figured that out because the same happened to `ignite-ui`. I will merge it now.

PR #343, opened 2017-03-13
Added readme files for data util and data container. Update demo for data container

m-hristov:


demos/app/data-operations/data-state-configurator.component.ts
```diff
@@ -0,0 +1,65 @@
+import { Component, Input, ViewChild, OnInit, EventEmitter, Output, ChangeDetectionStrategy } from "@angular/core";
+import { DataContainer, DataUtil, DataState, DataType,
```
m-hristov: import * as Infragistics from '../../src/main'; everywhere we are referencing classes, interfaces, components from src/main we should use the alias Infragistics(or whatever you are using).

kdinev: Yeah, I know. I was just asking whether that would be easier :)

PR #345, opened 2017-03-13
Updated snackbar.component to be auto hidden with setTimeout instead …

d-georgiev-91:


src/snackbar/snackbar.component.ts
```diff
@@ -100,18 +100,18 @@ export class IgxSnackbar {
...
-    private _intevalId;
+    private timeoutId;
 
     /**
      * Shows the IgxSnackbar component and hides it after some time span
      * if autoHide is enabled
      */
     public show(): void {
-        clearInterval(this._intevalId);
+        clearInterval(this.timeoutId);
```
kdinev: It's still using an interval instead of a timeout.

kdinev:
@d-georgiev-91 Could you please take a look?

PR #351, opened 2017-03-16
New grid samples

rkaraivanov:


kdinev:
@rkaraivanov You should also update the README. The grid is no longer in a "planned" state ;)

PR #367, opened 2017-04-25
Migrate to Angular 4

rkaraivanov:


JSMike:
FYI, this is throwing a lot of typescript errors when trying to build on windows.

rkaraivanov:
@JSMike hmm, well that's strange as my dev machine is windows and it builds fine.
Do you have `tsc` globally installed and if so what version is it?

JSMike:
I didn't have tsc installed globally, it's included in node_modules/.bin for the package. I tried installing globally with same result. global version is 1.5.3.

kdinev:
@JSMike We're using version 2.2.2 and we don't see any errors.

kdinev:
@JSMike We reproduced this on a different environment. We are looking into it.

kdinev:
@JSMike We found the problem. Could you manually delete the typings folder? We will include a script that deletes it as well, because that's what messing the tsc up.

JSMike:
Ok, yea I removed the global version. I was using the packaged version, 2.2.2, initially. I just tried the global version when @rkaraivanov asked.

FYI, removing the typings directory didn't solve the issue on it's own, you also need to update your tsconfig.json file to include the following inside compilerOptions:

```
    "typeRoots": [
      "node_modules/@types"
    ],
```

This resolved the issue for me.

PR #368, opened 2017-05-02
Migrate to Angluar 4

rkaraivanov:
Closes #367


rkaraivanov:
Added typeRoots based on @JSMike's PR

PR #453, opened 2017-08-22
Propagate calendar options to datepicker

### LINKED ISSUE #443: Propagate the igx-calendar options up to igx-datePicker
### Description  
Propagate the following igx-calendar option setting up to the igx-datePicker

## Inputs
| Name       |      Type      |  Description |
|:----------:|:-------------|:------|
| `weekStart`| `Number \| WEEKDAYS` | Sets on which day will the week start. |
| `locale` | `string` | Sets the locale used for formatting and displaying the dates in the calendar. For more information check out [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) page for valid formats. |
| `formatOptions` | `Object` | The format options passed along with the `locale` property used for formatting the dates. |

## Outputs
| Name | Return Type | Description |
|:--:|:---|:---|
| `onSelection` | `Date` | Fired when selection is made in the calendar.|

## Methods
| Name   | Arguments | Return Type | Description |
|:----------:|:------|:------|:------|
| `selectDate` | `date: Date` | `void` | Change the calendar selection. Calling this method will emit the `onSelection` event. |

Aleksandyr:
Closes #443


src/date-picker/README.md
```diff
@@ -12,9 +12,9 @@ Basic initialization
 ```html
 <igx-datePicker></igx-datePicker>
 ```
-Custom formatter function with passed initial date.
+Custom formatter function with passed initial date and locale.
 ```html
-<igx-datePicker [formatter]="customFormatter" [value]="dateValue">
+<igx-datePicker [formatter]="customFormatter" [value]="dateValue" [locale]="'en-US'">
```
kdinev: This is `regional` option, not `locale`. `locale` refers to language localization strings, while `regional` options refer to date and number formatting, as well as currency, etc.

kdinev: @Aleksandyr @rkaraivanov I suppose this should also be corrected in the calendar?

Aleksandyr: @kdinev  I'm quite sure, that the name of this property is locale, because when you pass a date or select one of the dates, it converts into string with appropriate type that you have passed, which means that this is the string type of the date.

Issue #496, opened 2017-12-19
igx-grid column width not applied


ddincheva:
### Description  
Can not set column width.

### Code-snippet
```
<igx-grid  [data]="localData" [autoGenerate]="false">
    <igx-column [sortable]="false" [field]="'EmployeeID'" [header]="'Id'" [width]="'1020px'"></igx-column>
</igx-grid>
```

### Result  
The Employee Id column does not have width 1020px.

### Expected result  
The Employee Id column should have width 1020px.

zdrawku:
With the next pull request we are going to update the HostBinding of the cellcomponent width:
```
@HostBinding("style.max-width")
get width() {
return this.column.width;
}

// TO

@HostBinding("style.width")
get width() {
return this.column.width;
}
```

PR #525, opened 2018-01-29
Grid sorting/filtering, events and ARIA attributes

### LINKED ISSUE #497: igx-grid cells aria-describedby is pointing to an id that doesn't exist
### Description  
The `igx-grid` `th` elements don't have ids.

### Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid.html
2. Inspect any grid cell and its `aria-describedby` attribute
3. Inspect the `th` the grid cell belongs to

### Result  
There is no id on the `th` element that corresponds to the `aria-describedby` attribute value.

### Expected result  
The `th` element id should be the same as the `aria-describedby` value.

### Attachments  
image

rkaraivanov:
- Added relevant ARIA attributes.
- Additional style classes exposed.
- Events are mirrored to the old API as close as possible.
- Removed the setter for the cell value. Updates should be done through the new exposed `update()` method.
- Row object now returns a reference to its native element (if available).
- "Streamlined" the API for filtering/sorting.

Also closes #497 #496


src/grid/api.service.ts
```diff
@@ -86,8 +88,25 @@ export class IgxGridAPIService {
         this.get(id).filteringExpressions = filteringState;
     }
 
+    public filter_multiple(id: string, expressions: IFilteringExpression[]) {
```
kdinev: Why are there methods not camel-cased?

src/grid/grid-filtering.spec.ts
```diff
@@ -253,86 +254,86 @@ describe("IgxGrid - Filtering actions", () => {
...
-        grid.clearFilter("ReleaseDate");
+        grid.clearFiltering("ReleaseDate");
         fix.detectChanges();
         grid.filter("ReleaseDate", null, DATE_FILTERS.notNull);
         fix.detectChanges();
         expect(grid.rowList.length).toEqual(6);
 
         // Empty filter
-        grid.clearFilter("ReleaseDate");
+        grid.clearFiltering("ReleaseDate");
```
kdinev: Is this a breaking change in the API?

src/grid/grid.component.ts
```diff
@@ -301,41 +302,47 @@ export class IgxGridComponent implements OnInit, AfterContentInit {
...
     public filterGlobal(value: any, condition?, ignoreCase?) {
-        // TODO: AND OR Filtering logic
         this.gridAPI.filterGlobal(this.id, value, condition, ignoreCase);
-        this.page = 0;
     }
 
-    public clearFilter(name: string) {
-        const col = this.gridAPI.get_column_by_name(this.id, name);
-        if (!col) {
+    public clearFiltering(name?: string) {
```
kdinev: Don't introduce such method naming changes. Revert this back to `clearFilter`

src/grid/grid.component.ts
```diff
@@ -301,41 +302,47 @@ export class IgxGridComponent implements OnInit, AfterContentInit {
...
+        if (!this.gridAPI.get_column_by_name(this.id, name)) {
             return;
         }
-
         this.gridAPI.clear_filter(this.id, name);
     }
 
-    public clearFilterAll() {
-
+    public clearSorting(name?: string) {
```
kdinev: To follow the `clearFilter` change this to `clearSort`

src/grid/cell.component.ts
```diff
@@ -192,6 +189,13 @@ export class IgxGridCellComponent {
         this.handleInlineEditMode(event);
     }
 
+    public update(val: any) {
```
kdinev: I guess changing a `value` setter to a modifier method introduces a breaking change. If this affects any of the documentation, readmes, etc. please pull updates to the related documentation. Check the getting-started page as well, for any code that may stop working.

src/grid/api.service.ts
rkaraivanov: Those are internal from the point of the grid and are not exposed directly to the user.

kdinev: Then why is `filterGlobal` camel-cased?

PR #547, opened 2018-02-06
IgxAvatar - Replace the canvas usage with svg when rendering initials

### LINKED ISSUE #136: Avatar component visual enhancement
Improve avatar visuals when zoomed in:
- [x] make the Initials avatar an SVG element
- [x] increase the native image size by 2 to give the avatar a crisper look on mobile devices

bkulov:
Closes #136  

Only the logic which renders initials is changed. No changes made on the image rendering size as it looks fine now.

### DIFF
--- a/src/avatar/avatar.component.html
+++ b/src/avatar/avatar.component.html
@@ -1,10 +1,10 @@
 <div class="igx-avatar igx-avatar--{{size}}" aria-label="avatar" role="img">
     <img #image *ngIf="src" src="{{src}}" class="igx-avatar--image igx-avatar--{{size}}" [class.igx-avatar--rounded]="isRounded"
-		[style.backgroundColor]="bgColor" [attr.aria-roledescription]="roleDescription" />
-	<svg #initialsImage *ngIf="initials && !src" xmlns="http://www.w3.org/2000/svg" [class.igx-avatar--rounded]="isRounded"
-		[style.backgroundColor]="bgColor" [attr.aria-roledescription]="roleDescription">
-		<text text-anchor="middle"/>
-	</svg>
+         [style.backgroundColor]="bgColor" [attr.aria-roledescription]="roleDescription" />
+        <svg #initialsImage *ngIf="initials && !src" xmlns="http://www.w3.org/2000/svg" [class.igx-avatar--rounded]="isRounded" [style.backgroundColor]="bgColor"
+            [attr.aria-roledescription]="roleDescription">
+            <text text-anchor="middle" />
+        </svg>
     <span *ngIf="!src && !initials" class="igx-avatar--{{size}} igx-avatar--icon" [class.igx-avatar--rounded]="isRounded" [style.backgroundColor]="bgColor"
         [style.color]="color" [attr.aria-roledescription]="roleDescription">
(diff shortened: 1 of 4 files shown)


src/avatar/avatar.component.ts
```diff
@@ -84,11 +85,12 @@ export class IgxAvatar {
...
     public ngAfterViewInit() {
-        if (this.initials && this.image) {
-            const src = this.generateInitials(
-                parseInt(this.image.nativeElement.width, 10)
-            );
-            this.image.nativeElement.src = src;
+        if (this.initials && this.initialsImage) {
+            // it seems the svg element is not yet fully initialized so give it some time
+            setTimeout(() => {
+                const size = parseInt(this.initialsImage.nativeElement.width.baseVal.value, 10);
```
simeonoff: Use Renderer2 to modify attributes on the element. This will allow server side rendering and will work in web workers too, where the DOM is not available.

src/avatar/avatar.component.ts
```diff
@@ -107,26 +109,23 @@ export class IgxAvatar {
...
+        if (svgText) {
+            svgText.textContent = this.initials.toUpperCase();
+            svgText.setAttribute("font-size", fontSize);
+            svgText.setAttribute("font-family", this.fontName);
+
+            const x = fontSize;
+            svgText.setAttribute("x", x);
+            const y = size - size / 2 + fontSize / 3;
+            svgText.setAttribute("y", y);
+        }
```
simeonoff: Attributes for width and height should be set with HostBinding decorators to support server side rendering and web workers. Also, don't apply font-size and font-family here, should be done with CSS. If custom font-size and font-family must be specified on component instance level use HostBinding here as well.

src/avatar/avatar.component.html
```diff
@@ -1,6 +1,10 @@
 <div class="igx-avatar igx-avatar--{{size}}" aria-label="avatar" role="img">
-    <img #image *ngIf="src || initials" src="{{src}}" class="igx-avatar--image igx-avatar--{{size}}" [class.igx-avatar--rounded]="isRounded"
-        [style.backgroundColor]="bgColor" [attr.aria-roledescription]="roleDescription" />
+    <img #image *ngIf="src" src="{{src}}" class="igx-avatar--image igx-avatar--{{size}}" [class.igx-avatar--rounded]="isRounded"
+         [style.backgroundColor]="bgColor" [attr.aria-roledescription]="roleDescription" />
+        <svg #initialsImage *ngIf="initials && !src" xmlns="http://www.w3.org/2000/svg" [class.igx-avatar--rounded]="isRounded" [style.backgroundColor]="bgColor"
+            [attr.aria-roledescription]="roleDescription">
+            <text text-anchor="middle" />
+        </svg>
```
simeonoff: Indentation misalignment for the SVG element - should be 4 instead of 8. Also bind a additional classes to the SVG element of "igx-avatar--initials and igx-avatar--{{size}}" so that I can use it as a selectors when styling the component.

zdrawku:
@simeonoff is going to make additional changes to the HostBindings and fonts and basically everything related to the styling.

PR #551, opened 2018-02-07
Simplify keyboard event handling in the cell object.

rkaraivanov:


ChronosSF:
Are we merging this for the beta? A fix for one of the issues with virtualization in the grid depends on this code so it will be good to know if we should base it on this new version.

PR #562, opened 2018-02-08
Calendar templating and localization

### LINKED ISSUE #498: Date Picker and Calendar - Localizing and formatting labels
### Description  
On a calendar, the place of date, month, year and week day are hard-coded in a template. It causes that I can't localize the header label(selected date) and the year-month label into valid Japanese date formats. I set locale "ja-JP" to the locale option then each of elements (for example month and weekday) is correctly localized but the order of the elements is not valid. It should be able to format the whole label since the order of them is different depending on a local culture.

image

rkaraivanov:
The calendar is now using directives where appropriate for most of its
behaviors
Minimized direct DOM  access and manipulation
Both the header and the sub header parts of the calendar can now be
template so users can correctly represent date labels according to their
local culture.
Updated the date picker component to reflect the changes to the calendar API
Updated README

Closes #498


kdinev:
@rkaraivanov Update the changelog with information related to this PR.

rkaraivanov:
@kdinev https://github.com/zdrawku/igniteui-angular/commit/22fbdc4fd78bfa17cdcfa022992bb114dc85f1b0

PR #567, opened 2018-02-10
Avatar implementation simplification

simeonoff:
Closes # .  

### DIFF
--- a/src/avatar/avatar.component.ts
+++ b/src/avatar/avatar.component.ts
@@ -1,13 +1,11 @@
 import { CommonModule } from "@angular/common";
 import {
     AfterContentChecked,
-    AfterViewInit,
     Component,
     ElementRef,
     HostBinding,
     Input,
     NgModule,
-    Renderer2,
     TemplateRef,
     ViewChild
 } from "@angular/core";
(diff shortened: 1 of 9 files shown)


src/avatar/avatar.component.ts
```diff
@@ -23,49 +21,42 @@ export enum Size {
     selector: "igx-avatar",
     templateUrl: "avatar.component.html"
 })
-export class IgxAvatarComponent implements AfterViewInit, AfterContentChecked {
+export class IgxAvatarComponent implements AfterContentChecked {
```
rkaraivanov: Does it really need to implement `AfterContentChecked`? The `getRole` method seems like something that should be called once and not every detection cycle

src/core/styles/components/avatar/_avatar-theme.scss
```diff
@@ -2,29 +2,45 @@ $component-name: 'igx-avatar';
...
+    $small-size: 40px;
+    $medium-size: 64px;
+    $large-size: 88px;
 
     %igx-avatar-display {
-        display: flex;
-        justify-content: center;
-        align-items: center;
         position: relative;
         pointer-events: none;
```
bkulov: Please restore pointer-events as the user may need to implement some logic based on them. Use user-select set to none to disallow text/icon selection.

src/avatar/avatar.component.ts
```diff
@@ -23,49 +21,42 @@ export enum Size {
...
-
     private _size: string;
     private _bgColor: string;
-    private _icon = "android";
-    private _isInitialsRepositionNeeded = false;
+    private _icon: string;
+    private _initials: string;
+    private _src: string;
+
+    @Input("roundShape") public roundShape = "false";
```
rkaraivanov: IMO this should be a boolean instead of a simple string.

src/avatar/avatar.component.ts
```diff
@@ -78,92 +69,61 @@ export class IgxAvatarComponent implements AfterViewInit, AfterContentChecked {
...
 
-    public get srcImage() {
-        return this.image ? this.image.nativeElement.src : "";
-    }
-
-    public set srcImage(value: string) {
-        this.image.nativeElement.src = value;
-    }
-
     get isRounded(): boolean {
```
rkaraivanov: If the `roundShape` property is boolean this is redundant.

src/avatar/avatar.component.ts
```diff
@@ -23,49 +21,42 @@ export enum Size {
...
     get size(): string {
-        if (this._isInitialsRepositionNeeded) {
-            this.repositionInitials();
-        }
-
         return this._size === undefined ? "small" : this._size;
     }
 
     set size(value: string) {
         const sizeType = this.sizeEnum[value.toUpperCase()];
```
rkaraivanov: Typescript supports string enumerations. Initialize the `Size` enumeration with the strings for the size and this check is kind of obsolete.

src/avatar/avatar.component.ts
```diff
@@ -23,49 +21,42 @@ export enum Size {
     selector: "igx-avatar",
     templateUrl: "avatar.component.html"
 })
-export class IgxAvatarComponent implements AfterViewInit, AfterContentChecked {
+export class IgxAvatarComponent implements OnInit {
     @ViewChild("image") public image: ElementRef;
```
rkaraivanov: Code formatting.
Put the property below the decorator and space them a bit.

src/avatar/avatar.component.ts
```diff
@@ -23,49 +21,42 @@ export enum Size {
...
     @ViewChild("initialsImage") public initialsImage: ElementRef;
-    @Input() public initials: string;
-    @Input() public src: string;
-    @Input("roundShape") public roundShape = "false";
-    @Input() public color = "white";
+    @ViewChild("imageTemplate", { read: TemplateRef }) protected imageTemplate: TemplateRef<any>;
```
rkaraivanov: Same as above.

src/avatar/avatar.component.ts
```diff
@@ -23,49 +21,42 @@ export enum Size {
...
+    @ViewChild("initialsTemplate", { read: TemplateRef }) protected initialsTemplate: TemplateRef<any>;
+    @ViewChild("iconTemplate", { read: TemplateRef }) protected iconTemplate: TemplateRef<any>;
 
     @HostBinding("attr.aria-label") public ariaLabel = "avatar";
     @HostBinding("attr.role") public role = "img";
```
rkaraivanov: Same as above.

PR #592, opened 2018-02-15
Adding events and sample for remote virtualization

### LINKED ISSUE #580: Add events and sample for remote virtualization
### Description  
Add remote virtualization sample for the directive implementation `igxFor`

### Expected result  
The demos should contain Remote forOf sample with igx-grid that shows how users can achieve this scenario

MayaKirova:
Closes #580


src/directives/for-of/for_of.directive.ts
```diff
@@ -358,14 +378,25 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
-            for (let i = this._currIndex; i < endingIndex && this.igxForOf[i] !== undefined; i++) {
-                const input = this.igxForOf[i];
-                const embView = embeddedViewCopy.shift();
-                const cntx = (embView as EmbeddedViewRef<any>).context;
-                cntx.$implicit = input;
-                cntx.index = this.igxForOf.indexOf(input);
+            if (!this.igxForRemote) {
+                const endingIndex = this._pageSize + this._currIndex;
+                for (let i = this._currIndex; i < endingIndex && this.igxForOf[i] !== undefined; i++) {
+                    const input = this.igxForOf[i];
```
ChronosSF: the two for loops seem to differ very slightly and their bodies are identical, therefore this can be simplified by applying borders of the for loop by the value of igxForRemote and leave only one for loop

src/directives/for-of/for_of.directive.ts
```diff
@@ -40,10 +41,18 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
     @Input() public igxForContainerSize: any;
     @Input() public igxForItemSize: any;
     public dc: ComponentRef<DisplayContainerComponent>;
+    public state: IState = {
```
ChronosSF: still not a fan of IState, change it to something like IForOfRemoteState, imo

PR #602, opened 2018-02-15
Stop event propagation for onDialogSelected event

### LINKED ISSUE #590: Nested igx-dialogs close both on click outside
### Description  
When nesting igx-dialog, there is an issue with the "close on click outside" feature. When clicking outside the inner dialog, I would expect that only that dialog is closed, and the parent dialog stays open. Currently, both dialogs are closed, even if the click is not outside the outer dialog.

### Steps to reproduce  

1. Create a igx-dialog
2. Create a igx dialog within the first dialog.
3. Open the inner dialog and click outside

### Result  
Both dialogs are closed.

### Expected result  
Only the inner dialog should close on the first click (at the very least if clicking on the point marked with a red x on the screenshot, as this is not outside the outer dialog).

### Attachments  
image

hanastasov:
Closes #590


kdinev:
@hanastasov Can you add a test for this?

kdinev:
@hanastasov TSLint ;)

Issue #633, opened 2018-02-21
Flaky IgxList Panning Tests


simeonoff:
### Description  
Panning tests fail due to flaky test implementation.

### Steps to reproduce  

1. Uncomment the panning tests in the list spec.
2. Run the tests.

### Result  
The panning tests fail.

### Expected result  
Tests should be passing.

### Attachments  
For instance you can simplify the tests for determining whether an item has been panned to the right like so:

```typescript
  itemNativeElement = fixture.debugElement.queryAll(By.css('igx-list-item'))[0];
  item = list.items[0] as IgxListItemComponent;

  itemNativeElement.triggerEventHandler("panstart", null);
  itemNativeElement.triggerEventHandler("panmove", {
        deltaX: 200
  });
  itemNativeElement.triggerEventHandler("panend", null);

 fixture.detectChanges();

 expect(item.panState).toBe(IgxListPanState.RIGHT);
```  
###Note
The before mentioned tests were commented out in [this pull request](https://github.com/IgniteUI/igniteui-angular/pull/635).

PR #635, opened 2018-02-21
Add themes for multiple components. Allow progress value positioning through CSS classes.

simeonoff:


simeonoff:
Coverage decreased because 3 tests were commented out due to this https://github.com/IgniteUI/igniteui-angular/issues/633

PR #670, opened 2018-02-26
Initial set of filtering UI tests

hanastasov:
#688 

Just initial set of UI tests for filtering. 
More tests are in progress, which will cover all types of columns (boolean, number, etc) and all filtering conditions per the given type.


zdrawku:
We have 4 failing tests because of grid issues. Once the issues are fixed, build will pass

Issue #679, opened 2018-02-27
Polish Column Pinning POC


ChronosSF:
1. change API so it matches spec
2. polish it visually with scrollbars updates, etc.

ChronosSF:
#677

ChronosSF:
@MayaKirova - one issue I see is that toggling the same column over and over leaves the unpinned area empty at some point. Check that one out.

ChronosSF:
Also check the tests that fail. If they are easy to fix, please do so.

PR #687, opened 2018-02-28
Various fixes related to column pinning.

MayaKirova:
Fixing styling issues.
Renaming Api methods.
Adding event onColumnPinning with args - column and insertAtIndex.
Making keyboard navigation work on visible indexes.
Restore original position (based on index) on unpinning the column.


MayaKirova:
Closes #679

Issue #686, opened 2018-02-28
NavDrawer doesn't get pinned when resizing from a small to large screen


kdinev:
### Description  
Resizing my screen from large to small causes the dialog to go into unpinned state. However resizing back to large doesn't bring it back to pinned state.

### Steps to reproduce  

1. Run the repo demos
2. Resize the window to < 1024px
3. Resize back to normal

### Result  
I cannot open the drawer anymore.

### Expected result  
The drawer should have appeared back.

### Attachments  
![image](https://user-images.githubusercontent.com/1472513/36802081-6fb0d092-1cbc-11e8-9fd8-5c42f8ac1e97.png)

damyanpetev:
Yeah, the hamburger button in the demos went missing at some point.
There should be a step before 3. "Click the overlay or select a sample" to hide the drawer as I can't reproduce it otherwise. If you can repro without that do let me know, otherwise I'm labeling this for the demos and I'll add back the button. (see below).

> Notes:
> Opening and closing the drawer is entirely up to the app. It's not even open by default - the initial open is explicit in the demos setup, and the auto-close on selecting a sample is app-specific as well.
> The only state the drawer should be auto-updating is the `pin` and if already closed it remains that way.



As far as how auto-pin works, I certainly don't mind adding more automatic open if needed.
@SlavUI @sbozalieva Thoughts?

P.S. Alternatively this can easily be achieved in the demos app by just binding the open state to the pin:
```html
<igx-nav-drawer [isOpen]="drawerState.pin" [(pin)]="drawerState.pin">
```
Though that'll also close it when it unpins, which I'm not sure is always desired.

sbozalieva:
@damyanpetev Let's not automate it for now.

damyanpetev:
Closing for now. If auto-pin + open combination comes up again in the future this can be reopened.

PR #701, opened 2018-03-02
chore: add back nav toggle button in demos

damyanpetev:
#686  

@simeonoff  The Icon button didn't position correctly vertically, so I had to set the line height in `rem`. Not sure that's the right way to do it.


simeonoff:
This should be fixed automatically with #703. Also, #703 adds the nav-drawer toggle button back, so this pull request is no longer needed.

damyanpetev:
Closing in favor of #703

PR #709, opened 2018-03-06
change calendar default locale to "en"

### LINKED ISSUE #653: Update default calendar locale property value
### Description  
The locale value currently defaults to "en-EN";. MAybe we need to change it to just "en", because there is no "EN country."

The language tag should be changed to default according to the [IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag)

hanastasov:
Closes #653


zdrawku:
@hanastasov make the changes to 'master' branch as well

hanastasov:
Closes #765

PR #710, opened 2018-03-06
feat(icon): implement icon service and multiple font support

### LINKED ISSUE #704: igx-icon should work with both glyph and ligature icon font sets.
### Description
The igx-icon component supports only the material-icon font set and can't be used with any other font icon set/family.

### Steps to reproduce  
1. Import another icon font family into your application (ex. font-awesome)
2. Set the `fontSet` input property to the class name of the newly imported font (ex. 'fa')
3. Set the `name` input property to the icon class name of the newly imported font (ex. 'fa-headphones')

### Result  
The font awesome fa-headphones icon is not shown.

### Expected result  
The igx-icon should provide a way to use any font icon family, with font ligatures, or glyphs.

simeonoff:
Allows users to use igx-icon with various icon fonts - both ligatures and glyphs.
Introduce the igx icon service to allow registering font aliases. Will also be able to create icon sets from svg files in due time.

Closes #704


src/tabbar/tabbar.component.ts
```diff
@@ -20,15 +20,15 @@ import {
     ViewChildren
 } from "@angular/core";
 import { IgxBadgeModule } from "../badge/badge.component";
-import { IgxIconModule } from "../icon/icon.component";
```
kdinev: This should be documented.

src/badge/badge.component.spec.ts
```diff
@@ -15,9 +15,9 @@ describe("Badge", () => {
                 InitBadgeWithDefaultsComponent,
                 InitBadgeWithIconComponent,
                 IgxBadgeComponent,
-                InitBadgeWithIconARIAComponent,
-                IgxIconComponent
-            ]
+                InitBadgeWithIconARIAComponent
+            ],
+            imports: [IgxIconModule.forRoot()]
```
kdinev: Document when singleton should be used.

Issue #693, opened 2018-03-01
Filtering a number column by "equals" returns wrong data


hanastasov:
1) Run the sample (npm install, ng serve)
2) Filter the Downloads column by value 100 and filter condition Equals (the default one)

Results:
Two records are returned, having values 100 and 1000. It seems filtering is done on "contains" filtering condition rather than on "equals"

[igxgrid.zip](https://github.com/IgniteUI/igniteui-angular/files/1770605/igxgrid.zip)

SlavUI:
@rkaraivanov  also to add to this it seems the filtering is Cut Off after you filter 👍 

![image](https://user-images.githubusercontent.com/33216749/36843792-b002355e-1d58-11e8-9479-8ad9c21bcdaf.png)

hanastasov:
Filteruin UI not visible is logged as another issue: https://github.com/IgniteUI/igniteui-angular/issues/691

SlavUI:
This is fixed and merged

PR #719, opened 2018-03-07
Equal date filter to compare date parts

### LINKED ISSUE #695: Filtering by date in IgxGrid returns wrong results
Filter the date column and pick the current day (there is a record in the grid populated with current date value) .

Filtering compares not only the date, but also the time part of the date value and returns 0 recors, while it should return 1 record.

[igxgrid.zip](https://github.com/IgniteUI/igniteui-angular/files/1770974/igxgrid.zip)

hanastasov:
Closes #695


zdrawku:
The problem that I observed is related to #693 after applying both fixes everything works fine.

PR #723, opened 2018-03-07
feat(icon): implement icon service and multiple font support 

### LINKED ISSUE #704: igx-icon should work with both glyph and ligature icon font sets.

simeonoff:
Allows users to use igx-icon with various icon fonts - both ligatures and glyphs.
Introduce the igx icon service to allow registering font aliases. Will also be able to create icon sets from svg files in due time.

Closes #704


src/avatar/avatar.component.spec.ts
```diff
@@ -14,11 +14,11 @@ describe("Avatar", () => {
                 InitAvatarComponent,
                 AvatarWithAttribsComponent,
                 IgxAvatarComponent,
-                IgxIconComponent,
                 InitIconAvatarComponent,
                 InitImageAvatarComponent,
                 InitAvatarWithAriaComponent
-            ]
+            ],
+            imports: [IgxIconModule.forRoot()]
```
kdinev: Same comment as in the previous PR. Document the usage of the `IgxIconService` as a singleton.

src/badge/badge.component.ts
```diff
@@ -1,6 +1,6 @@
 import { CommonModule } from "@angular/common";
 import { Component, HostBinding, Input, NgModule } from "@angular/core";
-import { IgxIconModule } from "../icon/icon.component";
+import { IgxIconModule } from "../icon";
```
kdinev: Document this change as well. In the changelog and in DocFX.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Ignite UI for Angular Change Log
 
 All notable changes for each version of this project will be documented in this file.
+## 5.3.0
```
kdinev: There's already a `5.3.0` section. Please move it under the existing one.

PR #724, opened 2018-03-07
IgxGrid and dynamic components with OnPush detection.

### LINKED ISSUE #631: IgxForOfDirective persists the state of components inside its views
### Description  
The instances of components are reused inside the view container of the directive. This is fine and offers really good performance for simple presentation containers/basic DOM elements.

However any component with complex internal state and behavior that depends on that state, should be recreated in the view.

rkaraivanov:
Added a decorator for the grid components created dynamically,
to trigger change detection.
For additional information check
https://github.com/angular/angular/issues/14087
Improved filtering dialog positioning.
Fixed a bug with `equals` filtering condition and improved
initial filtering condition when opening the filter UI.
Removed workarounds in the interaction between the grid component
and the virtualization directive.

Closes #631 #693


kdinev:
@rkaraivanov Anything for the changelog?

rkaraivanov:
@kdinev I don't think so. The changes are internal implementation logic. There are no breaking changes and no new functionality exposed.

ChronosSF:
@kdinev , there is a sort-of breaking change with the `dirty` property no longer being assigned for newly created views. It's also documented in docfx, so at the very least we should provide some new guidance on how to workaround the issue for users that use igxFor for non-grid scenarios.

src/directives/for-of/for_of.directive.ts
```diff
@@ -101,7 +101,7 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
                 const input = this.igxForOf[i];
                 const embeddedView = this.dc.instance._vcr.createEmbeddedView(
                     this._template,
-                    { $implicit: input, index: this.igxForOf.indexOf(input), dirty: false }
+                    { $implicit: input, index: this.igxForOf.indexOf(input) }
```
kdinev: As per Stamen's comment, please update the changelog.

PR #757, opened 2018-03-12
Fixing issue #744 - Last element rendered with igxFor is not always fully visible

### LINKED ISSUE #744: Last element rendered with igxFor is not always fully visible
### Description  
If the ratio of  the container height and the item height is not an integer, the very last row will not be fully visible.

### Steps to reproduce  

1. Open the Grid Performance sample in Demos
2. Scroll all the way down
3. Observe the very last column

### Result  
It's not fully visible

### Expected result  
The last row should be fully visible.

**Note:** This is to be resolved with a specific case logic that modifies the properties of the display container for the very last chunk so that the height diff is applied to the top instead of bottom.

MayaKirova:
Closes #744

### DIFF
--- a/src/directives/for-of/for_of.directive.ts
+++ b/src/directives/for-of/for_of.directive.ts
@@ -224,6 +224,17 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         if (this.igxForRemote) {
             return;
         }
+        if (endingIndex === this.igxForOf.length) {
+            // scrolled to bottom
+            const h = this.dc.instance._viewContainer.element.nativeElement.clientHeight;
+            const diffContent = this.state.chunkSize * parseInt(this.igxForItemSize, 10) - parseInt(this.igxForContainerSize, 10);
+            const totalDiff = h - parseInt(this.igxForContainerSize, 10) - diffContent;
+            if (totalDiff < 0) {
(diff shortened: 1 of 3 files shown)


kdinev:
@MayaKirova Is this fix applicable to `5.2.x`?

src/directives/for-of/for_of.directive.ts
```diff
@@ -224,6 +224,17 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         if (this.igxForRemote) {
             return;
         }
+        if (endingIndex === this.igxForOf.length) {
+            // scrolled to bottom
+            const h = this.dc.instance._viewContainer.element.nativeElement.clientHeight;
+            const diffContent = this.state.chunkSize * parseInt(this.igxForItemSize, 10) - parseInt(this.igxForContainerSize, 10);
+            const totalDiff = h - parseInt(this.igxForContainerSize, 10) - diffContent;
```
dkamburov: To calculate `totalDiff` there is addition and subtraction of one and the same parameter `igxForContainerSize`. It does not seem to take account for the final result of `totalDiff`. It is the same as(no need of `diffContent`):
`totalDiff = h - this.state.chunkSize * parseInt(this.igxForItemSize, 10);`
@MayaKirova

MayaKirova: Yeah. Makes sense.

Issue #548, opened 2018-02-06
Add semantic-release to the repo


zdrawku:
### Description  
[Semantic release](https://github.com/semantic-release/semantic-release) will help us to:
- generates a changelog
- enforce valid and consistent commit messages (have a look at the [VS code extension - commitizen](https://marketplace.visualstudio.com/items?itemName=KnisterPeter.vscode-commitizen))

- [x] Copy hooks into local .git file to the user.

kdinev:
I guess this is mostly here because we don't enforce commit messages for semantic release?

zdrawku:
Yes, I am also updating the description based on the sprint commitment related to this task

PR #829, opened 2018-03-23
consistent commit messages

Aleksandyr:
#548 

- [x] Still not determined how the hook files should be copied into the local .git directory of the user.


zdrawku:
Alex, update the [Contributing file](https://github.com/IgniteUI/igniteui-angular/blob/master/.github/CONTRIBUTING.md) and also send email to all web tools members with the new commit conventions.

PR #830, opened 2018-03-23
Fix and test for issue #826 

### LINKED ISSUE #826: When igxForContainerSize is initially 0/null/undefined so that no scrollbar is shown, then is updated to have actual height but has small amount of records and then is update to have more records, the scrollbar is not correctly positioned.
### Description  
When igxForContainerSize is initially 0/null/undefined so that no scrollbar is shown, then is updated to have actual height but has small amount of records and then is update to have more records, the scrollbar is not correctly positioned. 

Note: This happens in certain scenarios in igxGrid since the height that needs to be passed to the igxFor directive is calculated runtime.

### Steps to reproduce  

1. Open the virtual-for directive samples from the dev demos.
2. Set the size to be bound to an option ->[igxForContainerSize]='size', and set the variable to be 0/null/undefined initially.
3. Set the data to be an empty array on ngOnInit. No scrollbar should be shown.
4. On ngAfterViewInit set the size to be an actual size in pixels ->  this.size = "500px";
5. Add a button to the template and on click bind more data to the virtfor so that a scrollbar will be shown.

### Result  
The scrollbar is not correctly positioned.

### Expected result  
The scroll should be correctly positioned.

MayaKirova:
Making sure the igx-display-container--inactive class is applied only when the data in the container is not virtualized (for example when there's no igxForContainerSize or when there's not enought data to display a scrollbar) .

Closes #826  .

### DIFF
--- a/src/directives/for-of/for_of.directive.spec.ts
+++ b/src/directives/for-of/for_of.directive.spec.ts
@@ -20,6 +20,7 @@ import { By } from "@angular/platform-browser";
 import { IgxForOfDirective, IgxForOfModule} from "./for_of.directive";
 
 describe("IgxVirtual directive - simple template", () => {
+    const INACTIVE_VIRT_CONTAINER = "igx-display-container--inactive";
     beforeEach(async(() => {
         TestBed.configureTestingModule({
             declarations: [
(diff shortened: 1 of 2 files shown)


ChronosSF:
@MayaKirova  after binding a new array (if the container size is set on ngAfterViewInit) scrolling with the mouse wheel scrolls the page as well. This is not the case in the unmodified sample.

MayaKirova:
@ChronosSF The issue reproduces before the fix as well.
Seems to be due to the logic in the onWheel event. It uses this.vh.instance.height, which we don't update on _recalcScrollBarSize (we update the style.height directly). We should unify the way the height is updated/retrived. This should be handled as a separate issue.

PR #840, opened 2018-03-26
Fix for issue #838 - When setting the igxForContainerSize post init and binding a new array scrolling with the mouse wheel scrolls the page as well.

### LINKED ISSUE #838: When setting the igxForContainerSize post init and binding a new array scrolling with the mouse wheel scrolls the page as well.
### Description  
When setting the igxForContainerSize post init and binding a new array scrolling with the mouse wheel scrolls the page as well.

### Steps to reproduce  

1. Open the virtual-for directive samples from the dev demos.
2. Set the size to be bound to an option ->[igxForContainerSize]='size', and set the variable to be 0/null/undefined initially.
3. Set the data to be an empty array on ngOnInit. No scrollbar should be shown.
4. On ngAfterViewInit set the size to be an actual size in pixels -> this.size = "500px";
5. Add a button to the template and on click bind more data to the virtfor so that a scrollbar will be shown.
6. Try scrolling with the mouse wheel while hovering over the display container.

### Result  
The page is also scrolled.

### Expected result  
Just the display container should be scrolled.

MayaKirova:
Closes #838  .

### DIFF
--- a/src/directives/for-of/for_of.directive.ts
+++ b/src/directives/for-of/for_of.directive.ts
@@ -505,8 +505,7 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         if (this.igxForScrollOrientation === "vertical") {
             const count = this.totalItemCount || this.igxForOf.length;
             this.vh.instance.elementRef.nativeElement.style.height = parseInt(this.igxForContainerSize, 10) + "px";
-            this.vh.instance.elementRef.nativeElement.children[0].style.height =
-                (count * parseInt(this.igxForItemSize, 10)) + "px";
+            this.vh.instance.height = count * parseInt(this.igxForItemSize, 10);
         }
     }
 


ChronosSF:
@MayaKirova the fix only appears to work on Chrome. In FF, Edge and IE the vertical scrollbar does not appear at all when binding post init.

MayaKirova:
@ChronosSF The issue seems to be with the height limitations in IE/Edge/Firefox. The scrollbar is there, the height is set, however when the height is over a certain threshold the browser stops respecting it and doesn't show a scrollbar. The threshold depends on the browser (around 17,895,697 for FireFox, 10,737,418 for IE etc.). This seems to be related to another already logged issue - #841, which is also related to the height limitation. In any case this has nothing to do with the mouse wheel scrolling, so you should either add it as an additional scenario in issue #841 or log it separately.

ChronosSF:
@MayaKirova please add your observations to #841 and suggest logging this in the limitations section for igxFor if it's  not fixable.

PR #842, opened 2018-03-27
Fix grid rows

desig9stein:
1. Add 1fr to grid body
2. Add Grid paginator theme


src/grid/grid.component.html
```diff
@@ -65,7 +57,16 @@
         </ng-template>
     </div>
 </div>
```
simeonoff: Delete lines 44-59 if no longer needed.

PR #847, opened 2018-03-27
feat(igx-grid): add primaryKey to grid, #805

### LINKED ISSUE #805: igxGrid - primaryKey specification and implementation
### Description  
In order for the `igx-grid` component to accommodate additional features (e.g. row selection) while still using data virtualization, a new way to select data rows needs to be introduced.

Implement a new feature for the `igx-grid` component that allows a `primaryKey` attribute to be passed. The value specified in the `primaryKey` should specify a parameter from that data set that will be used as a unique identifier for the rows in the grid. This `primaryKey` can be used for selecting rows wherever the current method (using row index) is not applicable (e.g. addressing rows persistently through sorting and filtering) .

### Steps to reproduce  
1. Create a grid and try to programatically select a row using index.
2. Filter and/or sort the grid
3. Select the row at the same index as in step 1

### Result  
The row data is different.

### Expected result  
There should be a way to consistently select the data object corresponding to a row. This selection should not be related to the row's position in the viewport.

ViktorSlavov:
Closes #805.  

Add primaryKey attribute to igx-grid. Allow row selection through value for rowData[primaryKey].


src/grid/api.service.ts
```diff
@@ -51,12 +51,13 @@ export class IgxGridAPIService {
         }
     }
 
-    public get_row(id: string, index: number): IgxGridRowComponent {
-        return this.get(id).rowList.find((row) => row.index === index);
+    public get_row(id: string, rowSelector: any): IgxGridRowComponent {
+        const primaryKey = this.get(id).primaryKey;
+        return this.get(id).rowList.find((row) => primaryKey ? row.rowData[primaryKey] === rowSelector : row.index === rowSelector);
```
ChronosSF: I wouldn't do it this way - yes it looks cleaner as a one-liner but you are checking if primaryKey is set for each record instead of checking it once outside and running different finds based on the result.
@rkaraivanov - not sure if we support two dimensional arrays for igxGrid, but if we do and the primaryKey is the 0-index column, then the check will fail as well ( if (0) = false )

src/grid/cell.component.ts
```diff
@@ -68,7 +68,7 @@ export class IgxGridCellComponent implements IGridBus, OnInit {
         return this.gridAPI.get(this.gridID);
     }
 
-    get rowIndex(): number {
+    get rowIndex(): any {
```
ChronosSF: Why did you change the return type? Index should still be a number.

src/grid/grid.component.ts
```diff
@@ -318,12 +321,16 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         return this.rowList.toArray()[index];
     }
 
+    public getRowByKey(keyValue: any): IgxGridRowComponent {
+        return this.rowList.find((row) => row.primaryValue === keyValue);
```
ChronosSF: Is this needed _in addition_ to the API service method?

src/grid/cell.component.ts
ViktorSlavov: The return type was changed with the initial implementation idea, it was an oversight not to revert it, thanks for pointing it out!

src/grid/api.service.ts
ViktorSlavov: Reworked in the latest commit

src/grid/grid.component.ts
ViktorSlavov: I added it as a counterpart to the getRowByIndex method (as there is no other exposed method to get a row reference to a row). We could refactor getRowByIndex to incorporate the gridAPI get_row method and have it except both index or key?
```
public getRow(keyOrIndex: any): IgxGridRowComponent {
        return this.gridAPI.get_row(this.id, keyOrIndex);
    }
```

CHANGELOG.md
```diff
@@ -56,6 +56,9 @@ All notable changes for each version of this project will be documented in this
         - The `igxGrid` `onEditDone` now exposes arguments of type `IGridEditEventArgs`. The arguments expose `row` and `cell` objects where if the editing is performed on a cell, the edited `cell` and the `row` the cell belongs to are exposed. If row editing is performed, the `cell` object is null. In addition the `currentValue` and `newValue` arguments are exposed. If you assign a value to the `newValue` in your handler, then the editing will conclude with the value you've supplied.
         - The `igxGrid` `onSelection` now correctly propagates the original `event` in the `IGridCellEventArgs`.
     - Added `jsZip` as a Peer Dependency.
+- `primaryKey` attribute added to `igxGrid`
+    - `primaryKey` allows for a property name from the `data` source to be specified. If specified, `primaryKey` can be used instead of `index` to indentify grid rows from the `igxGrid.rowList`. As such, `primaryKey` can be used for selecting rows for the following `igxGrid` methods - `deleteRow`, `updateRow`, `updateCell`, `getCellByColumn`, `getRowByKey`
```
kdinev: Remove the ` ` from `data`. Explain that the `primaryKey` is better suited for uniquely identifying data members, especially when the DOM is virtualized, because indices are reused in the virtual DOM scenario.

ViktorSlavov: Thanks for pointing those out! They're addressed in the latest commit.

PR #864, opened 2018-03-29
Fix for issue #841 - Handling browser's height limitations.

### LINKED ISSUE #841: Virtualizing a lot of items in Edge/IE causes navigation issues
### Description  
When the scroll container exceeds a certain threshold IE and Edge produce navigation/scrolling errors with igxFor.

### Steps to reproduce  

1. Open the Grid Performance sample
2. Scroll down a bit then select a cell
3. Try to navigate up with the keyboard

### Result  
The grid navigates *down*

### Expected result  
The grid should navigate *up*

(there are other effects of the issue that can be observed easily)

MayaKirova:
Closes #841 .  

Handling browser's height limitations

Issues to resolve:
- [x] In Edge/IE11 - when navigating with the keyboard the selected cell is always somewhat out of view in the Performance Sample (basically it happens only if the height ratio code is executed)


src/directives/for-of/for_of.directive.ts
```diff
@@ -66,6 +66,8 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
     private _lastTouchY = 0;
     private _pointerCapture;
     private _gestureObject;
+    private _maxSize;
```
mpavlinov: @MayaKirova How about naming it `_maxHeight`.

src/directives/for-of/for_of.directive.ts
```diff
@@ -404,6 +410,17 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         }
     }
 
+    protected _calcMaxBrowserHeight() {
```
mpavlinov: @MayaKirova The new functions start with `_calc`(_calcMaxBrowserHeight, _calcHeight). The old one is `_calculate` (_calculateChunkSize). Maybe we should adopt a unified prefix for `_calc`.
There is no return type declared.

src/directives/for-of/for_of.directive.ts
```diff
@@ -497,8 +514,18 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         if (this.igxForScrollOrientation === "vertical") {
             const count = this.totalItemCount || this.igxForOf.length;
             this.vh.instance.elementRef.nativeElement.style.height = parseInt(this.igxForContainerSize, 10) + "px";
-            this.vh.instance.height = count * parseInt(this.igxForItemSize, 10);
+            this.vh.instance.height = this._calcHeight();
+        }
+    }
+
+    private _calcHeight() {
```
mpavlinov: @MayaKirova There is no return type declared.

src/directives/for-of/for_of.directive.ts
```diff
@@ -66,6 +66,8 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
     private _lastTouchY = 0;
     private _pointerCapture;
     private _gestureObject;
+    private _maxSize;
+    private _virtSizeRatio = 1;
```
mpavlinov: @MayaKirova How about renaming it `_virtHeightRatio`.

src/directives/for-of/for_of.directive.ts
```diff
@@ -196,7 +199,12 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
             }, 0);
             this.hScroll.scrollLeft = sumWidths;
         } else {
-            this.vh.instance.elementRef.nativeElement.scrollTop = parseInt(this.igxForItemSize, 10) * index;
+            const vhElem = this.vh.instance.elementRef.nativeElement;
+            const containerSize = parseInt(this.igxForContainerSize, 10);
+            const vcHeight = (vhElem.children[0].scrollHeight - containerSize) * this._virtSizeRatio + containerSize;
+            const count = this.totalItemCount || this.igxForOf.length;
+            const scrolllTop = ((index / count) * vcHeight) / this._virtSizeRatio;
```
mpavlinov: @MayaKirova `scrolllTop`. Too many "l"s?

MayaKirova:
@ChronosSF Please re-test the issues after the changes and the latest merge from master.

tbelcheva:
It seems like scrolling happens before selection in IE11 and when you scroll up or down using the arrow keys the selection jumps. Navigating up and down doesn't happen smoothly as in Edge.

MayaKirova:
@tbelcheva Selection would always happen after scrolling if the next cell is not in the currently loaded chunk of data. The grid would need to first scroll so that the next cell is loaded into the view and then select it. This applies to all browsers.

skrustev:
This is a performance issue in the navigation, because on IE11 it is a lot slower compared to other browsers and the way it works is more visible. Like @MayaKirova said it first scrolls then selects the correct cell and this is more visible on IE11.

PR #865, opened 2018-03-29
Fix for issue #841- Handling browser's height limitations. 5.2.x

### LINKED ISSUE #841: Virtualizing a lot of items in Edge/IE causes navigation issues

MayaKirova:
Closes #841.  

Handling browser's height limitations.

### DIFF
--- a/src/core/styles/components/_common/_igx-vhelper.scss
+++ b/src/core/styles/components/_common/_igx-vhelper.scss
@@ -5,7 +5,6 @@
 
 %vhelper--vertical {
     width: 18px;
-    height: calc(100% - 18px);
     float: right;
 }
 
--- a/src/directives/for-of/for_of.directive.spec.ts
+++ b/src/directives/for-of/for_of.directive.spec.ts
@@ -804,7 +804,7 @@ export class TestIgxForOfDirective<T> extends IgxForOfDirective<T> {
     }
 
     public testCalculateChunkSize(): number {
-        return super._calculateChunkSize();
+        return super._calcChunkSize();
     }
 
     public testInitHCache(cols: any[]): number {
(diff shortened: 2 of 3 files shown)


MayaKirova:
@kdinev We’ve been discussing this with the team and some are in favor of leaving this as a limitation of the igxFor (and igxGrid) instead of applying the changes in this PR, because of the increased code complexity.

For some context of the exact issue, here are the browsers current height limitations:
Chrome :    	134,217,726 px
Safari:    	134,217,726 px
IE:     		10,737,418 px
FF  :     		17,895,697 px

If an element has a bigger height then the max it renders no scrollbars (actual clientHeight becomes 0).
The limit is relatively easy to hit in IE and FireFox.

For instance in the igxGrid where the default row height is 50px limit will be hit with around 214-215k rows. If item height is increased, the limit will be hit faster.

Let me know if you’re ok with the changes or if you prefer to document this as a limitation and drop the PR.

kdinev:
@MayaKirova I'm fine with leaving this as a limitation.

ChronosSF:
@kdinev after discussing this further we decided to go for the fix as such solutions are already available in other popular grid implementations and ultimately we think we shouldn't have a situation where a customer that wants to display 500 000 rows for example decides to go for another library just because they workaround the limitations and we don't.

PR #869, opened 2018-03-29
Call preventDefault and stopPropagation in the wheel event handler

### LINKED ISSUE #862: When scrolling with the mouse wheel the whole page is scrolled as well
### Steps to reproduce  

1. Open any timepicker in a page with visible page scrollbar
2. Start to scroll with the mouse wheel on the minutes/seconds scrolling area

### Result  
The whole page is scrolled along with the time values

### Expected result  
Only time picker values should be scrolled

### Note:
Have a look at datePicker years scroll area, the issue do not persist there.

### Attachments  
image

sstoyanovIG:
Call preventDefault and stopPropagation in the wheel event handler

Closes #862 .


zdrawku:
@sstoyanovIG there are lint errors

zdrawku:
@npaunov there are failing timePicker test, could you have a look at them

PR #880, opened 2018-04-02
igxGrid remote virtualization implementation and dev sample.

### LINKED ISSUE #877: igxGrid support for remote virtualization
### Description  
igxFor supports remote virtualization. We need to add it in igxGrid as well.

MayaKirova:
Closes #877.


MayaKirova:
Current Issues:
- [ ] Summaries + Remote virtualization causes a second scrollbar to show.
- [ ] Handle scenarios where the service returns less items than the actual chunk size of the grid.

src/grid/grid.component.ts
```diff
@@ -181,6 +182,9 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
     @Input()
     public columnWidth: string = null;
 
+    @Input()
+    public remoteVirtualization: boolean;
```
kdinev: Why is this input needed? What is the difference between remote and local for the grid?

ChronosSF: Its linked to the `igxFor` property with the same purpose (because the user can't pass options to the internal `igxFor`) . The `igxFor` property is used mainly to distinguish between the situation where the data input is local and it has to show *m* of *n* items and the situation where it is remote and it has to show all items that came from the service.

This is just an explanation for the current situation. I will check if we can get rid of the igxFor prop and therefore remove the igxGrid one as well. It may be redundant and inferable from other input props.

ChronosSF: @kdinev - `igxFor` assumes that the user will set its `totalRecordsCount` public property (not exposed as an input) so that it knows how large its scrollbar should be. This means that remote virtualization may be inferred from having this property set to something different than the default. I think it's a good way to get rid of this additional input.

ChronosSF:
@kdinev , please review again

src/directives/for-of/for_of.directive.ts
```diff
@@ -45,7 +45,7 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         startIndex: 0,
         chunkSize: 0
     };
-    public totalItemCount: number;
+    public totalItemCount: number = null;
 
     @Input() public igxForRemote = false;
```
kdinev: If this is being deprecated, then set a deprecation warning on it.

PR #919, opened 2018-04-05
igxAutoFocus and igxAutoSelection implementation and integration

### LINKED ISSUE #800: When cell edit mode is activated I should be able to start typing
### Description  
I am not able to change the cell value when cell edit mode is invoked on double click.

### Steps to reproduce  

1. Configure a grid with editable column
2. Double click any cell from the column
3. Start typing something

### Result  
Cell value is not changed

### Expected result  
Cell value should be changed

### Attachments  

image

Aleksandyr:
Closes #800


rkaraivanov:
Also the `igxSelection` directive should be renamed to `igxTextSelection`.

PR #922, opened 2018-04-05
feat(switch): add label positioning, ripple, and accessibility

simeonoff:


hanastasov:
Good work and nice set of tests !

PR #923, opened 2018-04-05
chore(*): adding csv/excel services to exports, tslint warnings cleanup

ChronosSF:
- fixed two tslint unused variables warnings
- added the csv/excel exporters to main.ts exports so that the demos app work

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -54,3 +54,8 @@ export * from "./directives/mask/mask.directive";
 export * from "./time-picker/time-picker.component";
 export * from "./directives/ripple/ripple.directive";
 export * from "./directives/toggle/toggle.directive";
+
+// services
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -54,3 +54,8 @@ export * from "./directives/mask/mask.directive";
 export * from "./time-picker/time-picker.component";
 export * from "./directives/ripple/ripple.directive";
 export * from "./directives/toggle/toggle.directive";
+
+// services
+export * from "./services/csv/csv-exporter";
```
rkaraivanov: Those should not be exported in the main bundle because we make JSZip a must have dependency for our package. Some users will not care for the excel export functionality at all so I don't think we should force another dependency on them.

PR #938, opened 2018-04-10
trigger text selection when element is focused

### LINKED ISSUE #936: trigger text selection properly through igxTextSelection directive
### Description  
There is no indication of text selection whenever you focus an input element when igxTextSelection directive is applied.

### Steps to reproduce  

1. Apply igxTextSelection directive into input element which contains some text.
2. Focus the input

### Result  
There is no indication of text selection

### Expected result  
Text should be marked as selected.

Aleksandyr:
Closes #936


zdrawku:
If I focus a simple input element with igxTextSelection directive, the input value became selected automatically.

`<input value="test" [igxTextSelection]="true" />`

image

Issue #668, opened 2018-02-26
Adding Export to Excel, CSV and TSV exports to the igxGird


SlavUI:
### Description  
As a User i would like to Export my date from the igxGrid into an excel document. The data that will be exported should be formatted according to the content each cell contains.

DDavidkov:
Linking the feature spec:

https://github.com/IgniteUI/igniteui-angular/wiki/GridExcelExporter-Specification

hanastasov:
Will the ColumnExporting event fire for each cell under the given column, or will fire once when the last cell under this columns is being exported ?

SlavUI:
Feature completed

PR #946, opened 2018-04-10
Excel export enhancements

DDavidkov:
#668   

Added several enhancements to the exporters:

- Numeric values are no longer saved as strings
- Added ignoreSorting property, which specifies if the exporters should honor the current sort in the grid
- Added exportAsTable option to the excel exporters, so that you can specify if the exported excel file will contain a table. It default to true.
- The exporters will now also use the column's formatter when exporting


src/services/excel/excel-exporter-grid.spec.ts
```diff
@@ -20,7 +20,7 @@ import { JSZipFiles } from "./jszip-helper";
 import { JSZipWrapper, ObjectComparer } from "./jszip-verification-wrapper";
 import { ExportTestDataService, FileContentData, ValueData } from "./test-data.service";
 
-describe("Excel Exporter", () => {
+fdescribe("Excel Exporter", () => {
```
gedinakova: Remove the focus from the spec.

src/services/excel/excel-exporter-grid.spec.ts
```diff
@@ -148,7 +148,7 @@ describe("Excel Exporter", () => {
     //     });
     // }));
 
-    it("should honor 'ignoreFiltering' option.", async(() => {
+    fit("should honor 'ignoreFiltering' option.", async(() => {
```
gedinakova: Remove the focus from this test.

PR #974, opened 2018-04-12
fix(igxGrid): resolving width calc issue #969

### LINKED ISSUE #969: Pinned columns are misaligned in grid body
### Description  
In 5.3.x, Pinned columns in grid body are misaligned with the grid header, summaries, and horizontal scroll. 
I assume that the problem comes from the vertical scrollbar since the misaligned is about the same width as the scrollbar

### Steps to reproduce  

1. Open branch 5.3.x
2. Open grid summaries demo
3. Press the enable summary button (optional)
4. Press the Pin button

### Result  
Pinned columns in grid body are misaligned with about 18px with the grid header and summaries

### Expected result  
Pinned columns in grid body should be aligned with the grid header and summaries 

### Attachments  
screenshot-2018-4-12 igniteui for angular samples

ChronosSF:
Closes #969 .


src/grid/cell.component.ts
```diff
@@ -162,11 +160,12 @@ export class IgxGridCellComponent implements IGridBus, OnInit {
     @HostBinding("class.igx-grid__th--pinned-last")
     get isLastPinned() {
         const pinnedCols = this.grid.pinnedColumns;
-        if (pinnedCols.length === 0) {
-            return false;
-        } else {
-            return pinnedCols.indexOf(this.column) === pinnedCols.length - 1;
-        }
+        return pinnedCols[pinnedCols.length - 1] === this.column;
```
ChronosSF: Simplified this line because it is safe to access whatever element of an array, you just get undefined if the index is unavailable (as in the case when the array is empty)

PR #1009, opened 2018-04-16
refactor(grid): add column resizing themability

simeonoff:


SAndreeva:
Resize line does not cover grid body area for unpinned columns.
resizer

PR #1017, opened 2018-04-16
Fix for #965 - adding change detection for ngModel bound templates in igxGrid

### LINKED ISSUE #965: Custom cell template bound to ngModel doesn't update correctly after scrolling
### Steps to reproduce  

1. Open the attached sample
2. run npm install
3. run ng serve -o
4. navigate to custom grid
5. observe the last two columns and their data values
6. scroll the grid down and then back to top
7. Observe the values in the last two columns

### Result  
The dates as well as checked state has changed. If you click on the ckecxbox cell the values are correctly recalculated which makes a lot of dates and checkboxes change their state

### Expected result  
Last two columns keep their state when scrollin

### Attachments  
[igx-grid-scrolling.zip](https://github.com/IgniteUI/igniteui-angular/files/1899259/igx-grid-scrolling.zip)

ChronosSF:
Closes #965  .  

This needs unit tests but they seem quite hard to implement cleanly.

@rkaraivanov - please, try this without `setVerticalScrollSubscription` - I don't think it works. There is a sample that you can use in #965.

***TODO:*** - currently not possible to scroll for the test to be useful
~~- [ ] Add a bug test~~

### DIFF
--- a/src/grid/cell.component.ts
+++ b/src/grid/cell.component.ts
@@ -2,7 +2,6 @@
     ChangeDetectionStrategy,
     ChangeDetectorRef,
     Component,
-    DoCheck,
     ElementRef,
     HostBinding,
     HostListener,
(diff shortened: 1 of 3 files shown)


ChronosSF:
@Lipata @ViktorSlavov  - there is a test that throws a cleanup error: "Should properly move focus when loading new row chunk"

It uses `setTimeout` which is most likely why it fails. I don't think the test waits for the timeout to execute so it likely that it does not execute the asserts inside the `setTimeout` callback.

tbelcheva:
:white_check_mark: verified

Issue #866, opened 2018-03-29
Expose igxGrid option for default column width


ChronosSF:
The following is included in the task:

- expose either `columnWidth` or `defaultColumnWidth` property (default is `null`) for igxGrid. Assign it for columns that have no width set
- Update the following docs:
   - Grid spec
   - Grid ReadMe
   - Grid DocFX topic (relates to [#233 in the DocFX repo](https://github.com/IgniteUI/igniteui-docfx/issues/233))

PR #1038, opened 2018-04-17
Default col width

kdinev:
Related #866 .  

`columnWidth` being `136px` is from 1366 screen width being the most common screen size we're getting visits from, and 10 being the average columns in a grid sample we're getting.


ChronosSF:
AFAIK, this will require some doc updates as well.

Issue #947, opened 2018-04-10
igx -grid - headers and cells are not properly aligned


SAndreeva:
### Description  
Headers and cells are not properly aligned if width is not set on columns.

### Steps to reproduce  

1. Run the attached sample.
2. Observe igxGrid headers and cells.

### Result  
Headers and cells are not properly aligned - see the screenshot below.

### Expected result  
Headers and cells are aligned.

### StackBlitz sample
https://stackblitz.com/edit/grid-header-cell-misalignment

![igxgrid](https://user-images.githubusercontent.com/7972965/38564570-1129438c-3ce8-11e8-896f-7a402d500068.png)

tiliev:
The issue is removed with removed width and height.

SAndreeva:
@zdrawku , you can close the issue.

PR #1042, opened 2018-04-18
Grid cells misalignment bug fixing - #1034, #947

### LINKED ISSUE #1034: igxGrid cells are misaligned when a single column is pinned with specified width
### Description  
igxGrid cells are misaligned when a single column is pinned with specified width

**NOTES:**
The issue might be related to #1008 

### Steps to reproduce  

Download and run the attached sample application.

### Result  
All cells are misaligned.

### Expected result  
All cells are properly aligned under their respective headers.

### Attachments  

image
[igxgridSample.zip](https://github.com/IgniteUI/igniteui-angular/files/1919776/igxgridSample.zip)

SAndreeva:
Closes #1034, #947 .


ChronosSF:
@SAndreeva , I am taking a look at this currently. The increased complexity of the width prop is what concerns me. It's a property that gets called a lot and we should keep it simple for performance reasons. If #1045 addresses #947 and #1034 only happens when pinning columns when no horizontal scrollbar is shown we might choose to disable pinning in this case rather than applying this fix. (pinning doesn't make sense in the first place without a horizontal scrollbar)

ChronosSF:
Discussed this with @simeonoff  - this fixes the case with pinning indirectly by specifying pixel widths based on the column widths (which can be null or %-based) . This, however, will kill the idea of setting responsive column width as defaults, that we decided on today.

@rkaraivanov , please, let us know what you think. Me and @simeonoff believe that #1034 can be resolved in another way that I explore today. Otherwise we might even forbid pinning when no horizontal scrollbar is shown (which is the only time this misalignment issue is present).

PR #1045, opened 2018-04-18
fix(igxGrid): removing column misalignment when there is no vertical scrollbar

### LINKED ISSUE #1044: When there is no vertical scrollbar a slight misalignment occurs between the header and body columns
### Steps to reproduce  

1. Open the attached sample
2. Inspect the header and body cells

### Result  
They are misaligned slightly

### Expected result  
They should be aligned

### Attachments  
[igxgridSample.zip](https://github.com/IgniteUI/igniteui-angular/files/1923027/igxgridSample.zip)

ChronosSF:
Closes #1044 .


ChronosSF:
@SAndreeva , do you think this addresses #947 ? The stackblitz sample shows the same issue, the screenshot in the issue made it a little bit confusing as to what the cause is.

SAndreeva:
@ChronosSF, I am not really sure, I will take a look

PR #1048, opened 2018-04-18
fix(igxGrid): Column duplication when pinnning/unpinning

### LINKED ISSUE #1043: Unpinning a column that is already unpinned creates a duplicate of that column
### Description  
Unpinning a column that is already unpinned creates a duplicate of that column.

**NOTES:**
As many times we invoke the unpin() method for an unpinned column, that many duplicates of the column will be created.

### Steps to reproduce  

1. Go to the demo app for grid-column-pinning sample and modify the toggle logic for the pinning of a column to invoke the unpin() method only (instead of toggling).
**Example:**
		col.unpin();
		// col.pinned ? col.unpin() : col.pin();
2. Run the "Grid Column Pinning" demo.
3. Click the "PIN/UNPIN 'COMPANYNAME'" button.

### Result  
A duplicate column of the CompanyName column is created.

### Expected result  
Nothing should happen since the CompanyName is already unpinned.

### Attachments  

image

rkaraivanov:
Pinning/unpinnig a column already in one of the states creates a
duplicate of the column.

Closes #1043


src/grid/grid.component.ts
```diff
@@ -720,6 +720,9 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
     public pinColumn(columnName: string): boolean {
         const col = this.getColumnByName(columnName);
 
+        if (col.pinned) {
+            return;
```
skrustev: It should return false when no pinning is done. Also for the if can be combined with the if below.

src/grid/grid.component.ts
```diff
@@ -748,6 +751,10 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
 
     public unpinColumn(columnName: string) {
         const col = this.getColumnByName(columnName);
+
+        if (!col.pinned) {
+            return;
```
skrustev: It should return false when no unpinning is done. I noticed that at the end of the function there is no return true. That should also be added.

PR #1055, opened 2018-04-18
refactor(grid): update sort icon in column header

simeonoff:


simeonoff:
@SAndreeva I can't reproduce it.

PR #1074, opened 2018-04-19
feat(igxGrid): Auto sizing of the grid dimensions

### LINKED ISSUE #948: igx -grid - pinned columns are not visualized correctly if no width is set
### Description  
Pinned columns are not visualized correctly if no width is set on that columns.

### Steps to reproduce  

1. Run the attached sample.
2. Observe how pinned columns are displayed.

### Result  
Header titles are not shown, no text is displayed - see the attached screenshot.

### Expected result  
Text is displayed on pinned columns, they are properly sized.

### Attachments  
igxgrid
[igxgrid.zip](https://github.com/IgniteUI/igniteui-angular/files/1895263/igxgrid.zip)

### LINKED ISSUE #1034: igxGrid cells are misaligned when a single column is pinned with specified width

### LINKED ISSUE #1054: New Grid width/height defaults
# Description  
The general idea is that when a user provides data to the `igxGrid`, the grid should display the data in the best possible layout, if dimensions are not specifically provided. The current behavior of the grid is that the grid will grow in height and width according to the size of the provided data (columns and rows) without displaying scrollbars, and thus will display smaller sets of data good by default, but will have a very large display for bigger sets of data. In general, the data should be visualized within the viewport in the best possible way, without extending the grid outside of the viewport, and regardless of the data size. This means that if larger sets are detected, then the grid should display a virtualized chunk of the data by default, without asking the developer to explicitly provide grid, record and column dimensions.

## Test cases
 1. Default rendering of data with 5 columns and 5 rows.
 2. Default rendering of data with 10 columns and 30 rows.
 3. Default rendering of data with 30 columns and 1000 rows.
 4. Default rendering of data with 150 columns and 20000 rows.
 5. Rendering of data with 5 columns and 5 rows where 2 of the columns have width set.
 6. Rendering of data with 5 columns and 30 rows where 3 of the columns have width set.
 7. Rendering of data with 30 columns and 1000 rows where 5 of the columns have width set.
 8. Rendering of data with 150 columns and 20000 rows where 5 of the columns have width set.

Define new defaults for grid width/height:
### width 
 - 100% - the grid should, by default, be spanning the container it's placed in.

### columnWidth 
 - null by default. The grid should be inferring from the number of columns and the container it's placed in, what the default column width should be set to.
 - the grid should span all of its columns within the grid body, but not allowing columns to go below a certain width by default. The minimum width can be set to 136px, which is 10 columns displayed within a screen size of 1366px. 
   - e.g.
      - 5 columns in a 1600px wide viewport should be displayed by each column being 320px wide.
      - 15 columns in a 1600px wide viewport should be displayed with a horizontal scrollbar and each column will be 136px wide - this will also be virtualized.
 - when some, but not all, of the columns have explicit width set, and there's no defaultWidth provided, then the same calculations should be applied, but taking into account the width set for the columns with provided configuration.
   - e.g.
     - 5 columns in a 1600px wide viewport, with two of the columns set at 500px width, should be displayed by each column being 200px wide, except for the two that are 500px wide.
     - 15 columns in a 1600px wide viewport, with two of the columns set at 500px width, should be displayed with a horizontal scrollbar and each column will be 136px wide, except for the two that are 500px wide - this will also be virtualized.

### height 
 - 100% - if parent container = 0 then use the height of the viewport/window. 
 - if rowHeight * rowCount <  height then set height = null.

Rationale:
The grid should by default be virtualized.
The grid should render in any configuration related to width/height/columnWidth.

rkaraivanov:
- [x] Unit tests
- [x] Feature integration testing
- [x] Virtualization offset test/solution when resizing a percent based grid scrolled to the last column.

Closes #1054, Closes #1034 , Closes #948


ChronosSF:
Closes #1034

ChronosSF:
Closes #948

PR #1081, opened 2018-04-20
fix(grid-filtering): Push the filtering menu below the grid header

### LINKED ISSUE #1078: Grid: Pinned column headers appear on top of filter menu
### Description  
On a grid with pinned and filterable columns, when an unpinned column's filter is activated, it is displayed below the pinned column header.

### Steps to reproduce  
1. Pin a column
2. Make/Find an filterable column (pinned or unpinned)
3. Move the unpinned column as close as possible to the pinned one
4. Click on the icon to activate the filter menu

### Result  
the filter menu is displayed below the pinned column header

### Expected result  
the filter menu should be displayed above any column header.

### Attachments  
https://stackblitz.com/edit/crm-grid-sample-demo
screen shot 2018-04-19 at 17 18 09

simeonoff:
Because of the way the filtering menu is currently implemented as part
of the header, its visibility is affected by the z-index of the header
itself. So if a column header has a higher z-index than the next
sibling, that sibling will have its filtering UI covered by the previous
header. To avoid that I am pushing the filtering menu below the header
so it doesn't get covered by the headers.

Closes #1078

### DIFF
--- a/src/core/styles/components/grid-filtering/_grid-filtering-theme.scss
+++ b/src/core/styles/components/grid-filtering/_grid-filtering-theme.scss
@@ -111,7 +111,7 @@
     $filtering-menu-min-width: 240px;
     $filtering-menu-shadow: igx-elevation($elevations, 8);
     $filtering-menu-radius: 4px;
-    $filtering-menu-distance-to-toggle: 4px;
+    $filtering-menu-distance-to-toggle: 24px;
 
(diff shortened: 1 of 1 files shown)


mpavlinov:
Note: The issue is fixed. The drop-down opens below the header, but now it's noticeable away from the button.

PR #1082, opened 2018-04-20
Ensure smooth vertical scrolling

### LINKED ISSUE #1016: There is a gap between the last visible grid row and the column summaries
### Description  
In the Grid Column Pinning demo, when 600px height has been set to the grid, there is a gap between the last visible row and the summaries. 

### Steps to reproduce  

1. Open branch 5.3.x
2. Set the height of the grid to 600px and enable the column summaries for the column pinning demo
3. Open the Grid Column Pinning demo

### Result  
When scrolling down, there is a visible gap between the last grid row and the summaries

### Expected result  
There shouldn't be any gap between the last row and the summaries

### Attachments  
sketch2

bkulov:
Closes #1016   

Add one additional (non-visible) row in the virtual vertical scroller chunk. This will ensure the smooth experience when scrolling up and down. The additional row is removed when we scroll to the bottom.

### DIFF
--- a/src/directives/for-of/for_of.directive.spec.ts
+++ b/src/directives/for-of/for_of.directive.spec.ts
@@ -179,6 +179,23 @@ describe("IgxVirtual directive - simple template", () => {
         }
     });
 
+    it("should scroll render one row less when scrolled to bottom", () => {
+        const fix = TestBed.createComponent(VirtualComponent);
+        fix.detectChanges();
+        const container = fix.componentInstance.container;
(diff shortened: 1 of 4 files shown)


PavlovVasil:
The following issue is found: 
1. Select a cell from the CompanyName column
2. Press and hold the "down" arrow key on the keyboard in order to scroll to the last row
3. Press and hold the "up" arrow key to go back to the first grid row
4. The scrolling suddenly stops in the middle of the grid and the keyboard navigation is not working anymore.

bkulov:
@PavlovVasil The issue you described is a know issue and PwnJS are working on it.

PR #1084, opened 2018-04-20
Fix changing width of a column not being detected by the igxFor

### LINKED ISSUE #1056: Resizing a column and then using keyboard navigation right to scroll shifts the content.
### Description  
When making a column bigger using resizing and using selection with keyboard navigation to scroll to the right at some point shifts the whole content.

### Steps to reproduce  

1. Open the resizing sample
2. Make "Company Name" columns bigger.
3. Click on a cell
4. Press right arrow when so you reach the most right column

### Result  
At some point the whole content shifts and the row styling doesn't fill the whole row.

### Expected result  
The content shouldn't shift

### Attachments  
resizing

### LINKED ISSUE #1069: Resizing a column to the rightmost of the igxGrid hides all the columns that are on right of the resized column
### Description  
Resizing a column to the rightmost of the igxGrid hides all the columns that are on right of the resized column.

**NOTES:**
- Performing an operation like **pinning/unpinning** fixes the issue and the columns become visible again.
- The issue is **not** reproduced when resizing analogically to the left.

### Steps to reproduce  

1. Run the "Grid Column Resizing" demo app.
2. Scroll horizontally to the rightmost of the grid.
3. Resize the "Region" column to the rightmost of the grid.

### Result  
The columns that were next to the "Region" column on the right are now **not** visible.

### Expected result  
The columns that were next to the "Region" column on the right are still visible.

### Attachments  

columnresize

skrustev:
Resolves #1056, resolves #1069

### DIFF
--- a/src/directives/for-of/for_of.directive.ts
+++ b/src/directives/for-of/for_of.directive.ts
@@ -201,6 +201,11 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
 
     public ngDoCheck(): void {
         if (this._differ) {
+            if (this.igxForScrollOrientation === "horizontal") {
+                // Resize scrollbar and hCache in case width of a cell has changed or etc. This cannot be detected with differ.
+                this._recalcScrollBarSize();
+            }
+
             const changes = this._differ.diff(this.igxForOf);
             if (changes) {
                 this._applyChanges(changes);


mpavlinov:
The described issue is fixed, however there is another problem when shrinking columns
igniteui-angular-1069

mpavlinov:
@skrustev Looks good.

PR #1086, opened 2018-04-20
Control when the horizontal scrollbar is displayed

### LINKED ISSUE #925: The container for the horizontal scrollbar is always visible
### Description  
Describe the issue.
The horizontal scroll container is always visible, even if scroll bar is not necessary 

### Steps to reproduce  
Place a grid on the page

### Result
capture

Use the attached sample or any other to verify two scenarios:

- [x] Grid width is set
- [x] Grid width is not set

In both cases grid should display horizontal scrollbar if total columns width exceeds the grid width.

[igx-grid-app.zip](https://github.com/IgniteUI/igniteui-angular/files/1931811/igx-grid-app.zip)

hanastasov:
Closes #925


zdrawku:
@rkaraivanov, do you have any comments for this change?

hanastasov:
This may not be needed anymore now that width is set to 100% internally if no other value is set.

hanastasov:
This fix is not necessary anymore now that width is set to 100% internally if no other value is set.

PR #1094, opened 2018-04-20
Expose reflow() method so it can be called publicly.

### LINKED ISSUE #1023: Vertical scrollbar disappears when the grid has "width" set as a percentage
### Description  
In the Grid Performance demo the vertical scrollbar disappears if the "width" of the grid has been set as a percentage.

### Steps to reproduce  

1. Open branch 5.3.x.
2. In the Grid Performance demo set the grid width to"100%".
3. Open the Grid Performance demo.

### Result  
The grid vertical scrollbar disappears.

### Expected result  
The grid should have a vertical scrollbar.

### Attachments  
sketch

skrustev:
Closes #1023

### DIFF
--- a/src/grid/README.md
+++ b/src/grid/README.md
@@ -1,4 +1,4 @@
-# igx-grid
+﻿# igx-grid
 **igx-grid** component provides the capability to manipulate and represent tabular data.  
 A walkthrough of how to get started can be found [here](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid.html)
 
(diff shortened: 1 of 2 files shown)


src/grid/README.md
```diff
@@ -217,6 +217,7 @@ Below is the list of all inputs that the developers may set to configure the gri
 |nextPage()|Goes to the next page if paging is enabled and current page is not the last.|
 |paginate(page: number)|Goes to the specified page if paging is enabled. Page indices are 0 based.|
 |markForCheck()|Manually triggers a change detection cycle for the grid and its children.|
+|calculateGridSizes()|Recalculates grid width/height. Should be ran when changing DOM elements dimentions manually that affect the grid.|
```
ChronosSF: should be run

ChronosSF: the grid's size

src/grid/grid.component.ts
```diff
@@ -764,6 +764,18 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         return true;
     }
 
+    /**
+     * Recalculates grid width/height. Should be ran when changing DOM elements dimentions manually that affect the grid.
+     */
+    public calculateGridSizes() {
```
kdinev: The current name suggests that the grid size is calculated and returned by the method. Instead the method reflows the grid, so `reflow()` is a better name.

PR #1116, opened 2018-04-23
Adjust cell navigation on arrow up

### LINKED ISSUE #1100: Keyboard navigation is not working in the Column Pinning demo
### Description  
In the Grid Column Pinning demo when using keyboard navigation the currently selected cell looses focus and the scrolling stops. 

### Steps to reproduce  

1. Open branch 5.3.x.
2. Change the grid height to 600px.
3. Open the Grid Column Pinning demo.
4. Select the first cell of the CompanyName column.
5. Press and hold the "down" arrow key until you scroll to the last grid row.
6. Press and hold the "up" arrow key and scroll back to the top.

### Result  
The scrolling suddenly stops at the middle of the grid as the selected cell looses focus

### Expected result  
Holding the "up" arrow key should scroll to the top.

### Attachments  
sample

ViktorSlavov:
Closes #1100

Adjust how "arrow up" key navigation is handled to mirror "arrow down" key navigation.

### DIFF
--- a/src/grid/cell.component.ts
+++ b/src/grid/cell.component.ts
@@ -549,9 +549,8 @@ export class IgxGridCellComponent implements IGridBus, OnInit, OnDestroy {
                         cell.nativeElement.focus();
                     }
                 });
-            } else {
-                target.nativeElement.focus();
             }
+            target.nativeElement.focus();
         } else {
             this.row.grid.verticalScrollContainer.scrollPrev();
             this.row.grid.verticalScrollContainer.onChunkLoad.pipe(take(1)).subscribe({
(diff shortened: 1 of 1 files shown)


src/grid/cell.component.ts
```diff
@@ -535,20 +535,16 @@ export class IgxGridCellComponent implements IGridBus, OnInit, OnDestroy {
...
+                && verticalScroll.scrollTop // the scrollbar is not at the first item
+                && target.row.element.nativeElement.offsetTop < this.grid.rowHeight) { // the target is in the first row
+                verticalScroll.scrollTop -= this.grid.rowHeight;
                 this.row.grid.verticalScrollContainer.onChunkLoad.pipe(take(1)).subscribe({
                     next: (e: any) => {
-                        if (oldChunkIndex === e.startIndex) {
-                            target.nativeElement.focus();
-                        }
-                        this.row.cdr.detectChanges();
+                        const cell = this.gridAPI.get_cell_by_visible_index(this.gridID, rowIndex, this.visibleColumnIndex);
```
kdinev: Code repetition

Lipata: What @kdinev is requesting, is that we remove duplication of those chunkload subscribe functions. We need a centralized function for that. Any other logic, inside keydown/up handlers that is used on more than one place, should be refactored. Thanks!

PR #1163, opened 2018-04-24
Do not filter on not unary conditions if no value

### LINKED ISSUE #931: igx-grid - it's not possible to apply filtering with the same value
### Description  
When you filter a igx-grid column with a single character, then reset the filtering and then again filter with the same char, the filter button is still disabled and data is not filtered.

### Steps to reproduce  

1. Open the following igx-grid sample and turn off live data
2. Filter the column, with athlete names, with the letter "a".
3. Reset the filter.
4.Type again "a" in the filter dialog

### Result  
The filter button is disabled and the data is not filtered

### Expected result  
The gird data to be filtered and the buttons in the filtering dialog to be enabled

### LINKED ISSUE #1013: Choosing different than default filtering condition leads to filtered data
### Description  
When the user chooses another than default filtering condition, for example "Equals", the grid data is filtered and displays only the first record in the datasource

### Steps to reproduce  

1. Run the Grid Performance sample in demos
2. Open the filtering dialog 
3. Choose different than the default condition, for example "Equals"

### Result  
The data is filtered and the grid displays the first record from the datasource

### Expected result  
The grid is not filtered until a value is entered

### LINKED ISSUE #1118: Filter conditions apply filtering without value
### Description  
When filtering a numeric column, only changing the filtering condition applies filtering without a value, and respectively uses 0 by default. Related #1013.

### Steps to reproduce  

1. Open the selection sample in the samples.
2. Open the filter UI for the `Change` column.
3. Choose `Less than` for a filtering condition.
4. Repeat the same process for `Greater than`, `Less than, or equal`, `Greater than, or equal`, `Equal`, `Does not equal`. 

### Result  
The UI is filtered for anything less than 0. You also cannot clear the filter, which is related to #931.

### Expected result  
Filtering shouldn't be applied unless there is a value in the filtering input. The only conditions applying filtering without the input should be `Null`, `Not null`, `Empty`, `Not empty`. 

### Attachments  
image

hanastasov:
Closes #1118, Closes #1013, Closes #931 

[This PR](https://github.com/IgniteUI/igniteui-angular/pull/1163/commits/ee75b51f50804f9ff78b73df64f9178a31b05318) seperates the filtering tests into two files - one for the API tests and for the UI tests.


src/grid/grid-filtering.spec.ts
```diff
@@ -778,10 +778,10 @@ describe("IgxGrid - Filtering actions", () => {
         filterIcon.nativeElement.click();
```
kdinev: Is there a test that tests changing filtering condition without an input?

hanastasov: Nope, I am about to add tests.

hanastasov: tests are added: https://github.com/IgniteUI/igniteui-angular/pull/1163/commits/ee75b51f50804f9ff78b73df64f9178a31b05318

PR #1172, opened 2018-04-25
fix(igGrid): Columns are not sized correctly when some of them are left without width. #1143

### LINKED ISSUE #1143: Columns are not sized correctly when some of them are left without width
### Description  
When I have a Grid with 4 columns and only two of them are having defined width (in percent or pixels) the rest of the columns should fill the whole grid width.

This is when I have width of the Grid set to 100%.

### Steps to reproduce  

1. Open the sample
2. Click on the last cell, first row
3. There is unused space 

### Result  

### Attachments  
image

### Sample

https://stackblitz.com/edit/igx-grid-width-issue?file=app%2Fapp.component.html

### Additional note

When I use the docfx more complex sample, the issue is not present, but I cannot see what is the difference

https://stackblitz.com/angular/xedxrrmbxnx?file=app%2Fgrid%2Fgrid-paging-sample%2Fgrid-paging-sample.component.html

MayaKirova:
Also refactoring test to check the exact width of the columns based on the current grid width.

Closes #1143 .


tbelcheva:
The issue reproduces in Edge.
extraspace

MayaKirova:
@tbelcheva Cannot reproduce it outside of demos dev, where the navbar causes the grid to resize while it's being initialized.

PR #1174, opened 2018-04-25
fix(igxFor): ensuring correct state on size changes #1136

### LINKED ISSUE #1136: Exception is thrown when trying to pin column in certain scenarios
Cannot read property 'destroy' of undefined exception is thrown when trying to pin column by following the steps to reproduce.

### Steps to reproduce  

1.  Run the following sample:
https://staging.infragistics.local/products/ignite-ui-angular/angular/components/grid_column_pinning.html
2. Pin "Company Name" column.
3. Pin "Fax" column
4. Observe the console

### Result  
Cannot read property 'destroy' of undefined exception is thrown in the console. Additionally, there is a flickering because the lastly pinned column initially is not added to pinned columns collection. This happens when user horizontally scrolls unpinned columns area.

### Expected result  
No exception to be thrown. The "Fax" column, which is pinned lastly, to be instantly added to pinned are and no flickering to appears when trying to horizontally scroll the unpinned area.

ChronosSF:
Closes #1136


kacheshmarova:
The issue is still reproducible if you pin and unpin "Fax" column multiple times.

PR #1181, opened 2018-04-25
fix(igGrid): Misalignment when resize a column and then scroll horizontally. #1169

### LINKED ISSUE #1169: Misalignment when resize a column and then scroll horizontally
### Description  
When we have pinned columns and resize a column there is a misalignment when we scroll horizontally.

### Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular
2. Resize a column
3. Scroll horizontally

### Result  
Misalignment in the headers

### Expected result  
No misalignment

### Attachments  
grid

MayaKirova:
Closes #1169.  

Issue reproduces only if the column width is set as number (for example width="88"), in which case the existing HostBinding for flex-basis is not applied correctly. Fix adds binding for the scenarios when width is set as number.


MayaKirova:
@kacheshmarova Please re-test on Edge.

PR #1184, opened 2018-04-26
Updating Ignite UI Roadmap

SlavUI:
Updating Ignite UI Roadmap with current tasks until end of June and also marking all task that have been completed for the past 3 months.

t:


mpavlinov:
How about the custom Scroll? Is it on the queue?

SlavUI:
It will be after June release

mpavlinov:
@SlavUI: Ok. Can you add it to the docs in milestone 4.

PR #1189, opened 2018-04-26
Set type of buttons to be 'button'

### LINKED ISSUE #1107: When clicking buttons on igx-dialog, form is submitted unexpectedly when igx-dialog is in the form.
### Description  
When clicking buttons on igx-dialog, form is submitted unexpectedly when igx-dialog is in the form.

### Steps to reproduce  

1. Go to https://stackblitz.com/edit/igx-editors-issue
2. Click igx-time-picker to show dialog.
3. Click "CANCEL" / "OK" buttons.
4. Click "OPEN DIALOG" button.
5. Click "LEFT" / "RIGHT" buttons.

### Result  
3, 5. Form is submitted. 

### Expected result  
Form should not be submitted.

This root cause is because the default type of button is `submit`. Even though it is, this behavior  leads to unexpected result.

https://stackoverflow.com/questions/9824808/disable-form-auto-submit-on-button-click

Buttons that are on the igx-dialog should be explicitly set types to `button`.

sstoyanovIG:
Closes #1107  .  


src/dialog/dialog-content.component.html
```diff
@@ -8,11 +8,11 @@
         <ng-content *ngIf="!message"></ng-content>
 
         <div *ngIf="leftButtonLabel || rightButtonLabel" class="igx-dialog__window-actions">
-            <button *ngIf="leftButtonLabel" igxButton="{{ leftButtonType }}" igxButtonColor="{{ leftButtonColor }}" igxButtonBackground="{{ leftButtonBackgroundColor }}"
+            <button *ngIf="leftButtonLabel" type="button" igxButton="{{ leftButtonType }}" igxButtonColor="{{ leftButtonColor }}" igxButtonBackground="{{ leftButtonBackgroundColor }}"
```
kdinev: Shouldn't it be an `<input type="button" />`?

sstoyanovIG: The Ripple effect doesn't work on Input, but despite of this after I spoke with @simeonoff , we think it is good to expose the type as 'Input' of the dialog and set it to button by default.

PR #1198, opened 2018-04-27
fix(pinning): Fix and test for issue #1129

### LINKED ISSUE #1129: Too small unpinned area
### Description  
The unpinned area is too small when there are several pinned columns

### Steps to reproduce  

1. Run the demos and then start Grid Column Pinning sample
2. Initially ContactName and ContactTitle columns are pinned. Pin CompanyName, City and Region.
3. Unpin City. Pin Address

### Result  
The result is very small unpinned area which cannot be operated with.

### Expected result  
There should be some unpinned area width limitation I guess as for the igGrid - minimalVisibleAreaWidth. 
unpinnedareawidth

tbelcheva:
Closes #1129  .

### DIFF
--- a/src/grid/grid.pinning.spec.ts
+++ b/src/grid/grid.pinning.spec.ts
@@ -13,7 +13,7 @@ import { IgxGridHeaderComponent } from "./grid-header.component";
 import { IGridCellEventArgs, IgxGridComponent } from "./grid.component";
 import { IgxGridModule } from "./index";
 
-describe("IgxGrid - Column Pinning ", () => {
+fdescribe("IgxGrid - Column Pinning ", () => {
     const COLUMN_HEADER_CLASS = ".igx-grid__th";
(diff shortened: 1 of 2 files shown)


src/grid/grid.pinning.spec.ts
```diff
@@ -584,6 +584,24 @@ describe("IgxGrid - Column Pinning ", () => {
         expect(grid.unpinnedColumns.length).toEqual(unpinnedColumnsLength);
         expect(result).toBe(false);
     });
+
+    it("should minimal visible unpinned area width be 20% of the grid width", () => {
```
MayaKirova: Can you make the test name a coherent sentence?

Currently  the test name is:
IgxGrid - Column Pinning should minimal visible unpinned area width be 20% of the grid width.

A more coherent name would be something like:
igxGrid – Column Pinning should reject pinning a column if the unpinned area width becomes less than 20% of the grid width

PR #1231, opened 2018-05-02
Fix igxForOf startIndex is not being updated before calculating chunk size.

### LINKED ISSUE #1130: IgxFor start index is not updating appropriately with enabled summaries
### Description  
When scrolling to the end of the rows with enabled summaries and then disable the summaries
the start index of the virtualization is not updated leaving white space.

### Steps to reproduce  
1. Open the summaries demo. 
2. Enable summaries and scroll to the end of the rows.
3. Disable the summaries

### Result  
There is white space between the last row and the footer of the grid.

### Expected result  
There should not be white space between the last row and the footer of the grid.

skrustev:
Resolves #1130


kacheshmarova:
If you hide summaries from a column with a bigger summary (more rows) to a column with a smaller summary(with fewer rows) the issue is still reproducible. For example, if you start hiding summaries in the following order: ProductName ->Unit Price ->Discontinued -> Order Date the issue still persists.
If you try hiding summaries in the reversed order the issue is not reproducible.

Issue #1187, opened 2018-04-26
Filtering behavior and UI updates


sbozalieva:
### Description  
Enhance filtering behaviour and interactions.

### Steps to reproduce  

1. Replace the “more“ icon with the “filter” icon. 
2. Remove the "filter" icon in the dropdown's input.
3. Reset Button becomes active only when there is a value already in the input or some of the non value required options is chosen.
4. "Filter" is changed to "Close". "Close" button is always active and closes the dropdown.
5. When the filter has no records, display "No Records Found".
6. Add "Clear" icon to the input that appears on initial input and if clicked clears the value.

### Result  

1. Be explicit about the ONLY action provided by the dropdown.
2. We've already have the filter icon in the header. No need to duplicate it.
3. Reset button will be active only when there is filter to reset.
4. The filtering is set to filter the data immidiately, so there is no need to have a button. Having a "Close" provides UI to close it, aside from the expected behaviour to click outside the dropdown.
5. Provide a status of the operation instead of just leaving an empty grid container.

### Expected result  

<img width="729" alt="image" src="https://user-images.githubusercontent.com/23053821/39301694-b97591fe-4958-11e8-9b33-7f0d06374cbe.png">

<img width="614" alt="image" src="https://user-images.githubusercontent.com/23053821/39301535-07d4df2c-4958-11e8-828f-a3adf8bc867b.png">

Issue #984, opened 2018-04-13
igxDropDown feature 


bazal4o:
### Research 
- [x] [Research document](https://infragisticsinc297.sharepoint.com/:w:/g/Groups/ProdDev/EbluN08un85LiOgstnWnC0cBXV1cztCTpHgOXlrxVxQRsA?e=S1lmvh)
- [x] [Technical specification](https://github.com/IgniteUI/igniteui-angular/wiki/DropDown-Specification)
- [x] POC

- [x] Design
- [x] Implementation
- [x] Tests
- [x] Samples
- [x] Documentation

desig9stein:
Visual design:
<img width="760" alt="components-menus-specs1-desktop" src="https://user-images.githubusercontent.com/21260963/39303245-0d5f3dba-495e-11e8-9489-bfe3e690fd21.png">

PR #1234, opened 2018-05-03
fix(igxGrid): Add min-height to igx-grid__tbody #1014

### LINKED ISSUE #1014: Filtering dialog is cut off when there is only one filtering result
### Description  
When the grid has no width and height explicitly set and the user filters the grid, so that it returns only one filtering result, the filtering dialog is cut off

### Steps to reproduce  

1. Run Grid Performance sample and remove the width and height options (you may decrease the rows and columns as well)
2. Open the filtering dialog and filter by some value that would return a single record

### Result  
The grid height shrinks and the filtering dialog is cut off

### Expected result  
The filtering dialog is fully visible

tachojelev:
Resolves #1014

Adding min-height to igx-grid__tbody, so the filter dialog does not get cut off when the filtered IN data records are few or none at all.

### DIFF
--- a/src/grid/grid.component.html
+++ b/src/grid/grid.component.html
@@ -34,7 +34,7 @@
     </div>
 </div>
 
-<div class="igx-grid__tbody" role="rowgroup" [style.height.px]='calcHeight' [style.width.px]='calcWidth' #tbody>
+<div class="igx-grid__tbody" role="rowgroup" [style.min-height.px]='200' [style.height.px]='calcHeight' [style.width.px]='calcWidth' #tbody>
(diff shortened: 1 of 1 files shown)


hanastasov:
Maybe we should go for implementing an Empty rows template, which is also a feature requested in #1187. @SlavUI please take a decision on whether the change from this PR is going to be included, or an Empty Rows template should be implemented.

tachojelev:
A couple of _notes_:
After discussing this with @simeonoff , we believe that using min-height in combination with the current implementation for the grid and the filtering would probably cause some additional issues.
As suggested by @simeonoff , another approach would be to mark the current issue as _not to fix_ and wait for the [igxDropDown feature](https://github.com/IgniteUI/igniteui-angular/issues/984) , which might be a more applicable approach for the grid filter UI in this case.

PR #1243, opened 2018-05-03
Check if column is sortable/filterable when using API methods (5.3.x)

### LINKED ISSUE #1237: A column is sorted/filtered using the API methods even if sortable/filterable are set to false
### Description  
A column is sorted using the API methods even if sortable is set to false

### Steps to reproduce  

1. Run attached Sample
2. Click Toggle Sort button

### Result  
'ProductName' column is sorted

### Expected result  
'ProductName' column shuldn't be sorted

### Attachments  
[igxgridSample.zip](https://github.com/IgniteUI/igniteui-angular/files/1970178/igxgridSample.zip)

sstoyanovIG:
Closes #1237 .

### DIFF
--- a/src/grid/api.service.ts
+++ b/src/grid/api.service.ts
@@ -187,7 +187,8 @@ export class IgxGridAPIService {
 
     protected prepare_filtering_expression(id, state, fieldName, searchVal, condition, ignoreCase) {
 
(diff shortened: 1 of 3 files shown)


sstoyanovIG:
After I spoke with @rkaraivanov we agreed that this is not a bug.

PR #1266, opened 2018-05-08
fix(pinning): Adding unpinned area min width limitation - master. #1129

### LINKED ISSUE #1129: Too small unpinned area

tbelcheva:
Closes #1129

### DIFF
--- a/src/grid/grid.component.ts
+++ b/src/grid/grid.component.ts
@@ -510,6 +510,9 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         return (this.calcWidth * 80) / 100;
     }
 
+    get unpinnedAreaMinWidth(): number {
+        return (this.calcWidth * 20) / 100;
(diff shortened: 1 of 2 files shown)


ChronosSF:
@PavlovVasil , assigning you as you tested the 5.3.x PR

mpavlinov:
@tbelcheva : `getUnpinnedWidth` code is not the same as in branch 5.3.x. Did you do cherry pick?

PR #1268, opened 2018-05-09
chore(*): migrate to Angular 6

rkaraivanov:


src/snackbar/snackbar.component.ts
```diff
@@ -172,13 +171,13 @@ export class IgxSnackbarComponent {
         this.onAction.emit(this);
     }
 
-    public snackbarAnimationStarted(evt: AnimationTransitionEvent): void {
+    public snackbarAnimationStarted(evt: any): void {
```
kdinev: Since there's no type here anymore, we should create a type. The method expects specific key - `fromState`, which is a string enumeration in the `AnimationTransitionEvent`.

kdinev: @rkaraivanov The two method parameter types still need to be fixed,

PR #1270, opened 2018-05-09
Support keydown events for Internet Explorer and Edge

### LINKED ISSUE #1139: Carousel: Arrow keys do not work consistently for swiping left/right the images.
### Description  
Arrow keys do not work consistently for swiping left/right the images. 

### Steps to reproduce  

1. Open a caroucel sample like: https://staging.infragistics.local/products/ignite-ui-angular/angular/components/carousel.html
2. Try navigating via the keyboard arrows left and right after interacting with the caroucel's image/arrows/navigation (if enabled)

### Result  
**In Chrome:** After clicking on the image, only the first navigation by arrow key swipes the image.
**In Edge/IE:** The images do not swipe no matter where the last user interaction was (click on the image, click on an arrow or the bottom nav)

### Expected result  
Arrow keys navigation should work in all above cases.

hanastasov:
Closes #1139

### DIFF
--- a/src/carousel/carousel.component.spec.ts
+++ b/src/carousel/carousel.component.spec.ts
@@ -253,6 +253,16 @@ describe("Carousel", () => {
         carouselNative.nativeElement.dispatchEvent(new KeyboardEvent("keydown", {key: "ArrowRight"}));
         fixture.detectChanges();
         expect(carousel.next).toHaveBeenCalled();
+
+        spyOn(carousel, "prev");
+        carouselNative.nativeElement.dispatchEvent(new KeyboardEvent("keydown", {key: "Left"}));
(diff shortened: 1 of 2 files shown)


hanastasov:
Internet Explorer 9, 10, 11, Microsoft Edge, and Firefox 36 and earlier return "Left", "Right", "Up", and "Down" for the arrow keys, instead of "ArrowLeft", "ArrowRight", "ArrowUp", and "ArrowDown". - [reference](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#Navigation_keys)

zdrawku:
@hanastasov when you are ready with the fix, create PR to master as well.

PR #1274, opened 2018-05-10
Dropdown implementation

### LINKED ISSUE #984: igxDropDown feature

wnvko:
Closes #984


src/directives/toggle/toggle.directive.ts
```diff
@@ -65,21 +71,29 @@ export class IgxToggleDirective implements IToggleView, OnInit, OnDestroy {
         if (!this.collapsed) { return; }
 
         const player = this.animationActivation();
-        player.onStart(() => this.collapsed = !this.collapsed);
+        player.onStart(() => {
+            // this.collapsed = !this.collapsed;
```
bazal4o: If we'll leave it commented we need to remove onStart at all

src/drop-down/drop-down-item.component.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    @HostBinding("class.igx-drop-down__item")
+    get itemStyle(): boolean {
+        return !this.isHeader;
+    }
+
+    /**
+     * Gets if the given item is selected
+     */
+    get isSelected() {
```
bazal4o: That one is not clear at all. Which is the given item? we should either rename the getter, or update the description.

src/drop-down/drop-down-item.component.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    @HostListener("keydown.Space", ["$event"])
+    onSpaceKeyDown(event) {
+        this.dropDown.setSelectedItem(this.index);
+        this.dropDown.toggleDirective.close(true);
+    }
+
+    @HostListener("keydown.Enter", ["$event"])
+    onEnterKeyDown(event) {
+        this.dropDown.setSelectedItem(this.index);
```
bazal4o: Do we need to prevent default in case the the dropdown gets rendered in a form, to prevent form submitting, or we are safe?

src/drop-down/drop-down.component.ts
```diff
@@ -0,0 +1,375 @@
...
+export interface ISelectionEventArgs {
+    oldSelection: IgxDropDownItemComponent;
+    newSelection: IgxDropDownItemComponent;
+}
+
+/**
+ * **Ignite UI for Angular DropDown** -
+ * [Documentation](https://www.infragistics.com/products/ignite-ui-angular/angular/components/drop-down.html)
+ *
+ * TODO: ADD DESCRIPTION HERE, SAME AS IN THE DOCS
```
bazal4o: we have docs on PR, so we can update this section.

src/drop-down/drop-down.component.ts
```diff
@@ -0,0 +1,375 @@
...
+            this.close();
+        }
+    }
+
+    focusFirst() {
+        let focusedItemIndex = 0;
+        while (this.items[focusedItemIndex] && this.items[focusedItemIndex].isDisabled) {
+            focusedItemIndex++;
+        }
+        if (focusedItemIndex < this.items.length - 1) {
```
bazal4o: We have a little code repetition here and in focusLast. Please refactor

src/drop-down/drop-down.component.ts
```diff
@@ -0,0 +1,375 @@
...
+            this._focusedItem.isFocused = true;
+        }
+    }
+
+    focusNext() {
+        let focusedItemIndex = -1;
+        if (this._focusedItem) {
+            focusedItemIndex = this._focusedItem.index;
+        }
+        while (this.items[focusedItemIndex + 1] && this.items[focusedItemIndex + 1].isDisabled) {
```
bazal4o: Will this logic work if we have given selected item and two more disabled items which are the end of the list.

src/drop-down/drop-down.component.ts
```diff
@@ -0,0 +1,375 @@
...
+            focusedItemIndex++;
+        }
+        if (focusedItemIndex < this.items.length - 1) {
+            if (this._focusedItem) {
+                this._focusedItem.isFocused = false;
+            }
+            this._focusedItem = this.items[focusedItemIndex + 1];
+
+            const elementRect = this._focusedItem.element.nativeElement.getBoundingClientRect();
```
bazal4o: Is it possible to move this as separate function and use parameters for top/bottom ? In this way we'll be able to use it in focusPrev.

src/drop-down/drop-down.component.ts
```diff
@@ -0,0 +1,375 @@
...
+            this.focusFirst();
+        }
+        this.onOpened.emit();
+    }
+
+    onToggleClosed() {
+        if (this._focusedItem) {
+            this._focusedItem.isFocused = false;
+        }
+
```
bazal4o: Extra line :)

src/directives/toggle/toggle.directive.ts
wnvko: Done

src/drop-down/drop-down-item.component.ts
wnvko: Done

src/drop-down/drop-down-item.component.ts
wnvko: I put the drop down in a form and pressing Enter key, while drop down is dropped did not submit the form

src/drop-down/drop-down.component.ts
wnvko: Done

src/drop-down/drop-down.component.ts
wnvko: Done

src/drop-down/drop-down.component.ts
wnvko: yes, sure. We have manually tested this and we will add test for this.

src/drop-down/drop-down.component.ts
wnvko: Done

src/core/styles/components/drop-down/_drop-down-theme.scss
```diff
@@ -0,0 +1,260 @@
...
+///
+/// @param {Color} $selected-item-background [null] - The drop-down selected item background color.
+/// @param {Color} $selected-item-text-color [null] - The drop-down selected item text color.
+///
+/// @param {Color} $disabled-item-background [null] - The drop-down disabled item background color.
+/// @param {Color} $disabled-item-text-color [null] - The drop-down disabled item text color.
+
+/// @requires text-contrast
+@function igx-drop-down-theme(
+    $drop-down-background-color: null,
```
simeonoff: Can we shorten that to just `background-color` ?

src/core/styles/components/drop-down/_drop-down-theme.scss
```diff
@@ -0,0 +1,260 @@
...
+            color: map-get($this, 'focus-item-text-color');
+        }
+
+        &:hover {
+            background: map-get($this, 'hover-item-background');
+            color: map-get($this, 'hover-item-text-color');
+        }
+    }
+
+    [dir='rtl']{
```
simeonoff: This doesn't work if the direction is set on a parent item. RTL should work even if the direction is set on parent items.

src/drop-down/drop-down-item.component.ts
```diff
@@ -0,0 +1,167 @@
...
+     * Sets/gets if the given item is disabled
+     */
+    @Input()
+    @HostBinding("class.igx-drop-down__item--disabled")
+    public isDisabled = false;
+
+    @HostBinding("attr.tabindex")
+    get setTabIndex() {
+        const shouldSetTabIndex = this.dropDown.allowItemsFocus && !(this.isDisabled || this.isHeader);
+        console.log(shouldSetTabIndex);
```
simeonoff: Should we not include console.log calls in production code?

wnvko: We should not :)
Done.

src/directives/toggle/toggle.directive.ts
```diff
@@ -65,21 +71,26 @@ export class IgxToggleDirective implements IToggleView, OnInit, OnDestroy {
...
-        player.onStart(() => this.collapsed = !this.collapsed);
         player.onDone(() =>  {
             player.destroy();
             if (fireEvents) {
-                this.onOpen.emit();
+                this.onOpened.emit();
             }
         });
 
+        this.collapsed = !this.collapsed;
```
ChronosSF: I think collapsed should be changed when the animation completes, or at least _after_ onOpening as the -ing event is supposed to fire before any state is changed.

(this seems to be the case with `closing` and we should be consistent)

wnvko: Done.
I also changed a little how collapsed was set in open and close methods. In open I am setting it to false, not to !collapsed, and in close I am setting it to true. This allows me to have access to html elements inside the toggle box in onOpening event.

PR #1282, opened 2018-05-11
feat(themes): theming with css variables

simeonoff:
Add the ability to style any component instance using css variables. Additional Sass functions and mixins added to ease the process and allow conversion of Sass generated component themes to css variables.


simeonoff:
@npaunov the igx-theme mixin accepts a new parameter `$legacy-support`. By default it's set to `true`. Setting this parameter to `false` should enable theming through css variables. Theming with css variables is not expected to work with Internet Explorer 11.

PR #1296, opened 2018-05-14
refactor(grid): display empty grid message when no data

### LINKED ISSUE #1269: No records found message is not aligned and styled accordingly.
Filter a grid to result in 0 filtered records

### Result  
The "No records found." message is not centered.

### Expected result  
The "No records found." message is not centered horizontally and vertically according to the grid body
capture

simeonoff:
Closes #1269


simeonoff:
@hanastasov exactly why I assigned it to you.

PR #1335, opened 2018-05-16
fix(igxDatePicker): notify ngModel for updated value through API metho

### LINKED ISSUE #1334: NgModel bind value is not working as expected datepicker
### Description  
When you have bound ngModel value in datepicker and try to select some different date through selectDate API method, the property that is bound is not updated. 

### Steps to reproduce  

1. Create some sample with datepicker component
2. Bind some custom property with [(ngModel)] to datepicker component
3. Try to update the **date** through the API **selectDate** method of the datepicker component.

### Result  
The bound property is not updated.

### Expected result  
The property should contains the newly passed date.

Aleksandyr:
Closes #1334


ddincheva:
@Aleksandyr  please add unit test for this fix.

Issue #1305, opened 2018-05-15
igxGrid - extra white space when hiding summaries in certain situations


SAndreeva:
### Description  
White space when disabling summaries one (bigger) after another (smaller) after scroll to bottom and having horizontal virtualization.

### Steps to reproduce  

1. Run: https://stackblitz.com/edit/angular-wuoy9r
2. Scroll to bottom.
3. Disable 'UnitsInStock' summary.
4. Disable 'OrderDate' summary.

### Result  
There is some white space below the last visible row:

![sum](https://user-images.githubusercontent.com/7972965/40040681-de43c6fc-5823-11e8-8b3d-64a04f60db8f.png)

### Expected result  
There is no white space.
This is not reproducible if there is no horizontal scrollbar, for example if you make the grid wider in the sample.

zdrawku:
The issue is with status resolved

PR #1339, opened 2018-05-17
feat(grid): add display density option to the grid

### LINKED ISSUE #1320: igx-Grid visual density: Compact mode
### Description  
The igx-grid needs to have an easy way to be visually optimised for data-heavy enterprise (financial, medical, etc.) scenarios. 

The change should apply to all parts of the grid: 
 - Cells 
 - Header Cells
 - Summaries
 - Paging

### Suggested sizes/paddings at `compact` version
- left/right paddings for all cells: `12px`  *(down from 24px)*
- default row height: `32px`  *(down from 48px)*
- header height: `32px`
- summaries vertical padding: `0`  *(down from 5px)*

simeonoff:
Closes #1320


src/grid/grid.component.ts
```diff
@@ -482,6 +511,7 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         this.calcWidth = this._width && this._width.indexOf("%") === -1 ? parseInt(this._width, 10) : 0;
         this.calcHeight = 0;
         this.calcRowCheckboxWidth = 0;
+        this.rowHeight = this.rowHeight ? this.rowHeight : this.defaultRowHeight;
```
kdinev: In this case the `rowHeight` overrides the `displayDensity` option. Maybe we should actually make the `rowHeight` private and control the height only through the `displayDensity` internally? This way we would give less control, but be sure that the grid looks optimal.

simeonoff: I believe rowHeight should remain public, regardless of the displayDensity. When you set rowHeight explicitly as a user you change the height of the row and knowingly stray from the recommended default value we set. The displayDensity controls the cell padding as well.

ddincheva:
there is two failing test because of issue [1305](https://github.com/IgniteUI/igniteui-angular/issues/1305)

zdrawku:
we are currently working on the falling tests

PR #1359, opened 2018-05-18
fix(carousel): support for keyboard navigation #1139

### LINKED ISSUE #1139: Carousel: Arrow keys do not work consistently for swiping left/right the images.

hanastasov:
Closes #1139

### DIFF
--- a/src/carousel/carousel.component.ts
+++ b/src/carousel/carousel.component.ts
@@ -406,13 +406,13 @@ export class IgxCarouselComponent implements OnDestroy {
         }
     }
 
-    @HostListener("keydown.arrowright", ["$event"])
-    public onKeydownArrowRight(event) {
+    @HostListener("keydown.arrowright")
+    public onKeydownArrowRight() {
         this.next();
     }
 
-    @HostListener("keydown.arrowleft", ["$event"])
-    public onKeydownArrowLeft(event) {
+    @HostListener("keydown.arrowleft")
+    public onKeydownArrowLeft() {
         this.prev();
     }
 }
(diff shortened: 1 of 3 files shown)


src/carousel/carousel.component.ts
```diff
@@ -404,6 +405,16 @@ export class IgxCarouselComponent implements OnDestroy {
             }, this.interval);
         }
     }
+
+    @HostListener("keydown.arrowright", ["$event"])
+    public onKeydownArrowRight(event) {
```
Aleksandyr: I think that the event parameter is unnecessary here. Respectively the passage of the argument through the HostListener too.

src/carousel/carousel.component.ts
```diff
@@ -404,6 +405,16 @@ export class IgxCarouselComponent implements OnDestroy {
...
+        this.next();
+    }
+
+    @HostListener("keydown.arrowleft", ["$event"])
+    public onKeydownArrowLeft(event) {
```
Aleksandyr: I think that the event parameter is unnecessary here. Respectively the passage of the argument through the HostListener too.

PR #1379, opened 2018-05-22
feat(groupby): Paging integration with two pipes

ChronosSF:


ChronosSF:
I am PR-ing this instead of pushing directly so that @kdinev & @rkaraivanov can check how the data flows between the pipes when implementing the complex paging integration.

I can't see a clean way to achieve the functionality with just one grouping pipe but feel free to give it a try.

The only major drawback for this approach is having to pollute records with references to their parent groups. (the property name `__groupParent` used shouldn't be "magical" but probably a property of the sorting strategy so users may change it...)

Issue #1355, opened 2018-05-18
Update the UI Filtering per column


sbozalieva:
Add UI that supports more than one filtering conditions per column.

- [x] Implementation
- [x] Tests
- [x] Documentation
- [x] Samples

Designs including integration with the toolbar below:
![1](https://user-images.githubusercontent.com/23053821/40230145-2f61e4ce-5a9f-11e8-838d-455a735c2e60.png)
![2](https://user-images.githubusercontent.com/23053821/40230144-2f20f4fa-5a9f-11e8-9d2a-fa3fe684c7c7.png)
![3](https://user-images.githubusercontent.com/23053821/40231143-984a110c-5aa2-11e8-926a-bcfedc6c7d87.png)
![4](https://user-images.githubusercontent.com/23053821/40231148-9b72f63c-5aa2-11e8-9482-7e105df1210a.png)
![5](https://user-images.githubusercontent.com/23053821/40230139-2ea7465a-5a9f-11e8-85f3-cca8357a4092.png)
![6](https://user-images.githubusercontent.com/23053821/40230138-2e87616e-5a9f-11e8-827b-cb5d2bf97dc1.png)
![7](https://user-images.githubusercontent.com/23053821/40230137-2e68107a-5a9f-11e8-93fa-824761ef5229.png)

PR #1406, opened 2018-05-30
Implementing filtering operands

kdinev:
Related #1355


projects/igniteui-angular/src/lib/data-operations/data-util.ts
```diff
@@ -36,12 +40,18 @@ export class DataUtil {
             });
         return target;
     }
-    public static getFilteringConditionsForDataType(dataType: DataType):
-        {[name: string]: (value: any, searchVal?: any, ignoreCase?: boolean) => void} {
-        return FilteringCondition[dataType];
-    }
-    public static getListOfFilteringConditionsForDataType(dataType: DataType): string[] {
-        return Object.keys(DataUtil.getFilteringConditionsForDataType(dataType));
+    public static getFilteringConditionsForDataType(dataType: DataType): string[] {
```
ChronosSF: We should probably remove this from the API as it's not extendable anymore and not used anywhere either.

projects/igniteui-angular/src/lib/grid/column.component.ts
```diff
@@ -120,7 +121,7 @@ export class IgxColumnComponent implements AfterContentInit {
     public formatter: (value: any) => any;
 
     @Input()
-    public filteringCondition: (target: any, searchVal: any, ignoreCase?: boolean) => any;
+    public filteringCondition: IFilteringOperation;
```
mpavlinov: @kdinev This is breaking change. We should document it in the changelog.

projects/igniteui-angular/src/lib/data-operations/filtering-expression.interface.ts
```diff
@@ -9,7 +10,7 @@ export enum FilteringLogic {
  */
 export declare interface IFilteringExpression {
     fieldName: string;
-    condition: (value: any, searchVal?: any, ignoreCase?: boolean) => boolean;
+    condition: IFilteringOperation;
```
mpavlinov: @kdinev This is breaking change that we should document in changelog.

PR #1417, opened 2018-05-31
fix(igxGridSummary): column with summary outside of viewport not working

### LINKED ISSUE #1407: Summary not working when we have identical width and it is applied to column outside of the view port
### Description  
Summary not working correctly when we have column with identical widths and enabled summary per column which is not visible

### Steps to reproduce  

1. Open grid summary sample
2. Set hasSummary to true for OrderDate column
3. Try to add or delete row.

### Result  
There is no update into summary section.

### Expected result  
Summaries should be updated based on the added/deleted rows.

Aleksandyr:
when we have column with summary applied and it is outside of the
viewport and we have indentical width per column summary doesn't reflect
on any changes

Closes #1407

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grid/grid.component.ts
@@ -451,6 +451,10 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
     protected _filteringLogic = FilteringLogic.And;
     protected _filteringExpressions = [];
     protected _sortingExpressions = [];
+    protected _subscribeOnEditDone$;
+    protected _subscribeOnRowAdded$;
+    protected _subscribeOnRowDeleted$;
+    protected _subscribeOnFilteringDone$;
     private _filteredData = null;
     private resizeHandler;
     private columnListDiffer;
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -482,6 +486,10 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         this.calcWidth = this._width && this._width.indexOf('%') === -1 ? parseInt(this._width, 10) : 0;
         this.calcHeight = 0;
         this.calcRowCheckboxWidth = 0;
+        this._subscribeOnRowAdded$ = this.onRowAdded.subscribe(() => this.clearSummaryCache());
```
rkaraivanov: Use the `takeUntil(this.destroy$)` for unsubscribing when the grid is destroyed. Don't manage subscribtions 'manually'.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -539,6 +547,22 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         this.zone.runOutsideAngular(() => this.document.defaultView.removeEventListener('resize', this.resizeHandler));
         this.destroy$.next(true);
         this.destroy$.complete();
+
+        if (this._subscribeOnRowAdded$) {
```
rkaraivanov: The `unsubscribe` should be in the `onDestroy` hook. Use the approach in the previous comment and this whole code snippet is redundant.

PR #1454, opened 2018-06-04
Groupby and Sorting expressions sync 

ChronosSF:


projects/igniteui-angular/src/lib/grid/api.service.ts
```diff
@@ -278,21 +278,24 @@ export class IgxGridAPIService {
...
-            state.push({ fieldName, dir, ignoreCase });
-        } else {
-            Object.assign(expression, { fieldName, dir, ignoreCase });
-        }
+        states.forEach(state => {
+            const e = state.find((expr) => expr.fieldName === expression.fieldName);
+            if (!e) {
+                state.push(expression);
+            } else {
+                Object.assign(e, expression);
```
ChronosSF: This part of the code is the only one that annoys me. Because sorting and grouping exprs contain the same objects, if the expr is found in both collections the `assign` will be executed twice when it can be executed only once.

On the other hand if an expr is found in the sorting collection but not in the grouping one, the function must differentiate between the two to ensure it is `push`-ed to the grouping one. I couldn't find a way to have it work on a states collection (without differentiating between grouping and sorting ones), so that `forEach` can make the code pretty and also allow for grouping on previously sorted columns.

PR #1408, opened 2018-05-30
igxGrid GroupBy Feature

ChronosSF:
Closes #856 .

PR #1486, opened 2018-06-06
Removing indentation prop and moving it to the themes.

MayaKirova:
Related to #1408

@simeonoff or @desig9stein could you take a look at the sass files? I've added some very basic styles so that it does not look broken, but it can probably be styled better.


projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -693,6 +693,21 @@
...
+            min-width: #{$i*30}px;
+            flex-basis: #{$i*30}px;
+
+            .igx-button--icon {
+                min-width: 30px;
+                flex-basis: 30px;
+            }
+        }
+        %grid__group-indentation-padding-level-#{$i} {
+            padding-left: #{$i*30}px;
```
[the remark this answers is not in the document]
MayaKirova: This was originally  bound to the padding:
[style.padding-left]="padding"
So I just moved it as is to the theme.
If it makes more sense to be margin let me know and/or feel free to push the change directly to the branch.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -693,6 +693,21 @@
...
+    @for $i from 1 through 10 {
+        %grid__group-indentation-level-#{$i} {
+            min-width: #{$i*30}px;
+            flex-basis: #{$i*30}px;
+
+            .igx-button--icon {
+                min-width: 30px;
+                flex-basis: 30px;
+            }
+        }
```
desig9stein: The smallest grouping-indicator should be min-width:60px, flex-basis: 60px

desig9stein: Respectively the smallest group-indentation should be 60px

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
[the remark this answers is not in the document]
simeonoff: After discussing with @desig9stein we decided to keep it as padding as the background is part of the element.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
desig9stein: I will test and do that change

MayaKirova: Ok. Thank you  :+1:

PR #1523, opened 2018-06-12
Grid cell editingtemplates according column data type

### LINKED ISSUE #1071: The values of boolean type column cells can be modified to any custom value
### Description  
The values of boolean type column cells can be modified to any custom value.

**NOTES:**
A derivative issue from this one is that when the column is filtered by "true", the custom values are included in the filtered IN values. (The custom values should not be able to exist at all.)

### Steps to reproduce  

1. Download and run the attached sample application.
2. Try to change the value of any cell from the "Released"column to a custom string (for example "Hello").

### Result  
The change is successful.

### Expected result  
The change is **not** successful.

### Attachments  

[igxgridSample.zip](https://github.com/IgniteUI/igniteui-angular/files/1928020/igxgridSample.zip)

### LINKED ISSUE #1075: The values of number type column cells can be modified to any custom value
### Description  
The values of number type column cells can be modified to any custom value.

**NOTES:**
A derivative issue from this one is that when the value of a number type cell is set to a string and then we sort the respective number column - the records above and the records below the edited record are independently sorted.

### Steps to reproduce  

1. Run the "Grid" demo app.
2. Enable "Sorting" and "Editable" for the "ID" column.
3. Change the ID value of the third record from 3 to "TEST".

### Result  
The change is successful.

### Expected result  
The change is **not** successful.

### Attachments  

edit

### LINKED ISSUE #1083: Edit the cells of a date type column by using the igx-datePicker
### Description  
Edit the cells of a date type column by using the igx-datePicker as we currently can for the filter cell.

date

### LINKED ISSUE #1271: Implement different ng templates for the igxGrid cell component when it comes to inline editing
### Description  
Currently we have only string template with igxInput for iline editing:

Code snippet:
```
<ng-template #inlineEditor let-cell="cell">
    <igx-input-group>
        <input igxInput [value]="value" [igxFocus]="true" [igxTextSelection]="true" (keydown.enter)="cell.update($event.target.value)">
    </igx-input-group>
</ng-template>
```

We need to create additional templates for boolean, number and date inline editing:

- [x] Create boolean ng template
- [x] Create number ng template
- [x] Create date ng template
- [x] Make default template to be the string template
- [ ] Add styling for the new templates

Related to issues:
#1071 
#1075 
#1083

ddincheva:
Closes #1271 
Closes #1071
Closes #1075
Closes #1083


zdrawku:
@mpavlinov could you assist us with this PR, some style changes should be implemented in order to overcome the three issues below (you can use the dev demos. Grid cell editing sample):

- Pin some columns (with the Pin button) edit OrderDate cell. The edited cell borders will be visible on horizontal scrolling:
image

- Checkbox should be centered:
image

- Issue with the datepicker (this will be fixed once we are ready with the overlay)
image

desig9stein:
1. The one with the borders for editing cell is a bug that need's to be fixed
2. The one with the checkbox is normal behavior and I will center it only for the sample,
3. The one with the overlay will be fixed after the overlay component is ready

PR #1525, opened 2018-06-12
Adding constant chunk size for horizontal virtualization

mpavlinov:
Closes # .  

- [x] When a cell is in edit mode and you click another cell, the first one doesn't exit edit mode
- [x] Open the column resizing sample. Select a cell from the first column and use the right arrow key to scroll to the last column to the right. There is a blank space to the right, where the last "Fax" column should be.


MayaKirova:
@PavlovVasil please re-test the issue and also test GroupBy as it is now in master.

PR #1551, opened 2018-06-13
Fixed setting grid's width

### LINKED ISSUE #1430: Igx-grid does not reflect changes in its width
### Description  
Igx-grid does not reflect changes in its width/height.

### Steps to reproduce  

1. Go to https://stackblitz.com/edit/angular-qht2xz
2. Click button Set grid size in px
3. The grid has no scrollbar
4. Click button reflow to make the grid recalculite its sizes.
5. Click button Set grid size in %
6. IgxGrid does not look ok, its right end is cut off.

### Result  
step 3 and step 6

### Expected result  
Step 3 - scrollbars should be rendered in the grid
Step 6 - grid should be visible to the end

tiliev:
Closes #1430.  

- Reflowing when setting width.
- Refactored grid body's height recalculation because of a repetitive code.
- Added header's and footer's borders to the height calculations.


Aleksandyr:
@tiliev Currently this fix doesn't fix the issue I'm reproducing it again.

PR #1562, opened 2018-06-14
fix(igxfor): removing deprecated input #952

### LINKED ISSUE #952: Remove igxForRemote input from igxFor
The property is deprecated in 5.3.x and should be removed for 5.4.x

ChronosSF:
Closes #952  .  

Note sure if we need another entry in the update schematics (there is one for 6.0).


damyanpetev:
@ChronosSF There's a chance updates missed it for projects that already had 6.0.0 (migrations went out a bit late). I'm thinking it won't hurt if we re-run the same change just in case. Theoretically it should be possible to reuse the whole schematic with something like `context.addTask(new RunSchematicTask(path.join(__dirname, '../migration-collection.json'), 'migration-01', {}));`
but I haven't been able to make it work.

So we either chain the `Rule` or just add the `selectors.json` change to the 6.1 migration.

Issue #856, opened 2018-03-28
IgxGrid GroupBy Feature


ChronosSF:
### Description  
Backlog item with item tracking for igxGrid GroupBy feature implementation

### Research 
- [x] Research document
- [x] Technical specification
- [x] Design specification
- [x] POC

### Development
- [x] POC polish
- [x] Styling implementation
- [x] UI implementation
- [x] Expose grouping elements templates
- [ ] ~~GroupBy summaries~~ (available with templating)
- [x] Unit tests

### Documentation
- [x] ReadMe
- [x] DocFX
- [x] Sample creation
- [x] Changelog update

### Links
- development branch (https://github.com/IgniteUI/igniteui-angular/tree/igxgrid-groupby-dev)
- spec (https://github.com/IgniteUI/igniteui-angular/wiki/Group-By)

SDinev:
## Design screens ripe for comments: 

### Default view of the Group By Bar
![00-grouping_default](https://user-images.githubusercontent.com/13866657/40421472-bdcbca92-5e94-11e8-86ed-8d135d92ae41.png)

### Dragging a header changes the look of the group by bar
![01-grouping_draggingacolumn](https://user-images.githubusercontent.com/13866657/40421473-bdec3566-5e94-11e8-80c5-47e272078964.png)

### By default, all groups are minimized
![02-grouping_allgroupsminimized](https://user-images.githubusercontent.com/13866657/40421475-be293542-5e94-11e8-8e12-b882f4b47bb2.png)

### When the groups are expanded. 
Note there is an icon in the header that collapses/expands groups
![02-grouping_allgroupsexpanded](https://user-images.githubusercontent.com/13866657/40421474-be0b6792-5e94-11e8-8992-244d857f7a1c.png)

### While a second column is dragged
![03-grouping_dragasecondcolumn](https://user-images.githubusercontent.com/13866657/40421476-be471058-5e94-11e8-9c64-44d41cf80686.png)

### Two groups
![04-grouping_twogroups](https://user-images.githubusercontent.com/13866657/40421477-be69b644-5e94-11e8-858e-e63b49eb6a71.png)

### Two groups expanded
![04-grouping_twogroups_expanded](https://user-images.githubusercontent.com/13866657/40421795-c124cd5a-5e95-11e8-90f5-f850aadb7411.png)

### Pinning Included
[...]

PR #1564, opened 2018-06-14
feat(groupby): applying droppable on correct element, drop style #856

ChronosSF:
In dev


ChronosSF:
@kdinev, @rkaraivanov, @simeonoff  - adding you to reviewers so you can approve (or reject) how this looks with `IgxDensityEnabledComponent` that components that have density classes should extend.

I didn't do anything more complex as it looks quite ok to have an abstract `hostClassPrefix` property that TS requires you to implement in the deriving class. As long as components have only one prefix for density classes, this should allow us to provide density in an unified way without copy pasting.

ChronosSF:
@desig9stein - in `cosy` and `compact` modes there is a misalignment between body and header because the expand/collapse all cells are narrower in the body than their header cell.

PR #1567, opened 2018-06-15
Adding documentation to IgxMaskDirective

HristoP96:
Closes # .

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
@@ -60,7 +60,7 @@ export class IgxMaskDirective implements OnInit, ControlValueAccessor {
  * ```
  */
 @Output()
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -19,61 +19,132 @@ const noop = () => { };
...
-
+/**
+ * Emits an event each time the value changes.
+ * Provides `rawValue: string` and `formattedValue: string` as event arguments.
+ * ```html
+ * <input (onValueChange) = "onValueChange(rawValue: string, formattedValue: string)">
+ * ```
+ */
+@Output()
+    public onValueChange = new EventEmitter<{ rawValue: string, formattedValue: string }>();
```
kdinev: Shouldn't there be an interface for the event arguments on this `@Output()`?

PR #1572, opened 2018-06-15
fix(toolbar): showing toolbar only when there is content

### LINKED ISSUE #1571: igx-grid-toolbar is shown empty
### Description  
The IgxGrid's toolbar is shown as a small empty row on the top of the grid.

### Result  
The IgxGrid's toolbar is shown as a small empty row on the top of the grid when there are no elements enabled in it.

### Expected result  
The IgxGrid's toolbar should only be shown when there are visible UI controls in it.

rmkrmk:
Closes #1571  

My commit should fix the issue, however there is a minor side effect:

When you enable the toolbar and set a title only there toolbar has some height.
If you enable also some of the exporters the height become bigger because of some margins set on the buttons.
Additional fixes in the styles are required to fix this.


desig9stein:
@rmkrmk the toolbar should have, compact and cosy versions, which means that when the user is enabling the cosy ore compact version the class "igx-grid-toolbar" should become "igx-grid-toolbar--compact" or "igx-grid-toolbar--cosy" speak with @ChronosSF as he just did the same for another component.

PR #1584, opened 2018-06-18
Setting paging should update the grid

### LINKED ISSUE #1540: Enabling Paging and Summaries of the igxGrid through API does not work
### Description  
Enabling Paging and Summaries of the igxGrid does not work through API.

**NOTES:**
If the browser windows is resized, both the paging and the summary area get displayed.

### Steps to reproduce  

1. Download and run the attached sample application.
2. Click the button on the bottom.

### Result  
Paging and Summary area are **not** visible.

### Expected result  
Paging and Summary area are visible.

### Attachments  

refreshissue
[gridSample.zip](https://github.com/IgniteUI/igniteui-angular/files/2097002/gridSample.zip)

DDavidkov:
Closes  #1540.

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grid/grid.component.ts
@@ -193,7 +193,10 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
     set paging(value: boolean) {
         this._paging = value;
         this._pipeTrigger++;
-        this.cdr.markForCheck();
+
+        this.cdr.detectChanges();
+        this.calculateGridHeight();
+        this.cdr.detectChanges();
     }
 
     @Input()
--- a/projects/igniteui-angular/src/lib/grid/grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grid/grid.component.ts
@@ -201,9 +201,11 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         this._paging = value;
         this._pipeTrigger++;
 
-        this.cdr.detectChanges();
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/grid/grid.pagination.spec.ts
```diff
@@ -247,6 +247,23 @@ describe('IgxGrid - Grid Paging', () => {
         expect(grid.getCellByColumn(0, 'ID').value).toMatch('9');
         expect(gridElement.querySelector('.igx-paginator > span').textContent).toMatch('3 of 3');
     }));
+
+    fit('activate/deactivate paging', async(() => {
```
kdinev: There's a leftover `fit`

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -195,7 +195,10 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
     set paging(value: boolean) {
         this._paging = value;
         this._pipeTrigger++;
-        this.cdr.markForCheck();
+
+        this.cdr.detectChanges();
```
kdinev: Should `detectChanges()` really be called twice here?

DDavidkov: Yes. At least for the way the grid currently functions. The first detect changes updates the paginator field, which is a ViewChild of the grid (otherwise it will be null in the calculateGridHeight method) and the second one is to actually update the grid once we have changed its metrics due to the paging showing up or hiding. Similar pattern can be seen in the calculateGridSizes method.

PR #1588, opened 2018-06-18
Fix IE compatibility of IgxTextHighlight directive

### LINKED ISSUE #1531: In IE when grid has fixed width and columns widths and try to scroll an error is returned
### Description  
In IE when grid has fixed width and columns widths and try to scroll an error is returned

### Steps to reproduce  

1. Open in IE https://stackblitz.com/edit/angular-swnrj3?file=app%2Fgrid%2Fgrid-filtering-sample%2Fgrid-filtering-sample.component.html
2. Scroll horizontal

### Result  
An error Object doesn't support property or method 'forEach' is returned

### Expected result  
Grid should be scrolled without errors

DDavidkov:
Closes #1531.

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
@@ -81,14 +81,16 @@ export class IgxTextHighlightDirective implements AfterViewInit, OnDestroy, OnCh
 
         const callback = (mutationList) => {
             mutationList.forEach((mutation) => {
-                mutation.removedNodes.forEach((n) => {
+                const removedNodes = new Array(... mutation.removedNodes);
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
```diff
@@ -81,14 +81,16 @@ export class IgxTextHighlightDirective implements AfterViewInit, OnDestroy, OnCh
 
         const callback = (mutationList) => {
             mutationList.forEach((mutation) => {
-                mutation.removedNodes.forEach((n) => {
+                const removedNodes = new Array(... mutation.removedNodes);
```
bkulov: I would use Array.from here but I suppose this should work too.

PR #1597, opened 2018-06-19
Adding IForOfState to the public_api.ts

### LINKED ISSUE #1596: Files not exported in the public API
### Description  
Files not added to the public_api.ts:

- IForOfState

### Result  
These files cannot be directly referenced form igniteui-angular folder. However, they should be accessed from the folder where the file belongs to. For example:
import { IForOfState } from "igniteui-angular/esm5/lib/directives/for-of/IForOfState";

### Expected result  
To be able to reference these files directly from igniteui-angular folder. For example":
import {  IForOfState} from "igniteui-angular";

kacheshmarova:
Closes #1596

### DIFF
--- a/projects/igniteui-angular/src/public_api.ts
+++ b/projects/igniteui-angular/src/public_api.ts
@@ -16,7 +16,6 @@ export * from './lib/directives/button/button.directive';
 export * from './lib/directives/filter/filter.directive';
 export * from './lib/directives/focus/focus.directive';
 export * from './lib/directives/for-of/for_of.directive';
-export * from './lib/directives/for-of/IForOfState';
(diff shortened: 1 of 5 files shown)


projects/igniteui-angular/src/public_api.ts
```diff
@@ -16,6 +16,7 @@ export * from './lib/directives/button/button.directive';
 export * from './lib/directives/filter/filter.directive';
 export * from './lib/directives/focus/focus.directive';
 export * from './lib/directives/for-of/for_of.directive';
+export * from './lib/directives/for-of/IForOfState';
```
kdinev: Just move the `IForOfState` inside the `for_of.directive`

PR #1647, opened 2018-06-21
Ensure height on resize

### LINKED ISSUE #1641: Changing height makes items in igx-nav-drawer inaccessible
### Description  
When changing the height of your browser/screen the items in the igx-nav-drawer become inaccessible.

### Steps to reproduce  

1. Browse to [StackBlitz](https://stackblitz.com/angular/lvvklabgdvn)
2. Let everything load
3. Change the height of you browser/screen

### Result  
A scrollbar is shown so that the user can scroll trough all  items

### Expected result  
A wild vertical-scrollbar appears!

### Attachments  
image

sstoyanovIG:
Closes #1641 .


damyanpetev:
@sstoyanovIG  There's an else clause with unsubscribe call at https://github.com/IgniteUI/igniteui-angular/pull/1647/files#diff-38fd2d9d39a92519e220359884c09583R463 tied with pin only that you should probably remove, otherwise this looks fine :)

Issue #749, opened 2018-03-09
Add Toolbar to the igxGrid


sbozalieva:
- [ ] Column Pinning UI - #714

sbozalieva:
![base copy 7](https://user-images.githubusercontent.com/23053821/39819142-5c518b06-53ab-11e8-8326-bea5d88131db.png)
![base copy 8](https://user-images.githubusercontent.com/23053821/39819143-5c760a12-53ab-11e8-9a0d-6b1465ffaad9.png)
![base copy 9](https://user-images.githubusercontent.com/23053821/39819144-5c9750dc-53ab-11e8-8ce9-d8ed88a17760.png)
![base copy 10](https://user-images.githubusercontent.com/23053821/39819145-5cb5a6ae-53ab-11e8-8a7f-b87238287efe.png)

kdinev:
We should be very careful with the excel export dependency, and should make sure it and all of its dependencies are not loaded when the export button is not enabled in the toolbar.

PR #1653, opened 2018-06-21
feat(Toolbar): #749 Adding UI for ColumnPinning in the toolbar.

gedinakova:
Closes # 749.  

- Added UI for ColumnPinning to the toolbar.
- Extracted a ColumnChooserBase and ColumnChooserItemBase classes.


ChronosSF:
I guess this is still in development because of the overlay integration?

ChronosSF:
What do we need to do to have the overlay show correctly? Just enabling the toolbar in the pinning sample shows this:

pinning

gedinakova:
@ChronosSF I updated the branch to not load the overlay modal and position it properly. However, there's an issue with the z-index, so currently the dropdown is displayed below the grid. This is expected to be fixed by PR#1695. Finally, the actual pinning through the toolbar UI will start working once the MultiColumnHeaders PR is merged along with the changes to the pinning functionality.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -323,6 +323,9 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         this._columnHiding = value;
```
rmkrmk: Please add a check whether the new value is different than the current value and apply the new value only if there is need.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -399,6 +402,24 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
     @Input()
     public columnHidingTitle = '';
 
+    @Input()
+    get columnPinning() {
+        return this._columnPinning;
+    }
+
+    set columnPinning(value) {
+        this._columnPinning = value;
```
rmkrmk: Please add a check whether the new value is different than the current value and apply the new value only if there is need.

PR #1666, opened 2018-06-22
fix(igxGrid): Increase group area right and left padding #1665

### LINKED ISSUE #1665: When group area is filled, it is been added one area at the bottom of the grid
### Description  
When group area is filled(grouping by more columns), it is been added one area at the bottom of the grid.

### Steps to reproduce  

1. Group by some amount columns, depending on the grid configuration
See the gif for more info

### Result  
An area appears

### Expected result  
No area should appear

### Attachments  
area_at_the_bottom

dkamburov:
Closes #1665

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
@@ -369,9 +369,9 @@
     );
 
     $grouparea-padding: (
-        comfortable: rem(16px) rem(24px),
-        cosy: rem(8px) rem(16px),
-        compact: rem(4px) rem(12px)
+        comfortable: rem(16px) rem(40px),
+        cosy: rem(8px) rem(24px),
+        compact: rem(4px) rem(16px)
     );
 
     $grid-grouping-indicator-padding: (


dkamburov:
Increasing the padding seems to resolve the issue, let me know if you are fine with the fix.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -369,9 +369,9 @@
     );
 
     $grouparea-padding: (
-        comfortable: rem(16px) rem(24px),
-        cosy: rem(8px) rem(16px),
-        compact: rem(4px) rem(12px)
+        comfortable: rem(16px) rem(40px),
+        cosy: rem(8px) rem(24px),
+        compact: rem(4px) rem(16px)
     );
```
desig9stein: This breaks the compact and cosy inheritance between grid elements.

dkamburov: Ok, in that case, is it a better idea to close to pull request and assign the issue to you ?

desig9stein: I don't see how this is a styling issue. from what I see the gap under the grouped rows is caused by incorrect calc on grid body. Try to add one more chip and the problem disappear. Can you explain why do you think it's caused by CSS/SASS

desig9stein: Actually, this is caused because when you add the last chip there is an arrow indicator that points to the right and there is no space just for a millisecond. You can see a small jumping of that whole area caused by that arrow.

https://drive.google.com/file/d/1iudzzRe-VY9mzvzb8nGMdhIcm0BWBoxF/view

dkamburov: I'll investigate it further then, thought it's a css issue, because adding extra padding resolves it.

PR #1670, opened 2018-06-22
Fix IE grid body and headers misalignment

### LINKED ISSUE #1657: In IE column headers and content are misaligned.
### Description  
In IE column headers and content are misaligned.

### Steps to reproduce  

1. Build and start the dev demos. 
2. Open the Grid Column Pinning or Grid GroupBy sample on IE.
3. Notice that there is a slight misalignment between headers and content.

### Result  
There is a misalignment.

### Expected result  
There should be no misalignment.

### Attachments  
GroupBy:
image

Pinning:
image

desig9stein:
Closes #1657 


MayaKirova:
@desig9stein The misalignment in the pinned area is fixed, however there is still misalignment in the unpinned area between the header and content cells.
Inspect CompanyName and Address header and content cells and you'll notice they are misaligned.

image

This seems to happen in samples that don't have row selection enabled as well, for example the "Grid Column Resizing sample":

image

PR #1681, opened 2018-06-24
Show a message on loading

### LINKED ISSUE #1680: igx-list Show load message
### Description  
Show a (custom) message that the data is being retrieved by setting a `isLoading` to `true` (and setting `loadingListMessage`).

### Result  
A message is shown when the data is being retrieved.
It would be nicer to show the igx-circular-bar that would keep on running but I was not able to make that.

### Note
Important to differentiate that it's not the list that is loading but the data is being retrieved by the back-end.

Hypenate:
Closes #1680 .  

Show a message when [loading]="true"

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ ig start
 ## Building the Library
 ```
 // build the code
-ng build ignuteui-angular
+ng build igniteui-angular
 
 // build the css
 npm run build:style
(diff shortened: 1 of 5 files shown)


mpavlinov:
@Hypenate: Please, resolve the conflicts.

PR #1695, opened 2018-06-25
fix for issue #1690

### LINKED ISSUE #1690: Clicking second and each next time on filter button does not close previously opened filter dialogs in igxGrid
### Description  
Clicking second and each next time on filter button does not close previously opened filter dialogs in igxGrid.

### Steps to reproduce  

1. Run dev samples;
2. Open grid sample with filtering, e.g. Grid performance sample; 
3. Click on any of the filtering buttons;
4. Click on any other filtering button.

### Result  
First filter dialog does not close.

### Expected result  
First filtering dialog should close and only single filtering dialog should remain.

wnvko:
Closes #1690


projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -173,6 +162,27 @@ export class IgxOverlayService {
             overlay.hook.parentElement.removeChild(overlay.hook);
 
             const index = this._overlays.indexOf(overlay);
+
+            if (overlay.settings.closeOnOutsideClick) {
+                if (this._overlays.filter(x => x.settings.closeOnOutsideClick && !x.settings.modal).length <= 1) {
```
bazal4o: this should be equal to 1, not less then

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -87,32 +88,33 @@ export class IgxOverlayService {
...
 
         this._overlays.find(c => c.id === id).initialSize = size as DOMRect;
-        contentElement.style.width = size.width + 'px';
-        contentElement.style.height = size.height + 'px';
         overlaySettings.positionStrategy.position(contentElement, size, document, true);
+
+        if (overlaySettings.closeOnOutsideClick) {
+            if (overlaySettings.modal) {
+                fromEvent(wrapperElement, 'click').pipe(take(1)).subscribe(() => this.hide(id));
+            } else if (this._overlays.filter(x => x.settings.closeOnOutsideClick && !x.settings.modal).length <= 1) {
```
bazal4o: this should be equal to 1, not less then

PR #1750, opened 2018-06-27
#1740 Fixed the issues with importing enums and event args from 'igniteui-angular'.

### LINKED ISSUE #1740: Cannot import 'ColumnDisplayOrder' and 'IColumnVisibilityChangedEventArgs' from 'igniteui-angular'
### Description  
Cannot import 'ColumnDisplayOrder' and 'IColumnVisibilityChangedEventArgs' from 'igniteui-angular'.

### Steps to reproduce  

Try running the attached application.

### Result  
Running fails, because 
**import { IColumnVisibilityChangedEventArgs, ColumnDisplayOrder } from 'igniteui-angular';**
does not work.

### Expected result  
Running succeeds, because 
**import { IColumnVisibilityChangedEventArgs, ColumnDisplayOrder } from 'igniteui-angular';**
works.

### Attachments  
[gridsample.zip](https://github.com/IgniteUI/igniteui-angular/files/2140172/gridsample.zip)

gedinakova:
Closes #1740  .  

Fix to allow importing `ColumnDisplayOrder` enum and `IColumnVisibilityChangedEventArgs` from `igniteui-angular`.


projects/igniteui-angular/src/lib/grid/column-pinning.component.ts
```diff
@@ -41,16 +34,11 @@ export class IgxColumnPinningComponent extends ColumnChooserBase {
     constructor(public cdr: ChangeDetectorRef) {
         super(cdr);
     }
-    @Output()
-    public onColumnPinning = new EventEmitter<IPinColumnEventArgs>();
```
kdinev: Should this event emitter really go away?

gedinakova: It wasn't used at all because we already emit such an event when setting the pinned property.

Issue #1700, opened 2018-06-25
Add integration tests


IvayloG:
### Description  
Add integration testsfor igxOverlay

IvayloG:
Closed with "Overlay integration tests #1751"

PR #1751, opened 2018-06-27
Overlay integration tests

jackofdiamond5:
Related to #1700  .  

Adding integration tests.


projects/igniteui-angular/src/lib/services/overlay/overlay.spec.ts
```diff
@@ -1036,8 +1036,9 @@ describe('igxOverlay', () => {
             document.body.removeChild(dummy);
         }));
 
-        it('closingScrollStrategy: no scrolling possible. The component changes ' +
-            'state to closed when reaching the threshold (example: expanded DropDown collapses).', fakeAsync(() => {
+        // TODO: Should test with ConnectedPositioningStrategy
+        xit('closingScrollStrategy: The component changes state to closed when reaching the threshold' +
```
kdinev: Should these tests really be excluded?

PR #1777, opened 2018-06-28
fix #1759: Fix for scenarios where focus moves outside the visible cells.

### LINKED ISSUE #1759: Layout is broken when entering editMode on cell which is not fully visible
### Steps to reproduce  

1. Open cell editing sample
2. Click the button **Pin**
3. Scroll horizontally in order the column **ProductName** to be partially visible
4. Enter edit mode in one of the cells in column **ProductName**

### Result  
screenshot_16

### Expected result  
screenshot_17

MayaKirova:
Closes #1759 .

Other interactions like tabbing between content/header cells should also be fixed with this.


ChronosSF:
Similar issue happens vertically:
scroll

PR #1781, opened 2018-06-28
fix 1766: Autofocus the editing input only if cell still has focus.

### LINKED ISSUE #1766: Horizontal scroll is broken if columns are pinned while a cell is being edited for columns that are not in the immediate view
### Steps to reproduce  

1. Open the Column Moving dev sample
2. Enter edit mode for a cell in the last column
3. Pin all the columns through the buttons
4. Try to scroll back to the cell in edit mode

### Result  
Scrolling causes the pinned columns to offset

### Expected result  
Scrolling should not be broken

### Attachments  
updating-pinning

MayaKirova:
Closes #1766.

In scenarios where the cell has gone outside the view and no longer holds the focus there's no need to autofocus the input. Constantly changing the focus while scrolling would lead to flickering, which we should avoid.


ChronosSF:
@MayaKirova , seems like this PR breaks a few tests.

PR #1784, opened 2018-06-28
docs(IgxChip): add API docs for IgxChipComponent

alexalex972:


projects/igniteui-angular/src/lib/chips/chip.component.ts
```diff
@@ -156,10 +334,23 @@ export class IgxChipComponent implements AfterViewInit {
         }
     }
 
+    /**
+     * Sets the `IgxChipComponent` to be selected.
```
dkamburov: Should be: "Returns if the `IgxChipComponent` is the last one among its siblings."

PR #1786, opened 2018-06-28
Column moving - preserve selection with horizontal and vertical scroll

SAndreeva:
Moving a column should now preserve selection even when selected cells is out of the view port (horizontally or vertically).


projects/igniteui-angular/src/lib/grid/grid.common.ts
```diff
@@ -257,6 +260,15 @@ export class IgxColumnMovingDragDirective extends IgxDragDirective {
         this.column.grid.isColumnMoving = true;
         this.column.grid.cdr.detectChanges();
 
+        const currSelection = this.column.grid.selectionAPI.get_selection(this.column.gridID + '-cells');
+        if (currSelection && currSelection.length > 0) {
+            this.cms.selection = {
+                column: this.column.grid.columnList.toArray()[currSelection[0].columnID],
+                rowIndex: currSelection[0].rowIndex,
```
kdinev: You need only the `rowID`. The `rowIndex` is a thing of the past 👿

PR #1791, opened 2018-06-28
Fix drag and drop not working on Desktop Safari. 

### LINKED ISSUE #1712: Drag and drop is not working on Safari
### Description  
Drag and drop is not working on Safari. There is no PointerEvent defined in Safari.

### Steps to reproduce  

1. Open a grid sample with grouping enabled on Safari
2. Drag a column to group it

### Result  
Dragging is not working on Safari

### Expected result  
Dragging should be working on Safari

skrustev:
Resolves #1712


dkamburov:
When dragging chips in group area on Safari the text in cells gets selected

PR #1823, opened 2018-06-29
fix(igxfor): ensuring vertical chunksize is never NaN #1767

### LINKED ISSUE #1767: igx-grid - body becomes empty on igniteui-angular@6.1.0-beta.0
### Steps to reproduce  

1. Unzip the attached sample: [igx-grid-app.zip](https://github.com/IgniteUI/igniteui-angular/files/2141828/igx-grid-app.zip)
.2. npm install
3. Replace node_modules/igniteui-angular with [igniteui-angular.zip](https://github.com/IgniteUI/igniteui-angular/files/2141832/igniteui-angular.zip), which is a `igniteui-angular@6.1.0-beta.0` built code,
4. ng serve -o

### Result  
Grid width and height are 100% against its parent div(dimension is 800px x 500px). But grid body is empty.
image

### Expected result  
Grid body should not be empty.

ChronosSF:
Closes #1767 .


mpavlinov:
There is a problem with horizontal scrollbar not fully visible.

ChronosSF:
@desig9stein , please review the fix group-area min-height. We need that because height calculations may occur before the icon loads so they take a lower height of the group area than it becomes.

PR #1826, opened 2018-06-29
fix(grid): calculate positions considering current scroll #1778

### LINKED ISSUE #1778: Cannot move columns when the grid page is scrolled vertically
### Steps to reproduce  

1. Go to https://stackblitz.com/edit/angular-y51wxx-bnwdfc / https://stackblitz.com/angular/qmkbgexlmqy
2. Scroll the page a bit down, until the grid headers are at the top
3. Try to move some column - for example the "Contact Title", "Contact Name" or "CompanyName" column

### Result  
The element being dragged is not rendered at the place where dragging started. Also user is not able to drop it anywhere:
moving

hanastasov:
Closes #1778

### DIFF
--- a/projects/igniteui-angular/src/lib/chips/chip.component.html
+++ b/projects/igniteui-angular/src/lib/chips/chip.component.html
@@ -6,7 +6,8 @@
     <ng-content class="igx-chip__label"></ng-content>
 
     <igx-icon #removeBtn [tabindex]="removeBtnTabindex" #cancelContainer *ngIf="removable" igxButton="icon" igxRipple igxRippleCentered="true"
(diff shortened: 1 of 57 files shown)


skrustev:
Note: If PR #1827 is merged before this PR the changes in the `grid.common.ts` shouldn't be needed if I am correct since it will use pageX/pageY then.

hanastasov:
Whichever PR is merged first there will be a conflict so when resolving the conflict better  keep the #1827 version as you noticed.

SAndreeva:
I suggest merging this PR before #1827 and as mentioned above, only the changes in `dragdrop.directive.ts` will be needed.

PR #1857, opened 2018-07-02
fix(igxGrid): Force change detection on filter.

### LINKED ISSUE #1839: Filtering does not work after scrolling the grid
### Steps to reproduce  

1. Go to https://stackblitz.com/edit/angular-hhznuq-ts156r
2. Scroll the grid horizontally to the end
3. Scroll the grid back to first column
4. Filter the Reorder Level column

PS.: Removing the Supplier columns resolves the issue

### Result  
Filtering does not happen unless you close the filtering dialog

Seems related to #1770

MayaKirova:
Closes #1839.

Note: This happens if column in view changes from a column with filtering to column without filtering and vice versa. If all columns have filtering does not reproduce.

Same issue does not happen with other header operations (like sorting), which is rather strange.


ChronosSF:
@MayaKirova, while typing in the input works, clicking on the Reset button does not work instantly, nor do clicking Or / And - all these buttons require to fidget around so that change detection occurs and they become usable.

Issue #1846, opened 2018-07-02
Adding density to chip doesn't make the density style to apply when it is dragged


MartinKamenov:
### Description  
When compact or cosy density is added to the chip the drag ghost stays with the default(comfortable) density.

### Steps to reproduce  

1. Open chips sample.
2. Add [displayDensity]="'compact'" to igx-chip element in the template.
3. Drag one of the chips.

### Result  
The drag ghost doesn't have the compact density style when it is applied to the chip.

### Expected result  
The drag ghost should have the same density style as the chip.

### Attachments  
![issuechips11](https://user-images.githubusercontent.com/24592428/42162743-98f027f2-7e08-11e8-99dd-4c759b37c923.PNG)
![issuechips12](https://user-images.githubusercontent.com/24592428/42162747-9c0010c4-7e08-11e8-95d6-dcc79e6bd3d5.PNG)

MayaKirova:
@mpavlinov @ChronosSF In general the igxDrag directive copies the element on which it was instantiated (which makes sense), which in this case is the internal div. This would be easy to fix if angular allowed adding directives to host elements but it currently does not - https://github.com/angular/angular/issues/8785.
Due to this we have 2 possible solutions:
1) Change the chip's template so that it is wrapped in an additional div, which can take the related styles and hold the igxDrag directive.
2) Add an additional option to the igxDrag directive that would allow setting a different element to be used for the drag ghost.

Option 1) would be a somewhat breaking change and would require reworking the existing tests.
Option 2) would not be a breaking change since at this time those directives are used only internally.

Option 2 seems to have less impact so I'm more inclined to lean towards it as a solution.
Let me know what you think.

PR #1871, opened 2018-07-03
Fix cosy compact chip styling

desig9stein:
Related to #1846


kdinev:
@desig9stein Is this still in development?

desig9stein:
@kdinev  This fix is only for the styling part of the issue. There is other work that needs to be done in order #1846 to be closed.

mpavlinov:
@desig9stein We cannot verify this PR without the other changes we need to do, that's why I put "in-development" status. We will add commits to this PR in order to close the #1846 .

kdinev:
@desig9stein @MartinKamenov Can you guys move this to a branch in upstream, so you can both make the changes without switching remotes?

mpavlinov:
@skrustev I guess this one is good for testing. Let me know if it's not.

PR #1896, opened 2018-07-03
#1798 Should not persist combo scroll position

### LINKED ISSUE #1798: Combo sample, searching triggers poor UX functionality.
### Description  
Describe the issue.

### Steps to reproduce  
Run Sample:
https://github.com/IgniteUI/igniteui-angular/tree/master/src/app/combo
Drop down
Scroll down
Then type Con, to find Connecticut

### Result  
The scroll position stays at it's lower point, but the scrollbar disappears and Connecticut is hidden outside the top bounds.
Additionally you cannot recover from the functionality without reloading the page.
  Deleting the value
   or typing something else, just wipes the search entirely
Also the close drop down stops working at that point. It closes and pop's back open with the same problem.
### Expected result  
Filtering should probably reset the scroll position to the top as you change the filter conditions.
Also none of the other bad stuff... :P

### Attachments  
image

Lipata:
Closes #1798 

When you scroll combo drop down, and then close it, then the scroll position will not be preserved.

### DIFF
--- a/projects/igniteui-angular/src/lib/combo/combo-dropdown.component.ts
+++ b/projects/igniteui-angular/src/lib/combo/combo-dropdown.component.ts
@@ -16,7 +16,6 @@ import { IgxForOfDirective } from '../directives/for-of/for_of.directive';
 })
 export class IgxComboDropDownComponent extends IgxDropDownBase {
     private _children: QueryList<IgxDropDownItemBase>;
-    private _scrollPosition = 0;
     constructor(
         protected elementRef: ElementRef,
         protected cdr: ChangeDetectorRef,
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/combo/combo-dropdown.component.ts
```diff
@@ -322,7 +320,7 @@ export class IgxComboDropDownComponent extends IgxDropDownBase {
     }
 
     onToggleClosing() {
+        this.verticalScrollContainer.scrollTo(0);
```
kdinev: I thought we said that the scroll position will be reset on searching, instead of always?

Lipata: @kdinev read my last comment in the issue and please share your thoughts. Thanks!

kdinev: I thought what we said yesterday is that the scrolling is preserved to the last scroll position, unless searching is initiated?

PR #1902, opened 2018-07-03
Migration redactor and remove beta from 6.1.0

damyanpetev:


projects/igniteui-angular/migrations/update-6_1/index.ts
```diff
@@ -15,7 +15,7 @@ import { UpdateChanges } from '../common/UpdateChanges';
 
 export default function(): Rule {
     return (host: Tree, context: SchematicContext) => {
-        context.logger.info('Applying migration for Ignite UI for Angular to version 6.0.1');
+        context.logger.info('Applying migration for Ignite UI for Angular to version 6.1.0');
```
kdinev: Put the version in a variable and template it.

PR #1903, opened 2018-07-03
Add check for focusedItem in drop-down selectItem

### LINKED ISSUE #1895: Pressing Enter throws an error when no item in the dropdown is selected.
### Description  
Using the navigation directive with an igx-input-group throws an error when Enter is pressed and no item in the dropdown is selected.

### Steps to reproduce  

1. Open https://stackblitz.com/angular/paepernyqry
2. Select the input field
3. Press Enter
4. Observe the error.

### Result  
An error is thrown.

### Expected result  
No error should be thrown.

### Attachments  
Related to:
https://github.com/IgniteUI/igniteui-angular-samples/issues/478

ViktorSlavov:
Closes #1895  .  

SelectItem was expecting that the drop-down would always be visible (and therefore it would have a focusedItem).
Add check to selectItem function in IgxDropDownComponent

### DIFF
--- a/projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
+++ b/projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
@@ -450,14 +450,12 @@ export class IgxDropDownBase implements OnInit, IToggleView {
         }
     }
 
-    public selectItem(item?) {
-        if (!item) {
-            item = this._focusedItem ? this._focusedItem : null;
+    public selectItem(item: IgxDropDownItemBase) {
+        if (item === null) {
+            return;
         }
-        if (item) {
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -452,10 +452,12 @@ export class IgxDropDownBase implements OnInit, IToggleView {
 
     public selectItem(item?) {
         if (!item) {
```
kdinev: `item` should not be an optional parameter on an public API method. Make it mandatory, remove the checks inside, and do the necessary changes where the method is called.

PR #1915, opened 2018-07-04
Combo handles AddItem visibility on input

### LINKED ISSUE #1899: An existing item can be added as "custom value" when Filtering is disabled
### Description  
When you have Filtering disabled and type an existing value inside the input, the "ADD ITEM" button appears and you are allowed to add the existing value as a custom one. This is reproduced only the first time an existing item is entered. If you add this existing item as a custom one and try to do the same for any other of the existing items - the "ADD ITEM" button will not show.

**Note**: In addition, actual custom values cannot be submitted after the first time, which looks like the very same issue

### Steps to reproduce  

1. Navigate to https://staging.infragistics.local/products/ignite-ui-angular/angular/components/combo.html#usage
2. Disable the Filtering for the combos in the sample that is inside this section of the topic
3. Open the "Toggle Features" combo
4. Type "Maine"
5. Press "ADD ITEM"

### Result  
Maine is added at the end of the list

### Expected result  
"ADD ITEM" button should not be shown

### Attachments  
screenshot

ViktorSlavov:
Closes #1899.

checkMatch() is now called on input instead of on set filteredData (which will never be called if filterable = false)
If filtering is off, combo will use data instead of filtered data where needed.

### DIFF
--- a/projects/igniteui-angular/src/lib/combo/combo.component.spec.ts
+++ b/projects/igniteui-angular/src/lib/combo/combo.component.spec.ts
@@ -2857,6 +2857,31 @@ describe('igxCombo', () => {
         });
     }));
 
+    it(`Should properly display "Add Item" button when filtering is off`, fakeAsync(() => {
+        const fixture = TestBed.createComponent(IgxComboInContainerTestComponent);
+        fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


dafo:
When you add a new custom item, it is not actually added to the list of items.

PR #1918, opened 2018-07-04
Fix column moving interaction with the search

### LINKED ISSUE #1914: Column moving breaks the active highlight
### Description  
Column moving breaks the active highlight

### Steps to reproduce  

1. Run the attached sample
2. Type "1" in the search box
3. Drag move the ID column to another position

### Result  
The "1" in the ID column is no longer the active highlight

### Expected result  
The "1" in the ID column should remain the active highlight

### Attachments  
[gridsample.zip](https://github.com/IgniteUI/igniteui-angular/files/2163008/gridsample.zip)

DDavidkov:
Closes #1914.

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/grid.search.spec.ts
+++ b/projects/igniteui-angular/src/lib/grid/grid.search.spec.ts
@@ -1075,6 +1075,30 @@ describe('IgxGrid - search API', () => {
         });
     }));
 
+    it('Active highlight should be preserved when all rows are filted out', async(() => {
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/grid/grid.search.spec.ts
```diff
@@ -1075,6 +1075,30 @@ describe('IgxGrid - search API', () => {
         });
     }));
 
+    it('Active highlight should be preserved when all rows are filted out', async(() => {
```
gedinakova: The test name seem to be copied from a previous one :)

PR #1943, opened 2018-07-05
fix(igx-list): Creating an index export file for the list #1939

### LINKED ISSUE #1939: IgxListItemComponent and the two template directives are missing from public_api.ts.
### Description  
IgxListItemComponent, IgxEmptyListTemplateDirective and IgxDataLoadingTemplateDirective are not listed in public_api.ts.

 * browser: N/A

### Steps to reproduce  

1. Try to import IgxListItemComponent, IgxEmptyListTemplateDirective and/or IgxDataLoadingTemplateDirective from public_api.

### Result  
Those items are missing.

### Expected result  
They should be listed in public_api.

kdinev:
Closes #1939

### DIFF
--- a/src/app/drop-down/drop-down.sample.ts
+++ b/src/app/drop-down/drop-down.sample.ts
@@ -123,9 +123,9 @@ export class DropDownSampleComponent implements OnInit {
         this.igxDropDown.toggle(overlaySettings);
     }
 
-    onSelection() {
+    onSelection(event) {
     }
 
-    onOpening() {
+    onOpening(event) {
     }
 }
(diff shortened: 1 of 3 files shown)


src/app/drop-down/drop-down.sample.ts
```diff
@@ -123,9 +123,9 @@ export class DropDownSampleComponent implements OnInit {
         this.igxDropDown.toggle(overlaySettings);
     }
 
-    onSelection() {
+    onSelection(event) {
```
gedinakova: Is this change intentional?

kdinev: Yes, the demos weren't building with --prod --aot=true. This is just in the demos.

PR #1950, opened 2018-07-05
Combo remote selection

### LINKED ISSUE #1944: Remote binding combo doesn't store the selected fields when scrolled or collapsed
### Description  
When field from the combo remote sample is selected and then the field comes out of sight by scrolling or collapsing the selection state is lost.

 * browser: all

### Steps to reproduce  

1. Open staging ig combo sample.
2. Select the first option from the remote binding sample.
3. Scroll down until the first option is out of sight and then scroll up.

### Result  
The field doesn't stay selected.

### Expected result  
The field should stay selected after scrolling or collapsing.

### Attachments  
combo12
combo14
combo1 6

Lipata:
Closes #1944


dafo:
The selection is preserved when the dropdown list is closed/opened but is lost when scrolling is executed.

dafo:
An impact on the performance can be observed, but the functionality works as expected.

projects/igniteui-angular/src/lib/combo/combo-item.component.ts
```diff
@@ -41,7 +41,7 @@ export class IgxComboItemComponent extends IgxDropDownItemBase {
     }
 
     get isSelected() {
-        return this.parentElement.selectedItem.indexOf(this.itemID) > -1;
+        return this.parentElement.parentElement.isItemSelected(this.itemID);
```
kdinev: What is `this.parentElement.parentElement`?

kdinev: I believe the combo should be injected in the item and not retrieved this way.

Lipata: It is injected as 'parentElement', which is the combo drop down. In the same manner the combo is injected as a 'parentElement' in the combo drop down. Maybe we should rename them. Thanks!

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -665,6 +665,21 @@ export class IgxComboComponent implements AfterViewInit, ControlValueAccessor, O
...
 
+    private _compareFunc(itemID, equalCheck?: boolean) {
+        let compareFunc;
+        const key = this.valueKey;
+
+        // When there is remote data we need to compare valueKey values,
+        // instead of comparing entire itemID objects, because in that case they are not equal by reference.
+        if (this.totalItemCount > 0 && this.valueKey && this.dataType === DataTypes.COMPLEX) {
+            compareFunc = function(selectedItemID) {
+                return equalCheck ? selectedItemID[key] === itemID[key] :
```
bazal4o: Is it better to rename itemID to item and selectedItemID to selectedItem.

projects/igniteui-angular/src/lib/combo/combo-item.component.ts
Lipata: @kdinev I will refactor names in different pull request, not to mix it with this one?

Lipata: Here I made the changes: https://github.com/IgniteUI/igniteui-angular/pull/2028

projects/igniteui-angular/src/lib/combo/combo-item.component.ts
```diff
@@ -50,7 +50,7 @@ export class IgxComboItemComponent extends IgxDropDownItemBase {
      * @hidden
      */
     get isSelected() {
-        return this.parentElement.selectedItem.indexOf(this.itemID) > -1;
+        return this.parentElement.parentElement.isItemSelected(this.itemID);
```
kdinev: My comment is still not addressed.

bazal4o: Isn't it addressed into https://github.com/IgniteUI/igniteui-angular/pull/2028/ ?
I think the two comments came one after another and there was misunderstanding

kdinev: @bazal4o Why would a PR review be addressed in a different PR?

Lipata: Along with the combo item I also refactored the combo drop down, so I've decided to create different branch, because I wanted to separate the the logic for selection and the renaming of the injectable components. I can now merge both here, if you want.

kdinev: @Lipata Please do.

PR #1962, opened 2018-07-06
Fix horizontal scrolling of the scrollTo method

### LINKED ISSUE #1886: Search match highlight not always scrolled into view
### Description  
The search match highlight is not always scrolled into view when a horizontal scrolling is required.

### Steps to reproduce  

1. Run the attached sample app
2. Type "ne" in the "Search API" box on the right
3. Click the "Find next" button until the ORANGE highlight is on the "Wayne" word
4. Click one more time the "Find next" button

### Result  
The grid tries to scroll right to show the next match but it is not scrolling enough.

### Expected result  
The grid should scroll enough horizontally so that the "Company" column is visible as well as the orange highlight in it.

### Attachments  

[gridsample2.zip](https://github.com/IgniteUI/igniteui-angular/files/2159386/gridsample2.zip)

DDavidkov:
Closes #1886.


projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -2166,7 +2166,15 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
 
         const state = directive.state;
         const start = state.startIndex;
-        const size = state.chunkSize - 1;
+        const isColumn = directive.igxForScrollOrientation === 'horizontal';
+
+        let size;
+        if (isColumn) {
+            const firstDataRow = this.rowList.find(r => r.cells);
+            size = firstDataRow.cells.length - 2;
```
kdinev: Why `- 2`?

kdinev: @ChronosSF Please look into providing an API method that returns how many cells are fully in view, or returns a floating point for the cells fully in view - like `5.6` when 60% of the last cell is visible.

kdinev:
@ChronosSF Please look into adding the API method requested.

projects/igniteui-angular/src/lib/grid/grid.component.ts
MayaKirova: @DDavidkov Instead of branching the logic couldn't we just always scroll to the matching item index? 
For example: 
    private scrollDirective(directive: IgxForOfDirective<any>, goal: number): void {
        if (!directive) {
            return;
        }
        directive.scrollTo(goal);
    }

DDavidkov: @MayaKirova Not really. Technically, we can, but if we did, and you need to scroll just one row/column and we just call directive.scrollTo(goal) it would scroll the whole grid, so that the goal is the first visible row/column. We should try to scroll as little as possible, which means that if we need to scroll downwards or to the right, the cell with the active highlight should end up on the last visible row/column.

MayaKirova: @DDavidkov Thanks for the clarification. I've done some changes that should achieve the behavior you have described. Please review them.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -513,14 +513,39 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         return null;
     }
 
+    public getItemCountInView() {
```
ChronosSF: As a public method it should contain description and code snippet.

ChronosSF:
Closes #1977

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
ChronosSF: @DDavidkov , this should be pulled to master as well

Issue #1369, opened 2018-05-21
[Testing] Organize grid components and test data.


gedinakova:
Organize grid components and test data so they can be easily reused for testing.

PR #1968, opened 2018-07-06
#1369 Adding test data classes and sample grid components.

gedinakova:
#1369 .  
Added:
- SampleTestData class, containing different data collections.
- template-strings.spec.ts, containing different classes defining grids, columns and events templates.
- grid-interfaces.spec.ts which defines interfaces for grid events.


projects/igniteui-angular/src/lib/test-utils/grid-base-components.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    public rowsCount: number;
+    public columnsType = 'string';
+    public hasEditableColumns = false;
+    public startFromOne = false;
+    public columnNamePrefix = 'col';
+    public columns = [];
+
+    ngOnInit() {
+        this.columns = (this.hasEditableColumns) ? SampleTestData.generateEditableColumns(this.colsCount, this.columnNamePrefix)
+                                : SampleTestData.generateColumnsByType(this.colsCount, this.columnsType, this.columnNamePrefix);
```
Aleksandyr: The hasEditableColumns property has not reflect on the generated columns, because the **editable** property which is part of the generated column object never reaches the defined grid template.

projects/igniteui-angular/src/lib/test-utils/grid-base-components.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    data = SampleTestData.personJobData;
+}
+
+@Component({
+    template: GridTemplateStrings.declareGrid(` [rowSelectable]="rowSelectable"`,
+        '', ColumnDefinitions.productBasic)
+})
+export class SelectionComponent extends BasicGridComponent {
+    public rowSelectable = true;
```
Aleksandyr: Due to consistency I think here we should have default passed data.

Aleksandyr:
@gedinakova Also I think that some documentation of how to use all these test enhancements would be really helpful for others and building tests would be really faster instead of digging into the code and searching some templates, column definitions, predefined data and so on.

gedinakova:
@Aleksandyr I will add a readme once we finalize the approach and implementation.

projects/igniteui-angular/src/lib/test-utils/grid-base-components.spec.ts
gedinakova: Done!

Aleksandyr:
As we discussed with @gedinakova and @rkaraivanov both RemoteServiceBase and LocalServiceBase classes should be removed because the tests should not make/create any requests but instead Mocking the remote data, so we have to replace those classes with another one which would make those Mocks for us.

PR #1983, opened 2018-07-11
fix(tabs): Fixed selection when tabs collection is reset #1972

### LINKED ISSUE #1972: Remove tabs from collection -> TabCollectionChange Output
### Description  
I have an igx-grid, when I select a row I get the data and build my tabs.
When a row states 'I need 5 tabs' I can *ngFor it in HTML (see #1538).
When I click the next row and it only needs 2 tabs the fifth tab remains selected.

 * browser: Chrome

### Steps to reproduce  

1. Create an igx-grid that holds numeric data (1 - 5)
2. Based on the number generate tabs with *ngFor in the HTML 

### Result  
The 'highest number' tab stays selected

### Expected result  
The last tab should be selected and/or provide an output "TabCollectionChange"

### Attachments  
Note the blue line
image

sboykova:
Closes #1972. 

[Test](https://angular-xveqns.stackblitz.io)
1. Select Tab 5
2. Click Change collection button -> selection indicator is not under Tab 2

[Material Tabs Example](https://angular-5perhd.stackblitz.io)

Fixed behavior on tabs groups collection changed:

- On collection changed - selected index is persisted and applied to the new collection.
- In case the selected index is not valid for the new collection - the last tabs group is selected.
- In case the new collection is empty - no selection.


SAndreeva:
The following error is thrown if tabs collection is reset in run time:
blabla

sboykova:
New issue logged based on review #2011

kdinev:
@SAndreeva Moving back to `status: awaiting-test` since the testing has not been performed with the latest code in `6.1.x`.

PR #1989, opened 2018-07-12
Added documentation for igxDropdown

jackofdiamond5:


projects/igniteui-angular/src/lib/drop-down/drop-down-item.component.ts
```diff
@@ -61,18 +97,49 @@ export class IgxDropDownItemBase {
 
     /**
      * Sets/gets if the given item is header
+     * ```typescript
+     *  // get
+     *  let mySelectedItem = this.dropdown.selectedItem();
```
ViktorSlavov: There's a slight typo here - IgxDropDownBase.selectedItem is a getter, no need for ();

PR #2035, opened 2018-07-18
Add empty grid template

### LINKED ISSUE #2001: No message is displayed when there is empty grid data without filtering enabled.
### Description  
No message is displayed when there is empty grid data without filtering enabled.

### Steps to reproduce  

1. Open grid performance demo
2. Change the data to be an empty array
3. Observe the rendered grid

### Result  
There is no empty message displayed.

### Expected result  
There should be empty message displayed.

### Attachments
How it looks:
howitlooks

How it should look:
howitshouldlook

nrobakova:
Closes #2001 .

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/grid.component.html
+++ b/projects/igniteui-angular/src/lib/grid/grid.component.html
@@ -121,7 +121,7 @@
 </div>
 
 <ng-template #emptyFilteredGrid>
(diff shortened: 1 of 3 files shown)


hanastasov:
I have verified that the fix is good and it works, will add the verified label after the requested changes are done.

PR #2058, opened 2018-07-19
Clear overlays after test execution

### LINKED ISSUE #2029: Remove overlays after filter-ui tests execution
### Description  
Overlays remain after tests for filter-ui are executed

 * browser: all 

### Steps to reproduce  

1. Run tests for filter ui

### Result  
Overlays remain

### Expected result  
Overlays should be removed

### Attachments  

sstoyanovIG:
Closes #2029  .  


projects/igniteui-angular/src/lib/services/overlay/overlay.spec.ts
```diff
@@ -80,11 +73,11 @@ describe('igxOverlay', () => {
             imports: [IgxToggleModule, DynamicModule, NoopAnimationsModule],
             declarations: DIRECTIVE_COMPONENTS
         }).compileComponents();
-        clearOverlay();
+        HelperUtils.clearOverlay();
     });
 
-    afterAll(async () => {
-        clearOverlay();
+    afterAll(() => {
```
zdrawku: Shouldn't we use "afterEach" instead of afterAll?

sstoyanovIG: There is no need, because in the beforeEach, 'clearOverlay' is called.

PR #2063, opened 2018-07-19
docs(forOf): Add API docs for IgxForOfDirective

alexalex972:


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -32,21 +32,116 @@ import { VirtualHelperComponent } from './virtual.helper.component';
...
+     * ```html
+     * <ng-template igxFor let-item [igxForOf]="data"></ng-template>
+     * ```
+     */
+    @Input()
+    public igxForOf: any[];
+
+    /**
+     * An @Input property that specifies the scroll orientation.
+     * Only `vertical` and `horizontal` are valid.
```
mpavlinov: @alexalex972 Suggestion: "Scroll orientation can be `"vertical"` or `"horizontal"`.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -32,21 +32,116 @@ import { VirtualHelperComponent } from './virtual.helper.component';
...
+     * </ng-template>
+     * ```
+     */
+    @Input()
+    public igxForScrollContainer: any;
+
+    /**
+     * An @Input property that sets the px-affixed size of the container along the axis of scrolling.
+     * ```html
+     * <ng-template igxFor let-item [igxForOf]="data" [igxForContainerSize]="'500px'"></ng-template>
```
mpavlinov: @alexalex972 if `igxForScrollOrientation` is mandatory for igxFor to work then you should use it in every code snippet.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -32,21 +32,116 @@ import { VirtualHelperComponent } from './virtual.helper.component';
...
+     *           </ng-template>
+     *       </div>
+     * </ng-template>
+     * ```
+     */
+    @Input()
+    public igxForScrollContainer: any;
+
+    /**
+     * An @Input property that sets the px-affixed size of the container along the axis of scrolling.
```
mpavlinov: @alexalex972 Please add: "For `"horizontal"` orientation this value is the width of the container and for `"vertical"` is the height."

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -32,21 +32,116 @@ import { VirtualHelperComponent } from './virtual.helper.component';
...
+     * ```html
+     * <ng-template igxFor let-item [igxForOf]="data" [igxForContainerSize]="'500px'"></ng-template>
+     * ```
+     */
+    @Input()
+    public igxForContainerSize: any;
+
+    /**
+     * An @Input property that sets the px-affixed size of the item along the axis of scrolling.
```
mpavlinov: @alexalex972 Please add: "For `"horizontal"` orientation this value is the width of the column and for `"vertical"` is the height or the row."

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -32,21 +32,116 @@ import { VirtualHelperComponent } from './virtual.helper.component';
...
+    @Input()
+    public igxForItemSize: any;
+
+    /**
+     * @hidden
+     */
     public dc: ComponentRef<DisplayContainerComponent>;
+
+    /**
+     * The current state of the directive. Contains the startIndex and chunkSize.
```
mpavlinov: @alexalex972 This needs more explanation. We should explain what "state" is and why it's used.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -32,21 +32,116 @@ import { VirtualHelperComponent } from './virtual.helper.component';
...
+    /**
+     * The total count of the virtual data items, when using remote service.
+     * ```typescript
+     * this.parentVirtDir.totalItemCount = data.Count;
+     * ```
+     */
     public totalItemCount: number = null;
 
+    /**
+     * Event triggered after a new chunk has been loaded.
```
mpavlinov: @alexalex972 Use "An event that is emitted" instead of "Event triggered" for all the events.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -306,30 +432,62 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
+     * this.parentVirtDir.scrollNext();
+     * ```
+     * @param index
+     */
     public scrollNext() {
         this.scrollTo(this.state.startIndex + 1);
     }
 
+    /**
+     * Scrolls by one item into the appropriate previous direction.
```
mpavlinov: @alexalex972 "Scrolls by one item into the appropriate previous direction. For `"horizontal"` orientation that will be the the left column and for `"vertical"` that is the upper row."

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -306,30 +432,62 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
 
+    /**
+     * @hidden
+     */
     public getColumnScrollLeft(colIndex) {
         return this.hCache[colIndex];
     }
 
+    /**
+     * Returns a reference to the vertical scrollbar.
```
mpavlinov: @alexalex972 "Returns a reference to the vertical scrollbar DOM element."

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -306,30 +432,62 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
+     */
     public getVerticalScroll() {
         if (this.vh) {
             return this.vh.instance.elementRef.nativeElement;
         }
         return null;
     }
 
+    /**
+     * Returns a reference to the horizontal scrollbar.
```
mpavlinov: @alexalex972 "Returns a reference to the horizontal scrollbar DOM element."

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -558,10 +746,30 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         return false;
     }
 
+    /**
+     * Returns the current function by which the different data items are compared.
```
mpavlinov: @alexalex972 This needs more explanation. Why do we compare items? Why would the developer need to set this option? What use case it resolves?

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
MayaKirova: @mpavlinov How about something like this:
The current state of the directive. It contains startIndex and chunkSize.
state.startIndex - The index of the item at which the current visible chunk begins.
state.chunkSize - The number of items the current visible chunk holds.
These options can be used when implementing remote virtualization as they provide the necessary state information.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
MayaKirova: @mpavlinov I suggest something like this, might be too long for a description so let me know what you think: 

Gets/Sets the function used to track changes in the items collection. This function can be set in scenarios where you want to optimize or customize the tracking of changes for the items in the collection.
The trackBy function takes the index and the current item as arguments and needs to return the unique identifier for this item.
By default the object references are compared, however this can be optimized if you have unique identifier value that can be used for the comparison instead of the object ref or if you have some other property values in the item object that should be tracked for changes.
This option is similar to ngForTrackBy.

PR #2078, opened 2018-07-21
Implement test enhancements for Column hiding, pinning, resizing

### LINKED ISSUE #2042: igxGrid tests enhancement
Implement tests enhancements - related to [PR](https://github.com/IgniteUI/igniteui-angular/pull/1968)

- [ ] cell.spec.ts
- [ ] column-group.spec.ts
- [x] column-hiding.spec.ts
- [x] column-pinning.spec.ts
- [x] column-resizing.spec.ts
- [ ] column.spec.ts
- [x] grid-filtering-ui.spec.ts
- [x] grid-filtering.spec.ts
- [ ] grid-selection.spec.ts
- [ ] grid-summary.spec.ts
- [ ] grid-toolbar.spec.ts
- [ ] grid.component.spec.ts
- [ ] grid.crud.spec.ts
- [ ] grid.groupby.spec.ts
- [ ] grid.initialization.spec.ts
- [ ] grid.pagination.spec.ts
- [ ] grid.pinning.spec.ts
- [ ] grid.search.spec.ts
- [ ] grid.sorting.spec.ts

Aleksandyr:
Closes #2042


projects/igniteui-angular/src/lib/grid/column-hiding.spec.ts
```diff
@@ -130,7 +132,7 @@ describe('Column Hiding UI', () => {
             grid.columns[0].hidden = true;
             expect(columnChooser.hiddenColumnsCount).toBe(1);
 
-            getCheckboxInput('Released').click();
+            GridFunctions.getCheckboxInput('Released', columnChooserElement, fix).click();
```
gedinakova: In order to avoid duplicating the "GridFunctions." portion every time we call some util function, I suggest we define a constant for each such method in the top level describe, and use it throughout the tests as follows:
const getCheckboxInput = GridFunctions.getCheckboxInput;
then use it with the signature from the utils class:
getCheckboxInput('Released', columnChooserElement, fix).click();

Thoughts?

Aleksandyr: It is good proposal but it won't work for this case because with this approach we will change the context of the execution of the function and when those functions invoke other functions from the GridFunctions class they won't exist anymore.

PR #2081, opened 2018-07-23
Add isSelected setter to igxDropDownItemBase, #2061

### LINKED ISSUE #2061: IgxDropDown Item cannot be set as selected.
### Description  
There's no option that allows to pre-select an item from the igx-drop-down.
There's an `isSelected` option, however it currently only has a getter.

### Result  
Cannot pre-select a drop-down item.

### Expected result  
Users should be able to pre-select an item from the drop-down, for example something like:

    <igx-drop-down>
        <igx-drop-down-item  isSelected='true'>
            1
        </igx-drop-down-item>
        <igx-drop-down-item>
            2
        </igx-drop-down-item>
    </igx-drop-down>

wnvko:
Closes #2061   

Also added some tests for this setter and refactor all igxDropDown to use fakeAsync.


projects/igniteui-angular/src/lib/drop-down/drop-down-item.component.ts
```diff
@@ -204,10 +205,48 @@ export class IgxDropDownItemBase {
...
+     *  let mySelectedItem = this.dropdown.selectedItem;
+     *  let isMyItemSelected = mySelectedItem.isSelected; // true
+     * ```
+     */
+    get isSelected() {
+        return this._isSelected;
+    }
+    @Input()
+    set isSelected(value: boolean) {
+        if (this.isHeader || this.disabled) {
```
damyanpetev: Should `disabled` have an effect on programmatic selection? Usually disabled only refers to the user interaction side.

kdinev:
Should be called `selected`.

projects/igniteui-angular/src/lib/drop-down/drop-down-item.component.ts
wnvko: fixed. Developers can select disabled items at through code behind now.

projects/igniteui-angular/src/lib/drop-down/drop-down-item.component.ts
```diff
@@ -204,10 +205,48 @@ export class IgxDropDownItemBase {
...
+            return;
+        }
+
+        this._isSelected = value;
+
+        const oldSelectedItem = this.dropDown.selectedItem;
```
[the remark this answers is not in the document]
wnvko: Actually we ended up with handling ngDoCheck event of IgxDropDownItemComponent and updating the selection when necessary in this event.

wnvko:
@kdinev we are adding setter to an existing property. If we change the name of the property from `isSelected` to `selected` this will be a breaking change.

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -16,7 +16,9 @@ import {
     Optional,
     HostListener,
     Directive,
-    Inject
+    Inject,
+    AfterContentInit,
+    AfterContentChecked
```
damyanpetev: Leftovers?

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -215,7 +218,14 @@ export class IgxDropDownBase implements OnInit, IToggleView {
      */
     public get selectedItem(): any {
         const selection = this.selectionAPI.get_selection(this.id);
-        return selection && selection.length > 0 ? selection[0] as IgxDropDownItemComponent : null;
+        const selectedItem = selection && selection.length > 0 ? selection[0] as IgxDropDownItemComponent : null;
```
damyanpetev: If you initialize the selection along with the component you won't need to check if it's there.
At that point this can be just `const selectedItem = selection[0] as IgxDropDownItemComponent` as [0] of empty array will just return undefined and you already have a check for the selected item right after :)

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -215,7 +218,14 @@ export class IgxDropDownBase implements OnInit, IToggleView {
...
+        if (selectedItem) {
+            if (selectedItem.isSelected) {
+                return selectedItem;
+            }
+            this.selectionAPI.set_selection(this.id, []);
```
damyanpetev: Umm shouldn't this be somewhere in `changeSelectedItem` and/or the item check instead of this getter?
Then you only have `selectedItem && selectedItem.isSelected => return selectedItem` otherwise `null`.

projects/igniteui-angular/src/lib/drop-down/README.md
```diff
@@ -65,10 +67,12 @@ The following getters are available on the **igx-drop-down** component:
 
 | Name | Type | Description |
 | :--- | :--- | :--- |
-| `selectedItem` | `number` | Gets the selected item. |
+| `selectedItem` | `any` | Gets the selected item. |
+| `collapsed` | `boolean` | Gets if the drop down is collapsed. |
 | `items` | `IgxDropDownItemComponent[]` | Gets all of the items but headers. |
 | `headers` | `IgxDropDownItemComponent[]` | Gets header items. |
 | `element`| `ElementRef` | Get dropdown html element. |
+| `scrollContainer`| `ElementRef` | Get drop down's html element of its scroll container. |
```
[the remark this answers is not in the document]
wnvko: removed.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 All notable changes for each version of this project will be documented in this file.
 ## 6.1.3
+- `igxDropDownItem`:
+    - isSelected property has Input setter and can be set both in markup and in code behind.
```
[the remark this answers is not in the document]
wnvko: fixed

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
wnvko: yeah, not there anymore.

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
wnvko: Yes, it will be better in this way. Fixed.

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
wnvko: If you set isSelected = true to an item in the markup this will set the selectedItem in selection service. At this point if the developer set to the same item isSelected = false in code behind this will not force change detector and selected item in selection service will not update. At this point if the developer tries to get the selected item from the drop down he will end with item which isSelected is set to false. I do not see any other place where I can handle this absolutely edge scenario.

PR #2083, opened 2018-07-23
chore(chip): Add chip dev sample for moving from one chiparea to another

alexalex972:


PavlovVasil:
The following error gets thrown on Google Chrome after the chip has been dropped: 

sketch

On Mozilla Firefox the error is this one: 

sketch2


On Edge the error is the same as in Chrome, but also the green highlight drop area is missing: 

sketch3

ChronosSF:
Please, fix the "Drop here" message not being removed if the chip is dropped while not being accepted.

Also when no chip is in the bottom area, the To: label is misplaced.

PR #2088, opened 2018-07-23
Fix grid not checking if initially pinned columns exceed the minimum unpinned area

### LINKED ISSUE #1612: The grid enables all the columns to be declared as pinned in the template
### Description  
The grid enables all the columns to be pinned and doesn't have an unpinned area and the scrollbars disappear.

### Steps to reproduce  

1. Run the column pinning sample.
2. Add property [pinned] = 'true' in the template..

### Result  
All columns are pinned.
There are no scrollbars.
There is no unpinned area.

### Expected result  
The grid should have pinned and unpinned area and should't allow all the columns to be pinned.

### Attachments  
issue

skrustev:
Closes #1612


ChronosSF:
Awaits agreement on the approach for such cases.

PR #2092, opened 2018-07-23
Prevent default, when pressing space and enter

### LINKED ISSUE #1947: When groupby row is focused and spacebar is pressed the browser scrolls down, everywhere except Chrome, although it should only collapse the group
### Description  
When spacebar is pressed and focus is on group header row the browser scrolls down and collapses group, but it should only collapse the group.

 * browser: Firefox, Edge, IE

### Steps to reproduce  

1. Open staging groupBy sample in Firefox.
2. Focus on any header-row of the groups.
3. Press spacebar.

### Result  
The grid collapses the group and the browser scrolls down.

### Expected result  
The grid should only collapse the group.

### Attachments  
groupby21
groupby22

sstoyanovIG:
Closes #1947 .


npaunov:
The issue is still reproducible in IE.

PR #2103, opened 2018-07-24
Add 'label' property to IFilteringOperation

### LINKED ISSUE #2067: IgxBooleanFilteringOperand doesn't work if the operation 'all' is localized(modified).
### Description  
IgxBooleanFilteringOperand doesn't work if the operation 'all' is localized(modified).

This is because 'all' is hardcoded at:
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grid/grid-filtering.component.ts#L193
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grid/grid-filtering.component.ts#L312

 * browser: Chrome

### Steps to reproduce  

1. Localize strings in filtering operands like this
  constructor() {
    IgxStringFilteringOperand.instance().operations.forEach(operation => operation.name = this.localizeFilteringString(operation.name));
    IgxDateFilteringOperand.instance().operations.forEach(operation => operation.name = this.localizeFilteringString(operation.name));
    IgxBooleanFilteringOperand.instance().operations.forEach(operation => operation.name = this.localizeFilteringString(operation.name));
    IgxNumberFilteringOperand.instance().operations.forEach(operation => operation.name = this.localizeFilteringString(operation.name));

  }

  localizeFilteringString(targetString: string): string {
    let result: string = targetString;
    switch (targetString) {
        case 'contains':
          result = '含む';
        break;
        case 'doesNotContain':
          result = '含まない';
          break;
          case 'all':
            result = '全て';
            break;
        default:
          break;
    }
    return result;
  }

2. Run the app and filter by 'true' in a boolean column.
3. Filter by '全て'(means 'All' in this case) in the same boolean column.

### Result  
Can't filter by 'All'

### Expected result  
Should be filtered by 'All'

### Attachments  
N/A

sstoyanovIG:
Closes #2067  .


kdinev:
What are the cases where different `name` and `label` would be used?

sstoyanovIG:
Once the user localize the operators' names, internally we are no longer able to get a specific operator by name. Currently there are two places where we need to check the name of an operator.

kdinev:
I've suggested a different approach in the issue.

PR #2107, opened 2018-07-24
Fix for #1886

### LINKED ISSUE #1977: igxFor API method that returns the number of cells/rows in view
`igxFor` - Create an API method that returns how many cells are fully in view,

- or returns a floating point for the cells fully in view - like 5.6 when 60% of the last cell is visible.

MayaKirova:
Adding new API method that returns count of items fully visible in view (excluded partially visible items and items outside of the view).

Fix for #1886: Search match should scroll correct item in view.

Closes #1977


projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -3722,16 +3722,24 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
 
         let size;
```
DDavidkov: I would remove the whole if/else block here and simply substitute it with:

```typescript
const size = directive.getItemCountInView();
```
since the new method determines the item count for both rows.

PR #2113, opened 2018-07-25
slider - track min/maxValue when lower/upperBound is undefined #920

### LINKED ISSUE #920: igx-slider - slider comes to not work well after changing maxValue.
### Description  
slider comes to not work well after changing `maxValue`.

### Steps to reproduce  

1. Go to https://stackblitz.com/edit/igx-slider-issue
2. At first, max value is 10. Drag a slider and move to the right end.
3. Then, click a "Change MaxValue to 100" button so that max value is changed to 100.
4. Drag the slider and move back to the left then move to the right end again.

### Result  
At step 4, I can't move slider to the right end. 

### Expected result  
Slider should work well even if after changing `maxValue`.

SAndreeva:
Closes #920 .  

The fix covers the following scenarios:
1. If upper/lowerBound are undefined on initialization, then upperBound/lowerBound will change whenever maxValue/minValue change. 
If upperBound/lowerBound are set from outside however, we stop updating them when maxValue/minValue change.
2. Whenever maxValue/minValue change, we take care that **value** is updated respectively, so that we avoid thumb flickering on focus.


projects/igniteui-angular/src/lib/slider/slider.component.spec.ts
```diff
@@ -5,12 +5,13 @@ import {IgxSliderComponent, IgxSliderModule, IRangeSliderValue, SliderType} from
 
 declare var Simulator: any;
 
-describe('IgxSlider', () => {
+fdescribe('IgxSlider', () => {
```
bkulov: Please remove the `f`.

projects/igniteui-angular/src/lib/slider/slider.component.ts
```diff
@@ -700,6 +719,54 @@ export class IgxSliderComponent implements ControlValueAccessor, OnInit, AfterVi
         );
     }
 
+    private resetValue() {
```
bkulov: Reset means setting the value to initial/default value. This method does value validation. So I think a proper name would be `invalidateValue`.

projects/igniteui-angular/src/lib/slider/slider.component.ts
```diff
@@ -700,6 +719,54 @@ export class IgxSliderComponent implements ControlValueAccessor, OnInit, AfterVi
...
+        if (!this.isRange) {
+            if (this.value >= this._lowerBound && this.value <= this._upperBound) {
+                this.value = this.value;
```
bkulov: Do we need to reassign the value if its valid?

projects/igniteui-angular/src/lib/slider/slider.component.ts
```diff
@@ -700,6 +719,54 @@ export class IgxSliderComponent implements ControlValueAccessor, OnInit, AfterVi
...
+            } else if (this.value < this._lowerBound) {
+                this.value = this._lowerBound;
+            } else if (this.value > this._upperBound) {
+                this.value = this._upperBound;
+            }
+        } else {
+            if ((this.value as IRangeSliderValue).lower >= this._lowerBound &&
```
bkulov: `(this.value as IRangeSliderValue)` is use a lot in this method. It introduces additional complexity to the code. Consider introducing a local variable for this.

projects/igniteui-angular/src/lib/slider/slider.component.ts
```diff
@@ -700,6 +719,54 @@ export class IgxSliderComponent implements ControlValueAccessor, OnInit, AfterVi
...
+                (this.value as IRangeSliderValue).lower <= this._upperBound) {
+                    this.value = {
```
bkulov: Do we need to reassign the value if its valid?

projects/igniteui-angular/src/lib/slider/slider.component.ts
```diff
@@ -700,6 +719,54 @@ export class IgxSliderComponent implements ControlValueAccessor, OnInit, AfterVi
...
+            } else if ((this.value as IRangeSliderValue).lower > this._upperBound) {
+                this.value = {
+                    lower: (this.value as IRangeSliderValue).lower,
+                    upper: this._upperBound
+                };
+            }
+
+            if ((this.value as IRangeSliderValue).upper >= this._lowerBound &&
+                (this.value as IRangeSliderValue).upper <= this._upperBound) {
+                    this.value = {
```
bkulov: Do we need to reassign the value if its valid?

projects/igniteui-angular/src/lib/slider/slider.component.spec.ts
```diff
@@ -622,6 +623,149 @@ describe('IgxSlider', () => {
...
+        fixture.detectChanges();
+
+        expect(slider.value).toBe(5);
+        expect(slider.minValue).toBe(5);
+        expect(slider.maxValue).toBe(8);
+        expect(slider.lowerBound).toBe(5);
+        expect(slider.upperBound).toBe(8);
+    });
+
+    it('should track min/maxValue and reset the value if lower/upperBound are undefined (issue #920)', () => {
```
bkulov: Probably it's better to use `invalidate` instead of `reset`.

projects/igniteui-angular/src/lib/slider/slider.component.spec.ts
```diff
@@ -622,6 +623,149 @@ describe('IgxSlider', () => {
...
         expect(slider.maxValue).toEqual(fix.componentInstance.maxValue);
     });
+
+    it('should track min/maxValue if lower/upperBound are undefined (issue #920)', () => {
+        const fixture = TestBed.createComponent(SliderTestComponent);
+        fixture.detectChanges();
+
+        const slider = fixture.componentInstance.slider;
+
+        fixture.componentInstance.slider.value = 5;
```
bkulov: You should not set the value in this test. You're doing this in the next test. This test is meant to test only the tracking of the lower and upper bounds. Test only the functionality described in the test title.

projects/igniteui-angular/src/lib/slider/slider.component.spec.ts
```diff
@@ -622,6 +623,149 @@ describe('IgxSlider', () => {
...
+        expect(slider.upperBound).toBe(8);
+    });
+
+    it('should stop tracking min/maxValue if lower/upperBound is set from outside (issue #920)', () => {
+        const fixture = TestBed.createComponent(SliderTestComponent);
+        fixture.detectChanges();
+
+        const slider = fixture.componentInstance.slider;
+
+        fixture.componentInstance.slider.value = 5;
```
bkulov: Same as above. Don't set the value.

projects/igniteui-angular/src/lib/slider/slider.component.spec.ts
```diff
@@ -622,6 +623,149 @@ describe('IgxSlider', () => {
...
+
+    it('should track min/maxValue if lower/upperBound are undefined - range slider (issue #920)', () => {
+        const fixture = TestBed.createComponent(SliderTestComponent);
+        fixture.detectChanges();
+
+        const slider = fixture.componentInstance.slider;
+        fixture.componentInstance.type = SliderType.RANGE;
+        fixture.detectChanges();
+
+        fixture.componentInstance.slider.value = {
```
bkulov: Same as above. Don't set the value.

projects/igniteui-angular/src/lib/slider/slider.component.spec.ts
```diff
@@ -622,6 +623,149 @@ describe('IgxSlider', () => {
...
+
+        expect(slider.minValue).toBe(5);
+        expect(slider.maxValue).toBe(8);
+        expect(slider.lowerBound).toBe(5);
+        expect(slider.upperBound).toBe(8);
+        expect((slider.value as IRangeSliderValue).lower).toBe(6);
+        expect((slider.value as IRangeSliderValue).upper).toBe(8);
+    });
+
+    it('should track min/maxValue and reset the value if lower/upperBound are undefined - range slider (issue #920)', () => {
```
bkulov: Probably it's better to use invalidate instead of reset.

PR #2177, opened 2018-07-30
Cell - Key Navigation - Rework Tab / Shift + Tab handlers

### LINKED ISSUE #1937: Can't navigate from first row cell to selection checkbox with key combination
### Description  
In a Grid with Row Selection (multiple) the key combination of Shift+Tab doesn't move the focus from the first row cell to the selection checkbox before it.

### Steps to reproduce  

1. Open the Grid Selection dev demo or from the embedded sample on [staging](https://staging.infragistics.local/products/ignite-ui-angular/angular/components/grid_selection.html)
2. Enable Multiple Selection if not active
3. Select a cell in the first data column
4. Press Shift+Tab

### Result  
Nothing happens, focus remains on the cell.

### Expected result  
The selection checkbox before the cell in that row should've been selected, like it does [on live](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid_selection.html).

#### Note
The checkbox itself is focusable alright - can be verified by clicking the last column header to focus it -> Tab to focus the first row -> Tab again focuses the selection checkbox. That leaves the handler on the row/cell.

ViktorSlavov:
Closes #1937 .  

TAB / SHIFT + TAB HostListener now checks if cell is last/first in row:

 - if true, clear cell selection and execute default behaviour
 - if false, execute arrowRight / arrowLeft key handlers


kdinev:
Can you write some tests for this behavior?

PR #2190, opened 2018-07-31
chore(*): updating dev deps - 6.1.x

ChronosSF:
It seems this may resolve the failing tests.


ChronosSF:
@kdinev, i wanted a branch that only contains this particular commit to isolate the possible cause for tests to pass.

I'd also not merge this until late afternoon so we can run it when travis is heavily loaded.

PR #2198, opened 2018-07-31
Add a null check for the _moveColumns method

### LINKED ISSUE #2158: columnMove doesn't work if no data is loaded.
### Description  
columnMove doesn't work if no data is loaded.

 * igniteui-angular version: 
 * browser: 

### Steps to reproduce  
1. run sample here: https://stackblitz.com/edit/poc-hal180
2. Run specifically the No Data ColSize sample

(optionally) also try to drag columns around

### Result  
Even though there are API calls in the AfterInit to move columns the Columns are in the original order

### Expected result  
Columns should be in the expected order.

Additionally the moveColumn(colSource, colTarget) is clunky as hell. Idealistically the visibleIndex should just be settable on the column. Or at least a move column overload that takes a number as the target to put the target at that index. 

The intent from the customers perspective is to load in the customers column preferences from a previous visit.  But the grid starts off empty. All settings should be settable at this time.

DDavidkov:
Closes #2158.

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grid/grid.component.ts
@@ -2250,12 +2250,16 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         const fi = list.indexOf(from);
         const ti = list.indexOf(to);
 
-        const activeInfo = IgxTextHighlightDirective.highlightGroupsMap.get(this.id);
-        const activeColumnIndex = activeInfo.columnIndex;
         let activeColumn = null;
+        let activeColumnIndex = -1;
 
-        if (activeInfo.columnIndex !== -1) {
-            activeColumn = list[activeColumnIndex];
+        if (this.lastSearchInfo.searchText) {
+            const activeInfo = IgxTextHighlightDirective.highlightGroupsMap.get(this.id);
+            activeColumnIndex = activeInfo.columnIndex;
+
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -2250,12 +2250,16 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
         let activeColumn = null;
+        let activeColumnIndex = -1;
 
-        if (activeInfo.columnIndex !== -1) {
-            activeColumn = list[activeColumnIndex];
+        if (this.lastSearchInfo.searchText) {
+            const activeInfo = IgxTextHighlightDirective.highlightGroupsMap.get(this.id);
+            activeColumnIndex = activeInfo.columnIndex;
+
+            if (activeInfo.columnIndex !== -1) {
```
gedinakova: You may simply use activeColumnIndex.

SAndreeva:
There is an error in the console, coming from `hidden` setter in `IgxColumnComponent`.

DDavidkov:
@SAndreeva Using the attached modified version of the sample (as the original one is ridiculous and full of errors), I am not able to reproduce the hidden setter error with both the 6.1 branch for the fix and the master one. Am I missing something?

[ColmoveError.zip](https://github.com/IgniteUI/igniteui-angular/files/2253350/ColmoveError.zip)

PR #2233, opened 2018-08-01
Grid - Cell Navigation - Align Top cell to header on KeyUp

### LINKED ISSUE #1185: Top cell is not positioned aligned to the header, after keyboard navigation
### Description  
When navigating with keyboard through grid cells, and in the same time the top row is not fully visible, after the top cell is reached, then the top selected cell should be selected and in the same time to be fully visible and aligned with header.

### Steps to reproduce  

1. Load a sample with the grid - Grid Selection
2. Scroll down the grid so that the top cell is half visible
3. Select a cell from the second visible row.
4. Use arrow up to focus the half visible cell

### Result  
The top cell is correctly selected, but the new loaded cell at the top is still half visible.

### Expected result  
The top cell is selected, fully visible and aligned with the top.

ViktorSlavov:
Closes #1185 

Adjust scroll amount depending on cell visibility.

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grid/grid.component.ts
@@ -3634,11 +3634,13 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         if (target) {
             const containerTopOffset =
                 parseInt(row.grid.verticalScrollContainer.dc.instance._viewContainer.element.nativeElement.style.top, 10);
-            if (this.rowHeight > -containerTopOffset // not the entire row is visible, due to grid offset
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -4134,7 +4136,7 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         if (this.groupingExpressions && this.groupingExpressions.length) {
             const state = {
                 expressions: this.groupingExpressions,
-                expansion: this.groupingExpansionState,
+                expansion:  this.groupingExpansionState,
```
kdinev: There's an extra space here.

Issue #2157, opened 2018-07-26
Deselecting all rows when they are filtered makes the grid unresponsive and sometimes crashes the browser.


PavlovVasil:
### Description  
Deselecting all rows when there is a large number of filtered rows takes very long time. 

 * igniteui-angular version: 6.1.x; 6.2.x
 * browser: All

### Steps to reproduce  

1. Open the Grid Performance dev sample
2. Filter the second grid column (column "1") by "Contains" and type 1 so all rows get included in the filtered results - see the attached screenshot
3. Select All rows by clicking left of "ID" column
4. Deselect all column by clicking again

### Result  
The browser freezes - sometimes for a very long time. MS Edge becomes completely unresponsive.

### Expected result  
Deselecting should not take more time than selecting

### Attachments  

![sketch](https://user-images.githubusercontent.com/18643062/43275599-e5aa2e7a-910a-11e8-9c68-6fefe9c97603.png)

PR #2237, opened 2018-08-01
Selection service refactor using Set

### LINKED ISSUE #1258: Improve row selection performance
### Description  
Now selection uses array to save the selected items' identifiers. In some cases, when there is filtering the selection performance can be improved.
Research a different than an array collection, that can be used to preserve selected items' identifiers and will allow fast access to its items.

Lipata:
Closes #1258, #2157  .  

Selection service now uses Set instead of array. The main benefit is that Set.has() and Set.delete() are with greater performance than array.indexOf() and array.filter(), when locating and deleting element in the collection. This affects selection performance, especially in IE11.

The public methods and event arguments for grid row selection and combo are still array's which requires some transformation from Set -> array and from array -> Set, which is O(n) and doesn't affect performance.

The combo with remote data was using compare functions, when comparing their selected items' ID. Set collection in JavaScript doesn't support compare functions, therefore when combo is with remote data the ID is stringified.


projects/igniteui-angular/src/lib/core/selection.ts
```diff
@@ -1,141 +1,98 @@
...
-     * If not presented, then all items equal by reference to any item from the selection (using ===) are removed.
      *
      * @returns selection after the item is removed
      */
-    public deselect_items(componentID: string, itemIDs: any[], compareFunc?: (item) => boolean, ): any[] {
-        let selection: any[];
-        itemIDs.forEach((deselectedItem) => selection = this.deselect_item(componentID, deselectedItem, compareFunc, selection));
+    public deselect_items(componentID: string, itemIDs: any[]): Set<any> {
+        let selection: Set<any>;
+        itemIDs.forEach((deselectedItem) => selection = this.deselect_item(componentID, deselectedItem, selection));
```
ViktorSlavov: This does not seems right 
In deselect_item, it will not enter the `!sel` *if*, skipping `sel = this.get_selection(componentID)`, so it will `delete` items from an empty set.

projects/igniteui-angular/src/lib/core/selection.ts
```diff
@@ -1,141 +1,98 @@
...
-     * If not presented then items are compared by reference (using ===).
      *
      * @returns selection after the new item is added
      */
-    public select_items(componentID: string, itemIDs: any[], compareFunc?: (item) => boolean): any[] {
-        let selection: any[];
-        itemIDs.forEach((item) => selection = this.select_item(componentID, item, compareFunc, selection));
+    public select_items(componentID: string, itemIDs: any[], clearSelection?: boolean): Set<any> {
+        let selection: Set<any>;
+        itemIDs.forEach((item) => selection = this.select_item(componentID, item, selection, clearSelection));
```
ViktorSlavov: I think that calling `select_item` with `clearSelection === true` will always add clear selection to an empty Set, so iwhen calling `select_items('grid1', ['1','2','3'], true)`, the selection in the end will be `Set('3')`

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -1044,20 +1045,25 @@ export class IgxComboComponent implements AfterViewInit, ControlValueAccessor, O
...
-    protected triggerSelectionChange(newSelection) {
+    protected triggerSelectionChange(newSelectionAsSet: Set<any>) {
         const oldSelection = this.dropdown.selectedItem;
+        const newSelection = newSelectionAsSet ? Array.from(newSelectionAsSet) : [];
         if (oldSelection !== newSelection) {
             const args: IComboSelectionChangeEventArgs = { oldSelection, newSelection };
             this.onSelectionChange.emit(args);
-            this.selectionAPI.set_selection(this.id, newSelection);
+            newSelectionAsSet = new Set();
+            for (let i = 0; i < args.newSelection.length; i++) {
```
ViktorSlavov: `newSelection` already is an array, I think it would be easier to read if cycling through `newSelection` instead of `args.newSelection`

PR #2269, opened 2018-08-02
fix(carousel): fixing issue when a slide is shown after it gets removed

### LINKED ISSUE #2182: In carousel when call method remove for selected slide it is still previewed
### Description  
In carousel when call method remove for selected slide it is still previewed

### Steps to reproduce  

1. Open https://stackblitz.com/edit/angular-ncldde?file=app/carousel-sample-1/carousel-sample-1.component.ts
2. Click on button Remove Slide

### Result  
Slide 1 is removed, but the carousel still displays it.

### Expected result  
When the current slide is removed, next slide should be selected

### Attachments  

rmkrmk:
Closes #2182  

fixing issue when a slide is shown after it gets removed


projects/igniteui-angular/src/lib/carousel/carousel.component.ts
```diff
@@ -315,19 +315,26 @@ export class IgxCarouselComponent implements OnDestroy {
      * @memberOf IgxCarouselComponent
      */
     public remove(slide: IgxSlideComponent) {
-        this.slides.splice(slide.index, 1);
-        this._total -= 1;
+        if (slide === this.get(slide.index)) { // check if the requested slide for delete is present in the carousel
```
gedinakova: It would be nice if we add a check for slide as well.

rmkrmk: Done, check added.

PR #2271, opened 2018-08-02
Combo - Add Item - Do not fire on Keydown.Space - 6.1.x

### LINKED ISSUE #2266: Combo - Keyboard Navigation - Add Item button fires on Keydown.Space
### Description  
The Add Item button fires `addToCollection()` on `keydown.Space` instead of only on `keydown.Enter`

 * browser: All

### Steps to reproduce  

1. Open DevSamples/Combo
2. Drop-down the Combo and type "CUSTOM VALUE" in search
3. Hit ArrowDown and then Space

### Result  
"CUSTOM VALUE" is added to the combo collection/selection.

### Expected result  
"CUSTOM VALUE" should not be added to the combo collection/selection. Per spec, only ENTER should add.

ViktorSlavov:
Closes #2266 .

### DIFF
--- a/projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
+++ b/projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
@@ -590,7 +590,8 @@ export class IgxDropDownItemNavigationDirective {
      */
     @HostListener('keydown.Space', ['$event'])
     onSpaceKeyDown(event) {
-        this.target.selectItem(this.target.focusedItem, event);
+        // V.S. : IgxDropDownComponent.selectItem needs event to be true in order to close DD as per specification
+        this.target.selectItem(this.target.focusedItem, this.target instanceof IgxDropDownComponent);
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -590,7 +590,8 @@ export class IgxDropDownItemNavigationDirective {
      */
     @HostListener('keydown.Space', ['$event'])
     onSpaceKeyDown(event) {
-        this.target.selectItem(this.target.focusedItem, event);
+        // V.S. : IgxDropDownComponent.selectItem needs event to be true in order to close DD as per specification
+        this.target.selectItem(this.target.focusedItem, this.target instanceof IgxDropDownComponent);
```
Lipata: Using this.target instanceof IgxDropDownComponent ? event : null can prevent problems, if in the future event inside selectItem is used and not only checked for existence

PR #2284, opened 2018-08-02
Fix column width calculations when there are hidden columns

### LINKED ISSUE #2249: Blank space on the right of igxGrid when there is a hidden column and grid width is 100%
### Description  
There is a blank space on the right when there is a hidden column, grid width is 100% and columns do not have width set.

 * browser: All browsers

### Steps to reproduce  

1. Given the following grid configuration (taken from [here](https://staging.infragistics.local/products/ignite-ui-angular/angular/components/integration_excel.html#paste-from-excel))

```
    <igx-grid #grid1 [data]="data" [width]="'100%'" [height]="'505px'" [autoGenerate]="false" paste-handler (onDataProcessed)="dataPasted($event)" [primaryKey]="'ID'">
        <igx-column [field]="'ID'" [hidden]="true"></igx-column>
        <igx-column [field]="'Name'"></igx-column>
        <igx-column [field]="'Title'"></igx-column>
        <igx-column [field]="'Phone'"></igx-column>
        <igx-column [field]="'Country'"></igx-column>
    </igx-grid>
```

2. Use the configuration in [this](https://github.com/IgniteUI/igniteui-angular-samples/blob/master/src/app/grid/grid-paste/grid-paste-sample.component.html) sample
3. Run the sample and observe the result

Note: This is only reproduced in igniteui-angular version 6.1.2

### Result  
There is a blank space on the right

### Expected result  
There should be no blank space

### Attachments  
image

DDavidkov:
Closes #2249.


projects/igniteui-angular/src/lib/grid/column-hiding.spec.ts
```diff
@@ -1213,3 +1236,19 @@ export class GridWithGroupColumnsComponent implements AfterViewInit {
     }
 
 }
+
+@Component({
```
gedinakova: Following the test optimizations we recently introduced, we should avoid declaring grid components in spec files. You should either find a component that is already defined (in grid-base-components or grid-samples), or create your new one in one of those files. For this particular scenario, I believe you can use the BasicGridSearchComponent or the PagingComponent and execute the code in changeColumnVisibility in the test itself.

projects/igniteui-angular/src/lib/test-utils/template-strings.spec.ts
```diff
@@ -75,6 +75,13 @@ export class ColumnDefinitions {
         <igx-column field="HireDate" pinned="true"></igx-column>
     `;
 
+    public static idNameHiddenJobHire = `
```
gedinakova: You may consider using the columns definition just above this new one and simply unpin the last column.

PR #2286, opened 2018-08-02
Fix column width calculations when there are hidden columns

### LINKED ISSUE #2249: Blank space on the right of igxGrid when there is a hidden column and grid width is 100%

DDavidkov:
Closes #2249.

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/column-hiding.spec.ts
+++ b/projects/igniteui-angular/src/lib/grid/column-hiding.spec.ts
@@ -17,6 +17,7 @@ import { UIInteractions } from '../test-utils/ui-interactions.spec';
 import { SampleTestData } from '../test-utils/sample-test-data.spec';
 import { GridTemplateStrings, ColumnDefinitions } from '../test-utils/template-strings.spec';
 import { GridFunctions } from '../test-utils/grid-functions.spec';
+import { GridSearchHiddenColumnsComponent } from '../test-utils/grid-samples.spec';
 
 describe('Column Hiding UI', () => {
     let fix;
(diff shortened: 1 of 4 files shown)


projects/igniteui-angular/src/lib/grid/column-hiding.spec.ts
```diff
@@ -1051,6 +1052,28 @@ describe('Column Hiding UI', () => {
...
+        beforeEach(() => {
+            fix = TestBed.createComponent(GridWithHiddenColumnComponent);
+            fix.detectChanges();
+            grid = fix.componentInstance.grid;
+            fix.detectChanges();
+
+            grid.cdr.detectChanges();
+        });
+
+        fit('the grid properly resizes its columns width when column visibility is changed', () => {
```
ChronosSF: left-over `fit`

PR #2289, opened 2018-08-03
Grid cell keyboard navigation performance and spec test enhancement

### LINKED ISSUE #1923: Grid Keyboard navigation performance issue
### Description  
When you open the Propeller Grid and you select a cell and hold the right arrow key to start navigation the scroll is lagging. 

### Steps to reproduce  

1. Open Staging
2. Scroll the Grid with the keyboard
3. ...

### Result  
There is a lag on scrolling

### Expected result  
There should be no lag

ddincheva:
Closes #1923.


zdrawku:
4 tests are failing because a fixes for the keyboard navigation should be added

zdrawku:
The issue that I've described earlier was related to the sample itself and grid width boundaries. This is not a issue related to the keyboard navigation functionality

PR #2314, opened 2018-08-03
feat(grid): adding no inferrable height fix #2313

### LINKED ISSUE #2313: Improve scenario when no height can be set for igxGrid
### Description  
When igxGrid has no height set (or has % height set) inside a container without height set no height can be inferred for the body during calculations. 

### Expected result  
The suggestion is to make the grid init a body with 500px height (based on the most often used browser resolution and space needed for headers / toolbar / paginator). If the page/data source has less records than needed to fill the space, the grid will instead render all records in the view (without vertical scrollbar)

ChronosSF:
Closes #2313 .  

I'll update the grid spec upon thumbs up from a review.


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 # Ignite UI for Angular Change Log
 
 All notable changes for each version of this project will be documented in this file.
+
+## 6.2.0
+- `igxGrid`
+    - Initializing an `igxGrid` component without setting height inside a container without height defined now causes the grid's body to 
+    become either 500px high or as high as it is needed to show all records in the current view whichever is smaller.
```
kdinev: This should be in `6.1.3`.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -1909,6 +1909,8 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
     private _columnWidth: string;
     private _columnWidthSetByUser = false;
 
+    private _defaultTargetBodyHeight = 500;
```
kdinev: So, by using a default `height` we're not sure how many records would be displayed, nor whether all records would be fully in view. Correct?

ChronosSF: If the records in the data view are not enough to span the whole 500px (based on their row height, based itself on the displayDensity set) the height will be their total height instead (and no vertical scrollbar will appear).

We could always do it the other way around - let's say we want Math.min(<number of records in data view>, 10)  records in the view - then we set body height equal to that * rowHeight. In this way the arbitrary number is the number of records, not the height. The caveat is that it's either 
 different based on the displayDensity or we get different default heights for each displayDensity.

kdinev: If it works with a number of records, it shouldn't care about the displayDensity. The records would just render with the display density that is set.

ChronosSF: Sure, let me know if you want me to change it. Have in mind the grid will still provide height to igxFor as per the directive's spec. It is just going to derive it based on an arbitrary number of records (10 to make it 500px in default density) instead of arbitrary height.

CHANGELOG.md
ChronosSF: Ok, I'll change it and make a PR to 6.1.x then. I didn't consider this a bug and only fixed it for master because of that.

projects/igniteui-angular/src/lib/grid/grid.component.ts
ChronosSF: We agreed on rendering height that allows for 10 records to be fully visible. This height will therefore differ between `displayDensity` settings. I'll make another PR with the proper changes.

PR #2317, opened 2018-08-06
Added documentation and code snippets to the IgxTextSelectionDirective

jackofdiamond5:


projects/igniteui-angular/src/lib/directives/text-selection/text-selection.directive.ts
```diff
@@ -8,6 +8,36 @@ export class IgxTextSelectionDirective {
...
+     * @ViewChild('firstName',
+     *  {read: IgxTextSelectionDirective})
+     * public firstName: IgxTextSelectionDirective;
+     *
+     * public getFirstNameSelectionStatus() {
+     *  return this.firstName.selected;
+     * }
+     * ```
+     */
     @Input('igxTextSelection')
```
Aleksandyr: For better generation of the documentation it would be more accurate if we could have two separate comments about the getter and the setter accessors.
My proposal for getter accessor: Returns whether the input element is selectable through the directive.
for setter accessor: Determines whether the input element could be selected through the directive.
And we could separate the snippets for the getter and setter respectively.

projects/igniteui-angular/src/lib/directives/text-selection/text-selection.directive.ts
```diff
@@ -17,17 +47,62 @@ export class IgxTextSelectionDirective {
...
+     */
     @HostListener('focus')
     onFocus() {
         this.trigger();
     }
 
     constructor(private element: ElementRef) { }
 
+    /**
+     * Triggers the selection of the element.
```
Aleksandyr: Another proposa, it may sounds more accurate: Triggers the selection of the element if it is marked as though.

PR #2322, opened 2018-08-06
Create FilteringExpressionTree internally

### LINKED ISSUE #2302: When you add expressions directly to grid's FilteringExpressionTree, the filter ui is not updated
### Description  
When you add expressions directly to grid's FilteringExpressionTree, the filter ui is not updated

 * browser: all

### Steps to reproduce  

1. Run the attached Sample
2. Got to Grid section
3. Open 'Country' filter dialog

### Result  
Grid is filtered, but the filter ui dialog is not populated

### Expected result  
filter ui dialog should be populated

### Attachments  
[grid-filtering.zip](https://github.com/IgniteUI/igniteui-angular/files/2257260/grid-filtering.zip)

sstoyanovIG:
Closes #2302  .


npaunov:
The filter ui is not updated when a Date column is used.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -234,9 +234,9 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
 	 * @memberof IgxGridComponent
 	 */
     set filteringExpressionsTree(value) {
-        if (value) {
-            for (let index = 0; index < (value as FilteringExpressionsTree).filteringOperands.length; index++) {
-                const val = (value as FilteringExpressionsTree);
+        if (value && value instanceof FilteringExpressionsTree) {
```
bkulov: The good practice is to check if a value is instance of an interface (if any). In this case this should be `value instance of IFilteringExpressionsTree`.

PR #2330, opened 2018-08-06
Add templating functionality to the IgxButtonGroup

### LINKED ISSUE #620: ixg-buttongroup templating
### Description  
I've been looking at the examples for the igx-buttongroup component. I have to say that I don't understand the reasoning behind how this component is structured. In the examples, it seems that the actual buttons are defined in the component. In my opinion it's a lot more clean to define all DOM elements in the template.

### Result  
Have to create an array of buttons in the component, with reduced control over the content

### Expected result  
Something in the lines of

```
<igx-buttongroup [multiSelection]="false">
<button igxButton><igx-icon>some_icon</igx-icon>Button 1 text</button>
<button igxButton><igx-icon>some_otheri_con</igx-icon>Button 2 text</button>
</igx-buttongroup>
```

DDavidkov:
Closes #620.

### DIFF
--- a/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
+++ b/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
@@ -277,6 +277,10 @@ export class IgxButtonGroupComponent implements AfterViewInit, OnDestroy {
      *```
      */
     public selectButton(index: number) {
+        if (index >= this.buttons.length || index < 0) {
+            return;
+        }
+
         const button = this.buttons[index];
         const buttonElement = button.nativeElement;
 
@@ -320,6 +324,10 @@ export class IgxButtonGroupComponent implements AfterViewInit, OnDestroy {
      * ```
      */
     public deselectButton(index: number) {
+        if (index >= this.buttons.length || index < 0) {
+            return;
+        }
+
         const button = this.buttons[index];
         const buttonElement = button.nativeElement;
 
@@ -348,6 +356,8 @@ export class IgxButtonGroupComponent implements AfterViewInit, OnDestroy {
         const initButtons = () => {
             // Cancel any existing buttonClick subscriptions
             this.buttonClickNotifier$.next();
+
+            this.selectedIndexes.splice(0, this.selectedIndexes.length);
 
             // initial configuration
             this.buttons.forEach((button, index) => {
--- a/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
+++ b/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
@@ -11,11 +11,14 @@ import {
     Output,
     QueryList,
     Renderer2,
-    ViewChildren
+    ViewChildren,
+    OnDestroy
 } from '@angular/core';
+import { Subject } from 'rxjs';
 import { IgxButtonDirective, IgxButtonModule } from '../directives/button/button.directive';
 import { IgxRippleModule } from '../directives/ripple/ripple.directive';
 import { IgxIconModule } from '../icon/index';
+import { takeUntil } from '../../../../../node_modules/rxjs/operators';
 
 export enum ButtonGroupAlignment { horizontal, vertical }
 let NEXT_ID = 0;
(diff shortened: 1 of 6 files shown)


projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
```diff
@@ -250,18 +272,28 @@ export class IgxButtonGroupComponent implements AfterViewInit {
...
             return;
         }
-        const buttonElement = this.buttons.toArray()[index]._el.nativeElement;
+
         this.selectedIndexes.push(index);
-        buttonElement.setAttribute('data-selected', true);
-        this.onSelect.emit({ button: this.buttons.toArray()[index], index });
-        this.values[index].selected = true;
+
+        this._renderer.setAttribute(buttonElement, 'data-selected', 'true');
```
rkaraivanov: IMO we should just expose an input property on the button directive and use it instead of setting data- attributes.

projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
```diff
@@ -283,29 +315,48 @@ export class IgxButtonGroupComponent implements AfterViewInit {
...
-        // initial selection
-        setTimeout(() => {
-            this.buttons.forEach((button, index) => {
-                if (!button.disabled && button._el.nativeElement.getAttribute('data-selected') === 'true') {
-                    this.selectButton(index);
-                }
-            });
-        }, 0);
+        // initial configuration
+        this.buttons.forEach((button, index) => {
```
rkaraivanov: What happens when we have arbitrary number of buttons rendered? Consider the case when somebody templates the buttons with `ngFor` and the input array is being changed "runtime". I think none of the classes and handlers will be applied for new elements as this is initialize once and forget. This should be moved inside the [`changes`](https://angular.io/api/core/QueryList#changes) emitter for both query lists and appropriate updates should be applied when a button is added/removed from the template.

projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
```diff
@@ -283,29 +315,48 @@ export class IgxButtonGroupComponent implements AfterViewInit {
...
+
+            if (this.disabled) {
+                button.disabled = true;
+            }
+
+            if (!button.disabled && buttonElement.getAttribute('data-selected') === 'true') {
+                this.selectButton(index);
+            }
+
+            button.buttonClick.subscribe((ev) => this._clickHandler(ev, index));
```
rkaraivanov: We should unsubscribe at `ngOnDestroy` 😺 
`takeUntil` is also an option

tachojelev:
**Issue:**
Setting the **selected** property of a button that is defined in the html template at run time does **not** work.

(The issue is not reproduced if the **buttonGroup.selectButton(...)** and **buttonGroup.deselectButton(...)** methods are used instead)

tachojelev:
**Issue:**

Let's say the buttonGroup has buttons defined in its template and its values property is bound to a separate collection and there are selected buttons from both the template buttons and the collection buttons.
In this case, if we clear or reset the bound button collection, then the **selectedButtons** property of the buttonGroup is not updated (still contains information about the selectedButton(s) from the cleared collection).

DDavidkov:
@tachojelev The first issue is expected, as the selected input is simply there on the button to signalize to the button group that it needs to select this button and to allow the developers to define a button as selected from the template. You should use the select/deselect methods of the button group to select/deselect a button, instead of setting the selected property directly. I have improved the API comment of the input to clarify that. As for the second issue, it should be now fixed, please retest it.

projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
```diff
@@ -1,17 +1,24 @@
...
+    QueryList,
+    Renderer2,
+    ViewChildren,
+    OnDestroy
 } from '@angular/core';
+import { Subject } from 'rxjs';
 import { IgxButtonDirective, IgxButtonModule } from '../directives/button/button.directive';
 import { IgxRippleModule } from '../directives/ripple/ripple.directive';
 import { IgxIconModule } from '../icon/index';
+import { takeUntil } from '../../../../../node_modules/rxjs/operators';
```
rkaraivanov: Wrong import path. I guess it is VS Code automatic import resolution but this should just be `import .. from 'rxjs/operators`

rkaraivanov: Wrong import path. I guess it is VS Code automatic import resolution but this should just be `import .. from 'rxjs/operators`

kdinev:
@DDavidkov Update the changelog please.

DDavidkov:
@kdinev Should I label it as a 6.2.0 change or as 6.1.3?

kdinev:
@DDavidkov It would be `6.2.0` or `6.1.4`. I already tagged `6.1.3`. Did you have a PR to `6.1.x` with the change?

DDavidkov:
@kdinev No I don't have a 6.1.x PR, so I am labeling it as 6.2.0 in the changelog

PR #2336, opened 2018-08-07
 circular and linear bars - prevent progress exceeding, smooth update when operate with big nums 

### LINKED ISSUE #2163: circular and linear bars - prevent progress exceeding, smooth update when operate with big nums, allow floating point nums, expose step input
### Description  

- [x] prevent progress exceeding 
- [x] smooth update when operate with big nums (max input) 
- [x] allow floating point nums 
- [x] expose step input
- [x] when increase/decrease value exceeds it should return the same value, not one step lower
- [x] parse passed value or step to float number.
- [x] cancel animation frame correctly

 * browser: all

### Steps to reproduce  

1. Run: https://stackblitz.com/edit/angular-3fa9sj?file=app/linear-progressbar-sample-1/linear-progressbar-sample-1.component.html
2. Press the start button

### Result  
The updating is too slow

### Expected result  
The update should be faster.

Aleksandyr:
Closes #2163

### DIFF
--- a/projects/igniteui-angular/src/lib/progressbar/circularbar.component.spec.ts
+++ b/projects/igniteui-angular/src/lib/progressbar/circularbar.component.spec.ts
@@ -307,7 +307,7 @@ describe('IgCircularBar', () => {
             fixture.detectChanges();
 
             const componentInstance = fixture.componentInstance;
-            const progressBarElem = fixture.componentInstance.circularBar.elementRef.nativeElement
+            const progressBarElem = fixture.debugElement.nativeElement
                 .querySelector('.progress-circular');
 
             tick(tickTime);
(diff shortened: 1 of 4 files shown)


Aleksandyr:
@hanastasov When you have no defined value for step input it automatically gets 1% of the max value in order to keep identical progress update for any kind of values, no matter whether they are too big or too small. I think this is the only correlation between the values.

But when the user provides value for the step input the only restriction he has is:
When step value is bigger than the value itself there is only one update and this update gets the value you have passed to the progressbar. So for instance when you have step=50 and value=20 the progress update gets the value representation which is 20.

As for the animation update yes. When the step value is bigger than the (maxValu * 1%) representation the progress update happens faster and respectively when the step value is lower than this same representation the progress update happens more slowly.

In conclusion I agree with you that all this new enhancements should be documented into the topic.

PR #2358, opened 2018-08-08
fix(igxGrid): default render target for grids with no height #2313

### LINKED ISSUE #2313: Improve scenario when no height can be set for igxGrid

ChronosSF:
Closes #2313 .  

I'll update the grid's spec when I receive thumbs up from a review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,8 @@ All notable changes for each version of this project will be documented in this 
         - The footerTemplate property is removed.
     - `igxColumnGroup`
         - The footerTemplate property is removed.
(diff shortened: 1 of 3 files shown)


ChronosSF:
The grid can calculate height for its container because of the sample's styling so nothing from the defaultTargetHeight implementation is used.

It's another question if it should consider this height when rendering in this case, but that's a question for another issue/discussion.

PR #2363, opened 2018-08-09
#2274 add Overlay setting for attach target, IgxOverlayOutletDirective

### LINKED ISSUE #2274: Apply multiple themes to components shown in igxOverlay
### Description
We need a way to apply multiple themes in igxOverlay.

### Steps to reproduce  
Create application with default theme. Add to this application some components. Then add one more component and apply to this one some other theme. Show in igxOverlay some component, e.g. igxDialog, and try to force it to be styled with the custom theme.

### Result  
Right now the global theme always style the components shown in igxOverlay.

### Expected result  
Developer should be able to set the theme igxOverlay should use which any specific component.

damyanpetev:
Closes #2274 .  

The `OverlaySettings` interface (also exposed through toggle directives and drop-down API) now also accepts an optional ~`target`~`outlet` property that the service will attach overlays to. That still defaults to body.
For example:
```ts
this.toggle.open({ outlet: this.elementRef });
```
will append the overlay structure to the host element instead of body to preserve DOM hierarchy for styling.

Additional `igxOverlayOutlet` directive available to use in markup only configuration accepted by the `igxToggleOutlet` on the toggle action and looks like:
```html
<button [igxToggleAction]="toggle" [igxToggleOutlet]="outlet">Toggle</button>
<div igxToggle #toggle="toggle"><!-- content to toggle --></div>
<!-- other content -->
<div igxOverlayOutlet #outlet="overlay-outlet" class="overlay-container"></div>
```
<br>

@kdinev Can you take a look at the naming and let me know if those look good? ~Particularly if `target` makes sense (considering the position strategy also has one) or perhaps it should be just `outlet` or the good old `appendTo` 🤔~


Lipata:
You can add the change to changelog and overlay readme.

rkaraivanov:
@damyanpetev Maybe rename `target` to `outlet`? Seems to be more consistent with the OverlayOutlet export in the markup

PR #2404, opened 2018-08-13
Disabled and special dates

### LINKED ISSUE #1980: Disabled dates
### Description  
I would like to see the calendar supports disabled function  eg past-dates, single dates, holidays etc

so when the calender is displayed the default feature should be all past dates and future dates are disabled.  and the current date is selected with background color like its  now 

 * igniteui-angular version: 
 * browser: 

[Update] - Calendar will supports "special" dates as well. These dates are going to have the same behavior as Disabled dates, although the difference would be the color.

"special" dates examples:
- National holidays of a country
- PTO dates in a company calendar
- Recurring monthy Company/Team meetings

### Attachments  here is screenshort how it suppose to look 

good one

tiliev:
Closes #1980.


projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -37,6 +36,25 @@ export enum CalendarSelection {
...
+export enum DateRangeType {
+    After,
+    Before,
+    Between,
+    Specific,
+    Weekdays,
+    Weekends
+}
+
+export class DateRangeDescriptor {
```
kdinev: You don't need a class here. `interface` is sufficient, since all you do in the constructor is assign the two properties.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -296,6 +314,59 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
...
+        this._disabledDates = value;
+    }
+
+    /**
+     * Gets the special dates descriptors.
+     * ```typescript
+     * let specialDates = this.calendar.specialDates;
+     * ```
+     */
+    public get specialDates(): DateRangeDescriptor[] {
```
kdinev: Let's think about the name of this `@Input`. Off the top of my head: 'holidays', 'highlightDates`, `significantDates`... Throw in your suggestions as well!

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -996,8 +1129,90 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
...
         this._onChangeCallback(this.selectedDates);
     }
+    /**
+     *@hidden
+     */
+    private selectDateFromClient(value: Date) {
+        switch (this.selection) {
+            case 'single':
+            case 'multi':
+                this.selectDate(value);
```
kdinev: Why would the disabled dates be selected in the case of `milti`?

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -996,8 +1129,90 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
...
+                case (DateRangeType.Between):
+                    const dRange = dRanges.map(d => d.getTime());
+                    const min = Math.min(dRange[0], dRange[1]);
+                    const max = Math.max(dRange[0], dRange[1]);
+                    if (dateInMs >= min && dateInMs <= max) {
+                        return true;
+                    }
+
+                    break;
+                case (DateRangeType.Specific):
```
kdinev: This would fail if the provided specific date has a time portion.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -996,8 +1129,90 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
...
+                    for (const specificDateInMs of datesInMs) {
+                        if (dateInMs === specificDateInMs) {
+                            return true;
+                        }
+                    }
+
+                    break;
+                case (DateRangeType.Weekdays):
+                    const day = date.getDay();
+                    if (day !== 0 && day !== 6) {
```
kdinev: How about `day % 6 !== 0`

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -996,8 +1129,90 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
...
+                        return true;
+                    }
+
+                    break;
+                case (DateRangeType.Weekends):
+                    const weekday = date.getDay();
+                    if (weekday === 0 || weekday === 6) {
```
kdinev: Same here `weekday % 6 === 0`

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -996,8 +1129,90 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
...
+                    if (dateInMs < dRanges[0].getTime()) {
+                        return true;
+                    }
+
+                    break;
+                case (DateRangeType.Between):
+                    const dRange = dRanges.map(d => d.getTime());
+                    const min = Math.min(dRange[0], dRange[1]);
+                    const max = Math.max(dRange[0], dRange[1]);
+                    if (dateInMs >= min && dateInMs <= max) {
```
kdinev: This would sometimes fail if the provided `min` and `max` dates have a time portion.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
tiliev: The method `selectDateFromClient` is called per each clicked date from the UI. The disabled dates are not clickable, that's why for `single` and `multi` selection we call the API method directly. But if the selected dates are in a range, the disabled dates from that range should no be selected. Please elaborate, if this does not answer you question.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
tiliev: I strip the time portion from each date, passed from the client. I compare dates only. At the beginning of the switch you can find the code `descriptor.dateRange.map(r => new Date(r.getFullYear(), r.getMonth(), r.getDate()))`, which does that.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
tiliev: Updated.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
tiliev: Updated.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
tiliev: I strip the time portion of the dates.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
zdrawku: Occasions, occasionDates, otherOccasions, otherDates, observanceDates

tiliev: IMHO `specialDates` is an excellent name.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
kdinev: So, calling this as an API method will allow me to select disabled dates in the case of `single` and `multiple`, but will not allow me to select them in the case of `range`. It's inconsistent.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -37,6 +36,20 @@ export enum CalendarSelection {
     RANGE = 'range'
 }
 
+export enum DateRangeType {
```
kdinev: These should not belong to the calendar component. They should be put in `core`.

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -37,6 +36,20 @@ export enum CalendarSelection {
...
+    After,
+    Before,
+    Between,
+    Specific,
+    Weekdays,
+    Weekends
+}
+
+export interface DateRangeDescriptor {
```
kdinev: Also belongs to `core`

projects/igniteui-angular/src/lib/calendar/calendar.component.ts
kdinev: @tiliev I'm actually also starting to think that the name is good as is.

zdrawku: @kdinev, @tiliev, lets keep that name as it is. After making a little research it seems like a common usage for - Holidays, PTO days, National holidays, teambuildings, key days and any other Day of the year with special meaning.

PR #2422, opened 2018-08-15
 Improve performance in scenarios where row template changes on scroll.

### LINKED ISSUE #2421: Vertical scrolling performance is slower when grouping is applied.
### Description  
Scrolling performance is slower when there is grouping applied. 

The issue is most noticeable when each row has a group row so that each scroll causes a template change (from data row to group row and vice versa).
### Steps to reproduce  

1.  Open the dev Grid GroupBy sample.
2. Group by "CompanyName" so that each rec generates a group.
3. Start scrolling vertically and compare the performance.

### Result  
Performance is slower compared to grid with no group rows.

### Expected result  
Performance should be similar to grid with no group rows.

### Attachments  
Performance profile of scrolling with groups (worst case scenario) on Chrome for reference:
[Profile-gb-branch-6.1.x.zip](https://github.com/IgniteUI/igniteui-angular/files/2289722/Profile-gb-branch-6.1.x.zip)

Performance profile without groups:
[Profile-noGroups.zip](https://github.com/IgniteUI/igniteui-angular/files/2289917/Profile-noGroups.zip)

MayaKirova:
Caching and re-using views via a custom template outlet.

Closes #2421.

Performance profile of scrolling with groups (worst case scenario) on Chrome after the improvement:
[Profile-branch-improve-gb-performance.zip](https://github.com/IgniteUI/igniteui-angular/files/2289934/Profile-branch-improve-gb-performance.zip)


MayaKirova:
@skrustev 
In general the problem is that detached views detect their changes only after they are re-inserted in the view, hence any changes done on them while detached will trigger onChanges only after they are re-inserted, which is good (they’ll be checked just once when inserted = big perf. gain).
However there is some code added for resetting scroll position if horizontal container size changes:
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts#L1045 
Which gets triggered for cached rows after they are re-added in the view after grouping was applied (grouping changes widths since it applies an offset at the begging of each column), hence the scroll position resetting observed in the issue.
After discussing this with @ChronosSF , instead of forcing detection on detached views for this particular scenario we opted for just removing the code for resetting the scroll and persisting the scroll position.

So this fix changes the behavior when container resizing happens. Before it was resetting horizontal scroll to 0, now it will persist the scroll position.

Due to this change please test with % width with window resizing, pinning, hiding and other features that change the widths.
Particularly scenarios where the start index and/or chunk size would change as a result of the container size change.

For instance one such scenario that would be broken without the additional code:

1)	Run the sample column pinning dev sample (80% width, full screen, no navbar)
2)	Minimize the window
3)	Scroll to the right end of the grid
4)	Unpin ‘Contact Title’ column
5)	Maximize the window

What to look out for:
-	 missing columns or white spaces.
-	columns that are not fully visible (cut off).

Please test on different browsers.

ChronosSF:
Verified - looks good. I think you should make a PR to master and have these merged after successful reviews.

PR #2423, opened 2018-08-15
[DO NOT MERGE] Improve perceived scrolling performance by throttling/auditing scroll event.

MayaKirova:


ChronosSF:
@kdinev , @rkaraivanov - can you please test this and compare against no throttling so we can decide whether to implement this?

Personally, I don't like the throttling experience compared to the usual one. In Chrome it is relatively the same, except when you scroll down a lot of data the perceived result is just a tad more sluggishness in the response. In Edge the throttling solution is a bit worse again because it looks like scrolling "skips" consistently.

So, imo, we should leave the scroll as is.

PR #2425, opened 2018-08-15
IgxList templates for left and right panning

### LINKED ISSUE #728: igxList Templates for left and right panning.
Create list item's templates to support various actions when panning left and right.

rmkrmk:
Closes #728  

Add a working version of the list items' left and right panning templates and functionality.


tachojelev:
**Issue:**
When panning a list item to the right, the automatic panning gets triggered even if the mouse pointer has not reached the 50% default threshold.

panning

projects/igniteui-angular/src/lib/list/list-item.component.ts
```diff
@@ -167,11 +188,14 @@ export class IgxListItemComponent implements IListChild {
         const isPanningToLeft = ev.deltaX < 0;
 
         if (isPanningToLeft && this.isTrue(this.list.allowLeftPanning)) {
-            this.left = Math.max(this.maxLeft, ev.deltaX);
+            this.showLeftPanTemplate();
+            this.contentLeft = Math.max(this.maxLeft, ev.deltaX);
```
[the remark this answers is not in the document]
rmkrmk: The issue is fixed and the code is refactored so the `contentLeft` property does not exist anymore.

projects/igniteui-angular/src/lib/list/list.component.spec.ts
```diff
@@ -445,6 +446,42 @@ describe('List', () => {
         });
     }, 5000);
 
+    it('check the panLeftTemplate is visible when left-panning a list item.', () => {
```
gedinakova: Add a test for panning a header item.
Add tests for verifying the templates are not visible after releasing the item.

projects/igniteui-angular/src/lib/core/styles/components/list/_list-theme.scss
```diff
@@ -36,7 +36,7 @@
         header-text-color: igx-color($palette, 'secondary'),
         item-background: #fff,
         item-text-color: igx-color($palette, 'grays', 800),
-        item-background-active: igx-color($palette, 'grays', 100),
+        item-background-active: #fff,
```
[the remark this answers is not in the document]
rmkrmk: Until now a moving/active list-item had a semi-transparent background (maybe added for some visual effect). After adding the left/right panning feature the active list-item should not be (semi) transparent because when the user starts dragging a list item, the whole panning template under it will be visible and this is not an effect we want to have. I added a hard-coded solid white background just like the list and the list-item are having.

projects/igniteui-angular/src/lib/list/list.component.spec.ts
rmkrmk: More tests were added.

projects/igniteui-angular/src/lib/list/list.component.ts
```diff
@@ -32,6 +37,12 @@ export interface IListItemClickEventArgs {
     event: Event;
 }
 
+export interface IListItemPanningEventArgs {
+    item: IgxListItemComponent;
+    direction: IgxListPanState;
+    cancel: boolean;
```
gedinakova: We should probably consider renaming the "cancel" property to something like "keepItem"? Because what it does is to tell the list to bring the panned item back if view after the event is fired.

projects/igniteui-angular/src/lib/list/list.component.spec.ts
```diff
@@ -445,6 +446,95 @@ describe('List', () => {
...
+        const headers = list.headers;
+        let header;
+        for (let i = 0; i < headers.length; i++) {
+            header = headers[i] as IgxListItemComponent;
+            expect(header.leftPanningTemplateElement).toBeUndefined();
+            expect(header.rightPanningTemplateElement).toBeUndefined();
+            expect(header.contentElement).toBe(null);
+        }
+    });
+
```
gedinakova: Add tests verifying that setting the ILIstItemPanningEventArgs cancel property to true brings the item back in view.

rmkrmk: Requested tests have been added.

rmkrmk:
1. The panning feature is updated to use translateX.
2. The CSS class name is still under debate.
3. The list is not aware regarding the data source collection, removing functionality is implemented in a dev. sample.

projects/igniteui-angular/src/lib/list/list.component.ts
rmkrmk: Property renamed

rmkrmk:
The CSS classes are updated as suggested by @simeonoff

tachojelev:
**Regression issue:**

The list items no longer have gray active background when they are clicked:
image

projects/igniteui-angular/src/lib/list/list.component.spec.ts
```diff
@@ -445,6 +446,129 @@ describe('List', () => {
...
+        fixture.detectChanges();
+
+        const firstItem = list.items[0] as IgxListItemComponent;
+        const leftPanTmpl = firstItem.leftPanningTemplateElement;
+        const rightPanTmpl = firstItem.rightPanningTemplateElement;
+        const itemNativeElements = fixture.debugElement.queryAll(By.css('igx-list-item'));
+
+        /* Click and drag item left */
+        clickAndDrag(itemNativeElements[1], -0.3);
+        fixture.detectChanges();
```
gedinakova: Remove all fixture.detectChanges() after clickAndDrag(), panItem() and tick() as they are not necessary.

rmkrmk: All occurrences removed as suggested.

tachojelev:
Gray background of the active list item is restored. - **Verified**

PR #2427, opened 2018-08-15
fix(filter-ui): #2411 filter dialog is not displayed with Z-index igx…

### LINKED ISSUE #2411: igx-grid - Filter dialog is not displayed if igx-grid surrounded by div with set Z-index.
### Description  
Filter dialog is not displayed if igx-grid surrounded by div with set Z-index.

 * browser: Chrome, Edge

```
<div style="z-index: 1;background-color: red;position: relative; width: 1000px; height: 700px;" >
    <igx-grid #grid1 [data]="localData"   width="960px" height="550px">
      <igx-column field="EmployeeID" filterable="true" dataType="string"></igx-column>
    </igx-grid>
</div>
```

### Result  
Filter dialog is hide behind.

### Expected result  
Filter dialog is show front.

### Attachments  
image

npaunov:
Closes #2411

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/grid-filtering.component.html
+++ b/projects/igniteui-angular/src/lib/grid/grid-filtering.component.html
@@ -11,8 +11,7 @@
     (onOpened)="focusFirstInput()"
     (onClosed)="refresh()"
     #directive="toggle"
-    class="igx-filtering__options"
-    #toggle="toggle">
+    class="igx-filtering__options">
 
     <igx-grid-filter-expression
         name="firstExpr"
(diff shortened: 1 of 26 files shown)


projects/igniteui-angular/src/lib/grid/grid-filtering.component.html
```diff
@@ -11,7 +11,8 @@
     (onOpened)="focusFirstInput()"
     (onClosed)="refresh()"
     #directive="toggle"
-    class="igx-filtering__options">
+    class="igx-filtering__options"
+    #toggle="toggle">
```
sstoyanovIG: This identifier doesn't seem to be needed

PR #2448, opened 2018-08-20
Tabs and TimePicker tests reworked

sboykova:


projects/igniteui-angular/src/lib/tabs/tabs.component.spec.ts
```diff
@@ -13,12 +14,13 @@ describe('IgxTabs', () => {
             .compileComponents();
     }));
 
-    it('should initialize igx-tabs, igx-tabs-group and igx-tab-item', () => {
+    it('should initialize igx-tabs, igx-tabs-group and igx-tab-item', (async () => {
```
bkulov: I think most of the tests are synchronous. It's better to use fakeAsync with tick().

PR #2464, opened 2018-08-21
fix(chip): igxChip onSelection event doesn't have argument types #2461

### LINKED ISSUE #2461: igxChip onSelection event doesn't have argument types
### Description  
The igxChip component `onSelection` event has argument type of `any`, but the event is emitted with the following arguments: 

```typescript
        const onSelectArgs = {
            owner: this,
            nextStatus: false,
            cancel: false
        };
```

The selection event args should have an interface type describing them. 

 * browser: N/A

The Chip is also missing a README and doesn't have an API table in docs.

alexalex972:
Closes #2461 

- [x] Interface
- [x] README
- [x] Docs


mpavlinov:
@alexalex972 Please update the Changelog as well.

projects/igniteui-angular/src/lib/chips/chip.component.ts
```diff
@@ -528,3 +530,9 @@ export class IgxChipComponent implements AfterViewInit {
     }
     // End chip igxDrop behaviour
 }
+
+export interface ChipSelectEventArgs {
```
skrustev: Interfaces should start with `I`, for example `ISelectChipEventArgs`

projects/igniteui-angular/src/lib/chips/chip.component.ts
```diff
@@ -528,3 +530,9 @@ export class IgxChipComponent implements AfterViewInit {
...
+    owner: IgxChipComponent;
+    cancel: boolean;
+    nextStatus: boolean;
```
skrustev: Could you rename `nextStatus` to `selected` since the second one is probably easier to understand.

projects/igniteui-angular/src/lib/chips/chip.component.ts
```diff
@@ -311,13 +318,13 @@ export class IgxChipComponent implements AfterViewInit {
...
-        const onSelectArgs = {
+        const onSelectArgs: IChipSelectEventArgs = {
             owner: this,
-            nextStatus: false,
+            selected: false,
             cancel: false
         };
         if (newValue && !this._selected) {
-            onSelectArgs.nextStatus = true;
+            onSelectArgs.selected = true;
```
mpavlinov: @alexalex972 This is a breaking change, so you should document it in the changelog as breaking change.

PR #2480, opened 2018-08-22
Fix filtering with japanese symbols

### LINKED ISSUE #2316: Japanese character is redundantly inserted into textbox on filter dialog on Safari
### Description  
As Japanese character is being typed, typed characters are inserted repeatedly.

 * browser: Safari on iOS / OS X

### Steps to reproduce  
1. Build the attached sample with the following command:
 * ng build --base-href "/my-app1/" --prod
2. Host the attached sample on IIS (or serve it from your machine by `ng serve --host 10.14.1.129` replacing the ip address with yours)
3. Open the sample on iOS Safari
4. Open the filter dialog on "name" field
5. Type 'あ' (a on English keyboard)
6. Type 'い' (i on English keyboard)

### Result  
'ああい' is inserted into the filter textbox

### Expected result  
'あい' is inserted into the filter textbox.

### Attachments  
[my-app1.zip](https://github.com/IgniteUI/igniteui-angular/files/2261279/my-app1.zip)

### Note
iOS version is 11.4.1 (15G77). This is not reproduced on Chrome on Windows 10.

bkulov:
Closes #2316.


npaunov:
The filtering is not performed on typing. For example if you type "あ" the grid's rows are not filtered out, you have to press "Enter" key to see the result.

bkulov:
@npaunov The filtering is updated when the value of the input is changed. In case of Japanese, Arab, etc. alphabets this happens when the Composion End event is received. This usually happens when you press Enter key.

For more details see the link bellow:
https://developer.mozilla.org/en-US/docs/Web/API/CompositionEvent

PR #2490, opened 2018-08-24
fix(chip): Replace any with the interface for the event #2481

### LINKED ISSUE #2481: Replace EventEmitter<any> with the respective interface for the event
### Description  
In the source there are several places where "EventEmitter<any>" is used.

 * browser: *

### Steps to reproduce  

1. Search the source code for "EventEmitter\<any\>"
2. Observe the result

### Result  

image

### Expected result  
Introduce interfaces for the respective events and use them in the event emitters.

### Attachments  

alexalex972:
Closes #2481


projects/igniteui-angular/src/lib/chips/chips-area.component.ts
```diff
@@ -24,13 +24,32 @@ import { CommonModule } from '@angular/common';
...
+    IChipEnterDragAreaEventArgs,
+    IBaseChipEventArgs
+} from './chip.component';
 import { IgxDragDropModule } from '../directives/dragdrop/dragdrop.directive';
 import { IgxButtonModule } from '../directives/button/button.directive';
 import { IgxAvatarModule } from '../avatar/avatar.component';
 import { IgxIconModule } from '../icon';
 import { IgxConnectorDirective } from './connector.directive';
 
+export interface IBaseChipAreaEventArgs {
```
skrustev: The name of the component is `ChipsAreaComponent`, I think the names should be consistent so there is not confusion.
In this case would be `IBaseChipsAreaEventArgs`

projects/igniteui-angular/src/lib/chips/chips-area.component.ts
```diff
@@ -24,13 +24,32 @@ import { CommonModule } from '@angular/common';
...
+    owner: IgxChipsAreaComponent;
+}
+
+export interface IChipReorderEventArgs {
```
skrustev: Same as above

projects/igniteui-angular/src/lib/chips/chips-area.component.ts
```diff
@@ -24,13 +24,32 @@ import { CommonModule } from '@angular/common';
...
+    chipsArray: IgxChipComponent[];
+    isValid: boolean;
+}
+
+export interface IChipAreaSelectEventArgs extends IBaseChipAreaEventArgs {
```
skrustev: Again same as above

PR #2498, opened 2018-08-27
Fix for 2467

### LINKED ISSUE #2467: igxChip leaves an empty space after text
### Description  
The igxChip leaves an empty space for remove button, even when there's no remove button.

 * browser: N/A

### Steps to reproduce  

1. Run `ig` from the `igniteui-cli`
2. Create a new angular app
3. Add a view with chip area and chips
```html
<igx-chips-area>
    <igx-chip *ngFor="let weekday of player.availability" [selectable]="playerIsUser()" (onSelection)="daySelected($event)">
        <span class='igx-chip__text'>{{ dayName(weekday.Day) }}</span>
    </igx-chip>
</igx-chips-area>
```

### Result  
Every chip has an extra blank space behind it.

### Expected result  
There should be no blank space.

### Attachments  
image

desig9stein:
Closes #2467   

I remove the margin after the text, but if we want to centre the text inside the chip we need to check if there is only text and add some class to make it centre.

This is what happens if I add text-align: center. It happens, because we have min-width on the chip. It's ok when we don't have icons, but if we have, the text should be left aligned, otherwise it looks odd.
https://drive.google.com/file/d/1RWk9QvJ_tyg-vWD8h7jGSBA6PVLCPa2s/view

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
@@ -336,8 +336,7 @@
     %igx-chip__text {
         color: --var($theme, 'text-color');
         font-size: rem(12px);
-        flex: 1 0 0;
-        text-align: center;
+        flex: 1 0 0%;
 
         @include ellipsis();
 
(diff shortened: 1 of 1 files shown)


mpavlinov:
Note: The issue is fixed for the scenario where the chip text is larger than the min-width but smaller than the max-width.
For the scenario where the text is smaller than the min-width the user should apply additional CSS if he wants to center the text in the chip.

mpavlinov:
@desig9stein Please make PR for branch 6.2.x as well.

PR #2501, opened 2018-08-27
Feat - IgxExpansionPanel - Initial Implementation

### LINKED ISSUE #307: Expansion Panel - new component
image

image

The collapsible MVP component must have Header and Container that could be expanded or collapsed. The Header is always visible, The component could be put in list, card, etc. 

Accordion-like behavior could be implemented programmatically by developers if preferred.

### API

|  *Methods* | *Properties*| *Events* | 
|---|---|---|
| Collapse | Collapsed| OnCollapsed | 
| Expand | | OnExpanded |

ViktorSlavov:
Closes #307.  

[Issue](https://github.com/IgniteUI/igniteui-angular/issues/307)
[Spec](https://github.com/IgniteUI/igniteui-angular/wiki/Expansion-Panel-Specification)

Do not merge this PR as the work on this feature is not complete!


projects/igniteui-angular/src/lib/expansion-panel/expansion-panel-header.component.html
```diff
@@ -0,0 +1,16 @@
+
+<div tabindex="0" role = "button" [attr.aria-disabled] = "disabled" [attr.aria-expanded] = "isExpanded" [attr.aria-controls] = "controls">
+<ng-template #iconContainer>
+    <ng-content *ngIf="iconTemplate" select="igx-expansion-panel-icon"></ng-content>
+    <igx-icon *ngIf="!iconTemplate" fontSet="material" [name]="panel.collapsed? 'expand_more':'expand_less'"></igx-icon>
```
simeonoff: We could deprecate the name input property on the IgxIcon component. Avoid using it for setting ligature icons. Instead specify the name between the opening and closing tags of the igx-icon selector.

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel-header.component.html
```diff
@@ -0,0 +1,16 @@
...
+</ng-template>
+<ng-container *ngIf="iconPosition === 'left'">
+    <ng-container *ngTemplateOutlet="iconContainer"></ng-container>
+</ng-container>
+<ng-content select="igx-expansion-panel-title"></ng-content>
+<ng-content select="igx-expansion-panel-description"></ng-content>
+<ng-content></ng-content>
+<ng-container *ngIf="iconPosition === 'right'">
```
simeonoff: Let's handle the icon positioning using CSS by changing the class modifier on the icon container? For instance if we could have two classes for the icon container - `igx-expansion-panel__header-icon--start` and `igx-expansion-panel__header-icon--end`. I don't see the point of having two conditions that output the same result regardless of the provided input value they check against - in this case the icon position. Also, if we have to still check for the icon position inside the template, wouldn't it be better to use the enumeration members, instead of hard-coded string values to do the checks?

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel-header.component.html
```diff
@@ -0,0 +1,16 @@
...
+<div tabindex="0" role = "button" [attr.aria-disabled] = "disabled" [attr.aria-expanded] = "isExpanded" [attr.aria-controls] = "controls">
+<ng-template #iconContainer>
+    <ng-content *ngIf="iconTemplate" select="igx-expansion-panel-icon"></ng-content>
+    <igx-icon *ngIf="!iconTemplate" fontSet="material" [name]="panel.collapsed? 'expand_more':'expand_less'"></igx-icon>
+</ng-template>
+<ng-container *ngIf="iconPosition === 'left'">
+    <ng-container *ngTemplateOutlet="iconContainer"></ng-container>
+</ng-container>
+<ng-content select="igx-expansion-panel-title"></ng-content>
+<ng-content select="igx-expansion-panel-description"></ng-content>
```
simeonoff: Can you also add CSS classes for the header title and description? For instance `igx-expansion-panel__header-title` and `igx-expansion-panel__header-description`.

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel-header.component.html
```diff
@@ -0,0 +1,14 @@
+<div class="igx-expansion-panel__header-inner" tabindex="0" role="button" [attr.id]="id"
+[attr.aria-disabled]="disabled" [attr.aria-expanded]="isExpanded" [attr.aria-controls]="controls">
+    <div class="igx-expansion-panel__title-wrapper">
+        <ng-content select="igx-expansion-panel-title"></ng-content>
+        <ng-content select="igx-expansion-panel-description"></ng-content>
+    </div>
+    <!--<ng-content></ng-content>-->
```
damyanpetev: We should probably keep this one to allow general content to be placed in the header (media, etc). Maybe wrap it in a container so flex order can be applied on the custom content as a whole? 
I don't mind it being before the icon, though the `icon--end` class would be order 2 instead of 1 with the extra container.

My only concern is the extra container would be present regardless of content being provided. Maybe force it with a selector rather than container?

ViktorSlavov:
@simeonoff @desig9stein 
Could you please removed the default padding in place for the body class (`.igx-expansion-panel__body`)?
As discussed, it interferes with the open/close animation and could also be a hindrance is some use scenarios.

PR #2532, opened 2018-08-30
Deselect date methods (master)

### LINKED ISSUE #2424: Add deselect method to igxCalendar
### Description  
Currently we have selectDate method which is selecting date/dates based on the `selection` type. We need to implement similar method for date deselect.

Consider the method to be executed with:
- empty value - this will deselect all selected dates.
- date object - this will deselect the passed date.
- array of date objects - this will deselect all passed dates.

tiliev:
Closes #2424.  

Additional information related to this pull request: added to `igxCalendar` and `igxDatePicker`.


ddincheva:
In addition to the @zdrawku comment:
- in the console should not be thrown error when the calendar is in single selection mode and try to deselect date which has not been selected;

screenshot_10

- when the calendar is in multi selection mode and select for an example a three dates, and then deselect the same dates all of them should be deselected;

zdrawku:
@tiliev we can consider your suggestions for future enhancements, although for the time being we are going to make both select and deselect date consistent, by allowing them both to accept array of dates (when using range selectino type), and based on the min and max date in the array, to select/deselect the corresponding range of dates.

Issue #2516, opened 2018-08-29
All cells in the row enter in edit mode if igx-columns are recreated.


norikois:
### Description  
All cells in the row enter in edit mode if igx-columns are recreated.

 * igniteui-angular version: 6.1.4
 * browser: Chrome, Edge, Firefox, IE11
 * component: igx-grid

### Steps to reproduce  

1. Run the attached sample.
2. Click "Rerender" button.
3. Double click one of the cells to enter edit mode.

### Result  
All the cells in the row enter edit mode.

### Expected result  
The only clicked cell should enter edit mode.

### Attachments  
[my-app2.zip](https://github.com/IgniteUI/igniteui-angular/files/2330763/my-app2.zip)

PR #2544, opened 2018-09-03
fix(grid): refactor column hide logic to preserve column width (6.2.x) #2521

### LINKED ISSUE #2521: Column width is reset to min width after showing
### Description  
If a grid column width is set to 200px and column hiding is performed, after showing the column its width is set to 136px (grid's min column width) instead to the initial value - 200px.

### Steps to reproduce  

1. Open Grid Cell Editing sample in product demos
2. In the first grid (with PK) scroll to `OrderDate` column 
3. Click Hide/show `OrderDate` toggle button in order to **Hide** the column
4. Click Hide/show `OrderDate` toggle button in order to **Show** the column

### Result  
Column width is 136px

### Expected result  
Column width should remain 200px

image

SAndreeva:
Closes #2521  .  

**Additional information related to this pull request:** the fix might affect the behavior reported in #2249, so the latter should be double checked!

### DIFF
--- a/projects/igniteui-angular/src/lib/grid/column.component.ts
+++ b/projects/igniteui-angular/src/lib/grid/column.component.ts
@@ -193,9 +193,7 @@ export class IgxColumnComponent implements AfterContentInit {
                     this.grid.summariesHeight = 0;
                 }
 
-                if (!value) {
-                    this.grid.columnsWithNoSetWidths.push(this);
-                } else if (this.grid.columnsWithNoSetWidths.indexOf(this) !== -1) {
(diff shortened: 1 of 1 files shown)


DDavidkov:
The fix that caused this issue wasn't correct to begin with. I have scrapped my original approach for a new one in this [PR](https://github.com/IgniteUI/igniteui-angular/pull/2558). It should fix the reported issue and [this one](https://github.com/IgniteUI/igniteui-angular/issues/2516). Because of this the columnsWithNoSetWidths collection will no longer exist and this PR won't be necessary.

PR #2577, opened 2018-09-10
On item click check if there is focused item

wnvko:
We were wrongly assuming that there is always focused item on item click, and were calling elementRef over the focused item. However, if there is no focused item, error was thrown.

Now we have check if the focused item actually exist - if not we are skipping calling of elementRef.


projects/igniteui-angular/src/lib/drop-down/drop-down-item.component.ts
```diff
@@ -205,7 +205,7 @@ export class IgxDropDownItemBase {
             if (focusedItem) {
```
damyanpetev: Shouldn't this respect `allowItemsFocus` and not call the native focus if not enabled?

damyanpetev: Same applies for the `isFocused` setter TBH.

wnvko: fixed

PR #2585, opened 2018-09-12
Update allowItemsFocus usage and description

Lipata:
More changes related to the allowItemsFocus change of the default value, are needed.


projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -155,8 +155,11 @@ export class IgxDropDownBase implements OnInit, IToggleView {
     }
 
     /**
-     * Gets/sets whether items will be able to take focus. If set to true, default value,
-     * user will be able to use keyboard navigation.
+     * Gets/sets whether items will be able to take focus.
+     * Default value is false and during keyboard navigation
+     * the focus will stay at the element that the igxDropDownItemNavigation directive is applied on.
+     * If set to true drop down items will get focus, during keyboard navigation, and the element that the
+     * igxDropDownItemNavigation directive is applied on, will loose focus.
```
damyanpetev: Both `focus will stay on elem w/ igxDropDownItemNavigation` and `igxDropDownItemNavigation will loose focus` and not guaranteed scenarios and it's not relevant here anyway. Focused and `igxDropDownItemNavigation` element are not necessary the same element at all.

damyanpetev: Alternative:
"Gets/sets whether items take focus. Disabled by default.
When enabled, drop down items gain tab index and are focused when active - this includes activating the selected item when opening the drop down and moving with keyboard navigation.

Note: Keep that focus shift in mind when using the `igxDropDownItemNavigation` directive and ensure it's placed either on each focusable item or a common ancestor to allow it to handle keyboard events."

Issue #2548, opened 2018-09-04
The onOpened event of the IgxToggle is not fired when openAnimation is disabled


tachojelev:
### Description  
The **onOpened** event of the **IgxToggle** is not fired when openAnimation is disabled 

 * igniteui-angular version: 6.1.x, 6.2.x
 * browser: all

**NOTE:**
- If both **onOpening** and **onOpened** events are handled, the issue is not reproduced.

### Steps to reproduce  

1. Open the **Toggle Demo** in stackblitz:
https://www.infragistics.com/products/ignite-ui-angular/angular/components/toggle.html
2. Subscribe for the onOpened event, for example:
```html
   <div class="toggle-content" igxToggle #toggleRef="toggle"
    (onOpened)="test()">
```
```typescript
    test() {
      console.log("opened has fired");
    }
```
3. Disable the openAnimation:
```typescript
    public _positionSettings = {
        ...
        openAnimation: null
    };
```
4. Click the toggleAction button.

### Result  
The onOpened event has **not** fired.

### Expected result  
The onOpened event has fired.

PR #2587, opened 2018-09-13
Change the way animation works in igxOverlay

### LINKED ISSUE #2551: Cannot hide an element due to unfinished animation in the IgxOverlay
### Description  
Cannot hide an element due to unfinished animation in the IgxOverlay.

 * browser: all

### Steps to reproduce  

1. Run the **"Toggle Overview"** sample from the **igniteui-angular-samples** repo.
2. Modify the **template** and the **ts** respectively as follows:
```html
...
  <article class="sample-column">
    <div>
        <button (click)="open()">open</button>
        <button (click)="close()">close</button>
    </div>
    <button class="toggle-button" #button igxButton="raised" (click)="toggle()">Toggle</button>
...
```

```typescript
...
    open() {
        this._overlaySettings.positionStrategy.settings.target = this.igxButton.nativeElement;
        this.igxToggle.open(this._overlaySettings);
        console.log("OPEN invoked");
    }
    close() {
        this.igxToggle.close();
        console.log("CLOSE invoked");
    }
...
```
3. Click the **open** button and then immediately the **close** button. (do not wait for the open animation to finish)

### Result  
Even though close was invoked (as seen in the console), the box element is still visible.

### Expected result  
A way should be provided, so the user can both terminate the animation before it is finished and hide the element immediately.

### Attachments  

toggleissue

wnvko:
Closes #2551; #2548 

Up to now if one calls hide method of igxOverlay, while opening animation is in progress, the hide just exits and leave opening animation to finish. Same was happening of one calls show method while close animation was in progress.

Now if one calls hide method it will stop opening animation, if any, and will start hide animation from the point where open animation has stopped. Same will happen if one calls show while closing animation is in progress.

Additionally we are firing now onAnimation event just before any animation has started.


projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.ts
```diff
@@ -229,28 +227,29 @@ export class IgxToggleDirective implements IToggleView, OnInit, OnDestroy {
...
-                this._overlayClosingSub.unsubscribe();
-            }
-        }
+        this.unsubscribe();
     }
 
     private overlayClosed = () => {
         this._collapsed = true;
         this.cdr.detectChanges();
+        this._overlayId = undefined;
```
bazal4o: If you need to clear the overlayId just use -> delete this._overlayId

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -66,35 +69,45 @@ export class IgxOverlayService {
     /**
      * Shows the provided component.
      */
-    show(component: ElementRef | Type<{}>, settings?: OverlaySettings): string {
-        const id: string = (this._componentId++).toString();
-        settings = Object.assign({}, this._defaultSettings, settings);
+    show(component: ElementRef | Type<{}>, settings?: OverlaySettings, id?: string): string {
+        let info: OverlayInfo;
+        if (id) {
```
bazal4o: If we pass 0 as Id this will evaluate as false - is that expected? If not use if (id !== undefined)

wnvko: fixed

projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.ts
wnvko: fixed

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -66,35 +69,45 @@ export class IgxOverlayService {
...
-        const id: string = (this._componentId++).toString();
-        settings = Object.assign({}, this._defaultSettings, settings);
+    show(component: ElementRef | Type<{}>, settings?: OverlaySettings, id?: string): string {
+        let info: OverlayInfo;
+        if (id !== undefined) {
+            info = this.getOverlayById(id);
+        }
 
-        const info = this.getOverlayInfo(component);
+        if (!info) {
```
damyanpetev: Does this mean that if I pass a wrong ID with ref to element already in overlay, it'll still make a new one and move it there? Not a biggie, just curious.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -66,35 +69,45 @@ export class IgxOverlayService {
     /**
      * Shows the provided component.
      */
-    show(component: ElementRef | Type<{}>, settings?: OverlaySettings): string {
-        const id: string = (this._componentId++).toString();
-        settings = Object.assign({}, this._defaultSettings, settings);
+    show(component: ElementRef | Type<{}>, settings?: OverlaySettings, id?: string): string {
```
damyanpetev: Alright, alternative idea for the ID. Feels way more natural to go `show(id, options)` if you already have the ID, no?
And since you'll be making a check for the extra param anyway, you might as well be checking it for an overload.
Try [this on the Typescript playground](https://www.typescriptlang.org/play/index.html#src=class%20OverlaySettings%20%7B%20%7D%0D%0Aclass%20ElementRef%20%7B%20%7D%0D%0Aclass%20Type<T>%20%7B%7D%0D%0A%0D%0Aclass%20Test%20%7B%0D%0A%20%20%20%20show(id%3A%20string%2C%20settings%3F%3A%20OverlaySettings)%3A%20string%3B%0D%0A%20%20%20%20show(component%3A%20ElementRef%20%7C%20Type<%7B%7D>%2C%20settings%3F%3A%20OverlaySettings)%3A%20string%3B%0D%0A%20%20%20%20show(compOrId%3A%20ElementRef%20%7C%20Type<%7B%7D>%20%7C%20string%2C%20settings%3F%3A%20OverlaySettings)%3A%20string%20%7B%0D%0A%20%20%20%20%20%20%20%20if%20(typeof%20compOrId%20%3D%3D%3D%20"string")%20%7B%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20console.log('string%3A'%2C%20compOrId)%3B%0D%0A%20%20%20%20%20%20%20%20%7D%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20console.log('Ref%2FType%3A'%2C%20compOrId)%3B%0D%0A%20%20%20%20%20%20%20%20%7D%0D%0A%20%20%20%20%20%20%20%20%0D%0A%20%20%20%20%20%20%20%20return%20"id"%3B%0D%0A%20%20%20%20%7D%0D%0A%7D%0D%0A%0D%0Alet%20test%20%3D%20new%20Test()%3B%0D%0Atest.show("text"%2C%20%7B%7D)%3B%0D%0Atest.show(new%20ElementRef())%3B) - seems to work super nice and the intellisense finds the overloads.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
wnvko: At this point we do not support moving an element from overlay to overlay to overlay and so on. And yes it will move the element form one overlay to another

PR #2589, opened 2018-09-13
Changing the onGroupingDone event name and arguments

### LINKED ISSUE #2447: ixgGrid GroupBy feature should have the ability to hide column when grouped
### Description  
As a developer I want to hide my columns when grouped and show them again when ungrouped
Currently there is not property on the grid or column to say hidewhengrouped

There is a onGroupedDone event that would be viable to have the developer do this himself, but the eventArgs is an array of the currently sorted columns. 
The problem lies with ungrouping, the previously grouped column is not in that list.
So to achieve this you need to retain previous sortingExpressions and see which column is missing. 

This adds a lot of unneeded burden on the developer to manage what should be a simple property setting to handle this commonly used behavior.

PavlovVasil:
Closes #2447  .  

### DIFF
--- a/src/app/grid-groupby/grid-groupby.sample.ts
+++ b/src/app/grid-groupby/grid-groupby.sample.ts
@@ -88,4 +88,7 @@ export class GridGroupBySampleComponent implements OnInit {
     getRowsList() {
         console.log(this.grid1.rowList);
     }
+    onGroupingChangedHandler(event){
+        console.log(event)
+    }
 }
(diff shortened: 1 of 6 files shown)


src/app/grid-groupby/grid-groupby.sample.ts
```diff
@@ -88,4 +88,7 @@ export class GridGroupBySampleComponent implements OnInit {
     getRowsList() {
         console.log(this.grid1.rowList);
     }
+    onGroupingChangedHandler(event){
+        console.log(event)
+    }
```
MartinKamenov: Missing ; after console.log()

MartinKamenov:
@PavlovVasil The event is triggered when you groupBy same columns that have been grouped previously. The name of the event onGroupingChanged is a misleading one in this case.
#### How to test:
Press twice the "Group 3 Columns" button in the dev group by sample.
The second time the same column are grouped the event shouldn't be fired.

ChronosSF:
You should also update the CHANGELOG file.

PR #2596, opened 2018-09-17
Add text template to the circularbar

### LINKED ISSUE #2370: Circular bar text is clipped
### Description  
When the text value is longer than the circular bar width, the text is clipped and this cannot be changed so easily. Should we leave it as it is, or we need to change this? By my opinion using of text-overflow: ellipsis could improve this, although it seems that it is not working properly with SVG text element:

.progress-circular__text, .progress-circular__text--hidden
{
text-overflow: ellipsis;
}
ViewEncapsulation is set to None.

Result:
image

nrobakova:
Closes #2370 .  

Add textTemplate property to the circular bar in order user to can set text to be displayed as he want


projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
```diff
@@ -455,6 +456,24 @@ export class IgxCircularProgressBarComponent extends BaseProgress {
...
+     * ```
+     * ```typescript
+     * @ViewChild("'textTemplate'", {read: TemplateRef })
+     * public textTemplate: TemplateRef<any>;
+     * this.circularBar.textTemplate = this.textTemplate;
+     * ```
+     * @memberof IgxCircularProgressBarComponent
+     */
+    @Input()
+    public textTemplate: TemplateRef<any>;
```
rkaraivanov: The `@Input` is great, however it would be even better to be able to instantiate the template directly inside the `igx-circular-bar` tag. For example:
```html
<igx-circular-bar>
    <ng-template let-value>…</ng-template>
</igx-circular-bar>
```

projects/igniteui-angular/src/lib/progressbar/templates/circular-bar.component.html
```diff
@@ -2,7 +2,12 @@
     viewBox="0 0 100 100" preserveAspectRatio="none" role="progressbar" aria-valuemin="0" [attr.aria-valuemax]="max" [attr.aria-valuenow]="value">
     <circle class="progress-circular__innercircle" cx="50" cy="50" r="46" />
     <circle #circle class="progress-circular__circle" cx="50" cy="50" r="46" />
-    <text #text class="progress-circular__text" [class.progress-circular__text--hidden]="!textVisibility" id="myTimer" text-anchor="middle" x="50" y="60">
-        {{textContent ? textContent: valueInPercent + '%'}}
+    <text #text [class.progress-circular__text--hidden]="!textVisibility" id="myTimer" text-anchor="middle" x="50" y="60">
+        <ng-container *ngTemplateOutlet="textTemplate ? textTemplate : defaultTextTemplate;">
```
rkaraivanov: The passed user-defined template should get a context with the value of the progress bar.
Something along the lines of `{ $implicit: this.valueInPercent }` so the user can bind on it in the passed template.

nrobakova:
@rkaraivanov I have update the text template to can be instantiate under the circular bar

PR #2597, opened 2018-09-17
Add text template to circular bar 6.1.x

### LINKED ISSUE #2370: Circular bar text is clipped

nrobakova:
Closes #2370 .  

Add text template to the circular bar to can user manipulate text


nrobakova:
@rkaraivanov I have update the text template to can be instantiate under the circular bar

PR #2604, opened 2018-09-18
Fix - Selection - Undefined values should not be accepted as ids

### LINKED ISSUE #2581: IgxSelectionAPIService allows to add items with id which is undefined
### Description  

### Expected result  

DiyanDimitrov:
Closes #2581.


projects/igniteui-angular/src/lib/core/selection.ts
```diff
@@ -19,7 +19,9 @@ export class IgxSelectionAPIService {
      * @param newSelection The new component selection to be set.
      */
     public set(componentID: string, newSelection: Set<any>) {
-        this.selection.set(componentID, newSelection);
+        if (componentID) {
```
kdinev: Shouldn't there be an exception if the component ID is `undefined`? To me this scenario is not valid.

PR #2606, opened 2018-09-18
Fix - Selection - Undefined values should not be accepted as ids

### LINKED ISSUE #2581: IgxSelectionAPIService allows to add items with id which is undefined

DiyanDimitrov:
Closes #2581.


projects/igniteui-angular/src/lib/core/selection.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { IgxSelectionAPIService } from './selection';
+
+describe('IgxSelectionAPIService', () => {
+    let service;
+    beforeEach(() => {
+        service = new IgxSelectionAPIService();
+    });
+
+    afterEach(() => {
```
gedinakova: The afterEach is not necessary.

Issue #2612, opened 2018-09-18
IgxChip raises onSelection before onRemove


kdinev:
### Description  
I have an `IgxChipComponent` and I'm bound to both `onRemove` and `onSelection` events. The `onRemove` is fired only after an `onSelection` is raised, even when I click directly on the remove button.

 * igniteui-angular version: 6.1.5
 * browser: N/A

### Steps to reproduce  

1. Open the chip dev demo
2. Click the remove button
3. Debug the outputs

### Result  
`onSelection` is raised

### Expected result  
Only `onRemove` should be raised.

PR #2630, opened 2018-09-19
Fix igxChip raises onSelection when clicking on remove button. 

skrustev:
#2612


mpavlinov:
@skrustev With keyboard navigation, when the focus is on the "Delete" button and pressing "Space" key the onSelection event is triggered.

PR #2643, opened 2018-09-20
Adding the "hideGroupedColumns" igxGrid input

### LINKED ISSUE #2447: ixgGrid GroupBy feature should have the ability to hide column when grouped

PavlovVasil:
Closes #2447 .  
When testing: 
Run the Group By grid demo - [http://localhost:4200/gridGroupBy](url)
There is a switch at the top of the page that toggles the new option. Clicking on in when some columns are grouped should apply the setting and hide/show them at runtime.


MayaKirova:
If initial grouping expressions are defined, for example: 

```
public ngOnInit(): void {
     this.groupingExpressions =  [
            {
                fieldName: 'CompanyName',
                dir: SortingDirection.Asc
            }
        ];
}
```

`<igx-grid [hideGroupedColumns]="hideGroupedColumns"  [groupingExpressions]='groupingExpressions' ...`

Then an error is thrown: 
```
GridGroupBySampleComponent.html:2 ERROR TypeError: Cannot read property 'find' of undefined
    at IgxGridComponent$$1.push../dist/igniteui-angular/fesm5/igniteui-angular.js.IgxGridComponent$$1.getColumnByName (igniteui-angular.js:36548)
    at igniteui-angular.js:35389
    at Array.forEach (<anonymous>)
    at IgxGridComponent$$1.set [as hideGroupedColumns] (igniteui-angular.js:35388)
    at updateProp (core.js:9457)
    at checkAndUpdateDirectiveInline (core.js:9236)
    at checkAndUpdateNodeInline (core.js:10515)
    at checkAndUpdateNode (core.js:10477)
    at debugCheckAndUpdateNode (core.js:11110)
    at debugCheckDirectivesFn (core.js:11070)
```

An error should not be thrown and the initial grouped column should be hidden/shown depending on the specified hideGroupedColumns value.

npaunov:
1. @PavlovVasil if there are several columns grouped and hidden beforehand, for example:
    public ngOnInit(): void {
        this.groupingExpressions =  [
            {
                fieldName: 'CompanyName', dir: SortingDirection.Asc
            },
            {
                fieldName: 'Country', dir: SortingDirection.Desc
            }
        ];
        this.hideGroupedColumns = true;

npaunov:
..

MayaKirova:
@npaunov 1. should be resolved with the latest commit. 
2. appears to be a styling issue related to flex styles in IE and should be logged separately as it is not related to this PR.

PR #2648, opened 2018-09-21
Fixing the onGroupingDone event #2645 #2644 #2711

### LINKED ISSUE #2644: Passing the same grouping expressions to the API multiple times raises the onGroupingDone event every time
### Description  
When the API is used to group columns and the same expressions get passed, the onGroupingDone event gets raised every time, although the grid state actually does not change.

 * browser: any

### Steps to reproduce  

1. Run the attached sample
2. Go to [http://localhost:4200/grid](url)
3. Click the "Group 3 columns" button several times
4. Observe the browser console - it would log the event every time it gets raised

### Result  
The event gets raised every time the button gets clicked, although the state of the grid does not change.

### Expected result  
The event should get raised only when the sorting expressions change.

### Attachments  

[GroupByAPI.zip](https://github.com/IgniteUI/igniteui-angular/files/2404288/GroupByAPI.zip)

### LINKED ISSUE #2645: The onGroupingDone event doesn't get raised when columns get ungrouped.
### Description  
The onGroupingDone event currently gets raised only when columns get grouped, but by specification it should be raised when columns get grouped **or ungrouped**. ~~This makes the name ambiguous - it should be changed to onGroupingChanged. As this is a breaking change, a new onGroupingChanged event would be added **in addition** to the old one - onGroupingDone should be deprecated in later product versions.~~

Additionally, the clearGrouping grid method should have an overload to accept an array of column names and ungroup them all at once. Currently it should get called separately for each column about to be ungrouped - this would raise the onGroupingDone multiple times, which in many cases would **not** be what the user wants.

 * browser: any

### Steps to reproduce  

1. Run the provided sample 
2. Open [http://localhost:4200/grid](url)
3. Group columns and upgroup them afterwards
4. Observe the browser console - it would log the onGroupingDone event. 

### Result  
The onGroupingDone event doesn't get raised when columns get ungrouped, only when they get grouped.

### Expected result  
By specification the event should get raised when columns get grouped or ungrouped.

### Attachments  

[GroupByAPI.zip](https://github.com/IgniteUI/igniteui-angular/files/2404407/GroupByAPI.zip)

### LINKED ISSUE #2711: Ungrouping a column through the chip remove icon does not emit onGroupingDone output
### Description
Ungrouping a column through the chip remove icon does not emit onGroupingDone output
Go to https://stackblitz.com/edit/angular-uetflb
 * igniteui-angular version: 
 * browser: 

### Steps to reproduce  

1. Go to https://stackblitz.com/edit/angular-uetflb
2. Group a column and see the console --> onGroupingDone output is emitted
3. Ungroup a column by removing its chip via the X icon on the chip

### Result  
onGroupingDone output is NOT emitted

### Expected result  
onGroupingDone output to be emitted

PavlovVasil:
Closes #2645   
Closes #2644  
Closes #2711  
Notes when testing: 
1. Build and run the Group By demo - [http://localhost:4200/gridGroupBy](url)
2. There are two new buttons at the bottom of the page that perform multiple grouping and ungrouping by using the API.
3. Clicking on any of these two buttons several times in a row should **NOT** raise onGroupingDone event multiple times - it should be raised only ONCE (on the first click), when the grouping expressions change. 
4. The onGroupingDone event should get raised either when new columns get grouped **or ungrouped**. 
5. Ungrouping several columns at once by calling the clearGrouping method, and passing in an array of column names, should **NOT** raise onGroupingDone multiple times. clearGrouping now has an overload and should ungroup all the columns at once - hence the event should be raised only once.


MartinKamenov:
@PavlovVasil 
Some tests fail after the last merge of the master, so please check them and see how to fix it.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -1087,6 +1087,22 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+     *     const grouping = event;
+     * }
+     * ```
+     * ```html
+     * <igx-grid #grid [data]="localData" (onGroupingChanged)="groupingChanged($event)" [autoGenerate]="true"></igx-grid>
+     * ```
+	 * @memberof IgxGridComponent
+     */
+    @Output()
+    public onGroupingChanged = new EventEmitter<ISortingExpression[]>();
```
kdinev: I thought there's going to be more than just the sorting expressions in the grouping event arguments. Shouldn't there be column components?

mpavlinov: @PavlovVasil See https://github.com/IgniteUI/igniteui-angular/issues/2447#issuecomment-419945534 and https://github.com/IgniteUI/igniteui-angular/issues/2447#issuecomment-419947496

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -129,6 +129,11 @@ export interface IColumnMovingEndEventArgs {
     cancel: boolean;
 }
 
+export interface IGroupingDoneEventArgs {
```
ChronosSF: This is insufficient for the user to determine the state change. At the very least it should be:

```typescript
export interface IGroupingDoneEventArgs {
     expressions: Array<ISortingExpression> | ISortingExpression;
     ungroupedColumns: columns: Array<IgxColumnComponent> | IgxColumnComponent;
     groupedColumns: columns: Array<IgxColumnComponent> | IgxColumnComponent;
}
```

This is because the user can push new grouping expressions that result in ***both*** grouping and ungrouping in the same operation.

ChronosSF: In the docs we should also be clear that these are only the ***newly*** grouped and ***newly*** ungrouped columns. The columns that changed their grouping state.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -2909,24 +2926,43 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         }
         this.cdr.detectChanges();
         this.calculateGridSizes();
-        this.onGroupingDone.emit(this.sortingExpressions);
-
+        const newSortingExpressions: Array<ISortingExpression> = this.sortingExpressions;
+        if (JSON.stringify(oldSortingExpressions) !== JSON.stringify(newSortingExpressions)) {
```
ChronosSF: following on the comment regarding the interface this implementation won't do
you need a full diff similar to what Maya recommended in the original issue so that you can extract cols that had their state changed from grouped to ungrouped and vise versa.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -2909,24 +2926,43 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
      *
      */
-    public clearGrouping(name?: string): void {
+    public clearGrouping(name?: string | Array<string>): void {
+        const oldSortingExpressions: Array<ISortingExpression> = this.sortingExpressions;
         this.gridAPI.clear_groupby(this.id, name);
         this.calculateGridSizes();
-
+        const newSortingExpressions: Array<ISortingExpression> = this.sortingExpressions;
+        if (JSON.stringify(oldSortingExpressions) !== JSON.stringify(newSortingExpressions)) {
```
ChronosSF: see groupby comment

dkamburov:
1. Reordering chips into the grouping area does not emit the onGroupingDone event.
2. Also changing the direction through the chips is not emitting the onGroupingDone event either.(Not sure if this should be considered as an issue, because the group remain the same only sorted in the opposite direction)

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -354,6 +363,34 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
             // setter called before grid is registered in grid API service
             this.sortingExpressions.unshift.apply(this.sortingExpressions, this._groupingExpressions);
         }
+        if (JSON.stringify(oldExpressions) !== JSON.stringify(newExpressions)) {
```
ChronosSF: Even if this is check is needed, it should be moved up so that the actual grouping is not performed while the event is being prevented.

projects/igniteui-angular/src/lib/grid/api.service.ts
```diff
@@ -340,30 +340,39 @@ export class IgxGridAPIService {
...
-            }
-            this.get(id).groupingExpressions = groupingState;
-            this.get(id).sortingExpressions = sortingState;
-
-            /* remove expansion states related to the cleared group
-            and all with deeper hierarchy than the cleared group */
-            this.get(id).groupingExpansionState = grpExpandState
-                .filter((val) => {
-                    return val.hierarchy && val.hierarchy.length <= grExprIndex;
+            if (typeof name === 'string') {
```
ChronosSF: this can be written a lot shorter, also you missed the `groupingExpansionState` clear from the else block

You can do something like:

```typescript
const names = typeof name === 'string' ? [ name ] : name;
const leftOvers = groupingState.filter((state) => !names.includes(state.fieldName));
this.get(id).groupingExpressions = leftOvers
```

And include the groupingExpansionState mods for all code paths

PavlovVasil:
@ChronosSF @MartinKamenov : Please review again and test again.

projects/igniteui-angular/src/lib/grid/grid.component.ts
kdinev: This check should also include the check for the existence of `this.columnList`. I still am not sure why this check is needed though. What is the case?

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -342,6 +349,8 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
         if (value && value.length > 10) {
             throw Error('Maximum amount of grouped columns is 10.');
         }
+        const oldExpressions: Array<ISortingExpression> = this.groupingExpressions;
+        const newExpressions: Array<ISortingExpression> = value;
         this._groupingExpressions = cloneArray(value);
```
kdinev: This is not part of the change, but why is a deep copy of the array provided needed?

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -354,6 +363,34 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
             // setter called before grid is registered in grid API service
             this.sortingExpressions.unshift.apply(this.sortingExpressions, this._groupingExpressions);
         }
+        if (JSON.stringify(oldExpressions) !== JSON.stringify(newExpressions)) {
+            if (this.columnList) {
+                const groupedCols: Array<IgxColumnComponent> | IgxColumnComponent = [];
+                const ungroupedCols: Array<IgxColumnComponent> | IgxColumnComponent = [];
+                const groupedColsArr = newExpressions.filter(function(obj) {
```
kdinev: Use the arrow notation: `(obj) => { }`

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -354,6 +363,34 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+                    return !oldExpressions.some(function(obj2) {
```
kdinev: Again, use arrow notation `(obj) => { }`

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -354,6 +363,34 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+                        return obj.fieldName === obj2.fieldName;
+                    });
+                });
+                groupedColsArr.forEach(function(elem) {
+                    groupedCols.push(this.getColumnByName(elem.fieldName));
+                }, this);
+                const ungroupedColsArr = oldExpressions.filter(function(obj) {
```
kdinev: Arrow.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -354,6 +363,34 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+                    return !newExpressions.some(function(obj2) {
```
kdinev: Arrow.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -354,6 +363,34 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+                        return obj.fieldName === obj2.fieldName;
+                    });
+                });
+                ungroupedColsArr.forEach(function(elem) {
```
kdinev: Arrow.

projects/igniteui-angular/src/lib/grid/grid.component.ts
PavlovVasil: @kdinev  Without this check, passing the same expressions multiple times in a row, through the API, would raise onGroupingDone every time. This is issue #2644. I will change it so that it includes the other check as well.

projects/igniteui-angular/src/lib/grid/grid.component.ts
PavlovVasil: @kdinev If you refer to this line : `this._groupingExpressions = cloneArray(value);` - it doesn't create a deep copy, because the optional boolean parameter hasn't been set to "true".

kdinev: Ok, then why is a new reference created is the question? Why can't `value` be used directly?

Issue #2607, opened 2018-09-18
Async callback was not invoked within timeout specified by jasmine


tiliev:
Our tests are randomly failing with this error.
Here is a list of the things which might cause this:

## Not cancelled subscriptions ##
Memory leaks.

- [x] `igx-chips-area` has 5 subscriptions in `ngDoCheck` which we should unsubscribe from (`onMoveStart`, `onMoveEnd`, `onDragEnter`, `onKeyDown` and `onSelection`).
- [x] `igx-combo-drop-down` - missing `takeUntil` in `virtualContainer.onChunkLoad` subscription.
- [x] `igx-date-picker` - missing `takeUntil` in `calendar.onSelection` subscription.
- [x] `igx-radio-group` - `button.change` subscription not cancelled.
- [x] `igxTextHighlight` - `IgxTextHighlightDirective.onActiveElementChanged`
- [x] `igxToggle` - `overlayService.onOpened`. We could introduce `takeUntil` for all of the subscriptions.
- [x] `igx-grid-cell` - `row.virtDirRow.onChunkLoad`
- [x] `igx-column-hiding` - `item.valueChanged`
- [x] `igx-grid` - `virtualContainer.onChunkLoad`
- [x] `IgxOverlayService` - `fromEvent(wrapperElement, 'keydown')` and `fromEvent(info.elementRef.nativeElement.parentElement.parentElement, 'click')`

## Incorrect usage of  `takeUntil` ##

Possible memory leaks. More information [here](https://blog.angularindepth.com/rxjs-avoiding-takeuntil-leaks-fb5182d047ef).
- [x] `igxDrag` subscriptions in `ngOnInit`. Example:
```
pipe(takeUntil(this._destroy),
    throttle(() => interval(0, animationFrameScheduler)));
```
It should be:
```
pipe(throttle(() => interval(0, animationFrameScheduler)),
    takeUntil(this._destroy));
```

- [x] `igxResizer` constructor - should have `takeUntil` in `this.resizeStart.pipe` as last operator. I guess here it does not leak, because we have `takeUntil` in the observable in the `switchMap`.
- [x] `igxResizer` `ngOnInit` - `takeUntil` should be last operator in `mousemove`.

## Missing `removeEventListener` ##
- [x] `igxFor` - `this.vh.instance.elementRef.nativeElement` and `this.hvh.instance.elementRef.nativeElement` have `scroll` event listeners not removed.
[...]

PR #2655, opened 2018-09-26
Transaction service + Row Editing implementation

### LINKED ISSUE #566: igxGrid Row edit
### Description  
I guess you may already be working on this, as it's possible in the old ig grid, but I would really like to see "row mode" in the igx-grid.

### Result  
Can only select / edit one cell at a time.

### Expected result  
Ability to select a row and show editors for all cells in the row at once when editing.

wnvko:
Closes #566  

This PR also brings a ton of async test improvements related to #2607 as well as a configuration change that allows tests to reuse test module (credits to @gedinakova for finding the workaround for that one 👏).
That reduces memory leaks significantly - an sample of ~42 grid tests before:
image
and after (look at heap):
image

Also greatly improved test run time - from [~11min](https://dev.azure.com/IgniteUI/igniteui-angular/_build/results?buildId=1141&view=ms.vss-test-web.test-result-details) to [~3.5min](https://dev.azure.com/IgniteUI/igniteui-angular/_build/results?buildId=1142&view=ms.vss-test-web.test-result-details) 🎉


projects/igniteui-angular/src/lib/services/transaction/utilities.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Applies the last undone transaction if any
+     */
+    redo(): void;
+
+    /**
+     * Returns a map of aggregated state for all transactions
+     */
+    aggregatedState(): Map<any, IState>;
```
damyanpetev: I have a feeling this would be much better served for the API as `ITransaction[]` return.
The Map doesn't serialize out of the box and the `Array.from(map)` result is a jagged array which is no good.
That means the `aggregatedState()` will have to do some conversions so the end result is easily usable and can be sent to a server.

PS: No need to describe the type of result ("Returns a map"), there's a type right there. The test should describe what 'aggregated' means - all changes are aggregated resulting in a single Transaction per record that has a value reflecting the final state of the row.

wnvko: All this is fixed now

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -1310,4 +1374,9 @@
     %igx-grid__filtering-outlet {
         z-index: 99999;
     }
+
+    %igx-grid__row-editing-outlet {
+        z-index: 10000;
```
ViktorSlavov: @simeonoff Could you please confirm if this is okay?
We need the row-edit overlay to be above the pinned cells (z-index: 9999), below the filtering overlay (z-index: 99999) and below the virtual scrolls (z-index: 10001, [link](https://github.com/IgniteUI/igniteui-angular/blob/155ef109764d1d7bc5a27268ea1e64aaff6e60e1/projects/igniteui-angular/src/lib/core/styles/components/_common/_igx-vhelper.scss#L4))

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -4592,4 +4855,134 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+                this.rowEditingOverlaySettings.positionStrategy.settings.verticalStartPoint = VerticalAlignment.Bottom;
+            }
+
+            if (this.rowEditingOverlay.element.style.display === 'none') {
+                this.rowEditingOverlay.element.style.display = 'block';
+            }
+        }
+    }
+
+    public calculateRowChangesCount(rowID) {
```
damyanpetev: This one shold also be `@hidden`, right? Or not `public` at all if not needed for template bindings

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -4592,4 +4855,134 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+        this._pipeTrigger++;
+        this.cdr.detectChanges();
+    }
+
+    /**
+     * Finishes the row transactions on the current row.
+     * If `commit === true`, passes them from the pending state to the data (or transaction service)
+     * @param commit
+     */
+    public closeRowTransaction(commit?: boolean) {
```
damyanpetev: TBH, I'd skip that one for the first release, but while at it:
- Shouldn't it be just `endRowEdit(commit = false)`? Transactions are a separate *optional* module after all.
- If you do go with close/end I'd expect a start
- There's already API to enter cell edit that should trigger row edit, right? Assuming ending the cell edit will do the opposite, do you really need a row-specific method?

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -18,6 +18,12 @@ export function cloneArray(array, deep?: boolean) {
 export function cloneObject(object: any) {
     return JSON.parse(JSON.stringify(object));
 }
+/**
+ *@hidden
+ */
+export function isObjectEmpty(object: any) {
```
damyanpetev: This is only used to check if a row transaction has any changes, right?
Checking `Object.keys({}).length` should be enough, the transaction API guarantees it's an object already.
Plus checking constructor really isn't very generic as it won't work with instances that extend Object for example.

projects/igniteui-angular/src/lib/services/transaction/utilities.ts
```diff
@@ -0,0 +1,104 @@
...
+     * @param id The id of the record
+     * @returns State of the record if any
+     */
+    getState(id: any): State;
+
+    /**
+     * Returns whether transaction is enabled for this service
+     * @returns If transaction is enabled
+     */
+    transactionsEnabled(): boolean;
```
damyanpetev: @wnvko This should be just `enabled` as a property, where the full implementation would return true always and the  base only when pending collection is in progress.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -4489,4 +4861,148 @@ export class IgxGridComponent implements OnInit, OnDestroy, AfterContentInit, Af
...
+        const value = this.transactions.getAggregatedValue(rowInEdit.rowID, true);
+        const rowObj = rowInEdit ? this.getRowByIndex(rowInEdit.rowIndex) : null;
+        const emitter = commit ? this.onRowEditDone : this.onRowEditCancel;
+        emitter.emit({
+            newValue: Object.assign({}, this.data[rowObj.dataRowIndex], change),
+            oldValue: this.data[rowObj.dataRowIndex],
+            row: rowObj
+        });
+        this.transactions.endPending(commit);
+        if (commit && value && !isObjectEmpty(value) && !this.transactions.transactionsEnabled()) {
```
damyanpetev: With `enabled` change this should just check `this.transactions.enabled` after calling `endPending()`.

projects/igniteui-angular/src/lib/grid/api.service.ts
```diff
@@ -169,16 +205,30 @@ export class IgxGridAPIService {
...
+            const oldValueInTransaction = grid.transactions.getAggregatedValue(rowID, true);
+            if (oldValueInTransaction  && !isObjectEmpty(oldValueInTransaction)) {
+                oldValue = oldValueInTransaction[column.field];
+            }
+
+            //  if edit (new) value is same as old value do nothing here
+            if (oldValue && oldValue === editValue) { return; }
+
+            const transaction: Transaction = { id: rowID, type: TransactionType.UPDATE, newValue: { [column.field]: editValue } };
+            if (!grid.transactions.add(transaction, grid.data[rowIndex])) {
```
damyanpetev: Similarly, this should check if `grid.transactions.enabled` and then add or update the data

projects/igniteui-angular/src/lib/services/transaction/utilities.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export interface TransactionService {
+    /**
+     * Adds provided  transaction with recordRef if any
+     * @param transaction Transaction to be added
+     * @param recordRef Reference to the value of the record in the data source related to the changed item
+     *
+     * @returns If operation is successful.
+     */
+    add(transaction: Transaction, recordRef?: any): boolean;
```
damyanpetev: And then there's no need for this method to return a bool anymore

wnvko: done

projects/igniteui-angular/src/lib/grid/grid.component.html
```diff
@@ -140,5 +157,20 @@
...
 
-<div class="igx-grid__filtering-outlet" igxOverlayOutlet></div>
-
+<div *ngIf="rowEditable" igxToggle>
+    <ng-container *ngTemplateOutlet="rowEditContainer; context: {$implicit: this}"></ng-container>
+</div>
+<ng-template #defaultRowEditTemplate >
+    <div igxRowEdit [className]="bannerClass">
+        <div class="igx-banner__message">
+            <span class="igx-banner__text">{{ rowEditMessageValue }}</span>
```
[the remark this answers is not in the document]
Lipata: All changes related to row edit template are done, except this.

projects/igniteui-angular/src/lib/grid/grid.component.ts
```diff
@@ -139,6 +158,33 @@ export interface IFocusChangeEventArgs {
     cancel: boolean;
 }
 
+interface ContainerPositionSettings extends PositionSettings {
+    container?: HTMLElement;
+}
+
+class ContainerPositioningStrategy extends ConnectedPositioningStrategy {
```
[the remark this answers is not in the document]
Lipata: Done

projects/igniteui-angular/src/lib/grid/grid-navigation.service.ts
```diff
@@ -62,12 +67,48 @@ export class IgxGridNavigationService {
...
+
     public isColumnLeftPartiallyVisible(visibleColumnIndex) {
         const index = this.getColumnUnpinnedIndex(visibleColumnIndex + 1);
         return (!this.isColumnLeftFullyVisible(visibleColumnIndex)) && this.displayContainerScrollLeft <
             this.grid.dataRowList.first.virtDirRow.getColumnScrollLeft(index);
     }
 
+    public getCellElementByVisibleIndex(rowIndex, visibleColumnIndex) {
+        return this.grid.nativeElement.querySelector(
+            `igx-grid-cell[data-rowindex="${rowIndex}"][data-visibleIndex="${visibleColumnIndex}"]`) || { focus: () => {}};
```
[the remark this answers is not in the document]
ViktorSlavov: This was implemented instead of a null check on each of the places where the `getCellElementByVisibleIndex` method was called - due to performance or the browser handling the `tab` keydown, you'd get the focus "fly out" of the grid when holding down tab and it would result in errors. With the improvement to the tab handling (with the listener in the grid component) that should not be an issue - should I remove it?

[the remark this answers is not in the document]
rkaraivanov: Yes, it should not be needed anymore.

PR #2656, opened 2018-09-26
[In Development] Height/Width inputs deprecation, introducing visibleRows input

### LINKED ISSUE #2384: Research and implement number-of-records-based rendering for igxGrid
This items includes but is not limited to:

- add an input for the number of records the grid should display fully in the view
- the height of the grid should not depend on the parent elements
- the height of the grid should not be an input
- the grid should work nicely with height set through css
- possible igxforof changes related to points above

As discussed with @kdinev, most web apps flow vertically based on the items rendered and their height instead of have explicit height to fit into. This makes setting height of components a chore for developers. Ideally we'll want the grid height to naturally add up from the sum of its seperate areas with the body height being an user set value of number of records * their height _or_ a default number.

ChronosSF:
Closes #2384 .  

TODOs:

- [ ] Code clean-up
- [ ] Bottom scrollbar positioning
- [ ] Set height lower than rendered height polish
- [ ] Modify spec
- [x] Fix tests
- [x] Change igxCombo to use the new inputs
- [x] Fix tfoot scrolling
- [x] Fix horizontal keyboard navigation


MartinKamenov:
When setting visibleRows option to records minus on count it visualizes all records.
For example if user sets column moving sample visibleRows to 26 and the sample has 27 records it renders all 27 records instead 26.

PR #2691, opened 2018-10-04
replace toLocaleDateString in calendar with Intl.DateTimeFormat

### LINKED ISSUE #2675: Improve igxCalendar performance
### Description  
We need to cover the following:
- [x] [avoid using of `toLocaleString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString#Performance). Instead it is better to create `Intl.DateTimeFormat` object and use the function provided by its `format` property.

- [x]  `animationDone` is performing `find` every time and this is reducing the time execution of the Calendar, we should think of a way to improve that.

nrobakova:
Closes #2675 .

### DIFF
--- a/projects/igniteui-angular/src/lib/calendar/calendar.component.ts
+++ b/projects/igniteui-angular/src/lib/calendar/calendar.component.ts
@@ -133,20 +133,32 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
         this.calendarModel.firstWeekDay = value;
     }
 
-    /**
-     * Sets/gets the `locale` of the calendar.
-     * Expects a valid BCP 47 language tag.
+   /**
+     * Gets the `locale` of the calendar.
      * Default value is `"en"`.
-     * ```html
-     * <igx-calendar [locale] = "de"></igx-calendar>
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/calendar/calendar.component.ts
```diff
@@ -534,6 +547,30 @@ export class IgxCalendarComponent implements OnInit, ControlValueAccessor {
      *@hidden
      */
     private _monthAction = '';
+    /**
+    *@hidden
+    */
+    private _locale = 'en';
```
zdrawku: Do not use "_" as a prefix for private properties ([coding-guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names)). As I see there are other private properties that do not cover this requirement, make changes to them also.

nrobakova: Remove prefix for properties:
rangeStarted,  formatterMonth, formatterDay, formatterYear, formatterMonthday and formatterWeekday

PR #2706, opened 2018-10-05
feat(toggle-action): set `overlaySettings` position target if empty

damyanpetev:
Related IgniteUI/igniteui-cli#369

If there are bound `overlaySettings` with `positionStrategy` that doesn't define a target, provide the toggle action element before opening.
This allows the config object to be created for targets that are created dynamically (async, ngIf, etc..).


projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.spec.ts
```diff
@@ -5,6 +5,7 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { IgxToggleActionDirective, IgxToggleDirective, IgxToggleModule, IgxOverlayOutletDirective } from './toggle.directive';
 import { IgxOverlayService, OverlaySettings, ConnectedPositioningStrategy,
     AbsoluteScrollStrategy, AutoPositionStrategy } from '../../services';
+import { IPositionStrategy } from 'dist/igniteui-angular/public_api';
```
kdinev: This import shouldn't be from `dist/`

PR #2713, opened 2018-10-08
Refactor chips so that they can be easier to style.

### LINKED ISSUE #2712: Refactor igxChip to be easier to customize its styling.
### Description  
The chip components should be easier to style and when user decides to customize it it should follow the same guidelines in all cases.

So far what has been decided the changes would be:
- [x] 1. The user would need to handle the chip connectors implementation, since they are more limiting than providing easier solution.
- [x] 2. Implement the connectors from the change 1) in the igxGrid GroupBy.
- [x] 3. The remove button should be fully customizable. When the user needs removing logic he would need to define what would be rendered for the remove button. The rest the chip would handle and emit `onRemove` event.
- [x] 4. The `igxPrefix` and `igxSuffix` are still used, but they will need to apply different styles depending on where they are used - `igxInputGroup` or `igxChip`.
- [x] 5. The template for the `igxChip` needs to be revisited. The order of elements would be `prefix|label|suffix|remove button`.
- [x] 6. Expose `data` property  for the chip component.
- [x] 7. Update tests and demo to reflect the new changes.
- [x] 8. Styling for the updated `igxChip` should be updated accordingly as well.
- [x] 9. Styling for the igxGrid GroupBy connectors and remove button implementation.

skrustev:
Closes #2712 .


projects/igniteui-angular/src/lib/chips/chip.component.ts
```diff
@@ -62,48 +63,56 @@ export class IgxChipComponent implements AfterViewInit {
     public id = `igx-chip-${CHIP_ID++}`;
 
     /**
-     * An @Input property that defines if the `IgxChipComponent` can be dragged in order to change it's position.
-     * By default it is set to false.
+     * An @Input property that stores data related to the chip.
      * ```html
-     * <igx-chip [id]="'igx-chip-1'" [draggable]="true"></igx-chip>
+     * <igx-chip [data]={ value: 'Country' }></igx-chip>
```
kdinev: You're missing `""`.

PR #2719, opened 2018-10-09
Variable Heights implementation

MayaKirova:
Implementing variable heights for igxFor items and igGrid row heights.
Variable heights is always enabled, this means that if items height is changed (from the default size) when they are rendered in DOM the related size cache and scroll height will be updated.
Virtualization start index is now always calculated based on size cache for both vertical and horizontal virtualization.


MartinKamenov:
@MayaKirova 
The style of the cells is changed and the text is not aligned in center as before the change so please check it.
returnmaq2

MartinKamenov:
@MayaKirova 
The grid summary sample doesn't allow user to scroll to bottom.
maq3

MartinKamenov:
@MayaKirova The grid cannot scroll to last record on IE in the grid performance sample.
The records are 100 000, but the grid only shows 30 679
maya4

MartinKamenov:
In Grid Cell Editing sample when style is changed from comfortable to compact the grid doesn't fulfill the space  over the grid summaries.
comf to compact

MayaKirova:
@MartinKamenov Issues have been fixed and KB navigation changes have been merged from master. Please re-test.

MartinKamenov:
The scroll bar in the performance sample flickers when using arrow navigation downwards.
scrollvariablerowheights

MartinKamenov:
When user enters edit mode for cell the row height is changed.
See the grid cell editing sample.
changerowsizevarrowheights

MartinKamenov:
Selection grid sample is not changing data when scrolling.
selectionvariablerowheights

MartinKamenov:
Cannot scroll to the bottom of the grid remote virtualization sample.
The records are 77 not 76.
remotevarrowheights

MayaKirova:
@MartinKamenov Regarding the height of the row changing when in edit mode when using compact display density. The issue is due to the height of the editor (height + padding) being larger than the available space in the cell, hence since now we only set min-height allowing the content to span  to its full height the row height will be changed to fit the whole content. The same issue with a slightly different result can also be observed in master, where since fixed height is enforced the content overflows outside the bounds of the cell. You can observe the underline for the editor text is displayed outside the bounds of the cell:
image

This is a styling issue that should be handled as a separate bug as it is also reproducible in master probably by setting different height and/or padding for the editors depending on the set display density.

kdinev:
Shouldn't there be some VD update to center the cell contents?

PR #2725, opened 2018-10-11
fix(selection): accept 0 as itemID

### LINKED ISSUE #2581: IgxSelectionAPIService allows to add items with id which is undefined

DiyanDimitrov:
Closes #2581.

### DIFF
--- a/projects/igniteui-angular/src/lib/core/selection.ts
+++ b/projects/igniteui-angular/src/lib/core/selection.ts
@@ -19,7 +19,7 @@ export class IgxSelectionAPIService {
      * @param newSelection The new component selection to be set.
      */
     public set(componentID: string, newSelection: Set<any>) {
-        if (!componentID) {
+        if (componentID == null) {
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/core/selection.ts
```diff
@@ -60,7 +60,7 @@ export class IgxSelectionAPIService {
         if (sel === undefined) {
             sel = this.get_empty();
         }
-        if (!itemID) {
+        if (itemID == null) {
```
Lipata: Even though `undefined == null`, returns true, it's clear and easier to understand `itemID === null || itemID === undefined` (I've checked our source code and the second approach is used). In addition when `itemID === ''` then this will not be handled.

DiyanDimitrov: Modified to check for all falsy values except 0 `!itemID && itemID !== 0`.

PR #2733, opened 2018-10-11
Grid's filtering dialog fix

### LINKED ISSUE #2508: igxOverlay - the content is not shown in the foreground of igxGrid in Edge and IE
### Description  
igxOverlay - the content is not shown in the foreground of igxGrid in Microsoft Edge and IE.

 * browser: Microsoft Edge, IE

### Steps to reproduce  

1. Open the attached sample application in Microsoft Edge.
2. Open color-picker sample
3. Open Date column Filtering UI.
4. Click inside a date column cell to edit it.

### Result  
3. The Filtering UI is opened under the grid.
4. The DatePicker is opened under the grid.

### Expected result  
3. The Filtering UI should be opened on the foreground of the grid.
4. The DatePicker should be opened on the foreground of the grid.

**Note:** The issue is not reproducible in Chrome and FireFox.
**Note:** If you open the Filtering UI in IE it will open in the foreground of the grid, but if it contains a DatePicker (in date column), the DatePicker will be under the grid.

### Attachments  
[igx-test-sample.zip](https://github.com/IgniteUI/igniteui-angular/files/2324060/igx-test-sample.zip)

wnvko:
This is a known issue in MS Edge and MS Internet Explorer. MS said this was fixed in May 2018. However they still did not released this fix.

As a work around we are setting the position of filtering outlet now to fixed. Setting the position to fixed creates new stacking context and this allows filtering dialog to appears above the grid body. We should check when MS release their fix to remove this work around.

Closes #2508

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
@@ -1308,6 +1308,7 @@
     }
 
     %igx-grid__filtering-outlet {
-        z-index: 99999;
+        z-index: 2;
+        position: fixed;
     }
 }


npaunov:
There is one scenario when filter ui is used and it contains a pop up - e.g. DatePicker, it is positioned under the filter ui. This is reproducible only in IE 11

Issue #2031, opened 2018-07-18
Feature request: ability to use igx-drop-down as menu


Eralmidia:
### Description  
It seems like setting the allowItemsFocus property of igx-drop-down has no effect

 * igniteui-angular version:  6.1.1
 * browser: Chrome

### Steps to reproduce  
https://stackblitz.com/edit/angular-dq7d9o?file=app%2Fdropdown-sample-1%2Fdropdown-sample-1.component.html

### Result  
Items gets focus

### Expected result  
Items should not get focus

### Attachments  
Attach a sample if available, and screenshots, if applicable.  
![image](https://user-images.githubusercontent.com/12936286/42865682-b3defc00-8a6a-11e8-8b63-5e09bb3702b3.png)

ViktorSlavov:
@Eralmidia The items do not actually get focus - while the style is applied, the document.activeElement is still the button. The class is applied for better UX - the **selected** item has a different text color (in this case - black) from **selected** + **focused** item (white).
In the above sample, you do cannot have keyboard navigation in the drop down (items cannot be focused, so keyboard events will never fire on the navigation directive).
When you get a chance, you can check the [following sample](https://angular-dq7d9o-p8fxck.stackblitz.io/), where the navigation directive is put on the button instead of the drop down items (enabling keyboard navigation):
- Clicking the button opens the drop down and enables keyboard navigation on the drop down list
- Use the keyboard to navigate through and select items to see how both css classes interact
- Even though **focused** class is applied, they are *not* actually focused - document.activeElement remains the button

kdinev:
@Eralmidia Do you need to just disable certain items?

Eralmidia:
@ViktorSlavov Thank you for the clarification. I can see the difference in terms of keyboard navigation, yes.

@kdinev Well, I'm sure we well have a use for disabling certain items sooner or later ;) But actually, my use case right now, was using the drop down as a menu like component, when implementing a quick navigation between the main modules of out application:
![image](https://user-images.githubusercontent.com/12936286/42868773-7cab40dc-8a73-11e8-805c-8631dcbfb1f3.png)

It seems to me, that if we could disable all focus, selection ect on the dropdown, it could easily work as a menu as well, for instance replacing the material2 menu: https://material.angular.io/components/menu/overview

kdinev:
@Eralmidia I guess we never thought about this use-case. We will look into providing a way to not apply those styles and maybe add this as a configuration input. @simeonoff

[...]

PR #2749, opened 2018-10-17
fix(drop-down): Added onSelecting event in the igx-drop-down componen…

sboykova:
Related to #2031


projects/igniteui-angular/src/lib/drop-down/drop-down.component.spec.ts
```diff
@@ -267,11 +270,115 @@ describe('IgxDropDown ', () => {
...
+            fixture.detectChanges();
+            const lastListItem = list.items[3].element.nativeElement;
+
+            lastListItem.click({});
+            tick();
+            fixture.detectChanges();
+            expect(list.selectedItem).toEqual(list.items[3]);
+            expect(list.onSelecting.emit).toHaveBeenCalledTimes(1);
+            expect(list.onSelection.emit).toHaveBeenCalledTimes(1);
+            expect(fixture.componentInstance.onSelecting).toHaveBeenCalledTimes(1);
```
bkulov: I think we don't need to spy on `fixture.componentInstance.onSelecting`.

bkulov: Also you could remove these expects as they are the same as in the previous test.

PR #2766, opened 2018-10-18
Change scrollTo method to account for relative position

### LINKED ISSUE #2751: Change scrollTo method to account for relative position
### Description  
scrollTo always brings the required index at the start of the virtualization frame. Instead a more natural behavior is to bring it:

- at the end if the index coming from is lower 
- at the start if the index coming from is higher

Related to the the keyboard navigation improvements.

MayaKirova:
Closes #2751


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -410,15 +410,23 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
             return;
         }
         // this.state.startIndex = index;
+        const containerSize = parseInt(this.igxForContainerSize, 10);
+        const isPrevItem = index < this.state.startIndex;
         if (this.igxForScrollOrientation === 'horizontal') {
-            this.hScroll.scrollLeft = this.sizesCache[index];
+            this.hScroll.scrollLeft = isPrevItem ?
+            this.sizesCache[index] :
+            this.sizesCache[index] - containerSize + this.igxForOf[index].width;
```
rkaraivanov: I recall that `this.igxForOf[index].width` returns a string. Maybe a `parseInt`?
Also is there a case when this whole expression can be less than 0 in the case when it's not a prevItem? Maybe wrap it in `Math.abs` or some other check?

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -410,15 +410,23 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
         } else {
-            const containerSize = parseInt(this.igxForContainerSize, 10);
             const maxVirtScrollTop = this._virtHeight - containerSize;
-            let nextScrollTop = this.sizesCache[index];
+            let nextScrollTop = isPrevItem ?
+            this.sizesCache[index] :
+            this.sizesCache[index] - containerSize + this.heightCache[index];
```
rkaraivanov: Again a case where the expression evaluates less than 0? I see there is a check below but in that case the igxFor won't scroll the whole item into view.

rkaraivanov:
Will something like this which takes into account partially visible rows/columns break some behavior of the directive?
```typescript
public scrollTo(index) {
        if (index < 0 || index > (this.isRemote ? this.totalItemCount : this.igxForOf.length)) {
            return;
        }

        const containerSize = parseInt(this.igxForContainerSize, 10);
        let isPrevItem;

        if (this.igxForScrollOrientation === 'horizontal') {
            isPrevItem = index <= this.state.startIndex || this.hScroll.scrollLeft > this.sizesCache[index];
            this.hScroll.scrollLeft = isPrevItem ?
                this.sizesCache[index] :
                Math.abs(this.sizesCache[index] - containerSize + parseInt(this.igxForOf[index].width, 10));
        } else {
            const maxVirtScrollTop = this._virtHeight - containerSize;
            isPrevItem = index <= this.state.startIndex || this.vh.instance.elementRef.nativeElement.scrollTop > this.sizesCache[index];
            let nextScrollTop = isPrevItem ?
                this.sizesCache[index] :
                Math.abs(this.sizesCache[index] - containerSize + this.heightCache[index]);
            if (nextScrollTop > maxVirtScrollTop) {
                nextScrollTop = maxVirtScrollTop;
            }

            this._bScrollInternal = true;
            this._virtScrollTop = nextScrollTop;
            this.vh.instance.elementRef.nativeElement.scrollTop = this._virtScrollTop / this._virtHeightRatio;
        }
    }
```

ChronosSF:
> Will something like this which takes into account partially visible rows/columns break some behavior of the directive?
> 
> ```ts
> public scrollTo(index) {
>         if (index < 0 || index > (this.isRemote ? this.totalItemCount : this.igxForOf.length)) {
>             return;
>         }
> 
>         const containerSize = parseInt(this.igxForContainerSize, 10);
>         let isPrevItem;
> 
>         if (this.igxForScrollOrientation === 'horizontal') {
>             isPrevItem = index <= this.state.startIndex || this.hScroll.scrollLeft > this.sizesCache[index];
>             this.hScroll.scrollLeft = isPrevItem ?
>                 this.sizesCache[index] :
>                 Math.abs(this.sizesCache[index] - containerSize + parseInt(this.igxForOf[index].width, 10));
>         } else {
>             const maxVirtScrollTop = this._virtHeight - containerSize;
>             isPrevItem = index <= this.state.startIndex || this.vh.instance.elementRef.nativeElement.scrollTop > this.sizesCache[index];
>             let nextScrollTop = isPrevItem ?
>                 this.sizesCache[index] :
>                 Math.abs(this.sizesCache[index] - containerSize + this.heightCache[index]);
>             if (nextScrollTop > maxVirtScrollTop) {
>                 nextScrollTop = maxVirtScrollTop;
>             }
> 
>             this._bScrollInternal = true;
>             this._virtScrollTop = nextScrollTop;
>             this.vh.instance.elementRef.nativeElement.scrollTop = this._virtScrollTop / this._virtHeightRatio;
>         }
>     }
> ```

I think the only way you get negative scroll is if you call scrollTo for an index that is in the initial view. I don't think this is related to partially visible items. 

Such items can't obviously appear at the bottom/right edge of the container so calling them from a position above or to the left of them, imo, should not change the scroll at all. Neither abs nor 0 look ok to me. I'll push a change for review.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -409,30 +409,23 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
             return;
         }
-        // this.state.startIndex = index;
         const containerSize = parseInt(this.igxForContainerSize, 10);
         const scr = this.igxForScrollOrientation === 'horizontal' ?
-        this.hScroll.scrollLeft :
-        this.vh.instance.elementRef.nativeElement.scrollTop;
+            this.hScroll.scrollLeft : this.vh.instance.elementRef.nativeElement.scrollTop;
         const isPrevItem = index < this.state.startIndex || scr >  this.sizesCache[index];
+        let nextScroll = isPrevItem ? this.sizesCache[index] : this.sizesCache[index + 1] - containerSize;
```
ChronosSF:
```typescript
this.sizesCache[index + 1]
```

To reviewers - this is safe because the sizesCache always contains 1 more item than the igxForOf collection (the last item right/bottom edge position)

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -425,19 +425,23 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
         if (index < 0 || index > (this.isRemote ? this.totalItemCount : this.igxForOf.length)) {
             return;
         }
-        // this.state.startIndex = index;
+        const containerSize = parseInt(this.igxForContainerSize, 10);
+        const scr = this.igxForScrollOrientation === 'horizontal' ?
+            this.hScroll.scrollLeft : this.vh.instance.elementRef.nativeElement.scrollTop;
```
skrustev: This might not work on IE when triggering its scroll height limitation because the actual vertical scroll would be be scrollbarScrollTop * virtHeightRatio.

skrustev: This might not work when we are over the browser scroll height limitation because the actual scroll top would be
scrollbarScrollTop * virtualScrollHeightRatio.

PR #2779, opened 2018-10-19
Extract dropdown element indexes in a deterministic way

dkamburov:

### DIFF
--- a/projects/igniteui-angular/src/lib/combo/combo.component.spec.ts
+++ b/projects/igniteui-angular/src/lib/combo/combo.component.spec.ts
@@ -43,7 +43,7 @@ const CSS_CLASS_INPUTGROUP_BORDER = 'igx-input-group__border';
 const CSS_CLASS_HEADER = 'header-class';
 const CSS_CLASS_FOOTER = 'footer-class';
 
-fdescribe('igxCombo', () => {
+describe('igxCombo', () => {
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/combo/combo.component.spec.ts
```diff
@@ -15,6 +15,7 @@ import { BehaviorSubject, Observable } from 'rxjs';
 import { take } from 'rxjs/operators';
 import { RemoteService } from 'src/app/shared/remote.combo.service';
 import { UIInteractions, wait } from '../test-utils/ui-interactions.spec';
+import { element } from 'protractor';
```
ViktorSlavov: This `element` definition seems unused, no need for the import

projects/igniteui-angular/src/lib/combo/combo.component.spec.ts
```diff
@@ -876,6 +877,10 @@ describe('igxCombo', () => {
 
 
     describe('Selection tests: ', () => {
+        function getIndexOfVisibleItem(dropDownitems: any[], valueKey, value) {
+            const item = dropDownitems.filter((el) => el.value[valueKey] === value)[0];
```
ViktorSlavov: I think `find` is better suited here

Issue #2765, opened 2018-10-18
GroupBy Chip sorting direction indicator is not changed if sorting direction is changed


kacheshmarova:
### Description  
Sorting indicator in the groupby chip is not updated when sorting direction is changed . This happens only in case the column is first sorted and afterwards grouped

 * igniteui-angular version: 6.1.x, 6.2.x

### Steps to reproduce  

1. Run Grid GroupBy sample from dev samples
2. Sort Country column
3. Group by Country column
4. Click on the Country`s chip in the groupby area in order to change sorting direction

### Result  
Sorting indicator in the chip does not change in order to reflect the new sorting direction

### Expected result  
Sorting indicator in the chip to change in order to reflect the new sorting direction applied

### Attachments  
![chipsortingindicator](https://user-images.githubusercontent.com/13034836/47144196-9bdfd180-d2cf-11e8-8810-63e72118617e.gif)

PR #2786, opened 2018-10-19
Fix sorting and groupby expression not syncing when there are already sorted columns. 

skrustev:
#2765


PavlovVasil:
When several columns have been grouped, trying to reorder the chips by focusing on any of them and pressing Shift + arrow key does not work.

PavlovVasil:
When several columns have been grouped, trying to reorder the chips by focusing on any of them and pressing Shift + arrow key does not work.

PR #2805, opened 2018-10-23
Fix - Drop down - Prevent defaults only if drop down is opened

### LINKED ISSUE #2803: igxDropDownItemNavigation directive stops keyboard event propagation
Use drop down dev demo.
Open the second drop down by clicking the button then close it by clicking the button again.
Try to move the focus to the following button by pressing Tab key.

The keyboard events are not propagated.

ViktorSlavov:
Closes #2803 
`IgxDropDownItemNavigationDirective` will prevent the keydown event defaults only if the dropdown is visible. This is to stop the directive from interfering with default keyboard nav (`Tab`, `PageDown`, etc.) when the toggle element is also the navigation directive target.


IvayloG:
Currently space key does not select items (same dev sample).

PR #2815, opened 2018-10-24
Allow editing of dates in grid with transactions, #2811

### LINKED ISSUE #2811: Cannot edit dates in grid with transactions
### Description  
When transactions in grid are enabled, and there is editable column of dates, cells in this column are not editable.

### Steps to reproduce  

1. Run the dev sample;
2. Open Grid Row Edit sample;
3. Go to grid with transactions;
4. Try to edit any of the dates in Order Date column.

### Result  
Date dialog does not appear and error shows in the console.

### Expected result  
Dates should be editable.

wnvko:
Closes #2811  

In transaction pipe we do not call cloneArray with deep set to true, as this is not necessary. Actually in mergeTransactions method we are setting cloned value to each row already in transaction.

Also cloneObject is fixed and will be used in transaction service to allow cloning of objects containing dates.


projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -12,12 +12,89 @@ export function cloneArray(array, deep?: boolean) {
...
+}
+
+/**
+ * Checks if provided variable is Date
+ * @param value Value to check
+ * @returns true if provided variable is Date
+ *@hidden
+ */
+export function isDate(value: any) {
+    return value && Object.prototype.toString.call(value) === '[object Date]' && !isNaN(value);
```
damyanpetev:
```suggestion
    return Object.prototype.toString.call(value) === '[object Date]';
```
Should be enough here. This is a copy only, the validity of the date is not relevant.

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -12,12 +12,89 @@ export function cloneArray(array, deep?: boolean) {
...
+}
+
+/**
+ * Checks if provided variable is Object
+ * @param value Value to check
+ * @returns true if provided variable is Object
+ *@hidden
+ */
+export function isObject(value: any): boolean {
+    return value !== null && value !== undefined && value.toString() === '[object Object]';
```
damyanpetev:
```suggestion
    return value && typeof value === 'object';
```
?

projects/igniteui-angular/src/lib/core/utils.ts
damyanpetev: Umm no it won't.. unless it's different in IE

projects/igniteui-angular/src/lib/core/utils.ts
damyanpetev: True, but you already have a date and array checks before that one, right?

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -12,12 +12,89 @@ export function cloneArray(array, deep?: boolean) {
...
+ * Creates deep clone of provided value.
+ * Supports primitive values, dates and objects.
+ * If passed value is array returns shallow copy of the array.
+ * @param value value to clone
+ * @returns Deep copy of provided value
  *@hidden
  */
-export function cloneObject(object: any) {
-    return JSON.parse(JSON.stringify(object));
+export function cloneObject(value: any): any {
```
damyanpetev: `cloneValue`?

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -12,12 +12,89 @@ export function cloneArray(array, deep?: boolean) {
...
+
 /**
+ * Deep clones all first level keys of Obj2 and merges them to Obj1
+ * @param obj1 Object to merge into
+ * @param obj2 Object to merge from
+ * @returns Obj1 with merged cloned keys from Obj2
+ * @hidden
+ */
+export function mergeObjects(obj1: {}, obj2: {}) {
+    if (obj1 === null || obj1 === undefined) {
```
damyanpetev: If this is to remain somewhat comparable to `Object.assign` the first param is required, otherwise code might as well use clone directly

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -12,12 +12,89 @@ export function cloneArray(array, deep?: boolean) {
...
+    if (obj2 === null || obj2 === undefined) {
+        return obj1;
+    }
+
+    if (!isObject(obj1)) {
+        throw new Error(`Should provide objects to mergeObjects method. ${obj1} is not an object!`);
+    }
+
+    if (!isObject(obj2)) {
+        throw new Error(`Should provide objects to mergeObjects method. ${obj2} is not an object!`);
```
damyanpetev: And this can probably just 
```suggestion
        return obj1;
```
and skip the 2nd check

projects/igniteui-angular/src/lib/core/utils.ts
wnvko: done!

PR #2821, opened 2018-10-24
perf(igxGrid): Improve horizontal scrolling

### LINKED ISSUE #2820: [igxGrid] Possible improvements for horizontal scroll performance
There are several areas where we can improve the horizontal scrolling of the grid.

- remove the `igx-grid-row` getter for min-height
- rework the `igx-grid-cell` getter for min-height

Both of these read out the `offsetHeight/clientHeight` of the native elements triggering unnecessary layout reflows. We can improve it if we remove the getter from the row and leave it to the cell to expand its parent row in height.

- reduce the reads of `scrollLeft` in the horizontal scroll handler.

Each of the horizontal scroll handlers reads `event.target.scrollLeft` which again triggers unnecessary layout reflows. We can change it by reading once and passing the read value down to the handlers.

## Attachments
Attached are the measurements in Chrome before and after the changes

[scrollperformance.zip](https://github.com/IgniteUI/igniteui-angular/files/2510432/scrollperformance.zip)

rkaraivanov:
Closes #2820

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
@@ -1156,15 +1156,14 @@ export class IgxGridForOfDirective<T> extends IgxForOfDirective<T> implements On
         });
     }
 
-    onHScroll(event) {
+    onHScroll(scrollAmount) {
         /* in certain situations this may be called when no scrollbar is visible */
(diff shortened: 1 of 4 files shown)


MayaKirova:
Due to the changes in the min-height the variable heights functionality should be re-tested to make sure it is not affected. For reference for some test scenarios the original PR can be reviewed: https://github.com/IgniteUI/igniteui-angular/pull/2719

PR #2824, opened 2018-10-24
Theme engine fix

### LINKED ISSUE #2814: Samples do not work in StackBlitz
### Description  
Samples do not work in StackBlitz because of Sass error. Further investigation is needed. We should find what is the cause of this and decide whether there is an issue at our end or it is a third-party issue related to StackBlitz. As @hanastasov pointed out, this might be connected with the older angular 5 config file we are currently using (this is a pending issue - https://github.com/IgniteUI/igniteui-angular-samples/issues/387). Judging by the error there is some issue with the ripple, perhaps a typo somewhere in the Sass.

 * igniteui-angular version: 6.2.0-beta.4
 * browser: Chrome, Firefox and Edge. It works in IE because we use CSS instead of Sass there.

### Steps to reproduce  

1. Go to https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid_paging.html
2. Hit open in StackBlitz

### Result  
```
Error in igniteui-angular/lib/core/styles/base/utilities/_mixins.scss (228:12)
argument `$map` of `map-get($map, $key)` must be a map

Backtrace:
sass/igniteui-angular/lib/core/styles/base/utilities/_mixins.scss:228, in function `map-get`
sass/igniteui-angular/lib/core/styles/base/utilities/_mixins.scss:228, in mixin `igx-root-css-vars`
sass/igniteui-angular/lib/core/styles/components/ripple/_ripple-theme.scss:38, in mixin `igx-ripple`
sass/igniteui-angular/lib/core/styles/themes/_index.scss:111, in mixin `igx-theme`
sass/~/styles.scss:4
```

error

### Expected result  
The sample should work.

### Details
The issue is introduced in version `6.2.0-beta.4`. The samples work correctly with `6.2.0-beta.3` - https://stackblitz.com/edit/iggrid-working-beta-3. The issue might be connected with some of these changes:

diff-1
---
diff-2

### LINKED ISSUE #2823: Themes: Stackblitz compiler doesn't work with theme engine

zdrawku:
Closes #2823 
Closes #2814


zdrawku:
I am not sure if the 55 commits are from  --author git commit or because we checked out the branch from local master, not upstream/master.

The important here is that the changes to all 2 files are applied correctly.

Issue #2456, opened 2018-08-20
Improve keyboard navigation of the igx-Grid


StefanIvanov:
### Description  
Provide support for richer keyboard navigation to the Grid

Result | Key combination
-- | --
Move to top cell in column | Ctrl+↑
Move to bottom cell in column | Ctrl+↓
Move to leftmost cell in row | Ctrl+← or Home
Move to rightmost cell in row | Ctrl+→ or End
Move to top left cell in the grid | Ctrl+Home
Move to top right cell in the grid | Ctrl+End
Move to right cell (if last cell on the current row move to first cell of the next row)| Tab or →
Move to left cell (if first cell on the current row move to last cell of the previous row)| Shift+Tab or ←

### Attachments   
Related to #1420 

- [x] Implementation
  - [x] Arrow keys navigation
  - [x] Home, End keys navigation
  - [x] Tab, Shift + Tab navigation 

- [x] Bug fixing 

  - [x] Tab, Shift tab navigation 

- [x] Featured integration
   - [x] Tab navigation of Grouped rows

- [ ] Expose public API navigation methods.

- [ ] On mouse wheel, Blur the focused element.

- [ ] Active style of Grouped rows and Selected Cell - [Issue](https://github.com/IgniteUI/igniteui-angular/issues/2709)

- [ ] Tab index to be added only to the group row element

- [ ] Blur expose output. Throw cell blur event (to be cancelable)

- [ ] Remove space of the Grouped row; Add arrow left, right (expand/collapse)

- [ ] Wiki update

- [ ] Tests

zdrawku:
Pending work:
- Expose public API navigation methods.
- On mouse wheel scrolling , Blur the focused element.
- Active style of Grouped rows and Selected Cell - Issue
- Adding of more Tests
- Changelog update

PR #2826, opened 2018-10-24
feat(igxFocus): allow igxFocus to work with custom editors #2456

damyanpetev:
Closes # .  

Taking advantage of the fact that all composite editors are all provided as `NG_VALUE_ACCESSOR` (needed a common token) and making them optional DI for the `igxFocus` so the directive can receive potential host component and focus its inner field instead.


damyanpetev:
@rkaraivanov I've updated the checkbox and date/time pickers with the method as those made the most sense.
Searching for value accessors I also see radio, slider and switch that can potentially implement this as well. Let me know I yo think I should update those too.

PR #2827, opened 2018-10-25
Hide/show row editing dialog with group collapse

### LINKED ISSUE #2817: Row Editing overlay is not hidden, when group is collapsed
### Description  
Row Editing overlay is not hidden, when group is collapsed

### Steps to reproduce  

1. Enable row editing and grouping.
2. Group a column
3. Open row editing dialog for a row from specific group
4. Collapse the same group

### Result  
The row editing overlay is visible, even the group is collapsed

### Expected result  
The row editing overlay is hidden.

Lipata:
Closes #2817 .  

Move test data to test util.


projects/igniteui-angular/src/lib/grid/api.service.ts
```diff
@@ -388,17 +388,22 @@ export class IgxGridAPIService {
...
+            expanded = !grid.groupsExpanded;
             expansionState.push({
-                expanded: !grid.groupsExpanded,
+                expanded,
                 hierarchy: DataUtil.getHierarchy(groupRow)
             });
         }
         this.get(id).groupingExpansionState = expansionState;
+        if (grid.rowEditable) {
+            grid.toggleRowEditingOverlay(expanded);
```
kdinev: @Lipata Wouldn't the fix also show a row editing overlay after expanding a group if there is no row in edit mode?

Lipata: Yes, it will show the overlay when group is expanded, because the cell(row) is still in edit mode (this is also valid, when row editing is disabled). Basically the overlay is just hidden (like it is when scrolling grid and overlay goes outside visible area, but it is still in edit mode). The other thing is that collapsing and expanding is not data changing operation, so in that case we are not forced to finish editing operation and submit changes.

Lipata: No, it will not show overlay, because showRowEditingOverlay is just dealing with display style property of already opened overlay (the latter happens, when overlay.open() method is used). We use open and close methods of the overlay, only when we finish row editing, because they trigger detect changes. For cases, when cell(row) is still in edit mode, we only show/hide overlay.

kdinev: The overlay display will still be correct if after collapse a different row enters edit mode, right?

IvayloG: There are a couple of issues related to this.
1.	The overlay is not repositioned correctly. When edit a row and collapse the group, edit in another group’s row and expand the previous/collapsed one, the overlay (row editing buttons) is still there and not next to the currently edited row.
 
image

2.	The overlay is not repositioned correctly. When edit a row and collapse the group, then collapse and expand a different one, the overlay (row editing buttons) position is wrong.

image

Lipata: In addition to those scenarios, I've also added test for hierarchical groups.

PR #2832, opened 2018-10-25
Don't allow editing on deleted row

### LINKED ISSUE #2831: It's possible to edit a cell from a deleted row
### Description  
When row editing is enabled and row is deleted, then it is possible to enter edit mode of a cell from that row and edit its value

### Steps to reproduce  

1. Enable row editing and add a transaction provider for the grid
2. Delete a row
3. Double click on e cell from that row

### Result  
It is possible to enter edit mode in a cell from the deleted row

### Expected result  
Grid should not allow any edits on a deleted row

Lipata:
Closes #2831


projects/igniteui-angular/src/lib/grid/cell.component.ts
```diff
@@ -283,6 +283,9 @@ export class IgxGridCellComponent implements OnInit, AfterViewInit {
      * @memberof IgxGridCellComponent
      */
     set inEditMode(value: boolean) {
+        if (this.grid.rowEditable && this.row.deleted) {
```
kdinev: Can't the row get deleted via API even if there's no row editing enabled?

Lipata: No, there isn't a deleted row, when row editing is disabled and yes the check for rowEditable is not needed. Thanks!

kdinev: I'm pretty sure that there is:

https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grid/grid.component.ts#L3078

Lipata: Right! Please ignore the first part of my previous answer. The correct thing is that only the `this.row.deleted` check is enough.

PR #2839, opened 2018-10-25
Fix - Row Edit - Row Edit properly updates data entry

### LINKED ISSUE #2838: Bug - Row Edit - Row Editing + Group By does not update the correct data entry
### Description  
When grouping is in place, Row Editing does not update the correct record.

 * igniteui-angular version: 
 * browser: 

### Steps to reproduce  

1. Add 'rowEditing = true' to the 'Grid GroupBy' sample
2. Open row editing on a cell in a row
3. Edit the cell and commit
4. Dismiss row editing via the grouping chip

### Result  
Nothing happens to the edited row. A duplicate row is present.

### Expected result  
The edited row should be updated properly. No duplicates should appear

### Attachments  

ViktorSlavov:
Closes #2838.  

Was not properly updating data entries when data was grouped/sorted - method was using rowIndex to find the data entry, but rowIndex is different when grouping (additional rows inserted), sorting or filtering (different order / less items).
Implement `get_row_index_in_data` in grid API and use it to determine actual index of edited row in data.


projects/igniteui-angular/src/lib/grid/api.service.ts
```diff
@@ -136,6 +136,11 @@ export class IgxGridAPIService {
         }
     }
 
+    public get_row_index_in_data(id: string, rowID: any): number {
+        const grid = this.get(id) as IgxGridComponent;
+        return grid.primaryKey ? grid.data.map((record) => record[grid.primaryKey]).indexOf(rowID) : grid.data.indexOf(rowID);
+    }
+
```
wnvko: add check for the grid - if `this.get(id)` returns undefined next row will throw an error

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -162,6 +162,14 @@ export class GridBaseAPIService <T extends IgxGridBaseComponent> {
         }
     }
 
+    public get_row_index_in_data(id: string, rowID: any): number {
+        const grid = this.get(id) as IgxGridBaseComponent;
+        if (!grid) {
+            return -1;
+        }
+        return grid.primaryKey ? grid.data.map((record) => record[grid.primaryKey]).indexOf(rowID) : grid.data.indexOf(rowID);
```
damyanpetev: Creating a separate array of ID-s lokks like unnecessary overhead. 
Just find the record in question with `.find()`/`.findIndex()`

PR #2847, opened 2018-10-26
feat(grid, chip): add display density DI token to igxGrid and igxChip

### LINKED ISSUE #2804: Feature Request: Add display density DI Token to igxGrid
### Description  
Use the newly implemented DI token. See: https://github.com/IgniteUI/igniteui-angular/pull/2748

 * browser: *

kacheshmarova:
Closes #2804


projects/igniteui-angular/src/lib/core/displayDensity.ts
```diff
@@ -55,20 +62,31 @@ export class DisplayDensityBase {
             default:
                 this._displayDensity = DisplayDensity.comfortable;
         }
+        if (currentDisplayDensity !== this._displayDensity) {
+            const densityChangedArgs: IDensityChangedEventArgs = {
+                oldDensity: currentDisplayDensity,
+                newDensity: this._displayDensity
+            };
+        this.onDensityChanged.emit(densityChangedArgs);
```
bkulov: Styling issue - This line should be one tab to the right.

projects/igniteui-angular/src/lib/core/displayDensity.ts
```diff
@@ -55,20 +62,31 @@ export class DisplayDensityBase {
...
+        }
     }
 
+    @Output()
+    public onDensityChanged = new EventEmitter<IDensityChangedEventArgs>();
+    protected oldDisplayDensityOptions: IDisplayDensity = { displayDensity: DisplayDensity.comfortable };
```
bkulov: Do we need both `protected _displayDensity: DisplayDensity | string;` and `protected oldDisplayDensityOptions: IDisplayDensity = { displayDensity: DisplayDensity.comfortable };`? These two members are holding the exact same information and needs to be in sync.

bkulov: Also `protected displayDensityOptions: IDisplayDensity`.

kacheshmarova: _displayDensity is the value set for the control`s displayDensity input and oldDisplayDensityOptions holds the previous value of the global display density. This mechanism is needed in order to detect changes in the global displayDenistyOptions and trigger event only then for the grid in order to update its values

bkulov: Ok. Thanks.

projects/igniteui-angular/src/lib/core/displayDensity.ts
```diff
@@ -77,10 +95,19 @@ export class DisplayDensityBase {
...
+    constructor(protected displayDensityOptions: IDisplayDensity) {
+        Object.assign(this.oldDisplayDensityOptions, displayDensityOptions);
+    }
 
-    constructor(protected displayDensityOptions: IDisplayDensity) {}
+    public ngDoCheck() {
+        if (this.oldDisplayDensityOptions && this.displayDensityOptions &&
+            this.oldDisplayDensityOptions.displayDensity !== this.displayDensityOptions.displayDensity) {
+            this.oldDisplayDensityOptions = Object.assign(this.oldDisplayDensityOptions, this.displayDensityOptions);
+            this.onDensityChanged.emit();
```
mpavlinov: @kacheshmarova There are no arguments for this emit. Is this on purpose?

kacheshmarova: It was not on purpose,  I added these arguments.

projects/igniteui-angular/src/lib/core/displayDensity.ts
```diff
@@ -16,6 +17,11 @@ export interface IDisplayDensity {
     displayDensity: DisplayDensity;
 }
 
+export interface IDensityChangedEventArgs {
+    oldDensity: DisplayDensity | string | IDisplayDensity;
```
bkulov: It's a bit odd that we emit `onDensityChanged` with different argument types. It should be emitted only with `DisplayDensity` in my opinion.
In addition we could rename `IDisplayDensity` to `IDisplayDensityOptions` to avoid confusion.

PR #2881, opened 2018-10-31
Fix - Selection - Selection header properly checks against filteredData

### LINKED ISSUE #2793: Header checkbox is not updated after deselect and select a filtered-in row
### Description  
Header checkbox is not updated after deselect and select a filtered-in row

 * browser: any

### Steps to reproduce  

1. Open the "grid-selection.sample.html" file and make the "ProductID" column **filterable** with a **number dataType**.
```html
<igx-column [field]="'ProductID'" [editable]="true" [filterable]="true" dataType="number"></igx-column>
```
2. Run the **Grid Selection** dev demo and enable the multiple selection.
3. Filter by the **ProductID** column with any criterion.
4. Check the header checkbox, so all visible rows become selected.
5. Uncheck a single data row.
6. Recheck the same data row.

**NOTE:**
The issue is **not** reproduced if the filtering step is skipped.

### Result  
The header checkbox is **indeterminate**.

### Expected result  
The header checkbox is checked as **true**.

### Attachments  

selectionissue

ViktorSlavov:
Closes #2793.  

`checkHeaderCheckboxStatus` proprely checks if all rows are selected using `filteredData.length` instead of `dataLength` when filtering is in place

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -3659,12 +3659,13 @@ export abstract class IgxGridBaseComponent implements OnInit, OnDestroy, AfterCo
      */
     public checkHeaderCheckboxStatus(headerStatus?: boolean) {
         if (headerStatus === undefined) {
-            this.allRowsSelected = this.selection.are_all_selected(this.id, this.dataLength);
+            const dataLength = this.filteredData ? this.filteredData.length : this.dataLength;
+            this.allRowsSelected = this.selection.are_all_selected(this.id, dataLength);
             if (this.headerCheckbox) {
                 this.headerCheckbox.indeterminate = !this.allRowsSelected && !this.selection.are_none_selected(this.id);
                 if (!this.headerCheckbox.indeterminate) {
                     this.headerCheckbox.checked =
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -3659,12 +3659,13 @@ export abstract class IgxGridBaseComponent implements OnInit, OnDestroy, AfterCo
...
         if (headerStatus === undefined) {
-            this.allRowsSelected = this.selection.are_all_selected(this.id, this.dataLength);
+            const dataLength = this.filteredData ? this.filteredData.length : this.dataLength;
+            this.allRowsSelected = this.selection.are_all_selected(this.id, dataLength);
             if (this.headerCheckbox) {
                 this.headerCheckbox.indeterminate = !this.allRowsSelected && !this.selection.are_none_selected(this.id);
                 if (!this.headerCheckbox.indeterminate) {
                     this.headerCheckbox.checked =
-                        this.selection.are_all_selected(this.id, this.dataLength);
+                        this.selection.are_all_selected(this.id, dataLength);
```
damyanpetev:
```suggestion
                         this.allRowsSelected;
```

PR #2900, opened 2018-11-01
Feat - Grid - Add cancellable grid edit events

### LINKED ISSUE #2897: Grid - Add new events to IgxGrid. Make events Cancelable
### Description  
The grid currently emits the following:
`onEditDone` (both cell and row edit)
`onRowEditDone` (when row edit is confirmed)
`onRowEditCancel` (when row edit is discarded)

 * igniteui-angular version:  all
 * browser: N/A

Implement the following cancelable events:
#### Events (all cancellable)
+	onCellEnterEditMode – when entering edit mode.
-	onEditDone -> onCellEditDone – When edit mode is confirmed ( tab, enter, endEditMode(true) ). 
+	onCellEditCancel – When edit mode is canceled  ( esc, cancel button on row, endEditMode(false) ).
+	onRowEnterEditMode
+	onRowEditDone
+	onRowEditCancel

#### Methods
-	endRowEdit(commit: bool) -> endEditMode(commit: bool)
+	clearSelection

ViktorSlavov:
Closes #2897 .  

Adding the following editing events to grid:
#### Cell
- onCellEditEnter
- onCellEdit
- onCellEditCancel
#### Row
- onRowEditEnter
- onRowEdit
- onRowEditCancel


projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -250,9 +256,11 @@ export class GridBaseAPIService <T extends IgxGridBaseComponent> {
         }
     }
 
-    //  TODO: refactor update_cell. Maybe separate logic in two methods - one with transaction
-    //  and one without transaction
-    public update_cell(id: string, rowID, columnID, editValue) {
+    public get_grid_edit_state(id: string, rowID, columnID, editValue): {
```
ViktorSlavov: @rkaraivanov @damyanpetev 
`update_cell` method in the API is not responsible for calling `escape_cell_inEditMode`, but cancelling `onCellEditDone` should leave the cell in edit mode (**?**), so I've created a method to get the current 'edit state' and I'm using it to pass it from wherever `update_cell` is called, so that the event can be cancelled properly when called either through the API or as a result of UI interactions.
When you can, please take a look and let me know if this seems okay to you.

damyanpetev: I'd personally prefer if `update_cell` did just that - simple update and the control to be before that, but that's a talk for a future time.
So as long as you name it like what it does (get edit event args or something) the method is fine

Issue #2530, opened 2018-08-30
IgxTreeGridComponent


DDavidkov:
Research and develop a tree grid component.

- [x] Research
- [x] Refactor grid code so it can be reused in the treegrid
- [x] Development
- [ ] Test
- [ ] Documentation
- [ ] Samples

Link to [spec](https://github.com/IgniteUI/igniteui-angular/wiki/igxTreeGrid-Specification).

PR #2910, opened 2018-11-02
IgxTreeGrid - Adding more tests

rmkrmk:
#2530

Adding some more expand/collapse tests.


projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-expanding.spec.ts
```diff
@@ -214,6 +214,21 @@ describe('IgxTreeGrid - Expanding / Collapsing', () => {
...
+            treeGrid.expandAll();
+            rows = TreeGridFunctions.getAllRows(fix);
+            expect(rows.length).toBe(12);
+
+            treeGrid.collapseAll();
+            rows = TreeGridFunctions.getAllRows(fix);
+            expect(rows.length).toBe(4);
+        });
+
         it('should emits an event when expanding rows (API)', (done) => {
```
gedinakova: All "should emits" occurrences should be "should emit".

rmkrmk: Done

Issue #2921, opened 2018-11-05
igxTreeGrid Batch Update Implementation


bazal4o:
### Description  
Implement igxTreeGrid Batch update

 * igniteui-angular version: 
 * browser:

PR #2932, opened 2018-11-05
Updating the roadmap

SlavUI:
Closes # .  


bazal4o:
TreeGrid batch update is still in progress (#2921 ). If we mark Tree Grid as completed we might need to add some note about batch update.

PR #2948, opened 2018-11-06
style(row-edit): Adding bigger z-index for the marker

### LINKED ISSUE #2934: In batch editing when change Boolean value form true to false the row is not marked as updated
### Description  
In batch editing when change Boolean value form true to false the row is not marked as updated

 * browser: N/A

### Steps to reproduce  

2. In the second grid ("Grid with rowEditing and with transactions") change value in column ' PINNED DISCONTINUED' and row 1 from true to false and click button Done

### Result  
Row is not marked as updated

### Expected result  
Row should be marked as updated

### Attachments  

desig9stein:
Closes #2934


kdinev:
@desig9stein Please PR the same change to master as well.

desig9stein:
@kdinev There is a PR to master with the same change: https://github.com/IgniteUI/igniteui-angular/pull/3063

Issue #2941, opened 2018-11-06
When cell in first row is selected and press arrow up the selection is cleared


nrobakova:
### Description  
When cell in first wow is selected and press arrow up the selection is cleared

 * igniteui-angular version: 6.2.x
 * browser: N/A

### Steps to reproduce  

1. Open dev sample gridCellEditing
2. In first grid select cell in row 0 and column Product ID
3. Press arrow up key

### Result  
Cell selection is cleared

### Expected result  
Cell should stay selected

PR #2964, opened 2018-11-07
Improve keyboard navigation in filtering

### LINKED ISSUE #2951: Keyboard navigation issues
### Description  
Keyboard navigation issues

 * browser: all

### Steps to reproduce  

1. Open Grid Moving sample, select the first cell and hold ‘shift + tab’ ; the focus should not propagate to checkbox inside grid header;
2. Open Grid Moving sample, start apply filtering over some column; ‘tab’ to the ‘Cancel’ button, press ‘shift + tab’ - the focus should be moved to the previous button.
3. Open Grid Moving sample, start apply filtering over some column - when reset button is disabled should not be able to focus it with ‘tab’ and ‘shift + tab’
4. Open Grid Moving sample, filter the column 'Phone'; select the first cell in the grid; press ‘shift + tab’; notice that when the focus is on the chip inside "Phone" column and press ‘shift + tab’ again scroll is performed and the focus remains over the chip;

sstoyanovIG:
Closes #2951, #2941

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/filtering/grid-filtering-cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/filtering/grid-filtering-cell.component.ts
@@ -107,7 +107,7 @@ export class IgxGridFilteringCellComponent implements AfterViewInit, OnInit {
     @HostListener('keydown.tab', ['$event'])
     public onTabKeyDown(eventArgs) {
         const pinnedColumns = this.filteringService.grid.pinnedColumns;
-        let nextIndex = this.column.visibleIndex + 1 - pinnedColumns.length;
+        const nextIndex = this.column.visibleIndex + 1 - pinnedColumns.length;
 
         if (this.isLastElementFocused()) {
             if (nextIndex < this.filteringService.grid.unpinnedColumns.length &&
@@ -137,7 +137,7 @@ export class IgxGridFilteringCellComponent implements AfterViewInit, OnInit {
     @HostListener('keydown.shift.tab', ['$event'])
     public onShiftTabKeyDown(eventArgs) {
         if (this.isFirstElementFocused()) {
-            let prevIndex = this.column.visibleIndex - 1 - this.filteringService.grid.pinnedColumns.length;
+            const prevIndex = this.column.visibleIndex - 1 - this.filteringService.grid.pinnedColumns.length;
 
             if (this.column.visibleIndex > 0 && !this.navService.isColumnLeftFullyVisible(this.column.visibleIndex - 1)) {
                 eventArgs.preventDefault();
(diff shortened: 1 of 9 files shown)


projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts
```diff
@@ -37,14 +37,15 @@ export class IgxFilteringService implements OnDestroy {
...
     private datePipe = new DatePipe(window.navigator.language);
+    private columnStartIndex = -1;
 
     public gridId: string;
     public isFilterRowVisible = false;
     public filteredColumn = null;
     public selectedExpression: IFilteringExpression = null;
-    public columnToChipToFocus = new Map<string, boolean>();
+    public columnToFocus = null;
+    public focusNext = false;
```
bkulov: Perhaps more meaningful name would be something like `shouldFocusNext` or `shouldFocusRight`.

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -419,8 +420,20 @@ export class IgxGridNavigationService {
                     this.grid.rowEditTabs.last.element.nativeElement.focus();
                     return;
                 }
-                this.navigateUp(currentRowEl, rowIndex,
-                    this.grid.unpinnedColumns[this.grid.unpinnedColumns.length - 1].visibleIndex);
+                if (rowIndex === 0 && this.grid.allowFiltering) {
+                    this.grid.rowList.first.cells.first._clearCellSelection();
+                    const visColLength = this.grid.visibleColumns.length;
+                    if (this.grid.headerContainer.getItemCountInView() === visColLength) {
+                        this.grid.filterCellList.last.focusChip();
```
bkulov: This should be done through the new method in the filtering service.

projects/igniteui-angular/src/lib/grids/filtering/grid-filtering-cell.component.ts
```diff
@@ -309,4 +313,35 @@ export class IgxGridFilteringCellComponent implements AfterViewInit, OnInit {
             this.cdr.detectChanges();
         }
     }
+
+    private ScrollToChip(columnIndex: number, scrollRight: boolean) {
+        this.filteringService.grid.nativeElement.focus({preventScroll: true});
+        this.filteringService.columnToFocus = this.filteringService.grid.visibleColumns[columnIndex];
```
bkulov: Setting `columnToFocus` and `focusNext` happens only in filtering cell component and in navigation service. And in the both places these two fields are set together and are never read. So in order to make the code a little bit easier to use and error prone we could create a method in the filtering service to set those two fields. They are only used in the service so make them private.

sstoyanovIG: focusNext is read in the filtering cell when we decide in which direction we should move the focus.

npaunov:
Several scenarios:
1. Pinned Columns, Go to Column Moving Demo. Click Pin ContactName. Click on the first cell in the grid.
  - Click shift and tab. The focus does not move on the last chip.
  - Click on the last chip, hold shift and tab. The focus does not go to the first chip.
2. Grouped Columns. Group a column, click on the first cell of the grid. Click shift + tab. The focus does not move to the last chip.

PR #2969, opened 2018-11-08
fix(grid): dont exit edit mode when updating through api #2773

### LINKED ISSUE #2773: Template editor does not end edit by calling cell.update() on Enter pressed
### Description  
Template editor does not end edit by calling cell.update() on Enter key press.

 * igniteui-angular version: 6.2.0-beta.3
Note: This is not reproduced on 6.0.3

 * browser: Chrome, IE11

### Steps to reproduce  

1. Run the attached sample
2. Enter edit mode on "startDate" cell
3. Type some characters on the editor
4. Hit enter to end edit mode

### Result  
Edit mode does not end.

### Expected result  
Edit mode ends and edited value is set to the cell.

### Attachments
[igniteui-app1.zip](https://github.com/IgniteUI/igniteui-angular/files/2494483/igniteui-app1.zip)

hanastasov:
Closes #2773

- API update method no longer escapes edit mode in consistency with all other scenarios when cell is not exiting edit mode
- this results in only 1 call of escape_edit_mode, test expectation changed accordingly
- manually updating the cell.editValue when updating through the API


hanastasov:
Consider the following:
In order to be consistent maybe it's good idea when you update row, if there is a cell in edit mode to also stays in edit mode. Also when cancel the 'onEditCell', edit value should not be changed.

hanastasov:
Issue is fixed in version 7, closing.

PR #2978, opened 2018-11-09
Banner component implementation, #2672

### LINKED ISSUE #2672: New Component: Banner
### Description  
The Banner is a mildly interruptive interface that provides contextual information and may only be dismissed by the user (unlike Snackbar and similarly to a Dialog but in a non-modal fashion). The Banner can contain templatable content and a collection of up to two actions.
[Material Banner](https://material.io/design/components/banners.html#)

simple

complex

wnvko:
Closes #2672


projects/igniteui-angular/src/lib/banner/banner.component.html
```diff
@@ -0,0 +1,27 @@
+<igx-expansion-panel #expansionPanel [animationSettings]="animationSettings" (onCollapsed)="onExpansionPanelClose()" (onExpanded)="onExpansionPanelOpen()"
+    [collapsed]="collapsed">
+    <igx-expansion-panel-body>
+        <div class="igx-banner">
+            <div class="igx-banner__message">
+                <div *ngIf="bannerIcon" class="igx-banner__illustration">
```
ViktorSlavov: @desig9stein @simeonoff Could you take a look at the refactored template, when you get a chance?
Can the `igx-banner__illustration` class wrapper be removed? We don't need it to set a gray circle background for the image and that could simplify the template.

simeonoff: @ViktorSlavov Unfortunately the class is used to add other styles apart from setting a shape and background. @desig9stein Could you please remove the oval and the background?

PlamenaMiteva:
Events in the sample do not fire because their names are wrong (should be onOpend/onClosed).

projects/igniteui-angular/src/lib/banner/banner.component.ts
```diff
@@ -0,0 +1,165 @@
+import { Component, NgModule, EventEmitter, Output, Input, ViewChild, ElementRef, ContentChild } from '@angular/core';
+import { IgxExpansionPanelModule } from '../expansion-panel/expansion-panel.module';
+import { AnimationSettings } from '../expansion-panel/expansion-panel.component';
+import { BrowserModule } from '@angular/platform-browser';
```
rkaraivanov: Don't import `BroweserModule` directly in our components. This is an application level import and will break the user's app when this module is lazy loaded for example.

ViktorSlavov: Removed in the latest commit

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes for each version of this project will be documented in this
 
 ## 7.0.0
```
kdinev: This would be for `7.1.0`

ViktorSlavov: Resolved in the latest commit

projects/igniteui-angular/src/lib/banner/banner.component.ts
```diff
@@ -0,0 +1,165 @@
...
+    event?: Event;
+}
+
+export interface BannerCancelEventArgs extends BannerEventArgs, CancelableEventArgs {
+}
+@Component({
+    selector: 'igx-banner',
+    templateUrl: 'banner.component.html'
+})
+export class IgxBannerComponent implements IToggleView {
```
damyanpetev: @ViktorSlavov A top comment for entire component won't be bad, also small snippets on all the public properties.

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.component.ts
```diff
@@ -17,6 +17,10 @@ import { IGX_EXPANSION_PANEL_COMPONENT, IgxExpansionPanelBase } from './expansio
 
 let NEXT_ID = 0;
 
+export interface AnimationSettings {
```
damyanpetev: 👍

damyanpetev: Could event be moved to core at some point.

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.component.ts
```diff
@@ -121,9 +125,17 @@ export class IgxExpansionPanelComponent implements IgxExpansionPanelBase {
...
         if (this.collapsed) {
-            this.expand(evt);
+            this.open(evt);
         } else {
-            this.collapse(evt);
+            this.close(evt);
         }
     }
+
+    open(evt?: Event) {
```
damyanpetev: Are those supposed to be a part of a different PR?

ViktorSlavov: Yes, so that the expansion panel properly implements the IToggle interface.

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.component.ts
ViktorSlavov: Should I move it now? AFAIK the only components that expose animations of this type are the panel and the banner (using the panel), so maybe it's a bit unneeded

projects/igniteui-angular/src/lib/banner/banner.component.ts
ViktorSlavov: Thanks for pointing that out! I've added a top comment to the component. All of the public properties already had one (or were @hidden)

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.component.ts
damyanpetev: No need, this one is an improvement already :)

PR #2984, opened 2018-11-12
Fix blank space when navigating to next page

### LINKED ISSUE #2957: Blank space in rows area after vertical scrolling and navigating to next page
### Description  
Blank space in rows area after vertical scrolling and navigating to next page.

 * browser: any

**NOTES:**
- The issue is reproduced on both the **igxGrid** and the **igxTreeGrid**.
- Performing scrolling again after **step 5** fixes the issue.

### Steps to reproduce  

1. Open the **grid-performance.sample.html** dev demo template and enable paging for the first igx-grid (the one on the top).
```html
<igx-grid ... [paging]="true">
...
```
2. Run the **Grid Performance** dev demo.
3. Change the per page rows to 500.
4. Scroll the grid vertically, so the scroll thumb is a bit below the middle.
5. Go to the next page.

### Result  
There is a blank space in the bottom part of the rows area.

### Expected result  
There is **no** blank space in the bottom part of the rows area.

### Attachments  

gridissue

### LINKED ISSUE #2987: There is an empty area in the igxGrid after scrolling ungrouping and then grouping again
### Description  
There is an empty area (gap) in the igxGrid after scrolling ungrouping and then grouping again.

 * browser: *

### Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid.html#live-updating-demo
2. Scroll down with the mouse wheel
3. Click on the "Grouped" switch
4. Click on the "Grouped" switch again

### Result  
There is an empty area in the grid

### Expected result  
There should be no empty area in the grid and the rows should span the whole view port.

### Attachments  
image

nrobakova:
Closes #2957 
Closes #2987   

Update method _updateVScrollOffset to get correct scrollOffset


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -1098,9 +1098,11 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
     private _updateVScrollOffset() {
         let scrollOffset = 0;
         const vScroll = this.vh.instance.elementRef.nativeElement;
-        scrollOffset = vScroll && parseInt(vScroll.style.height, 10) ?
-        vScroll.scrollTop - this.sizesCache[this.state.startIndex] : 0;
-        this.dc.instance._viewContainer.element.nativeElement.style.top = -(scrollOffset) + 'px';
+        if (vScroll) {
+            scrollOffset = this.fixedUpdateAllRows(this._virtScrollTop);
```
MayaKirova: @nrobakova I don't think it is necessary to trigger a full update on all rows here. The problem seems to be that since the igxForOf data is changed (due to page change), the sizesCache is updated but the startIndex is not. Updating the startIndex on _applyChanges before the embedded views are updated should fix this.

mpavlinov:
NB: Please also verify [2987](https://github.com/IgniteUI/igniteui-angular/issues/2987).

PR #2991, opened 2018-11-12
CellID is reference

### LINKED ISSUE #2636: deselect_item does not remove the cell from the selected cells collection
### Description  
Describe the issue.

### Steps to reproduce  

1.Go to https://stackblitz.com/edit/angular-mkx9q5-cxqoh2
2. Select the first cell on the first row
3. Click the clear selection button

### Result  
Cell stays selected

### Expected result  
Cell to be deselected

Lipata:
Closes #2636

When requested, cellID is now reference, instead of new object. This also means that selection will work, when using cellID as identifier.


Lipata:
Using reference for a cellID (which is object) seems to introduce other problems. Either a new way of saving ID in the selection will be introduced or better solution will be found. This could happen while creating the multi cell selection feature and this problem will be taken into account.

PR #3006, opened 2018-11-13
feat(themes): add theme schemas

simeonoff:
Decouple palettes from the default component themes to make creating component themes more robust.


damyanpetev:
@simeonoff Is this change intended for 6.2 or 7? (master is the current target)

simeonoff:
> @simeonoff Is this change intended for 6.2 or 7? (master is the current target)

It's intended for the next release. Btw, I am not sure if setting the tag to `breaking change` is really needed as the `$preset` param was never documented and it was only introduced in 6.2, without apparent way to use it.

PR #3040, opened 2018-11-19
Prevenitng redundant method calls on value property when initialized with an empty string(6.2.x).

### LINKED ISSUE #3021: Error emitting when IgxDatePicker value property is initialized with empty string.
### Description  
The IgxDatePicker allows the  `value` property` to call getHours`/`getMinutes`/`getSeconds`/`getMilliseconds`  when it  is initialized with an empty string.

### Steps to reproduce  
Bind the `value` property of IgxDatePicker with an empty string and click on the IgxDatePicker to set Date

### Result  
When you click on the Calendar an error is thrown -`this.value.getHours is not a function` and the Date is not set.

### Expected result  
The Date should be set.

### Attachments  
 
errorcausing

datepickererror

HristoP96:
Closes #3021 .


tiliev:
This isn't a bug. `value` should be of type `Date`, not `string`, as described in our docs - [link](https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/classes/igxdatepickercomponent.html#value).

HristoP96:
@tiliev  This issue is reproduced in the [Reactive Forms Sample](https://stackblitz.com/edit/igniteui-reactive-forms?file=app%2Freactive-forms%2Freactive-forms.component.ts) where the datePicker `value` is set initially with `""`. The following screenshots are from debugging the sample when selecting a date:

errorcausing

datepickererror

Also in the sample there is a TimePicker which has also an empty string for initial value but throws no error on time selection. 

After a discussion with @hanastasov  and @rkaraivanov it was decided that this behavior should be logged as a bug.

PR #3132, opened 2018-11-23
Preventing a memory leak  #3033 - 6.2.x

### LINKED ISSUE #3033: igx-grid - Potential memory leak. There are event listeners that are not removed as expected.
### Description  
Some event listeners on `igx-grid` don't be removed as expected. So, it causes memory leak.

 * browser: N/A

### Steps to reproduce.
Here is the `addEventListener`s that have problem.
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.component.ts#L2146
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.component.ts#L2192
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.component.ts#L2196
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.component.ts#L2213
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.component.ts#L4306

### Result  
These event listeners are added with `bind()`.

### Expected result  
They should be implemented like this.
https://stackoverflow.com/questions/11565471/removing-event-listener-which-was-added-with-bind#answer-22870717

PavlovVasil:
Closes #3033.  

 - [x] Bug fix

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -4344,15 +4363,15 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         this.configureRowEditingOverlay(cell.rowID);
         this.rowEditingOverlay.open(this.rowEditSettings);
         this.rowEditPositioningStrategy.isTopInitialPosition = this.rowEditPositioningStrategy.isTop;
-        this.rowEditingOverlay.element.addEventListener('wheel', this.rowEditingWheelHandler.bind(this));
+        this._wheelListener = this.rowEditingOverlay.element.addEventListener('wheel', this.rowEditingWheelHandler.bind(this));
```
kdinev: Same issue here.

PR #3204, opened 2018-11-28
Provide IgxIconService in root

### LINKED ISSUE #3026: Feature request: Custom IgxIcons should be available across modules.
### Description  
Right now custom IgxIcon's are declared using `this.iconService.addSvgIcon("contains", "/assets/images/svg/contains.svg", "filter-icons");` and they are available to be used only at the module were the `this.iconService.addSvgIcon` function was called.

### Result  
Implementation across multiple modules requires to encapsulate and call the `this.iconService.addSvgIcon` method for every added icon even though they already available (at `igx-svg-container` div-node).

This seems to be related to the fact that current "cachedSvgIcons" variable exists only at the current module "IgxIconService" ([here](https://github.com/IgniteUI/igniteui-angular/blob/40f6a05fee2f57517d6f0717dbbce9db13e5a063/projects/igniteui-angular/src/lib/icon/icon.service.ts#L24)).

### Expected result  
Please one of the following:
1. Cached icons should be stored somewhere else where they would be available to be  used by any other module (something like "window")
2. Update your documentation ([here](https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/classes/igxiconservice.html#addsvgicon)) so it includes the note that custom icons will only work at the module were 'addSvgIcon' was called.

bkulov:
Closes #3026

- Provide IgxIconService in root.
- Remove IgxIconModule.forRoot() method.
- Remove deprecated glyphName property of IgxIconComponent.
- Add dev demo demonstrating how svg icons could be used in lazy loaded module.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [x] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes for each version of this project will be documented in this 
 ## 6.2.2
 - `igx-checkbox`:
     - Added a new input property - `disableTransitions`. It allows disabling all CSS transitions on the `igx-checkbox` component for performance optimization.
+- `IgxIconModule`:
+    - **Breaking change** `igxIconService` is now provided in root (providedIn: 'root') and `IgxIconModule.forRoot()` method is deprecated.
+    - **Breaking change** `glyphName` property of the `igxIconComponent` is deprecated.
 
 ## 6.2.1
 ### Features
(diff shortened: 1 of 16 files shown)


bkulov:
We will do this changes only in 7.0.x.

PR #3309, opened 2018-12-04
fix(grid): set min width to header groups programmatically

### LINKED ISSUE #3071: IgxGridHeaderGroupComponent should have preset min width
### Description  
IgxGridHeaderGroupComponent should have preset min width

 * browser:  N/A

### Steps to reproduce  

1.  Switch to  **SAndreeva/header-group-component-6.2.x**
2.  Run **Grid Column Moving** demo
3.  Inspect the the 'ID' column header group

### Result  
**IgxGridHeaderGroupComponent** is sized to 40px as this is the width set on the column. **IgxGridHeaderComponent** and **IgxGridFilteringCellComponent** are 48px as they have paddings.
Therefore cells and headers looked misaligned, the resize handler on the header is not shown at the expected place.

### Expected result  
The  **IgxGridHeaderGroupComponent** should not be able to go below a certain min width.

SAndreeva:
Closes #3071

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -2377,6 +2377,22 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
...
+    get defaultHeaderGroupMinWidth(): number {
+        if (this.isCosy()) {
+            return 32;
+        } else if (this.isCompact()) {
+            return 24;
+        } else {
+            return 48;
+        }
+    }
+
```
simeonoff: Do the listed sizes allow enough space so at least 1 character and the sorting icon is visible at all times?

SAndreeva: No, they only take header cell left and right padding into consideration. Sorting is considered only when resizing a column. I would rather leave it this way. I only want to guarantee that there will not be any cells misalignment on grid initialization, if the user has set set a column width that is smaller that the header paddings mentioned above.

PR #3448, opened 2018-12-14
Fix - Transactions - Gird Transactions + Paging interactions - Master

### LINKED ISSUE #3425: When transactions are enabled and delete a row page is changed to first page
### Description  
When transactions are enabled and delete a row page is changed to first page

 * igniteui-angular version: 7.1.x
 * browser: N/A

### Steps to reproduce  

2. In second grid go to page 2
3. Click on button delete for a row

### Result  
Page is changed to page 1

### Expected result  
Page should not be changed

### Attachments  

ViktorSlavov:
Closes #3425

 - [x] Bug fix

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -3049,7 +3049,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
     /** @hidden */
     public deleteRowById(rowId: any) {
         let index: number;
-        const data = this.gridAPI.get_all_data(this.id, this.transactions.enabled);
+        const data = this.gridAPI.get_all_data(this.id);
         if (this.primaryKey) {
             index = data.map((record) => record[this.primaryKey]).indexOf(rowId);
         } else {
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -3075,10 +3081,13 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         this.deleteRowFromData(rowId, index);
         this._pipeTrigger++;
         this.cdr.markForCheck();
-
+        const dataAfterDelete = !this.transactions.enabled ? data : this.dataWithAddedInTransactionRows;
```
damyanpetev: The api `get_all_data` method from above already does the same check. Also take `this.dataWithAddedInTransactionRows` only when transactions are not enabled?

damyanpetev: Also put a comment you need recalculated data with transactions (if any) after the delete.

projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -3075,10 +3081,13 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
...
         this.refreshSearch();
-        if (data.length % this.perPage === 0 && this.isLastPage && this.page !== 0) {
-            this.page--;
+        if (dataAfterDelete.length % this.perPage === 0 && this.isLastPage && this.page !== 0) {
```
damyanpetev: These checks and the one below could be simplified to something like 
```suggestion
        if (dataAfterDelete.length % this.perPage === 0 && dataAfterDelete.length / this.perPage - 1 < this.page) {
```
where `dataAfterDelete.length / this.perPage` is pretty much a recalculated `totalPages`.
Actually, when you have transactions, the `onStateChange` emitter will actually retrigger the pipes and `totalPages` is up to date at this point and you can probably use it

PR #3469, opened 2018-12-18
IgxForOf Performance Optimization

### LINKED ISSUE #3083: Scrolling igxGrid on iPad Poor Performance
### Description  

Scrolling performance on iPad very poor.  Open 
https://www.infragistics.com/angular-demos/finjs-sample on iPad, attempt to scroll.

 * igniteui-angular version: 6.2.0.
 * browser:  Chrome, Safari, Edge on iPad.

### Steps to reproduce  
1. run 
https://www.infragistics.com/angular-demos/finjs-sample on iPad.

### Result  

Expect smooth scrolling.  On Chrome on iPad, it is very poor.  On Safari on iPad, it appears to do well, but becomes choppy and chunky and doesn’t scroll well after several good scrolls.

### Expected result  

Buttery smooth, just like Desktop.

### Attachments  

ChronosSF:
Closes #3083 

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 
### Bugs:
- [x] Ctrl + Right/Left in igxGrid throws error (Column Resizing sample and others)
- [x] Combo keyboard navigation not going to the bottom of the dropdown
- [x] Combo focus style class applying on wrong rows when scrolling down
- [x] Disabling summaries and navigating down doesn't scroll to next cell (Grid Summary sample)
   > this is separate issues. The cache assumes 32 px height for the rows but the rows are expanded because of the long form date to 48 px which means that the ```scrollTo``` calculations are not correct


rkaraivanov:
It seems this breaks the igx-combo keyboard navigation.

1. Open the combo dev sample.
2. Focus the combo input and hold down the arrow down key.
3. At a certain point the navigation stops but no error is thrown.

skrustev:
There seem to be still wrongly applied focus style when opening the Combo dropdown that when scrolled stays with the row template even though the row is different..

ChronosSF:
If we end up having an `index` property, it should represent the data index and have another name for the visible index. Virtualization shouldn't be something users have to take into account. The first assumption is that index is index in the data as `ngForOf` and in turn `igxForOf` expose a property of the same name already.

Additionally `IgxRowComponent` already has an index property that is the data index so there will be discrepency in naming if we do it as you suggest. I'll mark the change as a breaking one.

damyanpetev:
@ChronosSF I sort of agree - the index normally refers to the data index of `ngForOf` and that's exactly what it is currently. Yes, visualization should be transparent, but the drop down doesn't really support one without additional changes anyway and a property split for visible index doesn't make it too transparent either. The change can be combo-specific too.

I don't mind actually removing the visual bit eventually (as @ViktorSlavov said) but I'd much prefer to have it deprecated first before an actual breaking change. 

How about we deprecate the current getter functionality and add the `@Input` at the same time (fallback to old functionality when not bound)?

ChronosSF:
@damyanpetev , @kdinev please do a final review

PR #3487, opened 2018-12-19
Add locale property to igx-grid

### LINKED ISSUE #799: Use browser locale when locale is not explicitly set
### Description  
When a locale is not explicitly set our components use the "en" locale, rather than checking for the browser locale and using it instead of the "en" one

### Steps to reproduce  

1. Create a calendar
2. Don't set an option for locale
3. Calendar shows with "en" locale

### Result  
"en" locale is used whenever the developer has not set a locale explicitly

### Expected result  
If no locale is set, the control should try to use the browser locale and only if this fails for some reason the locale should be set to "en"

### LINKED ISSUE #3455: Feature Request : locale property on a grid level
### Description  

Currently filteringService and grid's filtering UI are formatting dates using browser's window.navigator.language. IgxDatePicker and IgxInputGroup components, which are used in the filtering UI are flexible and support different locales. Currently there is no easy way to change the locale in the grid.

So whenever a user is using a language different from English on his browser, we need to provide the corresponding angular locale (as explained [here](https://angular.io/guide/i18n#i18n-pipes)) in order to avoid errors for missing pipes during the formatting.

To avoid the import of numerous amount of angular locales when a user may need only one of those and/or to avoid language inconsistency when the website/webapp isn't translated in the user's browser's language. It would be preferable to let the developer choose the language for the whole grid.

### Potential Solution (credits goes to Borislav Kulov)
Provide a locale property on a grid level that will ensure that all grid's child components will be using that same locale and the grid will stay in the desired language even if the browser's locale is different. 

### Expected result  
A `[locale]` input on the `<igx-grid>` tag that will behave as it does on the `<igx-date-picker>` tag
Eventually, the default value should be 'null' in order to prevent this change to be a "breaking change" and still allow people to custom the locale child by child

### Attachments
Initial Discussion
https://www.infragistics.com/community/forums/f/ignite-ui-for-angular/119335/manually-programatically-set-locale-for-igxgrid-s-filter-s-datepickers/529961#529961

sstoyanovIG:
Closes #3455, Closes #799   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes for each version of this project will be documented in this 
     - Added a new `igxToolbarCustomContent` directive which can be used to mark an `ng-template` which provides a custom content for the IgxGrid's toolbar ([#2983](https://github.com/IgniteUI/igniteui-angular/issues/2983))
     - Summary results are now calculated and displayed by default for each row group when 'Group By' feature is enabled.
     - `clearSummaryCache()` and `recalculateSummaries()` methods are deprecated. The grid will clear the cache and recalculate the summaries automatically when needed.
+	- `locale` property added. If not set, it returns browser's languge. All child components will use it as locale.
     - **Breaking change** `IgxSummaryOperand.operate()` method is called with empty data in order to calculate the necessary height for the summary row. For custom summary operands, the method should always return an array of `IgxSummaryResult` with proper length.
 - `IgxIconModule`:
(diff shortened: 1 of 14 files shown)


npaunov:
Consider the following scenario:
The igxGrid is included in an app. The grid contains date and/or number columns. If the app is opened in any browser with language not set to 'en', the data in those columns will not be shown and cells rendering will break on scrolling horizontally. 
The reason for this behavior is that the localized strings (e.g.  @angular/common/locales/fr) for this particular language are not included in the app.

PR #3515, opened 2019-01-03
fix(Grid): #3513 Fixed items' percentage width calculations.

### LINKED ISSUE #3513: Horizontal scrollbar is not shown when column's width is set to a percentage value.
### Description  
Describe the issue.

 * igniteui-angular version: 7.1.x, 6.2.x
 * browser: N/A

### Steps to reproduce  

1. Start Dev Demos
2. Go to Grid Selection Demo
3. Set the first column width to 80%

### Result  
The last column gets partially out of view but the horizontal scrollbar is not shown.

### Expected result  
Horizontal scrollbar should be shown as the sum of columns' widths becomes greater than the grid's width.

### Attachments  
image

gedinakova:
Closes #3513   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
@@ -903,7 +903,7 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
                 size = parseInt(this.igxForItemSize, 10) || 0;
                 this.heightCache.push(size);
             } else {
-                size = parseInt(items[i][dimension], 10) || 0;
+                size = this._calcItemWidth(items[i][dimension], this.igxForContainerSize) || 0;
             }
             totalSize += size;
             this.sizesCache.push(totalSize);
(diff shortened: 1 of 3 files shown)


MayaKirova:
@gedinakova The igxFor does not handle width in % since if widths are in % the expected result would be for all columns to fit in the available space, hence there would be no virtualization (everything will be rendered as is).

In the following setup of the sample:
<igx-grid [width]="'800px'" …
        <igx-column [field]="'ProductID'" [editable]="true" [width]="'80%'"></igx-column>
        <igx-column [field]="'ProductName'" ></igx-column>
        <igx-column [field]="'UnitsInStock'"></igx-column>

I would expect the grid to resolve this to the following widths: 80%, 10%, 10% (which if set renders correctly).
Currently somewhere the grid calculations fail and it renders: 80%, 360px, 360x. The fix should probably be applied to the grid so that it resolves the % widths correctly.

gedinakova:
@MayaKirova Thanks, I will review the calculations made by the grid. Please, note, though, that the calculated items' widths would still be wrong because with a value of '80%' the igxFor would get just the 80 and add it to the total width, which also seems wrong.

MayaKirova:
@gedinakova @ChronosSF
I don’t think % is a valid value for it in the first place.
In general none of the igxFor inputs allow setting a value in %, all of the size related inputs expect number or px (igxForContainerSize, igxForItemSize), which is explicitly stated in their API.
Probably a better way to handle this would be for the grid to transform the value to px before passing it to the igxFor (like it does for the igxForContainerSize in case the grid width is set in %). In that case it may be possible to also support mixed widths correctly (some values in pixels, some in %).

As for what the expected result should be if someone sets % value in a stand-alone igxFor directive, I suppose either it should throw an error as it is invalid setting, or simply set its size to 0. Since none of the other size calculation expect or handle scenarios where % is used it doesn’t really make sense for the widths to be an exception and convert the value to pixels.

 Let me know what you think.

gedinakova:
As discussed reassigning to @ChronosSF to review the grid's code as opposed to making such a change in IgxForOf.

PR #3521, opened 2019-01-04
Add locale property to igx-grid

### LINKED ISSUE #799: Use browser locale when locale is not explicitly set

### LINKED ISSUE #3455: Feature Request : locale property on a grid level

sstoyanovIG:
Closes #3455, Closes #799

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 All notable changes for each version of this project will be documented in this file.
 
 ## 7.1.1
+### Features
+- `igxGrid`
+    - `locale` property added. If not set, it returns browser's languge. All child components will use it as locale.
```
bkulov: Typo `languge`.

projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts
```diff
@@ -38,7 +38,7 @@ export class IgxFilteringService implements OnDestroy {
     private columnToExpressionsMap = new Map<string, ExpressionUI[]>();
     private filterPipe = new IgxGridFilterConditionPipe();
     private titlecasePipe = new TitleCasePipe();
-    private datePipe = new DatePipe(window.navigator.language);
+    private datePipe = new IgxDatePipeComponent(window.navigator.language);
```
bkulov: I think the date pipe should use `grid.locale` instead of `window.navigator.language`.

projects/igniteui-angular/src/lib/grids/grid.common.ts
```diff
@@ -553,9 +553,13 @@ export class IgxDatePipeComponent extends DatePipe implements PipeTransform {
         // D.P. constructor duplication due to es6 compilation, might be obsolete in the future
         super(locale);
     }
-    transform(value: any): string {
+    transform(value: any, locale: string): string {
         if (value && value instanceof Date) {
-            return super.transform(value);
+            if (locale) {
+                return super.transform(value, 'mediumDate', undefined, locale);
```
bkulov: You could extract the default format string as constant.

projects/igniteui-angular/src/lib/grids/grid/grid-filtering-ui.spec.ts
```diff
@@ -2752,6 +2754,33 @@ describe('IgxGrid - Filtering Row UI actions', () => {
         filteringRow = fix.debugElement.query(By.directive(IgxGridFilteringRowComponent));
         expect(filteringRow).toBeNull();
     }));
+
+    it('should correctly apply locale to datePicker.', fakeAsync(() => {
```
bkulov: Add similar test for the newly added `locale` property in the grid.

PR #3529, opened 2019-01-07
Adding ng-content to IgxSnackbar - 7.1.x

mtsvyatkova:
Closes #  3328

 - [ ] Bug fix
 - [X] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [X] This PR includes `feature/README.MD` updates for the feature docs

### DIFF
--- a/projects/igniteui-angular/src/lib/snackbar/snackbar.component.spec.ts
+++ b/projects/igniteui-angular/src/lib/snackbar/snackbar.component.spec.ts
@@ -11,7 +11,8 @@ describe('IgxSnackbar', () => {
     beforeEach(async(() => {
         TestBed.configureTestingModule({
             declarations: [
-                SnackbarInitializeTestComponent
+                SnackbarInitializeTestComponent,
+                SnackbarCustomContentComponent
             ],
             imports: [
                 BrowserAnimationsModule,
@@ -35,7 +36,6 @@ describe('IgxSnackbar', () => {
         expect(snackbar.displayTime).toBe(4000);
         expect(snackbar.autoHide).toBeTruthy();
         expect(snackbar.isVisible).toBeFalsy();
-        expect(snackbar.actionText).toBeUndefined();
 
         expect(domSnackbar.id).toContain('igx-snackbar-');
         snackbar.id = 'customId';
@@ -67,6 +67,7 @@ describe('IgxSnackbar', () => {
 
         snackbar.show();
 
+        fixture.detectChanges();
         expect(snackbar.isVisible).toBeTruthy();
         expect(snackbar.autoHide).toBeFalsy();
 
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/snackbar/snackbar.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="igx-snackbar" *ngIf="isVisible" (@slideInOut.start)="snackbarAnimationStarted($event)" (@slideInOut.done)="snackbarAnimationDone($event)"
     [@slideInOut]="isVisible">
     <span class="igx-snackbar__message" [@fadeInOut]="isVisible">{{ message }}</span>
+    <ng-content *ngIf="!message"></ng-content>
```
gedinakova: I am not sure if we should refrain from showing the content when a message is specified. Although, they can easily add some text content along with the other content they need in. @kdinev , any thought?
I believe, however, we should ngIf the message span, especially now that we can have content without a message!

mtsvyatkova: I made the changes suggested by @gedinakova.

projects/igniteui-angular/src/lib/snackbar/snackbar.component.spec.ts
```diff
@@ -90,6 +91,64 @@ describe('IgxSnackbar', () => {
...
+        expect(customContent).toBeTruthy('Custom content is not found');
+
+        // Verify the custom content element is on the left side of the button
+        const button = fixture.debugElement.query(By.css('.igx-snackbar__button'));
+        const buttonRect = (<HTMLElement>button.nativeElement).getBoundingClientRect();
+        const customContentRect = (<HTMLElement>customContent.nativeElement).getBoundingClientRect();
+        expect(customContentRect.right <= buttonRect.left).toBe(true, 'The custom element is not on the left of the button');
+    });
+});
+
```
gedinakova: We should probably add a test with both a message and some content.

projects/igniteui-angular/src/lib/snackbar/snackbar.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="igx-snackbar" *ngIf="isVisible" (@slideInOut.start)="snackbarAnimationStarted($event)" (@slideInOut.done)="snackbarAnimationDone($event)"
     [@slideInOut]="isVisible">
-    <span class="igx-snackbar__message" [@fadeInOut]="isVisible">{{ message }}</span>
+    <span class="igx-snackbar__message" [@fadeInOut]="isVisible" *ngIf="message">{{ message }}</span>
+    <ng-content></ng-content>
```
kdinev: Shouldn't the `ng-content` be wrapped by a container with the `igx-snackbar__message` class? or like a content class, which provides default alignment and positioning?

gedinakova: Not sure about that ... We may want to have some igx-snackbar__content class, but assigning the message class to some custom content may not be the best idea in all scenarios.

kdinev: @gedinakova @mtsvyatkova You should sync this with @simeonoff

simeonoff: @kdinev I am with Konstantin on this one. Regardless of whether you want to use the message input attribute, or simply place any content between the opening and closing brackets(preferred), the _message_ should be wrapped in an element with class igx-snackbar__message.

If you have time, you can also do away with the whole wrapping div in the template, hostbinding the 'igx-snackbar' class. Animations will have to be reimplemented, though.

projects/igniteui-angular/src/lib/snackbar/snackbar.component.spec.ts
mtsvyatkova: The tests are updated so they check if the elements are positioned in the correct order when a message, a custom content and a button are set, and also that the message span is not in the DOM when the message is empty.

projects/igniteui-angular/src/lib/snackbar/snackbar.component.html
mtsvyatkova: The message is now wrapped in a container with the custom content with the igx-snackbar__message class. 
The removal of the wrapping div element will be researched and probably included in another PR.

PR #3539, opened 2019-01-08
fix(time picker): IE cursor flickering on spinning

SAndreeva:
 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -708,6 +712,18 @@ export class IgxTimePickerComponent implements
         this._dialogOverlaySettings = {};
     }
 
+    /**
+     * @hidden
+     */
+    public ngAfterViewInit(): void {
+        if (this.mode === TimePickerInteractionMode.dropdown) {
+            fromEvent(this.input.nativeElement, 'keydown').pipe(
+                throttle(() => interval(0, animationFrameScheduler)),
```
bkulov: There is no sense to throttle events for 0 ms. I think setting the interval to 300 ms would make more sense.

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -1600,7 +1616,10 @@ export class IgxTimePickerComponent implements
             displayVal = this._formatTime(this.value, this.format);
         }
 
-        this.displayValue = this.inputFormat.transform(displayVal);
+        // minor hack for preventing cursor jumping in IE
+        const normalizedValue = this.inputFormat.transform(displayVal);
+        this._renderer.setElementProperty(this.input.nativeElement, 'value', normalizedValue);
```
bkulov: Please update the displayValue private field here too. This will ensure that if someone is using the getter the value will be correct.

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
SAndreeva: **interval(0, animationFrameScheduler)** creates an observable that emits one event every time the browser is ready to display a new frame and then the throttle operator will emit a value from that source observable.

PR #3547, opened 2019-01-08
Add prompt for enabling polyfills 

### LINKED ISSUE #3528: ng add improvements
### Description  
When a user runs ng add igniteui-angular command the following improvements might be done. 
- Notify if the ng/cli versions are lower than the one coming from the package.json of igniteui-angular package
- Prompt to enable polyfills for IE

jackofdiamond5:
Closes #3528 

 - [x] New functionality

### Checklist:


projects/igniteui-angular/package.json
```diff
@@ -63,6 +63,7 @@
   "peerDependencies": {
     "@angular/common": "^7.0.3",
     "@angular/core": "^7.0.3",
+    "@angular/cli": "~7.0.5",
```
damyanpetev: Why would the Angular CLI be a peer dep?

projects/igniteui-angular/schematics/ng-add/schema.json
```diff
@@ -0,0 +1,15 @@
...
+    "$schema": "http://json-schema.org/schema",
+    "id": "igniteui-angular-ng-add",
+    "title": "IgniteUI for Angular ng-add schematic",
+    "type": "object",
+    "properties": {
+        "polyfills": {
+            "type": "boolean",
+            "description": "Add polyfills for IE.",
+            "default": "false",
+            "x-prompt": "Would you like to add polyfills for IE?"
```
damyanpetev: those are already present in the project, correct? So more like "enable" rather than "add".

projects/igniteui-angular/schematics/ng-add/index.ts
```diff
@@ -1,8 +1,9 @@
 import { WorkspaceSchema } from '@angular-devkit/core/src/workspace';
-import { chain, Rule, SchematicContext, Tree, SchematicsException } from '@angular-devkit/schematics';
+import { chain, Rule, SchematicContext, Tree, SchematicsException, FileDoesNotExistException } from '@angular-devkit/schematics';
 import { NodePackageInstallTask, RunSchematicTask } from '@angular-devkit/schematics/tasks';
 // tslint:disable-next-line:no-submodule-imports
 import { getWorkspace } from '@schematics/angular/utility/config';
+import { DependencyNotFoundException } from '@angular-devkit/core';
 
 export interface Options {
   [key: string]: any;
```
damyanpetev: You can add the `polyfills` prop to the options schema btw

projects/igniteui-angular/schematics/ng-add/index.ts
```diff
@@ -49,6 +52,25 @@ function LogIncludingDependency(context: SchematicContext, pkg: string, version:
...
 
+function promptAddPolyfills(tree: Tree, options: Options) {
+  if (options['polyfills']) {
+    const targetFile = 'src/polyfills.ts';
+    if (!tree.exists(targetFile)) {
+      throw new Error(`${targetFile} not found in the src folder.`);
+    }
+
+    // Match all commented import statements that are core-js/es6/*; core-js/es7/*; web-animations-js
+    const pattern = /\/{2}\s{0,}\w+\s{0,}\'(\w+\-\w+\/(\w+)(6|7)\/.+|\w+\-\w+\-\w+\';)/g;
```
damyanpetev: That's way too broad. You _can_ limit it by using core-js in the regex and a super simple text replace for the web animations plugin.

Keep in mind there's a `'core-js/es6/array'` import that's not related and there might be more. For this specific usage even a pattern list of the specific imports doesn't seem like a bad option.

projects/igniteui-angular/schematics/ng-add/index.ts
```diff
@@ -49,6 +52,25 @@ function LogIncludingDependency(context: SchematicContext, pkg: string, version:
   context.logger.log('info', `Including ${pkg} - Version: ${version}`);
 }
 
+function promptAddPolyfills(tree: Tree, options: Options) {
+  if (options['polyfills']) {
+    const targetFile = 'src/polyfills.ts';
+    if (!tree.exists(targetFile)) {
+      throw new Error(`${targetFile} not found in the src folder.`);
```
damyanpetev: This could be a warning - you know, on the off chance we can't get to the file, just let the user know the polyfils might need to be tweaked and carry on, rather than erring out the whole schematic

projects/igniteui-angular/package.json
damyanpetev: Nvm, saw you use it to check versions. Still a nope, though.
You can still infer the CLI version is outdated if core & common are.

projects/igniteui-angular/schematics/ng-add/index.ts
```diff
@@ -145,6 +167,47 @@ function addPackageToJsonDevDependency(tree: Tree, pkg: string, version: string)
...
+  const ngKey = '@angular/core';
+  const ngCliKey = '@angular/cli';
+  const ngProjVer = getDependencyVersion(ngKey, tree);
+  const ngCliProjVer = getDependencyVersion(ngCliKey, tree);
+
+  const igAngularVer = igPackageJson.peerDependencies[ngKey];
+  const igAngularCliVer = igPackageJson.peerDependencies[ngCliKey];
+
+  if (ngProjVer < igAngularVer) {
+    context.logger.log('warn',
```
damyanpetev: There should be a `context.warn` shortcut?

projects/igniteui-angular/schematics/ng-add/index.ts
```diff
@@ -145,6 +167,47 @@ function addPackageToJsonDevDependency(tree: Tree, pkg: string, version: string)
...
+      `Angular version mismatch. Installed version: ${ngProjVer} - igniteui-angular version: ${igAngularVer}`);
+  }
+  if (ngCliProjVer < igAngularCliVer) {
+    context.logger.log('warn',
+      `Angular CLI version mismatch. Installed version: ${ngCliProjVer} - igniteui-angular version: ${igAngularCliVer}`);
```
damyanpetev: I'd check `@angular/common` instead of CLI.
The message could be a bit more descriptive too. Something explaining the added Ignite UI for Angular package is build against a newer version of angular[/core] and lower versions were found in the current project.
Could also suggest a way to remedy that would be to use `ng update`

projects/igniteui-angular/schematics/ng-add/index.ts
```diff
@@ -21,6 +23,25 @@ function logSuccess(options: Options): Rule {
   };
 }
 
+function LogIncludingDependency(context: SchematicContext, pkg: string, version: string): void {
```
damyanpetev: Would appreciate lower cased function name :)

projects/igniteui-angular/schematics/ng-add/index.ts
```diff
@@ -41,12 +62,20 @@ function addDependencies(options: Options): Rule {
...
 
-function LogIncludingDependency(context: SchematicContext, pkg: string, version: string) {
-  context.logger.log('info', `Including ${pkg} - Version: ${version}`);
+/**
+ * The import is needed so that the igxGrid can render under IE.
+ * - https://github.com/IgniteUI/igniteui-cli/issues/344
+ */
+function addIgxGridSupportForIe(polyfillsData: string): string {
+  const targetImport = 'import \'core-js/es6/set\';';
+  const lineToAdd = 'import \'core-js/es7/object\';';
```
damyanpetev: I'd actually include the comment about why this poly is needed as well

projects/igniteui-angular/schematics/ng-add/index.spec.ts
```diff
@@ -99,4 +129,19 @@ describe('schematics', () => {
     expect(Object.keys(pkgJsonData.devDependencies).filter(k => k.includes('igniteui-cli')).length).toBeGreaterThan(0);
     expect(Object.keys(pkgJsonData.dependencies).filter(k => k.includes('igniteui-cli')).length).toBe(0);
   });
+
+  it('should properly add polyfills', () => {
+    expect(tree.readContent('src/polyfills.ts')).toBe(polyfills);
+    runner.runSchematic('ng-add', { polyfills: true }, tree);
+    expect(tree.readContent('src/polyfills.ts')).not.toBe(polyfills);
```
damyanpetev: While technically true the file shouldn't remain the same, it's not really an exact check the schematic did the expected job :)

projects/igniteui-angular/schematics/ng-add/index.spec.ts
```diff
@@ -19,16 +40,25 @@ describe('schematics', () => {
...
     tree = new UnitTestTree(new EmptyTree());
     tree.create('/angular.json', JSON.stringify(ngJsonConfig));
     tree.create('/package.json', JSON.stringify(pkgJsonConfig));
+    tree.create('src/polyfills.ts', polyfills);
+    populatePackageJson();
   });
 
+  function populatePackageJson() {
+    const pkgJson = JSON.parse(tree.read('/package.json').toString());
+    pkgJson.dependencies['@angular/common'] = '^7.0.3';
```
damyanpetev: Probably not a great idea to hardcode those versions (you can grab them from the current project dependencies).
Also might want to have one test with a lower versions to verify the warning.

jackofdiamond5:
The version mismatch message is moved in the [schematic](https://github.com/IgniteUI/igniteui-cli/pull/436) of igniteui-cli.

projects/igniteui-angular/schematics/tsconfig.json
```diff
@@ -4,6 +4,6 @@
         "module": "commonjs",
         "sourceMap": false,
         "declaration": true,
-        "outDir": "../../../dist/igniteui-angular/schematics/ng-add"
+        "outDir": "../../../dist/igniteui-angular/schematics/"
```
damyanpetev: Umm, @jackofdiamond5  Are you sure about this one?
`collection.json` still says `"factory": "./ng-add/index"`, i.e. this should break the schematic.

jackofdiamond5: The path should not lead to the `ng-add` folder as the gulp `copy-schematics` task would place the built files explicitly within the `ng-add` folder. Which basically means that if you were to add an additional schematic, to the project and then build it, it would go under `ng-add`.

PR #3559, opened 2019-01-09
igxGrid - fix last column visual cut off

### LINKED ISSUE #1115: The selection in the last grid column does not span in the whole cell.
### Description
The selection in the last grid column does not span in the whole cell.

### Steps to reproduce
Place a grid, where columns have their width set, but the columns' total width is less than the grid width.

### Expected result  
Selection to span the whole space.

### Attachments  
39047063-3baa78e0-44a1-11e8-92cb-a11eaba4130e

### LINKED ISSUE #3396: Last (right-aligned) column is cut off when no widths are set for the columns
### Description  
Last (right-aligned) column is cut off when no widths are set for the columns.

 * igniteui-angular version: 7.1.x, 7.0.x
 * browser: any

**NOTES:**
-  The issue is reproduced for both the IgxGrid and the IgxTreeGrid.

### Steps to reproduce  

1. Open the **grid-cellEditing.component.html** file and do the following changes for the top-most grid:
- set width to 900px instead of 800px
- comment the templated column
- remove the width setting for every column
2. Run the "Grid Cell Editing" dev demo.
3. Move the UnitsInStock column to the right-most position.

### Result  
Column cells are cut off and there is no horizontal scrollbar.

### Expected result  
Column cells are **not** cut off.

### Attachments  

cutoff

SAndreeva:
Closes #3396 
Closes #1115 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### Scenarios to be tested:
- grid with small number of records when no vertical scrollbar is displayed;
- grid with columns which widths are not defined;
- grid where the sum of the columns' widths equals the overall width of the grid;
- keyboard navigation - data cells, summary cells, filtering cells;


ChronosSF:
I still stand by my opinion that #1115 is not a bug and the simplest resolution for #3396 is to have right padding on the last cell based on some check like 

*if (vertical-scrollbar is visible && horizontal-scrollbar is invisible) then padding-right: 18px*

What is the reason to add a new scrollTo method and a service for this fix?

SAndreeva:
@ChronosSF,  I do not insist that #1115 is a bug, especially considering the fact that the grid has changed, since this behavior was initially reported. However, from UX point of view, that might seem a bit inconsistent as the header and data cells have different widths. If I have set the column width to be 200px, I expect that all artifacts belonging to that column have that width. And there are still minor cases where we have undesirable behaviors, for example when auto sizing the last column via double click or in the case reported in #3396. 
As for the **ScrollTo()** method, KB navigation calls it. With the newly introduced change of adding 18px to the horizontal  **sizesCache**, the  **scrollLeft** is not calculated correctly in the original method. I haven’t actually added a new  **ScrollTo()** method, I have only overridden the base one in the derived  **IgxGridForOfDirective** directive as all these changes are meant only for the grid and I do not want to touch anything in the base  **IgxForOfDirective**. Most of the logic is the same as in the base method, I only added those 18px in certain cases.
The service is used to allow communication between the horizontal and vertical instances of the directive, because I want to add 18px to the horizontal cache only when the grid is virtualized vertically and a vertical scrollbar is present. 

I do not mind changing the approach, especially if there is a cleaner and easier way. Moreover my change may lead to other undesirable 'effects'.

SAndreeva:
Closing this PR after discussion - a new approach will be implemented.

PR #3635, opened 2019-01-16
feat(grid): Add isLoading option and progress indicator for remote data

### LINKED ISSUE #3600: IgxGrid should allow column auto-generation with remote data (data unavailable on init)
### Description  
Currently column autogeneration (```autogenerate``` option) is attempted only once on init which does not work if the data is to be loaded on demand after grid initialization.

 * browser: n/a

### Steps to reproduce  

1. Set `autogenerate` to true on any of the remote samples and remove the column definitions

### Expected result  
Columns to be generated and data to be bound when the remote call is completed.

### Attachments  
n/a

***Note***: the exact required behavior will be available in the grid's spec

dkamburov:
Closes #3600

 - [ ] Bug fix
 - [x] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


mpavlinov:
I guess you need to update the Changelog.md, feature/README.MD and the spec.

projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -3936,6 +3991,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         });
 
         this.columnList.reset(columns);
+        this.shouldGenerate = false;
```
ChronosSF: Shouldn't we only do this if data.length > 0 ?

This is called on content init when there may not be data yet. You will make the flag false forcing users to have to set it to true again when just having a single bind from a remote source. In this case simply putting autoGenerate to true should suffice. 

The flag should only be raised again when **re**-generating the columns.

dkamburov: Ok, I agree. Will add it.

PR #3701, opened 2019-01-23
Toggle action should provide its element as target

### LINKED ISSUE #3636: Issue with reusing a positioning strategy instance across drop downs
### Description  
When I try to reuse a positioning strategy instance across drop downs, the target of the positioning is always set to the element that triggers the first drop down opening.

 * igniteui-angular version: 7.1.1
 * browser: N/A

### Steps to reproduce  

1. Define multiple drop downs
2. Define multiple elements that toggle each drop down
3. Define a single overlaySettings instance and provide it to all elements
4. Open the drop downs one by one

### Result  
Depending on which drop down is opened first, all the rest open against the first drop down target.

### Expected result  
Each drop down opens according to its own target,

wnvko:
If you add igxToggleAction on some element it will provide the element as target for the position startegy.

Closes #3636

 - [x] Bug fix

### Checklist:


projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.spec.ts
```diff
@@ -464,14 +501,13 @@ export class IgxOverlayServiceComponent {
...
-    constructor(public overlay: IgxOverlayService) {}
+    constructor(public overlay: IgxOverlayService) { }
 }
 
-
 @Component({
     template: `
         <button igxToggleAction="toggleID">Open/Close Toggle</button>
-        <div igxToggle id="toggleID">
+        <div style igxToggle id="toggleID">
```
damyanpetev: Leftover `style` ?

projects/igniteui-angular/src/lib/directives/toggle/README.md
```diff
@@ -53,7 +53,7 @@ handlers when the toggle is opened and respectively closed.
 
 # IgxToggleAction Directive
 
-The **IgxToggleAction** provides a way for user to Open/Close(toggle) every Component/Directive which implements **IToggleView** interface by providing the reference to this particular Component/Directive or ID which is registered into **IgxNavigationService**. It is also applicable upon **IgxToggle**
+The **IgxToggleAction** provides a way for user to Open/Close(toggle) every Component/Directive which implements **IToggleView** interface by providing the reference to this particular Component/Directive or ID which is registered into **IgxNavigationService**. It is also applicable upon **IgxToggle**. When applied to a Component/Directive **IgxToggleAction** forces **IToggleView**'s target to be the element where **IgxToggleAction** is set.
```
damyanpetev: It could be placed on plain HTML elements too, right? With much the same effect. So this should probably state that the toggle action will always set its host element as the position strategy target.

damyanpetev:
I have one question though - would this account for sharing the overlay settings/position strategy (per the original issue) if multiple items are open at once?
Not sure if it's an issue at all as the toggle action directive is there for convenience and its usage is completely optional, but just realized this might cause issues with the shared strategy and had to ask.

projects/igniteui-angular/src/lib/directives/toggle/README.md
wnvko: done!

PR #3811, opened 2019-02-01
fix(drop-down): Calling open() only on a collapsed toggle #3810

### LINKED ISSUE #3810: Calling open() on an already opened IgxDropDown replays the opening animation
### Description  
The title describes the whole issue.

 * igniteui-angular version: 7.1.3
 * browser: N/A

kdinev:
Closes #3810 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them-


bazal4o:
As discussed we need to return back the check into igxToggle and remove it from all the components which are using the toggle and make such check into the open/close methods.

projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.ts
```diff
@@ -173,6 +173,10 @@ export class IgxToggleDirective implements IToggleView, OnInit, OnDestroy {
      * ```
      */
     public open(overlaySettings?: OverlaySettings) {
+        if (!this._collapsed) {
```
kdinev: OK, question: isn't this required for the tooltip to actually replay its animation, and if so, wouldn't this break the tooltip behavior?

wnvko: igxTooltip should work correctly now

projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts
```diff
@@ -195,6 +198,12 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
 
         this._overlayDefaults.positionStrategy = new AutoPositionStrategy(positionSettings);
         this._overlayDefaults.closeOnOutsideClick = false;
+        this.target.onClosed.subscribe(() => {
```
tachojelev: I believe we should unsubscribe from the tooltip's onClosed event when the respective targets gets destroyed.

projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts
```diff
@@ -195,6 +198,12 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
...
+            this.closeAnimationInProgress = false;
+        });
+        this.target.onOpened.subscribe(() => {
```
tachojelev: I believe we should unsubscribe from the tooltip's onOpened event when the respective targets gets destroyed.

wnvko: done

projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts
wnvko: done

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -508,12 +531,17 @@ export class IgxOverlayService implements OnDestroy {
         }
     }
 
-    private getOverlayById(id: string): OverlayInfo {
+    /** @hidden */
```
damyanpetev: If this is intended to be internal, mark it `@internal`

wnvko: added :)

PR #3822, opened 2019-02-05
Reposition correctly element after its size was changed

### LINKED ISSUE #3743: igxOverlay should propose a way to recalculate its height, when overlay content height has changed
### Description  
igxOverlay should propose a way to recalculate its height, when overlay content height has changed

### Attachments  
The following attachment shows when drop down is filtered, then even the `reposition()` method of the `toggle` is called, the height is not recalculated and drop down is not positioned aligned to the target element.
image

wnvko:
We were measuring the element we should show in the overlay and was caching its size. If the size of the element was changed while element is shown on the overlay, e.g. filtering a list of items, the size of the content element (div which wraps shown element ) did not change. Now we are no longer cache the size of the element, but only its width. The height of the element gets calculated on each call to position method of the position strategy. 

Closes #3743

 - [x] Bug fix

### Checklist:


projects/igniteui-angular/src/lib/core/styles/components/overlay/_overlay-theme.scss
```diff
@@ -79,6 +79,10 @@
         pointer-events: initial;
     }
 
+    %igx-toggle--hidden {
+        display: none !important;
```
simeonoff: Why do we use important here?

src/app/overlay/overlay.sample.ts
```diff
@@ -261,10 +221,131 @@ export class OverlaySampleComponent {
...
+                this.items.push(`Item ${item}`);
+            }
+            this.cdr.detectChanges();
+            this.onChange2();
+            this._overlaySettings.positionStrategy.settings.target = this.button.nativeElement;
+        }
         this.igxDropDown.toggle(this._overlaySettings);
     }
 
+    public resizeAndSoOn() {
```
kdinev: Seriously?

projects/igniteui-angular/src/lib/core/styles/components/overlay/_overlay-theme.scss
simeonoff: Ok, disregard for now.

src/app/overlay/overlay.sample.ts
wnvko: removed
shame on me :confused:

PR #3913, opened 2019-02-13
Refresh search highlights when a row is loaded from the cache

### LINKED ISSUE #3895: Some cells are not highlighted on scrolling a grid with grouping and search applied
## Description  
When you scroll a grid that has grouping and search applied, the highlights are not correctly updated.

 * browser: NA

## Steps to reproduce  

1. Open 'Grid Search' dev demo.
2. Group by the 'ContactTitle' column.
3. Type 'a' in the search box,
4. Scroll down

## Result  
Some cells that contain 'a' are not highlighted.

## Expected result  
All cells containing 'a' should be highlighted.

## Attachments  
image

DiyanDimitrov:
Closes #3895

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -4734,6 +4723,25 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
     protected get isAttachedToDom(): boolean {
         return this.document.body.contains(this.nativeElement);
     }
+
+    /**
+     * @hidden
+     */
+    public cachedViewLoaded(args: IViewChangeEventArgs) {
+        const view: EmbeddedViewRef<any> = args.view;
```
MayaKirova: Instead of comparing the view’s native element perhaps it would be better to add the index in the context that is passed to the igxTemplateOutlet:
   [igxTemplateOutletContext]='getContext(rowData, rowIndex)
…
   public getContext(rowData, rowIndex): any {
        return {
            $implicit: rowData,
            index: rowIndex,
            templateID: this.isGroupByRecord(rowData) ? 'groupRow' : this.isSummaryRow(rowData) ? 'summaryRow' : 'dataRow'
        };
    }
…
So that you can then get the row by index, check if it is a data row, check if there is any search applied, maybe even check if this row’s data is in the matchInfoCache etc., for example something like this:

public cachedViewLoaded(args: IViewChangeEventArgs) {
const index = args.context.index;
        const row = this.getRowByIndex(index);
        if (this.lastSearchInfo.searchText && row instanceof IgxGridRowComponent) {
           row.cells.forEach((c) => {
                c.highlightText(
                this.lastSearchInfo.searchText,
                this.lastSearchInfo.caseSensitive,
                this.lastSearchInfo.exactMatch);
            });
        }
}
Otherwise I think this will be called too often and may have a performance impact.

DiyanDimitrov: I have tried these suggestions and here are my findings:
1. adding `index` to the context doesn't work since the row component is still not updated with its new context.
2. getRowByIndex should have similar performance since it also uses `find` but compares numbers instead of object references - `rowList.find((row) => row.index === rowIndex)`.
3. checking `this.lastSearchInfo.searchText` is not an option since a row with highlights may be loaded from the cache and its highlights should be cleared if now the searchText is empty.

Also I just want to mention that `cell.highlightText` won't do anything if the search criteria is not changed.

However I agree with you that this code may be called too often so I will try to optimize it. I will add the old context to the onCachedViewLoaded event in order to catch only the cases when the new context is the same as the old one:
```
        if (args.context['templateID'] === 'dataRow' &&
            args.context['$implicit'] === args.oldContext['$implicit'] &&
            view.rootNodes.length) {
            const row = this.dataRowList.find(r => r.nativeElement === view.rootNodes[0]);
            if (row) {
                row.cells.forEach((c) => {
                    c.highlightText(
                        this.lastSearchInfo.searchText,
                        this.lastSearchInfo.caseSensitive,
                        this.lastSearchInfo.exactMatch);
                });
            }
        }
```

Any other suggestions are welcome.

PR #3972, opened 2019-02-21
fix(grid): Fix calculate height when initially grid has no data #3950

### LINKED ISSUE #3949: igx-grid: row virtualization doesn't work when setting height in percent if you fetch and bind data after initial rendering.
## Description  
I want to fit `igx-grid` height to its parent's height. So I'm setting `height` to `100%`. But, if I initially bind no data to `igx-grid` then fetch remote data, all rows of data are rendered. Row virtualization doesn't work in this scenario. 

 * igniteui-angular version: 7.1.7
 * browser: N/A

## Steps to reproduce  

1. Run the attached sample.
2. Click "load data" button

## Result  
Row virtualization doesn't work. If I set `height` in px, there is no problem.

## Expected result  
Row virtualization should work.

## Attachments  
[my-app.zip](https://github.com/IgniteUI/igniteui-angular/files/2878814/my-app.zip)

### LINKED ISSUE #3950: igx-grid: Incorrect height calculation when setting height in percent and binding empty data.
## Description  
There is a problem with height calculation when setting height in percent and binding empty data.

 * igniteui-angular version: 7.1.7
 * browser: N/A

## Steps to reproduce  

1. Run the attached sample.

## Result  
Although I set `igx-grid` height to `100%`, `igx-grid` doesn't fit to its parent.

## Expected result  
`igx-grid` should fit to its parent.

## Attachments  
[my-app.zip](https://github.com/IgniteUI/igniteui-angular/files/2878857/my-app.zip)

### LINKED ISSUE #4745: Grid height is calculated wrongly as grid width narrows
## Description  
Grid height changes as grid width changes by narrowing browser window width.

 * igniteui-angular version: 7.2.8
 * browser: n/a (Chrome, Firefox and IE11)

## Steps to reproduce  

1. Run the attached sample
2. Narrow window width

## Result  
Chrome and Firefox: A white space appears at the bottom of the grid. Vertical scrollbar is calculated wrongly.
IE11: Grid height changes.
IgxGrid_height_issue

## Expected result  
Grid's height stays the same height.

## Attachments  
[igx-grid-scrollbar.zip](https://github.com/IgniteUI/igniteui-angular/files/3170799/igx-grid-scrollbar.zip)

mpavlinov:
Closes #3950 
Closes #3949 
Closes #4745

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@mpavlinov  This affects the HierarchicalGrid. In master by default if no height is specified for rowIslands in remote data scenario, they extend to their full content.
With this change now shrink to about 100px:
image

This will probably affect existing samples. 
Perhaps we should add special handling for child grids, so that by default when no height is set it will act as height = null instead of height = 100% since child containers are small by default and won't look good in the default use-case.

projects/igniteui-angular/src/lib/grids/grid/grid.component.html
```diff
@@ -123,7 +123,7 @@
     <div class="igx-grid__tfoot-thumb" [hidden]='!hasVerticalSroll()' [style.height.px]='summariesHeight' [style.width.px]="scrollWidth"></div>
 </div>
 
-<div class="igx-grid__scroll" [style.height]="'18px'" #scr [hidden]="unpinnedWidth - totalWidth >= 0">
+<div class="igx-grid__scroll" [style.height]="'18px'" #scr [hidden]="unpinnedWidth - (totalWidth - 18) >= 0">
```
ChronosSF: This seems to allow the vertical scrollbar to cover the last cells right edge before a horizontal scrollbar is shown. Why is this part of this PR?

I believe we moved the v scrollbar out of the tbody so that it never covers the cells.

ChronosSF:
@mpavlinov , if you don't have empty data but rather have initial data that can fit under what height 100% would usually be and then load the full data, you still get the grid to disable virtualization. I guess we need to derive the height on the data prop setter as well and not lose the initial setting. Not if this won't break something else though.

projects/igniteui-angular/src/lib/grids/grid/grid.component.html
mpavlinov: @ChronosSF There were failing tests. The issues was with the horizontal scroll that opened up a gap on the bottom. I fixed it by clearing the cache when the columns get recalculated.
Note: The issue is when the columns don't have widths set.

PR #3973, opened 2019-02-21
Open toggle with same id when there is any

### LINKED ISSUE #3971: IgxToggle open method always tries to get id even when it has one
## Description  
When you call open method of the ixgToggle it calls attach method of igxOverlay to retrieve unique id. If at this point you call close method of the igxToggle and immediately open method the last one will generate a new id. In the same time close is closing the toggle, and open opens it with new id. 

 * igniteui-angular version: 7.2.0
 * browser: all

## Steps to reproduce  

1. Start dev samples;
2. Go to overlay sample;
3. Choose connected positioning strategy. This should set close on outside click to true and modal to false;
4. Click toggle button.
5. Click toggle button again, after drop down is shown

## Result  
Drop down remains somehow in the middle of the screen

## Expected result  
Drop down should try to close and then should open again - first comes close on outside click then comes toggle method which calls open.

wnvko:
When calling open method of igxToggle it calls internally attach method of igxOverlay. This should happen only if igxToggle has not retrieve yet id. If there is id it should use it and should not call attach.

note: why was testing this I have updated overlay sample in dev samples. Update is included in this PR too.

Closes #3971

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.ts
```diff
@@ -182,7 +182,9 @@ export class IgxToggleDirective implements IToggleView, OnInit, OnDestroy {
             return;
         }
 
-        this._overlayId = this.overlayService.attach(this.elementRef, overlaySettings);
+        if (!this._overlayId) {
```
damyanpetev: Saw the ID gets reset on close, but since we use the `info` above - can this check use it instead?
Just in case the id becomes invalid somehow.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -279,7 +279,7 @@ export class IgxOverlayService implements OnDestroy {
             this.playOpenAnimation(info);
         } else {
             //  to eliminate flickering show the element just before onOpened fire
-            info.elementRef.nativeElement.parentElement.style.visibility = null;
+            info.elementRef.nativeElement.parentElement.visibility = '';
```
damyanpetev: Ahem. Where did the `.style` go?

projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.ts
wnvko: much better in this way. Thank you!

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
wnvko: it used to disappear :(
Fixed

PR #4011, opened 2019-02-25
chore(ExelFiltering): Restore scroll postion after drop-down is opened.

MayaKirova:
Suggestion on how to restore scroll top after drop-down has lost it as a result of moving content in overlay.


bkulov:
I've integrated those changes into my branch manually as I did some more stuff and there were conflicts.
Good job :)

PR #4077, opened 2019-02-27
igxSelect - select-positioning-strategy code cleanup

### LINKED ISSUE #4019: igxSelect - select-positioning-strategy code cleanup
## Description  
igxSelect - select-positioning-strategy code cleanup. 
Use utilities methods, cache variables, etc.

 * browser:

IvayloG:
Closes #4019  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [x] CI/CD

### Checklist:


ViktorSlavov:
The default behavior seems to be working fine (drop down height < view port height, with and without scroll), but when the select height is > view port height, then it behaves weird:
### To reproduce:
 1. Set max height of select item container to be more than the view port height
 2. Set the number of items so that they are more than LESS than the max height and MORE than the viewport height (for standard 1080px < **n** * 48 < max height, where n is the number of items)
 3. Select an item on the select

### Results:
The item container flips to the opposite direction (so, if it would normally drop down, it flips and is displayed upwards) and only the end of the list is visible.

### Image:
ff

### Resolution
If this is considered expected behavior, feel free to mark this as `verified` - I could not find any other issues with the positioning

IvayloG:
In my opinion this could be considered as an application scenario and thus i will mark this as verified

PR #4162, opened 2019-03-05
Don't clear sorting if column is grouped

### LINKED ISSUE #4119: When change sorting from the excel filter it is not applied for the grouped column
## Description  
When change sorting from the excel filter it is not applied for the grouped column

 * browser: N/A

## Steps to reproduce  

1. Open dev sample gridColumnMoving
2. Group by a column
3. Open excelfilter for this column and remove sortig
4. Click on DESCENDING button

Note: after that cannot change sort order from the grouped column

## Result  
The change is not applied for the grouped column 

## Expected result  
The change should be applied

## Attachments  
image

sstoyanovIG:
Closes #4119   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-sorting.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-sorting.component.ts
@@ -6,7 +6,7 @@ import {
 } from '@angular/core';
 import { IgxColumnComponent } from '../../column.component';
 import { IgxButtonGroupComponent } from '../../../buttonGroup/buttonGroup.component';
-import { IgxGridBaseComponent, IgxGridComponent } from '../../grid';
+import { IgxGridBaseComponent } from '../../grid';
 
 /**
  * @hidden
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-sorting.component.ts
```diff
@@ -31,8 +31,16 @@ export class IgxExcelStyleSortingComponent {
     constructor() {}
 
     public onSortButtonClicked(sortDirection) {
+        let isColumnGrouped;
+        if (this.grid instanceof IgxGridComponent) {
```
bkulov: Excel style filtering is applicable to all grids we currently have. Casting the current grid instance to one grid type is not a good solution.
In my opinion you should add a method in the base class which will return if a given column from the current grid instance is grouped or not. This method will return actual result only for the grid and for the hierarchical and tree-grid it will return false.
That way we wont make a dependency to one grid type in the feature which is applicable to all grid types.

sstoyanovIG: The reason for this is that only IgxGrid supports grouping and only it has 'groupingExpressions' property.

bkulov: Sure, but my comment is still valid. We shouldn't introduce dependency to one grid type here. Tomorrow some of the other grids may support grouping and this code will fail.

sstoyanovIG: Ok, I got your point, I will add a getter in the base class that will return false and I will override it in the IgxGrid only.

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
```diff
@@ -638,6 +638,17 @@ export class IgxGridComponent extends IgxGridBaseComponent implements IGridDataB
         this._gridAPI.sort_multiple(this.id, this._groupingExpressions);
     }
 
+    /**
+     * @hidden
+     */
+    public isColumnGrouped(fieldName: string): boolean {
+        if (this.groupingExpressions.find(exp => exp.fieldName === fieldName)) {
```
bkulov: Just return the expression here.

PR #4166, opened 2019-03-05
Add property that returns if all columns are in view

### LINKED ISSUE #4111: Can not navigate with tab in filtering row if grid has no horizontal scroll
## Description  
 * browser: all

## Steps to reproduce  

1. Open Grid Toolbar Custom Content Dev Sample
2. Click over the first cell in the grid
3. Press shift + tab twice
4. Press tab

## Result  
The focus element is lost.

## Expected result  
The next filter chip should be focused.

sstoyanovIG:
Closes #4111   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/filtering/grid-filtering-cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/filtering/grid-filtering-cell.component.ts
@@ -356,6 +356,10 @@ export class IgxGridFilteringCellComponent implements AfterViewInit, OnInit, DoC
     }
 
     private isColumnRightVisible(columnIndex: number): boolean {
+        // If displayContainerWidth is 0 than there is no horizontal scrollbar, which means that all columns are fully visible.
+        if (this.filteringService.displayContainerWidth === 0) {
+            return true;
+        }
         let currentColumnRight = 0;
         for (let index = 0; index < this.filteringService.unpinnedColumns.length; index++) {
             currentColumnRight += parseInt(this.filteringService.unpinnedColumns[index].width, 10);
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/grids/filtering/grid-filtering-cell.component.ts
```diff
@@ -356,6 +356,10 @@ export class IgxGridFilteringCellComponent implements AfterViewInit, OnInit, DoC
     }
 
     private isColumnRightVisible(columnIndex: number): boolean {
+        // If displayContainerWidth is 0 than there is no horizontal scrollbar, which means that all columns are fully visible.
+        if (this.filteringService.displayContainerWidth === 0) {
```
bkulov: Could you create a property in filtering service to return true if all columns are visible. If someone in the future changes the behavior of `displayContainerWidth` this code will fail.

projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts
```diff
@@ -67,6 +67,10 @@ export class IgxFilteringService implements OnDestroy {
         return parseInt(this.grid.parentVirtDir.getHorizontalScroll().scrollLeft, 10);
     }
 
+    public get areAllColumnsInView() {
+        return parseInt(this.grid.parentVirtDir.dc.instance._viewContainer.element.nativeElement.offsetWidth, 10) === 0;
```
bkulov: In the previous implementation you were using something else to get the view port size. Is it possible to not use private properties?

sstoyanovIG: The code is used in the quick filtering as well and I think there shouldn't be any issues

PR #4180, opened 2019-03-05
fix(GroupBy): Fix error thrown when expanding a groupby row. 

### LINKED ISSUE #4179: Expanding a group row while at the bottom of the grid throws error
## Description  
Error is thrown when having two collapsed group rows at the bottom and we try to expand the upper one.

 * browser: All

## Steps to reproduce  

1. Run the Grid Search sample.
2. Group by `Contact Title` column.
3. Scroll to the bottom of the grid using mouse wheel.
4. Collapse `Sales Representative` group.
5. Collapse `Sales Manager` group.
6. Expand the `Sales Manager` group.

**Note**: There are other scenarios that this is reproduced by also expanding while having collapsed groups, but this seems to be the one of the more consistent scenarios.

## Result  
ERROR TypeError: Cannot read property 'forEach' of undefined

## Expected result  
There should be no error thrown.

skrustev:
Closes #4179

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -4855,8 +4855,8 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
     public cachedViewLoaded(args: ICachedViewLoadedEventArgs) {
         if (args.context['templateID'] === 'dataRow' && args.context['$implicit'] === args.oldContext['$implicit']) {
             args.view.detectChanges();
-            const row = this.getRowByIndex(args.context.index);
-            if (row && row instanceof IgxRowComponent) {
+            const row = this.dataRowList.find((r) => r.index === args.context.index);
+            if (row) {
                 row.cells.forEach((c) => {
                     c.highlightText(
                         this.lastSearchInfo.searchText,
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -4856,14 +4856,14 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         if (args.context['templateID'] === 'dataRow' && args.context['$implicit'] === args.oldContext['$implicit']) {
             args.view.detectChanges();
             const row = this.getRowByIndex(args.context.index);
-            if (row) {
+            if (row && row instanceof IgxRowComponent) {
```
DiyanDimitrov: The reason for this issue is that we have a groupBy row and a grid row with the same index. We have to search only in the `dataRowList` like this:
`const row = this.dataRowList.find((r) => r.index === args.context.index);`

DiyanDimitrov: Also I have noticed that there is no test added for this scenario. It would be nice if we have such.

skrustev: Well yes, this is the other way that could be fixed. I can change it.
There is already a test that covers this that for some was passing.

DiyanDimitrov: It looks ok, but I was thinking if a similar issue comes up with the horizontal scrolling. It would be better to fix the row finding there as well.

skrustev: There is no issue for the horizontal scrolling because there is basically the same check I used earlier that covers these scenarios instead of parsing the dataRowList directly. Either way will work.

PR #4293, opened 2019-03-12
Including summary row cells in tab sequence for HierarchicalGrid navigation.

### LINKED ISSUE #4280: Keyboard Navigation Issues in the Hierarchical Grid
## Description  

 * browser: all

## Steps to reproduce  

- [x] Can not navigate to root summary row when press **Tab** over the last cell in the grid's body - should be fixed with https://github.com/IgniteUI/igniteui-angular/pull/4293

- [x] Error is thrown when click over summary cell and press **Tab** or **Shift + Tab** - should be fixed with https://github.com/IgniteUI/igniteui-angular/pull/4293

- [x] When child grid is filterable and the focus is on the last filtering chip of that grid and press **Tab** the error is thrown in the console - should be fixed with https://github.com/IgniteUI/igniteui-angular/pull/4267

## Result  
Errors are thrown in the console and navigation is not performed.

## Expected result  
Should not be thrown errors in the console and the navigation should be performed as expected
## Attachments  

MayaKirova:
Closes #4280

Tab navigation should go through summary rows, arrow key navigation should skip them.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -192,11 +197,22 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
             // handle scenario where last child row might not be in view
             // parent should scroll to child grid end
             const childContainer = this.grid.nativeElement.parentNode.parentNode;
-            const diff =
+            const diffBottom =
             childContainer.getBoundingClientRect().bottom - this.grid.rootGrid.nativeElement.getBoundingClientRect().bottom;
-            const endIsVisible = diff < 0;
+            const row = this.grid.getRowByIndex(rowIndex).element.nativeElement;
```
skrustev: I think it should take also summary row list if End key is pressed while a summary is focused.
Also isn't there a chance to get a child grid row by using getRowByIndex?

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -192,11 +197,22 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
             // handle scenario where last child row might not be in view
             // parent should scroll to child grid end
             const childContainer = this.grid.nativeElement.parentNode.parentNode;
-            const diff =
+            const diffBottom =
             childContainer.getBoundingClientRect().bottom - this.grid.rootGrid.nativeElement.getBoundingClientRect().bottom;
```
skrustev: I see that it takes into account only the grid boundaries, but will this scroll the actual child grid if the cell is partially visible in it?
Also this is not handled for `Home` button, both imo should either use default logic or override it this way.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -192,11 +197,22 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
...
-                this.scrollGrid(this.grid.parent, diff, () => super.onKeydownEnd(rowIndex));
+                this.scrollGrid(this.grid.parent, diffBottom, () => super.onKeydownEnd(rowIndex));
+            } else if (!topVisible) {
+            const scrGrid = this.grid.verticalScrollContainer.getVerticalScroll().scrollTop !== 0 ? this.grid :
+             this.getNextScrollable(this.grid).grid;
+            const topGrid = scrGrid.tbody.nativeElement.getBoundingClientRect().top >
+            this.grid.rootGrid.tbody.nativeElement.getBoundingClientRect().top ? scrGrid : this.grid.rootGrid;
+            this.scrollGrid(topGrid, diffTop, () => super.onKeydownEnd(rowIndex));
             } else {
                 super.onKeydownEnd(rowIndex);
```
skrustev: `onKeydownEnd` takes `isSummary` as well. I think it should be implemented for this method override too.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -252,14 +291,44 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
         } else if (visibleColumnIndex === 0 && currentRowEl.previousElementSibling &&
             currentRowEl.previousElementSibling.children[0].tagName.toLowerCase() === 'igx-child-grid-row') {
             const gridElem = currentRowEl.previousElementSibling.querySelector('igx-hierarchical-grid');
```
skrustev: If there are more than one rowIsland defined it would select only the first but it should be the last one.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -226,16 +242,39 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
...
+            super.performTab(currentRowEl, rowIndex, visibleColumnIndex, isSummaryRow);
         }
     }
-    public performShiftTabKey(currentRowEl, rowIndex, visibleColumnIndex) {
-        if (visibleColumnIndex === 0 && rowIndex === 0 && this.grid.parent) {
+    public performShiftTabKey(currentRowEl, rowIndex, visibleColumnIndex, isSummary = false) {
+        if (visibleColumnIndex === 0 && rowIndex === 0 && this.grid.parent && !isSummary) {
             if (this.grid.allowFiltering) {
                 this.moveFocusToFilterCell();
             } else {
```
skrustev: In this scenario there should be also a check if the sibling child grid has summaries when pressing Shift + Tab to focus the last summary.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -226,16 +242,39 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
...
+        const isLastDataRow = rowIndex === this.grid.verticalScrollContainer.igxForOf.length - 1;
+        const nextIsDataRow = this.grid.dataRowList.find(row => row.index === rowIndex + 1) ;
+        const isLastColumn =  this.grid.unpinnedColumns[this.grid.unpinnedColumns.length - 1].visibleIndex === visibleColumnIndex;
+        const isLastSummaryRow = hasSummaries && isSummaryRow;
+        if (!nextIsDataRow && !(isLastDataRow && hasSummaries) && isLastColumn && !isSummaryRow) {
+            // navigating in child, next is not summary
             this.navigateDown(currentRowEl, rowIndex, 0);
+        } else if (isLastSummaryRow && isLastColumn && this.grid.parent) {
+            // navigating in child summary, next is parent summary or next parent row
+            const parent = this.grid.parent;
```
skrustev: If there is a next sibling child grid it should move to the first cell inside it, otherwise here would focus the parent summary cell.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
MayaKirova: It is also used when goToLastBodyElement is called, which is executed when performShiftTabKey is called on summary row first cell. In that case the previous cell may not be visible (may be above the current visible viewport) so it may need to be scrolled up so that the target data cell is fully in view hence the additional handling.

skrustev: In this scenario though could't `navigateUp` be used instead?

MayaKirova: Not without modifying navigateUp to handle elements that are not row siblings (summary row is not in tbody but in footer) and since that function is mainly used for arrow key navigation (which should skip headers/footers and only check for sibling rows inside tbody) I'd rather not have summaries handling in it.

skrustev: Well it might not be the best place to handle it there because still would create some slight inconsistency when it will scroll but this is not big anyway.

MayaKirova:
Please test again with sibling layouts that have summaries.

MartinKamenov:
@MayaKirova shift+tab not working correctly when we have row-island siblings:
1. Open hierarchical grid dev sample.
2. Add this as a sibling to the inner-most row-island
<igx-row-island [key]="'childData2'" [autoGenerate]="false" [rowSelectable]='isRowSelectable'>
    <igx-column field="ChildLevels" [groupable]='true' [hasSummary]='true'></igx-column>
    <igx-column field="ProductName" [groupable]='true' [hasSummary]='true'></igx-column>
</igx-row-island>
3. Expand last row of every hGrid.
4. Try navigating from parent grid cell to the previous summary using shift+tab.
(Focus skips couple of layers and goes directly to the innermost - last sibling.)

navigation1

PR #4295, opened 2019-03-12
Fixing #1861 and #1858 7.2.x

### LINKED ISSUE #1858: igx-grid - setting height in percent doesn't work when the sum height of all rows is lower than parent height
### Steps to reproduce  

1. Go to https://stackblitz.com/edit/igx-grid-height-in-percent

### Result  
Although height is set in percent, the sum height of all rows is taken priority. If the height is set in pixel, there is no problem.
image

### Expected result  
When height is specified no matter what it is pixel or percent, it should be taken priority over the sum height of all rows.

### LINKED ISSUE #1861: Grid with 100% height will incorrectly exceed its parent container if the total row height is smaller than it
### Description  
The grid bases its percent height rendering only on the total row height which means that it will exceed the container it is put in by potentially the full height of its headers, pager, summaries, group area, toolbar, etc as long as it doesn't have enough ***data*** rows to fill in that parent container.

### Steps to reproduce  

1. Open the Summaries dev sample, wrap the grid in a container with height 500px. Set the height to 100%

```html
<div style="height: 500px">
     <igx-grid #grid1 [data]="data" [autoGenerate]="false" displayDensity="compact" height="100%" width="600px" (onColumnInit)="initColunm($event)" [paging]="true">
     ...
</div>
```

2. Observe the sample

### Result  
The grid renders in a total of 611 pixels instead of the expected 500 to fit its parent container

### Expected result  
The grid to render with 500px height

### Attachments  
untitled

mpavlinov:
Closes #1861 
Closes #1858

### Description
The grid now won't set height=null internally in some conditions. The default height=100% will be set to the grid as it is by [specification](https://github.com/IgniteUI/igniteui-angular/wiki/igxGrid-Specification#height). The developer should explicitly set height=null if he wants the grid to render all rows without a vertical scrollbar.

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


ChronosSF:
@mpavlinov , please check the case where the grid is wrapped in no height container and it itself has no height set and the data amount is smaller than the default height. It seems the chunk size is wrong in this case and the rows get repeated.

mpavlinov:
@ChronosSF Talking about the Grid Summaries sample, it's a sample "bug". See the constructors code:
```
    constructor() {
        for (let i = 0; i < 10; i++) {
            this.data = this.data.concat(this.data);
        }
        this.data = this.data.slice(0, 19);
        for (let i = 0; i < this.data.length; i++) {
            this.data[i]['Index'] = i;
        }
    }
```

ChronosSF:
I didn't manage to break it with my other tests so I guess it's good for a real PR.

mpavlinov:
Alternative fix is already applied in the product. I'm adding the tests in a separate PR. Closing this one.

PR #4364, opened 2019-03-20
Public enumerations should not be constants

### LINKED ISSUE #4363: Public enumerations should not be constants
## Description  
Public exported igniteui-angular enumerations (like `InteractionMode`) should not be constants,  because during [JIT compilation](https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime) `const enum` aren't available in the [transpiled JS code](https://www.typescriptlang.org/play/index.html#src=export%20enum%20InteractionMode%20%7B%0D%0A%20%20%20%20DropDown%2C%0D%0A%20%20%20%20Dialog%0D%0A%7D%0D%0A%0D%0Avar%20test%3B%0D%0A%0D%0Aif%20(test%20%3D%3D%3D%20InteractionMode.Dialog)%20%7B%0D%0A%0D%0A%7D%20else%20if%20(test%20%3D%3D%3D%20InteractionMode.DropDown)%20%7B%0D%0A%0D%0A%7D). While this is not a problem in the AOT compilations, because the d.ts files are available and enumerations are properly replaced, it is a problem when a JIT compiled application is trying to use those publicly exposed enumerations, which in that case are missing.

## Steps to reproduce  

1. Open the [following](https://www.infragistics.com/products/ignite-ui-angular/angular/components/autocomplete.html) sample and go the Check movie availability demo. The demo is working, because it is AOT compiled.
2. Click the StackBlitz demo (or use that [link ](https://stackblitz.com/edit/angular-m4jmt7)with all exported CONST enumerations). In order to work you need to comment them.

## Result  
There is an exception when trying to access enumerations.

## Expected result  
No exceptions are thrown.

## Attachments  
[How do the different enum variants work in TypeScript?](https://stackoverflow.com/questions/28818849/how-do-the-different-enum-variants-work-in-typescript)

Lipata:
Closes  #4363 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/date-picker/date-picker.utils.ts
```diff
@@ -5,7 +5,7 @@ import { isIE } from '../core/utils';
  *
  *@hidden
  */
-export const enum DateState {
+export enum DateState {
```
damyanpetev: That one is hidden, though. As in, not intended to be exposed in public API at all.
There's a bit of a problem with that as the public API exposes everything like so:
[`export * from './lib/date-picker/date-picker.utils';`](https://github.com/IgniteUI/igniteui-angular/blob/7.2.x/projects/igniteui-angular/src/public_api.ts#L96)
@SAndreeva @sboykova Should this be excluded from public exports? More over, I don't see anything in the date picker util file that looks like public API. Should the entire export be dropped?

sboykova: @hidden can be removed from enum `DateState`. In case the date picker is retemplated, the user may need some of the methods in date-picker.utils for parsing masks/formats.

damyanpetev: You mean the static methods of the `DatePickerUtil` class that is also hidden?
I see a lot of `any`-s and types using the other non-exposed enums in there. IMHO, unless the need arises by a customer, this should probably remain unexposed.

PR #4382, opened 2019-03-21
Use parent or null row editing outlet when 1-2 rows in grid

### LINKED ISSUE #4240: Row editing overlay is not visible when grid has either 1 or 2 rows and height is not set.
## Description  
When grid is rendered with now height set and the data has only 1/2 rows and when the first row is row edit mode the overlay containing the Done/Cancel buttons is not visible.

 * igniteui-angular version: 7.1.x
 * browser: All

## Steps to reproduce  

1. Open the Grid Row Editing demo.
2. Remove the `height` property of the first grid and set the data to 1 or 2 rows.
3. Try to edit the first row.

## Result  
The Done/Cancel overlay is not visible to the user.

## Expected result  
The Done/Cancel overlay should be visible.

## Attachments  
row-editing-1
row-editing-2

Lipata:
Closes #4240

Use parent outlet for row editing overlay, when the grid has 1-2 rows. In the case of the grid, in such cases the outlet is `null`.

The current behavior of the grid is that it will open the overlay above the row in cases where there isn't enough space below. This happens on the last two rows of the grid because the row editing row is higher than a row and there will be no space below not only for the last row but also for the one before that.

In cases with only one row, this will mean that the row editing overlay will appear above the header and will hide it. Considering the fact that is an edge case this will not be seen so often (more in the case of a hierarchical grid).

>Note: by design, the row editing overlay is not closed on a scroll, and this is the only overlay to do so in the context of the hierarchical grid.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


ViktorSlavov:
Fix works as intended for the base grid (regardless of filtering)
Fix does not work for Hierarchical Grid with the following setup:
 - In `Hierarchical Grid` sample, turn on row editing and change the data to be 1 entry
 - In the markup, add a primary key to the grid, otherwise row editing will crash
 - Start editing a cell - the row editing overlay is displayed "below" the filtering row and is not visible

projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -4907,7 +4921,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
     }
 
     openRowOverlay(id) {
-        this.configureRowEditingOverlay(id);
+        this.configureRowEditingOverlay(id, this.rowList.length <= 2);
```
kdinev: What is `2`?

Lipata: I've added a comment about that.

kdinev: Well, according to this comment, there should be something like `private MIN_ROW_EDITING_COUNT_THRESHOLD = 2`

The comment should also go above the constant, rather than above where it's used.

PR #4386, opened 2019-03-22
Adding *Change events for state inputs for two-way binding 

### LINKED ISSUE #3323: Expose *Change events for state inputs so that users can two-way bind their models
### Description  
Most of the state inputs (e.g. pinned, hidden, etc. for columns, width/height for most components and others) do not have a *Change output and cannot be two-way bound for when that state change needs to be propagated back to the user's app.

 * browser: n/a

### Steps to reproduce  

1. Open a sample from the demos app and take some sample that sets certain state props e.g.:
```html
<igx-grid #grid1 [data]="data" [width]="'1200px'" [height]="'800px'" [rowSelectable]="true">
        <igx-column *ngFor="let c of columns" [sortable]="true" [field]="c.field" [header]="c.field" [width]="c.width" [pinned]='c.pinned'
                [hidden]='c.hidden'>
        </igx-column>
</igx-grid>
```

2. Check the model after a prop such as hidden or pinned is changed.

### Result  
As expected the prop is not updated. While it is possible to update it yourself by binding to certain events, the disconnect between the model objects and the component instances will make it harder than it has to be.

### Expected result  
Subscribing to changes in the props should be as simple as using the banana in a box syntax.

```html
<igx-column *ngFor="let c of columns"[field]="c.field" [(width)]="c.width" [(pinned)]='c.pinned'
                [(hidden)]='c.hidden'>
```

### Attachments  
n/a

Assigning to @SlavUI  for distribution. Discussed with @kdinev and based on a user question from the forums.

_**Note:**_ We should first identify which properties should allow such two-way binding. The change is by no way limited to `pinned`, `hidden`, `width` and `height` .

MayaKirova:
Closes #3323

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


kdinev:
@mpavlinov This has been assigned to you for a really long time now. What are we going to do with it?

mpavlinov:
@kdinev I'll plan it for the next sprint.

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.component.ts
```diff
@@ -97,7 +98,16 @@ export class IgxExpansionPanelComponent implements IgxExpansionPanelBase, AfterC
...
+    public get collapsed(): boolean {
+        return this._collapsed;
+    }
+    public set collapsed(value) {
+        this._collapsed = value;
+        this.collapsedChange.emit(this._collapsed);
+    }
+
+    @Output()
+    public collapsedChange = new EventEmitter<any>();
```
mpavlinov: @MayaKirova There are 2 other events: onCollapsed/onExpanded that duplicate this event. Maybe we should remove them. What do you think?

MayaKirova: @mpavlinov That would be a breaking change. onCollapsed/onExpanded emit additional information - the event and panel that was collapsed/expanded, while the Change events emit just the current value. I don't think they can replace each other.

projects/igniteui-angular/src/lib/chips/chip.component.ts
```diff
@@ -165,12 +165,24 @@ export class IgxChipComponent extends DisplayDensityBase {
...
+     *
+     * Two-way data binding:
+     * ```html
+     * <igx-chip #myChip [id]="'igx-chip-1'" [selectable]="true" [(selected)]="model.isSelected">
+     * ```
      */
     @Input()
     public set selected(newValue: boolean) {
         this.changeSelection(newValue);
+        this.selectedChange.emit(this._selected);
```
mpavlinov: @MayaKirova This is not working if I have a selectable chip and change its state from the UI.
Here is a code:
```
<igx-chip [selectable]="true" [(selected)]="isSelected" [selectIcon]="mySelectIcon">
	With a custom select icon.
</igx-chip>
<span>Is Selected: {{isSelected}}</span>
```

projects/igniteui-angular/src/lib/drop-down/drop-down-item.base.ts
```diff
@@ -135,8 +140,15 @@ export abstract class IgxDropDownItemBase implements DoCheck {
             return;
         }
         this._selected = value;
+        this.selectedChange.emit(this._selected);
```
mpavlinov: @MayaKirova This code doesn't work, because drop-down-item overrides the setter. As a result when the selection is changed from the UI the model is not updated.

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
```diff
@@ -288,6 +294,12 @@ export class IgxGridComponent extends IgxGridBaseComponent implements IGridDataB
         }
     }
 
+    /**
+     *@hidden
+     */
+    @Output()
+    public groupingExpressionsChange = new EventEmitter<IGroupingExpression[]>();
```
mpavlinov: @MayaKirova It doesn't update the model if "Expand All/Collapse All" button is clicked from the UI.

MayaKirova: @mpavlinov It appears that groupingExpansionState is empty in both cases - when all are expanded or collapsed and the default state is controlled via an additional option (groupsExpanded) which is toggled on expand/collapse all. Hence the binding works, the values are perhaps misleading.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
```diff
@@ -125,6 +127,15 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseCompone
     *      <igx-column field="Description"  [dataType]='string'></igx-column>
     * </igx-hierarchical-grid>
     * ```
+    *
+    * Two-way data binding.
+    * ```html
+    * <igx-hierarchical-grid [primaryKey]="'ID'" [data]="Data" [autoGenerate]="false" [(hierarchicalState)]="hgridState">
```
mpavlinov: @MayaKirova Setting two-way data binding for `hierarchicalState` throws 'Maximum call stack size exceeded' exception.

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
```diff
@@ -238,6 +238,11 @@ export class IgxGridComponent extends IgxGridBaseComponent implements IGridDataB
      *     dir: SortingDirection.Asc,
      *     ignoreCase: false
      * }];
+     * ```
+     *
+     * Two-way data binding.
+     * ```html
+     * <igx-grid #grid [data]="Data" [autoGenerate]="true" [(groupingExpressions)]="model.groupingExpressions"></igx-grid>
```
mpavlinov: @MayaKirova Setting both `groupingExpressions` and `groupingExpansionState` two-way data binding throws 'Maximum call stack size exceeded' error.

MayaKirova:
@mpavlinov please re-test.

PR #4399, opened 2019-03-25
Fix auto-generate columns for TreeGrid

### LINKED ISSUE #4396: Auto-generate columns is not working for TreeGrid
## Description  
Setting the autoGenerate to true does not work for TreeGrid.

 * browser: NA

## Steps to reproduce  

1. Go to the `tree-grid.sample.html` page in the dev demos.
2. Remove the columns and set the `autoGenerate` property to true.
3. Run the sample.

## Result  
The grid is empty.

## Expected result  
The grid columns should be auto-generated.

DiyanDimitrov:
Closes #4396

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/api.service.ts
@@ -585,6 +585,12 @@ export class GridBaseAPIService <T extends IgxGridBaseComponent & IGridDataBinda
         return column.dataType === DataType.Number;
     }
 
+    public get_data(id: string): any[] {
+        const grid = this.get(id);
+        const data = grid.data ? grid.data : [];
+        return data;
+    }
+
     public get_all_data(id: string, includeTransactions = false): any[] {
         const grid = this.get(id);
         let data = grid.data ? grid.data : [];
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -4048,7 +4048,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
      * @hidden
      */
     protected autogenerateColumns() {
-        const data = this.gridAPI.get_all_data(this.id);
+        const data = this.gridAPI.get_data(this.id);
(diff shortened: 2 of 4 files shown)


projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -585,6 +585,12 @@ export class GridBaseAPIService <T extends IgxGridBaseComponent & IGridDataBinda
         return column.dataType === DataType.Number;
     }
 
+    public get_data(id: string): any[] {
```
skrustev: This is pretty much the same as `get_all_data` bellow. 
If this is specific to the Tree Grid shouldn't it be handled in the IgxTreeGridAPIService?

DiyanDimitrov: It is in the base api service because it is used in the grid-base class. The idea is that get_all_data returns the child rows as well. However in the case of treeGrid it uses flatData which is not populated before the pipes are processed. For the auto-generate we don't need the child rows that's why we can use the grid.data value, but since the grid-base doesn't have data property, I have added this method to the service.

skrustev: I mentioned the 'get_all_data' because in the context of the other grids it already does the same thing.
Is there are a reason for the override of the `get_all_data` in the Tree Grid to return always flatData or empty array even when it is not populated yet and data is set?

DiyanDimitrov: We could do that, but the idea of get_all_data is to return the all of the data including the child rows, that's why I prefer to add a separate method just for the grid.data.

skrustev: Yep and the grid.data initially also has the child rows, but it is not flattened. That's why I think it might be better to be handled by the tree grid to avoid confusion.

At least in order to avoid duplicate code the 'get_all_data' can use the 'get_data' in the base class if you decide to keep the current logic.

Issue #4440, opened 2019-04-02
Hierarchical Grid: scrolled child views remain after the root grid has been destroyed


damyanpetev:
## Description  
Turned up from a test case
hierarchical-grid.navigation.spec > IgxHierarchicalGrid Complex Navigation > should allow navigating up from parent into nested child grid
that scrolls child instances by a significant size (to last).

Already looked into it with @MayaKirova and as far as we can see after the root Hierarchical Grid instance had been destroyed, 3 child grids in `IgxTemplateOutletDirective`'s `_embeddedViewsMap` cache remained not destroyed, thus not going through `OnDestroy` hook and removing their handlers.
The error in the tests is merely a side-effect of that.

I've commented out the test that reproduces this behavior and added a repro one:
https://github.com/IgniteUI/igniteui-angular/blob/49d45e7fb221700fa92ff30a2bbbd2046fe89f7a/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.navigation.spec.ts#L747-L785

 * igniteui-angular version: 7.2.x (7.2.2)
 * browser: N/A

## Steps to reproduce  

1. [Optional] If https://github.com/IgniteUI/igniteui-angular/pull/3916 is not yer merged, switch to the branch
2. `fit` the `"should allow navigating up from parent into nested child grid"` and the repro test after it in `hierarchical-grid.navigation.spec.ts`
3. Run the tests
4. Observe the console

## Result  
The remaining resize handlers trigger actions on the destroyed fixture and throw:
![image](https://user-images.githubusercontent.com/3198469/55421559-2bc5bf00-5582-11e9-9ee0-c77993d94123.png)

## Expected result  
After the Hierarchical Grid is destroyed, there shouldn't be any child instances left behind.

## Attachments  
Attach a sample if available, and screenshots, if applicable.

damyanpetev:
Note: Reproducible in 7.2.x as well (if you debug the resize handler in the base grid), except the test lacks the `onChunkLoaded` helper that would throw.

PR #4444, opened 2019-04-03
Fixing failing test and fixing issue with scrollbar misplacement in child grid when width is in %.

MayaKirova:
…for pinned/unpinned widths are calculated once after ViewInit. In case expander native element is still not updated use the default expander width based on display density.

The failing test is not related to this, still in some scenarios if child grids have with in % (or no width) the pinned/unpinned width calculations pass before widths for grid and expander element are available resulting in incorrect horizontal scrollbar calculation, which results in the vertical scrollbar not displaying in the correct place.


zdrawku:
This change is related to issue https://github.com/IgniteUI/igniteui-angular/issues/4440
For the time being we are going to remove skip the test and implement a solution later this week.

MayaKirova:
This will be handled as a separate bug fix after mcs is merged to master.

PR #4471, opened 2019-04-05
hide Overlays when scroll grid

### LINKED ISSUE #4099: Calendar should be closed when scrolling is initiated
## Description  
Calendar should be under the edited cell when scroll 

 * browser: N/A

## Steps to reproduce  

1. Open dev sample gridCellEditing
2. Open cell in column OrderDate in edit mode
3. Open calendar
4. Start scroll vertical or horizontal

Note: when scroll horizontal and the cell is no longer visible the calendar should be closed(Now it is done only for vertical scroll)

## Result  
The calendar is no longer under the edited cell

## Expected result  
Calendar should be closed when scrolling is initiated

## Attachments  

image

nrobakova:
Closes #4099  

 - [ x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -26,7 +26,7 @@ import {
     Optional
 } from '@angular/core';
 import { Subject } from 'rxjs';
-import { takeUntil, first } from 'rxjs/operators';
+import { takeUntil, first, filter } from 'rxjs/operators';
 import { IgxSelectionAPIService } from '../core/selection';
 import { cloneArray, isEdge, isNavigationKey, CancelableEventArgs, flatten, mergeObjects } from '../core/utils';
 import { DataType } from '../data-operations/data-util';
@@ -207,6 +207,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
     private _locale = null;
     private _observer: MutationObserver;
     private _destroyed = false;
+    private overlayIDs = [];
     /**
      * An accessor that sets the resource strings.
      * By default it uses EN resources.
@@ -2430,9 +2431,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
             this.disableTransitions = false;
         });
 
-        if (this.allowFiltering && this.filterMode === FilterMode.excelStyleFilter) {
-            this.closeExcelStyleDialog();
-        }
+        this.hideOverlays();
     }
 
     private horizontalScrollHandler(event) {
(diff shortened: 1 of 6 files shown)


MayaKirova:
Note: Should be tested with all possible overlays the grid may display, including:
- filtering overlays (simple and excel style).
- edit row overlay (this one should stay open on scroll)
- toolbar overlays
- editor overlays (in case anything other than the date picker uses an overlay).

This includes all grids as the change is in the base class.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
```diff
@@ -676,29 +663,13 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseCompone
 
     private hg_verticalScrollHandler(event) {
         this.scrollTop = event.target.scrollTop;
-        this.hideOverlays();
     }
 
-    public onContainerScroll() {
-        this.hideOverlays();
```
MayaKirova: This should not be removed. In scenarios where the child grid has larger width than the parent a scrollbar is rendered. Currently  when scrolling via that scrollbar overlays will not close, as a result they may go outside the scope of their parent container:

image

projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -2448,22 +2447,23 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
...
-        }
+    /**
+    * @hidden
+    * @internal
+    */
+    public hideOverlays() {
+        this.overlayIDs.forEach(overlayID => {
+            this.overlayService.hide(overlayID);
+            // blur in case some editor somewhere decides to move focus back
+            this.overlayService.onClosed.pipe(
```
bkulov: There is another subscription for `this.overlayService.onClosed` event in method `_setupListeners()`. I suggest removing this subscription from here and moving `this.nativeElement.focus()` there. I know this code is only moved to a different location but I think it could be improved.

nrobakova: @bkulov  I tried this, but focusing the this.nativeElement in all the cases will break the keyboard navigation and in the hgrid there is flickering. We do not need to focus native element in all the cases when overlay is closed, only in case when hide it from hideOverlays. The other way is to empty the overlayIDs in hideOverlays, but when the overlay is not closed by the hideOverlays, the array will not be correct. That is why I will leave it this way.

PR #4485, opened 2019-04-08
Fix completing filter in the filtering row

### LINKED ISSUE #4198: Completing filter in the filtering row
## Problem:
When users are adding filters for numeric or string column **using mouse**, they need to confirm completion of the current filter. Now, when they click outside of the input or current filter chip, the input goes out of focus but the filter chip stays selected, so the user may stay with the impression that the filter is not completed.

ezgif com-video-to-gif 1

## Solution:
Proposed solution:
- having additional UI for confirmation filter chip completion
- clicking (using mouse) out of the input will unfocus the input and the current chip

Solution 3

Similar to the Erase action icon, the Confirm action icon is:
- not shown if there is nothing typed
- inactive if there is at least 1 symbol typed

In that way we are not going to add more interactions, this Confirmation Action would serve as a reminder to select Enter when the filtering condition is done.

Currently considering other alternative solution.

DiyanDimitrov:
Closes #4198 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


DiyanDimitrov:
@simeonoff, Should we add a focus style to the commit and cancel icons? Currently they are using the browser's default focus style:
image

simeonoff:
@DiyanDimitrov Yes. Will add it tomorrow.

PR #4486, opened 2019-04-08
 feat(grid-mrl): MRL initial implementation #4384

### LINKED ISSUE #4384: Grid Multi Row Layout feature
Multi-row layouts allow developers to create intricate cell arrangements within the same row by defining a template layout that is applied to header and body cells. This provides flexibility of the arrangement and makes it possible to create responsive web apps where grids would adapt and avoid horizontal scrolling when the available space becomes too narrow.

## Additional context
Specification: https://github.com/IgniteUI/igniteui-angular/wiki/Multi-row-layouts

dkamburov:
Closes #4384

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -836,6 +836,23 @@ export class IgxColumnComponent implements AfterContentInit {
 
         return pinnedCols.indexOf(this) === pinnedCols.length - 1;
     }
+    get gridRowSpan() {
+        return (this.rowEnd && this.rowEnd.indexOf('span') !== -1) ?
```
ChronosSF: parseInt excludes letters afaik, so you don't need the replace

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -836,6 +836,23 @@ export class IgxColumnComponent implements AfterContentInit {
...
+            parseInt(this.rowEnd.replace('span', ''), 10) : 1;
+    }
+    get gridColumnSpan() {
+        return (this.colEnd && this.colEnd.indexOf('span') !== -1) ?
+            parseInt(this.colEnd.replace('span', ''), 10) : 1;
+    }
+
+    @Input()
```
ChronosSF: write API descriptions for the initial merge

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -1025,6 +1042,54 @@ export class IgxColumnComponent implements AfterContentInit {
...
+        return isIE ?
+        generatedSizes || `(1fr)[${templateItems}]` :
+            generatedSizes || `repeat(${templateItems},1fr)`;
+    }
+
+    protected generateColumnSizes(children) {
+        const columnSizes = [];
+        // find the smallest col spans
+        children.forEach(col => {
+            const actualWidth = parseFloat(col.calcWidth) + 'px';
```
ChronosSF: I'd avoid floats in widths because browsers handle them differently

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -1025,6 +1042,54 @@ export class IgxColumnComponent implements AfterContentInit {
             }
         }
     }
+
+    /**
+     * @hidden
+     */
+    getGridTemplate(isRow, isIE) {
```
ChronosSF: add return type

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -1025,6 +1042,54 @@ export class IgxColumnComponent implements AfterContentInit {
...
+            (acc, val) => Math.max(val.rowStart + val.gridRowSpan - 1, acc) :
+            (acc, val) => Math.max(val.colStart + val.gridColumnSpan - 1, acc);
+        const templateItems = this.children && this.children.reduce(itemAccum, 1) || 1;
+        const generatedSizes = !isRow ? this.generateColumnSizes(this.children) : null;
+        return isIE ?
+        generatedSizes || `(1fr)[${templateItems}]` :
+            generatedSizes || `repeat(${templateItems},1fr)`;
+    }
+
+    protected generateColumnSizes(children) {
```
ChronosSF: add return type

on that note, the word generate isn't usually associated with methods that return something

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -1025,6 +1042,54 @@ export class IgxColumnComponent implements AfterContentInit {
...
+                    (columnSizes[i].widthSetByUser ?
+                        parseFloat(columnSizes[i].width) / columnSizes[i].colSpan + 'px' :
+                        columnSizes[i].width);
+                }
+                i += columnSizes[i].colSpan - 1;
+            } else if (columnSizes[i]) {
+                result += ' ' + columnSizes[i].width;
+            }
+        }
+        return result.trim();
```
ChronosSF: at a glance something called generateColumnSize**s** which returns a single string is very strange.

please, refactor

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -1502,17 +1571,33 @@ export class IgxColumnGroupComponent extends IgxColumnComponent implements After
...
-                return acc;
-            }
-
-            if (typeof val.width === 'string' && val.width.indexOf('%') !== -1) {
-                isChildrenWidthInPercent = true;
-            }
-            return acc + parseInt(val.width, 10);
-        }, 0)}`;
+        let isChildrenWidthInPercent = false, width;
+        if (this.grid && this.grid.enableMRL) {
```
ChronosSF: this getter is probably used a lot and the calculation doesn't look very quick - consider caching

ChronosSF: I also see the two if branches are quite similar so you can reuse more code
this might not be an issue if we extend the column group component for MRL

projects/igniteui-angular/src/lib/grids/column.component.ts
dkamburov: `parseInt('span 3', 10)` would be `NaN`

gedinakova:
@dkamburov Column moving issue
Currently, column moving seems possible, because both drag and drop indicators are shown, but on releasing the mouse columns do not change their positions. Only columns that do not have any of the layout related properties set can be rearranged.

ChronosSF:
@gedinakova , we are not targeting feature integration with this PR. It implements MRL in a grid without features so it is easier for review. We'll then PR feature integration implementation separately for each group of features.

dkamburov:
@gedinakova In this PR we don't have any feature integration with MRL. And ColumnMoving will only be able to move column layout blocks. Basically single column won't be allowed to move (unless it is the only one into the column layout)

gedinakova:
@dkamburov What do you think of making the `colEnd` and `rowEnd` inputs accept number values as well as string? Currently, they support specifying the end index but as a string only, so we need to have `[colEnd]="'5'"`. It would be nice if we can have: `[colStart]="1" [colEnd]="5"`.

gedinakova:
@dkamburov Currently, it is allowed to have overlapping columns on a single row? Do we plan to throw an error when the layout is not setup correctly or are we letting developers to make sure the layout is OK?

dkamburov:
@gedinakova Our idea was to adopt [Grid Layout](https://www.w3.org/TR/css-grid-1/). And `colEnd` represents [grid-column-end](https://www.w3.org/TR/css-grid-1/#propdef-grid-column-end) property and the value for that property can be `span n`. Although we can consider changing  `colEnd` and `rowEnd` to `colSpan` and `rowSpan` and then their type can be `number`.

We don't plan to have an error. If the layout is wrong, the browser will still render it.

gedinakova:
Thanks, I'll review the Grid Layout. However, in the current version [colEnd]="'5'" produces the same result as [colEnd]="'span 4'", hence my suggestion to allow just [colEnd]="5".

dkamburov:
Good point! We'll discuss it and probably change them to be numbers.

gedinakova:
@dkamburov Currently, it seems impossible to have a column layout block with a single column that spans multiple rows, even if the neighbor layout spans multiple rows.
<igx-column-layout>
    ... 3 rows layout
</igx-column-layout>
<igx-column-layout>
    <igx-column field="Fax" [rowStart]="1" [rowEnd]="'span 3'" [colStart]="1"></igx-column>
</igx-column-layout>
The 'Fax' column header would be on a single row while the cells are on all 3 rows.

gedinakova:
Headers are not properly arranged for layouts like (while cells seem fine):
```html
<igx-column-layout>
    ... 3 rows layout
</igx-column-layout>
<igx-column-layout>
    <igx-column field="PostalCode" [rowStart]="1" [rowEnd]="'span 3'" [colStart]="1"></igx-column>
    <igx-column field="Fax" [rowStart]="1" [rowEnd]="'span 2'" [colStart]="2"></igx-column>
</igx-column-layout>
```
Result:
image

OR:
```html
<igx-column-layout>
    ... 3 rows layout
</igx-column-layout>
<igx-column-layout>
    <igx-column field="PostalCode" [rowStart]="1" [rowEnd]="'span 3'" [colStart]="1"></igx-column>
    <igx-column field="Fax" [rowStart]="1" [rowEnd]="'span 3'" [colStart]="2"></igx-column>
</igx-column-layout>
```
Result:
image

gedinakova:
@dkamburov Currently, column layout cannot be used in igxTreeGrid - no columns are shown without any errors thrown.

MayaKirova:
@gedinakova @ChronosSF  As MRL will only be initially supported for normal igxGrid (not supported for TreeGrid/HierarchicalGrid) we have explicitly excluded group layout columns in those grids with the following commit: 
https://github.com/IgniteUI/igniteui-angular/pull/4486/commits/79065abf977003594a0c1dd7a3f02ba7d881e447

At this time we don't plan to throw errors, just ignore invalid configurations, like setting mrl config + column groups (mch) config or setting mrl in treegrid or hierarchical grid.

gedinakova:
The layout breaks if even a single column has no [colStart] specified but [rowStart] can be omitted. 
E.g. the following setup (no rowStart): 
```html
<igx-column-layout [movable]="false" [pinned]="false">
       <igx-column [colStart]="1" [colEnd]="5" field="ContactName"></igx-column> - no rowStart
       <igx-column [rowStart]="1" [colStart]="5" field="ContactTitle"></igx-column>
       <igx-column [rowStart]="1" [colStart]="6" field="Country"></igx-column>
       <igx-column [rowStart]="2" [colStart]="1" [colEnd]="3" field="Phone"></igx-column>
       <igx-column [rowStart]="2" [colStart]="3" [colEnd]="5" field="City"></igx-column>
       <igx-column [rowStart]="2" [colStart]="5" [colEnd]="7" field="Address"></igx-column>
       <igx-column [rowStart]="3" [colStart]="1" [colEnd]="7" sortable="true" field="CompanyName"></igx-column>
</igx-column-layout>
```
Produces the following layout:
image

While the following setup (no colStart): 
```html
<igx-column-layout [movable]="false" [pinned]="false">
       <igx-column [rowStart]="1" [colEnd]="5" field="ContactName"></igx-column> - no colStart
       <igx-column [rowStart]="1" [colStart]="5" field="ContactTitle"></igx-column>
       <igx-column [rowStart]="1" [colStart]="6" field="Country"></igx-column>
       <igx-column [rowStart]="2" [colStart]="1" [colEnd]="3" field="Phone"></igx-column>
       <igx-column [rowStart]="2" [colStart]="3" [colEnd]="5" field="City"></igx-column>
       <igx-column [rowStart]="2" [colStart]="5" [colEnd]="7" field="Address"></igx-column>
       <igx-column [rowStart]="3" [colStart]="1" [colEnd]="7" sortable="true" field="CompanyName"></igx-column>
</igx-column-layout>
```
Produces the following layout:
image

PR #4511, opened 2019-04-10
Focusing on parent grid using keyboard navigation in IE.

### LINKED ISSUE #4488: In IE parent grid scrolls when using ctr + right navigation.
## Description  
Currently when using ctr + right for navigating to the last cell parent grid scrolls and child grid flips on IE.
 * browser: IE

## Steps to reproduce  

1. Open hierarchical grid dev sample on IE.
2. Select the first cell of the first child grid.
3. Use ctr + right for navigation.

## Result  
Child grid flips and parent grid scrolls down even though the cell is visible without scrolling.

## Expected result  
Grid should stay in same position without scrolling.

## Attachments  
KeyboardNavIE

MartinKamenov:
Closes #4488
Internet explorer doesn't support [preventScroll](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) option. Try this sample [here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#Focus_prevent_scroll) (Click on the "Click me to focus on the button without scrolling!" in IE).
This is why we should not focus on a child grid but rather on the parent grid in some cases when using keyboard navigation in IE.
Please test the solution for IE with ctr + right, ctr + left, home, end, arrowLeft and arrowRight navigation keys.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -553,6 +555,15 @@ export class IgxGridNavigationService {
             });
         this.horizontalScroll(rowIndex).scrollTo(unpinnedIndex);
     }
+
+    private focusGridWithoutScrolling() {
```
MayaKirova: The base grid navigation service should not contains checks for hierarchicalgrid specific properties.
HierarchicalgGrid specific code should be in hierarchical-grid-navigation.service.
Make this function protected and overwrite it in the hierarchical grid navigation service with its specific code.

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -553,6 +555,15 @@ export class IgxGridNavigationService {
...
+        if(isIE() && (<IgxHierarchicalGridComponent>this.grid).parent) {
+            (<IgxHierarchicalGridComponent>this.grid).parent.nativeElement.focus();
```
MayaKirova: This may not work when navigating in nested children. Use the hierarchical grid’s rootGrid  prop instead to get the root grid.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -194,6 +195,18 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
         }
     }
 
+    public onKeydownArrowRight(element, rowIndex, visibleColumnIndex, isSummary = false) {
```
MayaKirova: I don’t see a need to add additional parameters and overwrite the onKeydownArrowRight, onKeydownArrowLeft, onKeydownHome methods.
This will also not cover all possible scenarios where the grid may flicker on IE as if you make a global search and check where this.grid.nativeElement.focus({ preventScroll: true }) is used you’ll see that it is called in other functions as well (in the base grid navigation and also in the hierarchical grid navigation).

The way it was implemented in the first commit to this PR was better. Just apply the changes suggested there so that each grid implements its own focusGridWithoutScrolling method in their related navigation service, move the hierarchical grid specific code in the method of the hierarchical grid service and make sure to replace all instances where this.grid.nativeElement.focus({ preventScroll: true }) is used with the this.focusGridWithoutScrolling() method call.

For example:

In grid-navigation.service.ts

protected focusGridWithoutScrolling() {
	this.grid.nativeElement.focus({ preventScroll: true });
}

In hierarchical-grid-navigation.service.ts

protected focusGridWithoutScrolling() {
	if(isIE()) {
		this.grid.rootGrid.nativeElement.focus();
} else {
	super.focusGridWithoutScrolling();
}
}

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -122,6 +122,7 @@ export class IgxGridNavigationService {
                 element.nextElementSibling.focus({ preventScroll: true });
             }
         } else {
+            this.getFocusableGrid().nativeElement.focus({ preventScroll: true });
```
ChronosSF: as ```performHorizontalScrollToCell``` does the same, this line is not needed

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -134,6 +135,7 @@ export class IgxGridNavigationService {
         if (!element.previousElementSibling && this.grid.pinnedColumns.length && index === - 1) {
             element.parentNode.previousElementSibling.focus({ preventScroll: true });
         } else if (!this.isColumnLeftFullyVisible(visibleColumnIndex - 1)) {
+            this.getFocusableGrid().nativeElement.focus({ preventScroll: true });
```
ChronosSF: as ```performHorizontalScrollToCell``` does the same, this line is not needed

VMihalkov:
The original issue seems to be fixed, but when using Ctrl + down there is some up/down jumping. Here are the steps to reproduce:

1. Open hierarchical grid dev sample on IE.
2. Select the first cell of the first child grid.
3. Use ctr + down for navigation.
CtrlDownKeysJumping

MayaKirova:
@VMihalkov Please re-test Ctrl+Down with nested child grids in Chrome as well to make sure there are no regressions.

PR #4540, opened 2019-04-15
Slider labels feature

### LINKED ISSUE #2622: Range Slider Thumps collapsing
### Description  
The Range Slider has a left and a right thumb that appear during interaction to both ease the end user navigation and also display the current slider value. These need to stay expanded whenever the user interacts with the control but the range thumbs currently hide after several seconds in Internet Explorer 11.
This is a duplicate of https://github.com/IgniteUI/igniteui-angular/issues/463 although the issue still seems to exist in IE 11.

 * igniteui-angular version: 6.1.5
 * browser: Internet Explorer 11

### Steps to reproduce  

1. Adjust slider min/max values in IE 11.

### Result  
the label disappears while still dragging the dot

### Expected result  
Thumbs to stay visible during the whole interaction

### LINKED ISSUE #3449: Number inside the slider component balloon
Inside the balloon of the slider component , can the number there be bound to data? I know it makes sense to have equal numbers of steps when you are sliding, however sometimes timelines for some projects makes no sense and have no equal range steps number. 

for example for this timeline i have 1 day - 3 days - 7 days - 4 weeks - 12 weeks. obviously this time frame makes no sense. so I will bind the value of the slider to a label to make this possible, and I will hide the balloon. 

Since the balloon is not showing. can the number inside that balloon be bound if I decide to show it ?

Thanks

### LINKED ISSUE #4559: When change the slider type at run time the slider is not updated correctly
## Description  
When change the slider type at run time the slider is not updated correctly 

 * browser: n/a

## Steps to reproduce  

1. Open dev sample slider
2. Change the type of the slider at run time

Note: Same is valid when change step.

## Result  
The slider interval is not correct

## Expected result  
The slider should be updated correctly when change its type

## Attachments  

image

### LINKED ISSUE #4562: In slider with type Range when change the lower value to be equal or greater than the upper the range is not correct
## Description  
In slider with type Range when change the lower value to be equal or greater than the upper the range is not correct

 * browser: n/a

## Steps to reproduce  

1. Open dev sample slider
2. In the second sample (this with type Range) start moving the lower value

## Result  
When the lower becomes equal to upper value the range is changed to 59/61

## Expected result  
When the lower becomes equal to upper value the range is should be to 60/60, and after one step the range should be 60/61.
It should be allowed to lower and upper values to be equal and when moving only one range the other should not be changed

Aleksandyr:
Closes #3449 
Closes #4562
Closes #4559
Closes #2622

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


Aleksandyr:
This PR needs further updates related to the thumb labels, due providing users with ability to implement custom templates representing those labels. The main goal for exposing them was to allow users cut off the text that the label contains. As we've already discussed with @simeonoff, below are the changes that need to be added.
%igx-thumb-label {
        max-width: rem(80px);
        overflow: hidden;
}

Aleksandyr:
@simeonoff has updated the [slider sample](https://github.com/IgniteUI/igniteui-angular/pull/4540/commits/cc58889f533b5bfa7042d5cb0b3252e7288e88c5) in order to show how the user can manipulate the label in order to fit the text within the thumbs.

zdrawku:
@Aleksandyr update che changelog as well

projects/igniteui-angular/src/lib/slider/thumb/thumb-slider.component.ts
```diff
@@ -0,0 +1,197 @@
...
+    @Input()
+    public step: number;
+
+    @Output()
+    public onThumbValueChange = new EventEmitter<number>();
+
+    @HostBinding('attr.tabindex')
+    public tabindex = 0;
+
+    @HostBinding('class.igx-slider__thumb-to')
```
nrobakova: Class name have to change to igx-slider__thumb, because it same for to and from thumb

nrobakova:
I tested this PR and found some issues:
1. When slider type is Range and move with the key board the next trump is not previewed, and if I move over it and continue navigation it does not hides
2. When change slider type from Slider to Range, the first trump cannot be moved
3. When remove labels for a slider, then add new with 2 items and then add again labels with 6 items the  max value of the slider becomes 1

Aleksandyr:
API documentation should be updated.

Issue #4367, opened 2019-03-20
Excel style filtering - display density design and UX changes


DaniMarinov:
## Description  
Due to the various actions within the ESF dialog, it has to be optimized for comfortable, cozy, compact display density which leads to some UX and design changes.
Check specification here:
https://github.com/IgniteUI/igniteui-angular/wiki/Excel-Style-Filtering

* [x] Tests
* [x] Development
* [x] CHANGELOG
* [x] Samples
* [x] Localization Strings (https://github.com/IgniteUI/igniteui-angular-i18n)

bkulov:
@DaniMarinov could you please add these designs to the spec too?

DaniMarinov:
@bkulov The spec is updated, so you can find them there, too. 
https://github.com/IgniteUI/igniteui-angular/wiki/Excel-Style-Filtering#42-display-density

tachojelev:
Related to #4300

PR #4614, opened 2019-04-22
Add display density to Excel Style Filtering

### LINKED ISSUE #4300: Use displayDensity of the igxGrid in ESF dialog
## Description  
Excel style filtering dialog doesn't use igxGrid's display density. The components that need to have 'displayDensity' input are:
    - igx-list
    - igx-checkbox
    - igx-button

 * browser: all

## Steps to reproduce  

1. Open 'Grid Cell Editing' demo
2. Enable Filtering and set it to be excel style
3. Open Filter menu for some column

## Result  
The components inside of the menu doesn't have 'Compact' density

## Expected result  
The components inside of the menu should have 'Compact' density

tachojelev:
#4367 
Closes #4300

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-default-expression.component.html
```diff
@@ -33,7 +33,7 @@
     />
 </igx-input-group>
 
-<igx-input-group #inputGroupValues type="box" [displayDensity]="'compact'" [supressInputAutofocus]="true">
```
bkulov: Changing the hardcoded display density to the one from the grid will make this input to be bigger then before, regardless of grid's display density. The dialog itself will have the same height and when you add new filter condition it won't be fully visible. My suggestion is to scroll to the end so the newly added filtering condition will be fully visible.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
```diff
@@ -31,6 +31,9 @@ export class IgxExcelStyleSearchComponent {
     @ViewChild('input', { read: IgxInputDirective })
     public searchInput: IgxInputDirective;
 
+    @Input()
+    public grid: any;
```
bkulov: I don't think you need the whole grid here. You only need displayDensity.
Also a good practice is to give specific type to properties, methods, parameters, etc. and don't use `any`. I suppose you've used `any` to avoid circular dependency but if you just pass displayDensity there won't be any.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-sorting.component.html
```diff
@@ -1,12 +1,23 @@
-<header class="igx-typography__overline"><span>{{ grid.resourceStrings.igx_grid_excel_filter_sorting_header }}</span></header>
+<header [ngClass]="(grid.displayDensity==='compact')? 'igx-typography__subtitle-1': 'igx-typography__overline'">
+    <span>{{ grid.resourceStrings.igx_grid_excel_filter_sorting_header }}</span>
+</header>
 <igx-buttongroup #sortButtonGroup [multiSelection]="false">
-    <span tabindex="0" igxButton type="button" [attr.data-togglable]="true" (click)="onSortButtonClicked(1)">
+    <span tabindex="0" igxButton  type="button" [displayDensity]="grid.displayDensity" [attr.data-togglable]="true" (click)="onSortButtonClicked(1)">
```
bkulov: Extra space.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/grid.excel-style-filtering.component.html
```diff
@@ -25,6 +48,7 @@ <h4 class="igx-typography__h6">{{ column.header || column.field }}</h4>
         <ng-template #defaultExcelStyleMovingTemplate>
             <igx-excel-style-column-moving
                 class="igx-excel-filter__move"
```
bkulov: Shouldn't you remove this class as it will be conditionally applied by `[ngClass]`?

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-default-expression.component.html
tachojelev: @bkulov I agree that whenever a new filtering condition is added, we should scroll to the bottom/end so both the new filtering condition and the "Add Filter" button are visible.
However since the newly added filtering condition was not fully visible up until now as well, we discussed that we should probably implement the **scrolling to bottom after adding new filtering condition** functionality in separate a issue (If would like, you can create it by including all the information that you think is necessary).

gedinakova:
@tachojelev @simeonoff Currently, the header showing the column's name is always the same irrespective of the specified density. According to the spec, the header should be the same for Comfortable and Cosy, but should be smaller when Compact is applied.
image

gedinakova:
@tachojelev We should better align the sort and move buttons vertically when Compact density is applied.
image

gedinakova:
@tachojelev Scrolling issue when density is Cosy or Compact
1. Set display density to Cosy or Compact
2. Drag the scrollbar thumb of the list to the bottom. 
Actual result: The tumb moves further down and when Compact it may even disappear.
Expected result: Scrollbar should be sized according to the items.
image

tachojelev:
@gedinakova Scrolling issue for **compact** and **cosy** was fixed.

gedinakova:
@tachojelev @simeonoff Scrolling is now working fine. Sort and move buttons are properly aligned. List class is OK for all densities.

PR #4671, opened 2019-05-02
time picker: dropdown mode retemplating ability 

### LINKED ISSUE #4656: Time picker component fails on dropdown mode in combination with igxTimePickerTemplate modifications
## Description  
If igx-time-picker [mode]="dropdown" and the igxTimePickerTemplate directive is used to modify the layout of the igx-time-picker, the time picker crashes, and does not render any values. 

 * igniteui-angular version: 7.2.6
 * browser: Chrome

## Steps to reproduce  

1. Open the following stackblitz application containing the code i'm trying to run https://stackblitz.com/edit/angular-xepcq4
2. Click the suffix icon to open the dialog 
3. Click the suffic icon again

## Result  
After suffix is clicked a second time, values will be rendered, and can be edited, but dialog cannot be closed and values will not be saved to the input field

## Expected result  
The dropdown opens and render values on first click, when the user clicks outside the bounds of the dialog or exits in any other way, the dialog is closed and specified value are displayed in the input field.

SAndreeva:
Closes #4656 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/time-picker/README.md
+++ b/projects/igniteui-angular/src/lib/time-picker/README.md
@@ -68,6 +68,45 @@ mode = InteractionMode.DropDown;
 </igx-time-picker>
 ```
 
+In order to re-template the TimePicker in `dropdown` mode, you should either mark an element with `dropDownTarget` template reference variable or set the overlaySettings appropraitaly:
+
+```html
+<igx-time-picker>
+    <ng-template igxTimePickerTemplate let-openDialog="openDialog" let-value="value" let-displayTime="displayTime">
+        <igx-input-group (click)="openDialog()" #dropDownTarget> 
+            <label igxLabel>Time</label>
+            <input igxInput [value]="displayTime"/>
+        </igx-input-group>
+    </ng-template>
+</igx-time-picker>
+```
+or
+
+```html
+<igx-time-picker>
+    <ng-template igxTimePickerTemplate [overlaySettings]="myOverlaySettings" let-openDialog="openDialog" let-value="value" let-displayTime="displayTime">
+        <igx-input-group (click)="openDialog()" #myTarget> 
+            <label igxLabel>Time</label>
+            <input igxInput [value]="displayTime"/>
+        </igx-input-group>
+    </ng-template>
+</igx-time-picker>
+```
+```typescript
+    myOverlaySettings = {
+        modal: false,
+        closeOnOutsideClick: true,
+        positionStrategy: new AutoPositionStrategy()
+    };
+
+    @ViewChild('myTarget')
+    public target: IgxInputDirective;
+
+    ngAfterViewInit() {
+        this.myOverlaySettings.positionStrategy.settings.target = this.target.nativeElement;
+    }
+```
+
 # API
 
 ###### Inputs
--- a/projects/igniteui-angular/src/lib/time-picker/README.md
+++ b/projects/igniteui-angular/src/lib/time-picker/README.md
@@ -68,10 +68,10 @@ mode = InteractionMode.DropDown;
 </igx-time-picker>
 ```
 
-In order to re-template the TimePicker in `dropdown` mode, you should either mark an element with `dropDownTarget` template reference variable or set the overlaySettings appropraitaly:
+In order to re-template the TimePicker in `dropdown` mode, you should either mark an element with `dropDownTarget` template reference variable or set the overlaySettings appropriately:
 
 ```html
-<igx-time-picker>
+<igx-time-picker [mode]="'dropdown'">
     <ng-template igxTimePickerTemplate let-openDialog="openDialog" let-value="value" let-displayTime="displayTime">
         <igx-input-group (click)="openDialog()" #dropDownTarget> 
             <label igxLabel>Time</label>
--- a/projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts
+++ b/projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts
@@ -16,7 +16,6 @@ export interface IgxTimePickerBase {
     format: string;
     promptChar: string;
     cleared: boolean;
-    collapsed: boolean;
     mode: InteractionMode;
     nextHour();
     prevHour();
--- a/projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts
+++ b/projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts
@@ -28,7 +28,7 @@ export interface IgxTimePickerBase {
     scrollHourIntoView(item: string): void;
     scrollMinuteIntoView(item: string): void;
     scrollAmPmIntoView(item: string): void;
-    hideOverlay(): void;
+    close(): void;
     parseMask(preserveAmPm?: boolean): string;
 }
 
--- a/projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts
+++ b/projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts
@@ -28,7 +28,7 @@ export interface IgxTimePickerBase {
     scrollHourIntoView(item: string): void;
     scrollMinuteIntoView(item: string): void;
     scrollAmPmIntoView(item: string): void;
-    close(): void;
+    hideOverlay(): void;
     parseMask(preserveAmPm?: boolean): string;
 }
 
--- a/projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
+++ b/projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
@@ -1165,21 +1165,22 @@ export class IgxTimePickerComponent implements
         if (this.collapsed) {
             this.collapsed = false;
 
-            let settings;
+            let settings, posStrategy;
             if (this.mode === InteractionMode.Dialog) {
                 settings = this.overlaySettings || this._dialogOverlaySettings;
             }
 
             if (this.mode === InteractionMode.DropDown) {
                 settings = this.overlaySettings || this._dropDownOverlaySettings;
+                posStrategy = settings.positionStrategy;
 
-                if (this.group && settings.positionStrategy) {
-                    settings.positionStrategy.settings.target = this.group.element.nativeElement;
-                } else if (this.templateDropDownTarget && settings.positionStrategy) {
-                    settings.positionStrategy.settings.target = this.templateDropDownTarget.nativeElement;
-                } else if (settings.positionStrategy && !settings.positionStrategy.settings.target) {
+                if (this.group && posStrategy) {
+                    posStrategy.settings.target = this.group.element.nativeElement;
+                } else if (this.templateDropDownTarget && posStrategy) {
+                    posStrategy.settings.target = this.templateDropDownTarget.nativeElement;
+                } else if (!posStrategy || (posStrategy && !posStrategy.settings.target)) {
                     throw new Error('There is no target element for the dropdown to attach.' +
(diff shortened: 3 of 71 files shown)


src/app/time-picker/time-picker.sample.html
```diff
@@ -72,5 +72,37 @@ <h4 class="sample-title">Templated Time Picker</h4>
...
+                    </ng-template>
+                </igx-time-picker>
+            </div>
+        </article>
+        <article class="timepicker-column">
+            <h4 class="sample-title">Templated Time Picker</h4>
+            <p class="sample-description">Time picker with dropdown mode + overlay settings </p>
+            <div class="preview">
+                <igx-time-picker [overlaySettings]="myOverlaySettings" format="HH:mm" mode="dropdown">
+                    <ng-template igxTimePickerTemplate let-openDialog="openDialog" let-value="value" let-displayTime="displayTime">
```
bkulov: For some reason the dropdown opens bellow the input and goes out of the screen. This is applicable to both samples.

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -1166,8 +1172,14 @@ export class IgxTimePickerComponent implements
 
             if (this.mode === InteractionMode.DropDown) {
                 settings = this.overlaySettings || this._dropDownOverlaySettings;
-                if (settings.positionStrategy && settings.positionStrategy.settings) {
+
+                if (this.group && settings.positionStrategy) {
```
bkulov: `settings.positionStrategy` is included in all if clauses. Consider extracting it in a wrapper if clause.

sboykova:
Please, add a note for retemplating in the time-picker README file. It is good to have a sample and a note in docs, as well.

src/app/time-picker/time-picker.sample.html
sboykova: AutoPositionStrategy can be used instead of ConnectedPositioningStrategy for the drop-down positionStrategy.

sboykova:
Tests can be added to cover the new retemplating functionality.

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -1159,15 +1165,22 @@ export class IgxTimePickerComponent implements
         if (this.collapsed) {
             this.collapsed = false;
 
-            let settings;
+            let settings, posStrategy;
```
bkulov: `posStrategy` is not assigned more than once so no need to declare it here. You could declare it in line 1175.

src/app/time-picker/time-picker.sample.html
```diff
@@ -72,5 +72,37 @@ <h4 class="sample-title">Templated Time Picker</h4>
...
             </div>
         </article>
+        <article class="timepicker-column">
+            <h4 class="sample-title">Templated Time Picker</h4>
+            <p class="sample-description">Time picker with dropdown mode</p>
+            <div class="preview">
+                <igx-time-picker [value]="today" format="HH:mm" mode="dropdown">
+                    <ng-template igxTimePickerTemplate let-openDialog="openDialog" let-value="value" let-displayTime="displayTime">
+                        <igx-input-group>
+                            <input #dropDownTarget igxInput class="date-picker" [value]="displayTime" (keyup.enter)="enterKeyUp(dropDownTarget.value, value)"/>
```
bkulov: It's better to update the value on input blur instead of on Enter key up. This will ensure that when the user modifies the value in the input and then opens the drop-down, the drop-down will show the same value present in the input.

src/app/time-picker/time-picker.sample.html
```diff
@@ -72,5 +72,37 @@ <h4 class="sample-title">Templated Time Picker</h4>
...
+            </div>
+        </article>
+        <article class="timepicker-column">
+            <h4 class="sample-title">Templated Time Picker</h4>
+            <p class="sample-description">Time picker with dropdown mode + overlay settings </p>
+            <div class="preview">
+                <igx-time-picker [value]="today" [overlaySettings]="myOverlaySettings" format="HH:mm" mode="dropdown">
+                    <ng-template igxTimePickerTemplate let-openDialog="openDialog" let-value="value" let-displayTime="displayTime">
+                        <igx-input-group>
+                            <input #target igxInput class="date-picker" [value]="displayTime"/>
```
bkulov: Apply the same logic as the other sample with retemplated input group.

src/app/time-picker/time-picker.sample.html
wnvko: The `container` of the time picker has `hidden` set to true in the markup [here](https://github.com/IgniteUI/igniteui-angular/blob/3660f146cb956636aa79fe26cb1cbce223ab27df/projects/igniteui-angular/src/lib/time-picker/time-picker.component.html#L27)
When the element is hidden in the DOM it has no size. As a result you are passing sizeless element to the `overlay`. `Overlay` checks if element size goes out of the screen when auto position strategy is in use. In your case element has no size so `overlay` decides it should not fit the element in the view port.
To solve this you should either remove `hidden` and hide element some other way, or you may use `igxToggle` directive where this is solved.
Note that you are also incorrectly setting `collapsed` to `false` in [openDialog](https://github.com/IgniteUI/igniteui-angular/blob/3660f146cb956636aa79fe26cb1cbce223ab27df/projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts#L1160). This should be set in `onOpened` event of the `overlay` as you are doing it [here](https://github.com/IgniteUI/igniteui-angular/blob/3660f146cb956636aa79fe26cb1cbce223ab27df/projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts#L692).

projects/igniteui-angular/src/lib/time-picker/README.md
```diff
@@ -68,6 +68,45 @@ mode = InteractionMode.DropDown;
 </igx-time-picker>
 ```
 
+In order to re-template the TimePicker in `dropdown` mode, you should either mark an element with `dropDownTarget` template reference variable or set the overlaySettings appropraitaly:
```
wnvko: appropriately maybe :)

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -542,7 +549,6 @@ export class IgxTimePickerComponent implements
     private _minuteView = [];
     private _ampmView = [];
 
-    private _overlayId: string;
     private _dateFromModel: Date;
     private _destroy$ = new Subject<boolean>();
     private _positionSettings: PositionSettings;
```
wnvko: you do not use this field and may remove it
you also do not need _dialogOverlaySettings field

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -704,16 +682,13 @@ export class IgxTimePickerComponent implements
             this._generateAmPm();
         }
 
-        this._positionSettings = {
-            horizontalStartPoint: HorizontalAlignment.Left,
-            verticalStartPoint: VerticalAlignment.Bottom
-        };
+        this._positionSettings = { };
```
wnvko: you do not need this

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -704,16 +682,13 @@ export class IgxTimePickerComponent implements
...
 
         this._dropDownOverlaySettings = {
             modal: false,
             closeOnOutsideClick: true,
             scrollStrategy: new AbsoluteScrollStrategy(),
-            positionStrategy: new ConnectedPositioningStrategy(this._positionSettings)
+            positionStrategy: new AutoPositionStrategy(this._positionSettings)
```
wnvko: this is same as:
positionStrategy: new AutoPositionStrategy()

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -704,16 +682,13 @@ export class IgxTimePickerComponent implements
...
         };
 
         this._dialogOverlaySettings = {};
```
wnvko: you can remove this line too

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -1156,29 +1159,34 @@ export class IgxTimePickerComponent implements
...
-            this.collapsed = false;
-
+        if (this.toggleRef.collapsed) {
             let settings;
             if (this.mode === InteractionMode.Dialog) {
                 settings = this.overlaySettings || this._dialogOverlaySettings;
             }
 
             if (this.mode === InteractionMode.DropDown) {
                 settings = this.overlaySettings || this._dropDownOverlaySettings;
```
wnvko: you can remove _dialogOverlaySettings and assign it like this:
settings = this.overlaySettings;

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -1156,29 +1159,34 @@ export class IgxTimePickerComponent implements
...
             if (this.outlet) {
                 settings.outlet = this.outlet;
             }
 
-            this._overlayId = this.overlayService.attach(this.container, settings);
-            this.overlayService.show(this._overlayId);
+            this.toggleRef.open(settings);
 
         } else if (this.mode === InteractionMode.DropDown) {
+            this.toggleRef.close();
```
wnvko: Calling close() in openDialog() looks strange to me. Probably you should refactor or at least rename this.

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -1451,7 +1459,7 @@ export class IgxTimePickerComponent implements
      * @hidden
      */
     public hideOverlay(): void {
```
wnvko: close or closeDialog maybe

wnvko:
One more point - you may change #tp in all code snippets to something more descriptive

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
SAndreeva: I would rather keep the name the way it is. We expose **openDialog**  in the context object and changing the name will be a breaking change for the users. I agree that calling **close()** in **openDialog()** does not make much sense but it is necessary for toggling the dropdown when clicking on the icon. I was thinking of using the **toggle()** method but it does not work well when having a retemplated input group.

projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts
```diff
@@ -29,7 +28,7 @@ export interface IgxTimePickerBase {
     scrollHourIntoView(item: string): void;
     scrollMinuteIntoView(item: string): void;
     scrollAmPmIntoView(item: string): void;
-    hideOverlay(): void;
+    close(): void;
```
kdinev: Please add a migration for this method.

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
bkulov: We won't change this name for now as it will introduce breaking change and we don't have migration for method renaming. Writing a migration will take some time and we can do it later.

PR #4707, opened 2019-05-07
fix(esf): Excluding the SelectAll from search results #4020

### LINKED ISSUE #4020: "Select All" should not be treated as a match when searching.
## Description  
When performing search in Excel-style filtering UI, "Select All" option should not be treated as a match similar to the behavior in MS Excel.

 * browser: N/A

## Steps to reproduce  

1. Run Grid Column Moving dev demo
2. Open Excel-style filtering dropdown on the first column
3. Type 'se' in the search box
4. Note the results

## Result  
"Select All" option remains as the only match. Checking/unchecking it an applying the filter does nothing.

## Expected result  
"Select All" should not be treated as a match.

astaev:
Closes #4020 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
@@ -45,19 +45,19 @@ export class IgxExcelStyleSearchComponent {
     }
 
     public onCheckboxChange(eventArgs: IChangeCheckboxEventArgs) {
-        const selectAll = this.column.grid.resourceStrings.igx_grid_excel_select_all;
+        // const selectAll = this.column.grid.resourceStrings.igx_grid_excel_select_all;
         const selectedIndex = this.data.indexOf(eventArgs.checkbox.value);
-        if (eventArgs.checkbox.value.value === selectAll && selectedIndex === 0) {
+        if (selectedIndex === 0) {
             this.data.forEach(element => {
                 element.isSelected = eventArgs.checked;
                 this.data[0].indeterminate = false;
             });
         } else {
             eventArgs.checkbox.value.isSelected = eventArgs.checked;
-            if (!this.data.filter(el => selectedIndex !== 0).find(el => el.isSelected === false)) {
+            if (!this.data.slice(1, this.data.length).find(el => el.isSelected === false)) {
                 this.data[0].indeterminate = false;
                 this.data[0].isSelected = true;
-            } else if (!this.data.filter(el => selectedIndex !== 0).find(el => el.isSelected === true)) {
+            } else if (!this.data.slice(1, this.data.length).find(el => el.isSelected === true)) {
                 this.data[0].indeterminate = false;
                 this.data[0].isSelected = false;
             } else {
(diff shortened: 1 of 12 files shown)


projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
```diff
@@ -46,17 +46,18 @@ export class IgxExcelStyleSearchComponent {
 
     public onCheckboxChange(eventArgs: IChangeCheckboxEventArgs) {
         const selectAll = this.column.grid.resourceStrings.igx_grid_excel_select_all;
-        if (eventArgs.checkbox.value.value === selectAll) {
```
bkulov: This code will work fine without this modifications. But if you want to do them I suggest several modification. See my comments bellow.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
```diff
@@ -46,17 +46,18 @@ export class IgxExcelStyleSearchComponent {
...
+        const selectedIndex = this.data.indexOf(eventArgs.checkbox.value);
+        if (eventArgs.checkbox.value.value === selectAll && selectedIndex === 0) {
```
bkulov: You can remove `eventArgs.checkbox.value.value === selectAll` as `Select All` will always be the first item in the list and it duplicates the second check in the if clause.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
```diff
@@ -46,17 +46,18 @@ export class IgxExcelStyleSearchComponent {
...
             this.data.forEach(element => {
                 element.isSelected = eventArgs.checked;
                 this.data[0].indeterminate = false;
             });
         } else {
             eventArgs.checkbox.value.isSelected = eventArgs.checked;
-            if (!this.data.filter(el => el.value !== selectAll).find(el => el.isSelected === false)) {
+            if (!this.data.filter(el => selectedIndex !== 0).find(el => el.isSelected === false)) {
```
bkulov: You can use `this.data.slice(1, this.data.length)` instead of `this.data.filter(el => selectedIndex !== 0)` similar to what you're doing in the next file. It would be much faster for large arrays.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
```diff
@@ -46,17 +46,18 @@ export class IgxExcelStyleSearchComponent {
...
                 this.data[0].indeterminate = false;
                 this.data[0].isSelected = true;
-            } else if (!this.data.filter(el => el.value !== selectAll).find(el => el.isSelected === true)) {
+            } else if (!this.data.filter(el => selectedIndex !== 0).find(el => el.isSelected === true)) {
```
bkulov: You can use `this.data.slice(1, this.data.length)` instead of `this.data.filter(el => selectedIndex !== 0)` similar to what you're doing in the next file. It would be much faster for large arrays.

bkulov:
You could also add unit test to cover the change.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
astaev: The check for selected index is necessary, since if there is another checkbox with 'Select All' value it behaves like the main select all.

PR #4757, opened 2019-05-13
Multi Row Layout keyboard navigation

### LINKED ISSUE #4708: Cell navigation does not work along with Multi Row Layout group
## Description  
Cell navigation does not work when cell focus tries to moves to other group.

 * igniteui-angular version: 7.3.0-beta.1
 * browser: n/a

## Steps to reproduce  

1. Run the attached sample
2. Set focus on 'ID' cell
3. Hit Tab key 3 times

## Result  
Chrome and IE11:
Active cell disappears after navigation moves to unpinned group.

Firefox:
Active cell does not move to cells on unpinned group

## Expected result  
On all supported browsers, cell navigation works.

## Attachments  
[igx-grid-multi-row-layout.zip](https://github.com/IgniteUI/igniteui-angular/files/3155649/igx-grid-multi-row-layout.zip)

dkamburov:
Closes #4708
Related #4384

TODO:
 - [x] Automate scenarios with Tab navigation, including cell, row editing, and navigating with Ctrl
 - [x] Fix Ctrl + Up/Down navigation
 - [x] Fix navigation from pinned or unpinned area

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -235,7 +235,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
     @HostBinding('attr.data-visibleIndex')
     @Input()
     get visibleColumnIndex() {
-        return this.column.parent && this.column.parent.columnLayout ? this.column.visibleIndex : this._vIndex;
+        return this.column.columnLayoutChild ? this.column.visibleIndex : this._vIndex;
     }
 
     set visibleColumnIndex(val) {
@@ -506,7 +506,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
     protected get selectionNode(): ISelectionNode {
         return {
             row: this.rowIndex,
-            column: this.column.parent && this.column.parent.columnLayout ? this.column.parent.visibleIndex : this.visibleColumnIndex };
+            column: this.column.columnLayoutChild ? this.column.parent.visibleIndex : this.visibleColumnIndex };
     }
 
     protected isInCompositionMode = false;
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -861,7 +861,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
                     this.grid.navigation.onKeydownHome(node.row);
                     break;
                 }
-                this.grid.navigation.onKeydownArrowLeft(this.nativeElement, node.row, node.column, this);
+                this.grid.navigation.onKeydownArrowLeft(this.nativeElement, node.row, node.column, false, this);
                 break;
             case 'arrowright':
             case 'right':
@@ -869,7 +869,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
                     this.grid.navigation.onKeydownEnd(node.row);
                     break;
                 }
-                this.grid.navigation.onKeydownArrowRight(this.nativeElement, node.row, node.column, this);
+                this.grid.navigation.onKeydownArrowRight(this.nativeElement, node.row, node.column, false, this);
                 break;
             case 'arrowup':
             case 'up':
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -506,7 +506,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
     protected get selectionNode(): ISelectionNode {
         return {
             row: this.rowIndex,
-            column: this.column.parent && this.column.parent.columnLayout ? this.column.parent.index : this.visibleColumnIndex };
+            column: this.column.parent && this.column.parent.columnLayout ? this.column.parent.visibleIndex : this.visibleColumnIndex };
     }
 
     protected isInCompositionMode = false;
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -785,7 +785,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
         if (ctrl) {
             this.grid.navigation.goToLastCell();
         } else {
-            this.grid.navigation.onKeydownEnd(this.rowIndex);
+            this.grid.navigation.onKeydownEnd(this.rowIndex, false, this.rowStart);
         }
     }
 
@@ -794,7 +794,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
         if (ctrl) {
             this.grid.navigation.goToFirstCell();
         } else {
-            this.grid.navigation.onKeydownHome(this.rowIndex);
+            this.grid.navigation.onKeydownHome(this.rowIndex, false, this.rowStart);
         }
     }
 
@@ -863,7 +863,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
             case 'arrowleft':
             case 'left':
                 if (ctrl) {
-                    this.grid.navigation.onKeydownHome(node.row);
+                    this.grid.navigation.onKeydownHome(node.row, false, this.rowStart);
                     break;
                 }
                 this.grid.navigation.onKeydownArrowLeft(this.nativeElement, node.row, node.column, false, this);
@@ -871,7 +871,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
             case 'arrowright':
             case 'right':
                 if (ctrl) {
-                    this.grid.navigation.onKeydownEnd(node.row);
+                    this.grid.navigation.onKeydownEnd(node.row, false, this.rowStart);
                     break;
                 }
                 this.grid.navigation.onKeydownArrowRight(this.nativeElement, node.row, node.column, false, this);
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -43,7 +43,7 @@ import { DeprecateProperty } from '../core/deprecateDecorators';
     templateUrl: './cell.component.html'
 })
 export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
-
+    private _vIndex;
     /**
      * Gets the column of the cell.
      * ```typescript
@@ -233,7 +233,12 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
      */
     @HostBinding('attr.data-visibleIndex')
     @Input()
-    visibleColumnIndex = -1;
+    get visibleColumnIndex() {
+        return this.column.parent && this.column.parent.columnLayout ? this.column.visibleIndex : this._vIndex;
+    }
+    set visibleColumnIndex(val) {
+        this._vIndex = val;
+    }
 
     /**
      * Gets the ID of the cell.
@@ -497,7 +502,9 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
     }
 
     protected get selectionNode(): ISelectionNode {
-        return { row: this.rowIndex, column: this.visibleColumnIndex };
+        return {
+            row: this.rowIndex,
+            column: this.column.parent && this.column.parent.columnLayout ? this.column.parent.index : this.visibleColumnIndex };
     }
 
     protected isInCompositionMode = false;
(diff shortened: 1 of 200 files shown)


projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -497,7 +504,9 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
     }
 
     protected get selectionNode(): ISelectionNode {
-        return { row: this.rowIndex, column: this.visibleColumnIndex };
+        return {
+            row: this.rowIndex,
+            column: this.column.parent && this.column.parent.columnLayout ? this.column.parent.visibleIndex : this.visibleColumnIndex };
```
ChronosSF:
```typescript
this.column.parent && this.column.parent.columnLayout
```
looks repeated enough to be added as a getter

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -1768,25 +1793,67 @@ export class IgxColumnGroupComponent extends IgxColumnComponent implements After
...
+    public childrenVisibleIndexes = [];
     /**
      * Gets the width of the column layout.
      * ```typescript
      * let columnGroupWidth = this.columnGroup.width;
      * ```
      * @memberof IgxColumnGroupComponent
      */
-    get width() {
+    get width(): any {
```
ChronosSF: do you really need to change it from nothing to ```any``` ?

also, it seems you are parseInt-ing inside so it most likely can only return a number

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -3,6 +3,7 @@ import { IgxGridBaseComponent, FilterMode } from './grid-base.component';
 import { first } from 'rxjs/operators';
 import { IgxColumnComponent } from './column.component';
 import { IgxGridGroupByRowComponent } from './grid/groupby-row.component';
+import { IgxGridCellComponent } from './tree-grid';
```
ChronosSF: possibly wrong import

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
```diff
@@ -0,0 +1,466 @@
+import { Injectable } from '@angular/core';
+import { IgxGridBaseComponent, FilterMode } from './grid-base.component';
+import { first } from 'rxjs/operators';
+import { IgxColumnComponent } from './column.component';
+import { IgxGridGroupByRowComponent } from './grid/groupby-row.component';
+import { IgxGridCellComponent } from './tree-grid';
```
ChronosSF: this shouldn't be imported from there

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
```diff
@@ -0,0 +1,466 @@
...
+
+    public onKeydownArrowRight(element, rowIndex, visibleColumnIndex, isSummary = false, cell?) {
+        this.focusNextCellFromLayout(cell);
+    }
+
+    public onKeydownArrowLeft(element, rowIndex, visibleColumnIndex, isSummary = false, cell?) {
+        this.focusPrevCellFromLayout(cell);
+    }
+
+    public performTab(currentRowEl, rowIndex, visibleColumnIndex, isSummaryRow = false, cell?) {
```
ChronosSF: performTab and performShift tab look so similar - have you tried handling them with the same code with a shift flag determining if you search for +1 or -1?

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
```diff
@@ -0,0 +1,466 @@
...
+    public navigateUp(rowElement, currentRowIndex, visibleColumnIndex, cell?) {
+        if (cell) {
+            this.focusCellUpFromLayout(cell);
+        } else {
+            super.navigateUp(rowElement, currentRowIndex, visibleColumnIndex);
+        }
+    }
+
+    public navigateDown(rowElement, currentRowIndex, visibleColumnIndex, cell?) {
+        if (cell) {
```
ChronosSF: Why would this function only care about a single parameter. Can't these be integrated better? Maybe extend the SelectionNode interface with information that's needed for MRL so that you don't have to pass a MRL-only flag at the end of all MRL-enabled functions? This ```cell?``` looks very excessive. Besides the other params derive from properties coming from the Cell or its Column component...

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
```diff
@@ -0,0 +1,466 @@
...
+                } else if (prevRow) {
+                    prevRow.nativeElement.focus({ preventScroll: true });
+                }
+            }
+            return;
+        }
+        this._focusCell(upperElement);
+    }
+
+    private focusCellDownFromLayout(cell, isSummary = false) {
```
ChronosSF: all four functions (down, up, left, right) look similar. I am pretty sure locating the next cell can be done with more granular functions and the overarching flow can then be put into a single one saving code. This should be true at least for the Up/Down Next/Prev pairs.

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
```diff
@@ -0,0 +1,466 @@
...
+        this._focusCell(prevElement);
+    }
+
+    public onKeydownEnd(rowIndex, isSummary = false, cellRowStart?) {
+        const layouts = this.grid.columns.filter(c => c.columnLayout && !c.hidden).length;
+        const lastLayout = this.grid.columns.filter(c => c.columnLayout && !c.hidden)[layouts - 1];
+        const lastLayoutChildren = lastLayout.children;
+        const layoutSize =  lastLayout.getInitialChildColumnSizes(lastLayoutChildren).length;
+        const currentRowStart =  cellRowStart || this.grid.multiRowLayoutRowSize;
+        const nextElementColumn = lastLayout.children.find(c =>
```
ChronosSF: these finds are very likely repeated in other places. e.g. wouldn't End look like shift + tab from the start? At least regarding columns. It just looks like each function has calculations on its own and the calculations are extremely similar from one another yet repeated everytime.

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
MayaKirova: @ChronosSF The function checks if navigation comes from a cell (could come from something that is not a cell  - for example group row). The cell is also used to get all the mrl specific options - rowStart, colEnd etc.. Extending the SelectionNode  with the mrl data and using just as a single parameter that holds all relevant information is a good idea, however since currently each cell calls directly the navigation service with the related params (rowElement, currentRowIndex, visibleColumnIndex), changing this will require a refactoring in all navigation services and in all components that call them. Since this would be a big change I would prefer we don't do it as part of this PR - rather log a separate issue so that we can work on it separately.

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
MayaKirova: @ChronosSF There are similar in the sense that both find a target element, scroll it in view if necessary and then do something specific to the action ( navigateUp or navigateDown, moveNextEditable or movePreviousEditable etc.).  We cannot handle them completely with the same code without some branching. If you prefer we can extract the common action (find and scroll in view) in a function (something like _scrollToVisibleIndex).

projects/igniteui-angular/src/lib/grids/column.component.ts
MayaKirova: Base component defines type string for width. This was probably added to prevent incompatible type conversion. Otherwise it would resolve it as type number (due to the parseInt) and would throw an error.

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
MayaKirova: Tab/Shift+Tab works on visible indexes only, so there the check is simpler - just gets the largest visible index and passes it to the base navigateUp function.

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -1082,15 +1091,15 @@ export class IgxColumnComponent implements AfterContentInit {
...
-    getGridTemplate(isRow: boolean, isIE: boolean): string {
-        const colSpanAccum = (acc, val) => Math.max(val.colStart + val.gridColumnSpan - 1, acc);
-        const templateItems = !isRow ?
-            this.children && this.children.reduce(colSpanAccum, 1) || 1 :
-            this.grid.multiRowLayoutRowSize;
-        const generatedSizes = !isRow ? this.getColumnSizesString(this.children) : null;
-        return isIE ?
-        generatedSizes || `(1fr)[${templateItems}]` :
-            generatedSizes || `repeat(${templateItems},1fr)`;
+    getGridTemplate(isRow, isIE): string {
```
ChronosSF: why remove the parameter types?

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -569,16 +594,29 @@ export class IgxGridNavigationService {
...
-    private performHorizontalScrollToCell(rowIndex, visibleColumnIndex, isSummary = false) {
+    public performVerticalScrollToCell(rowIndex, visibleColIndex, cb?) {
+        this.grid.verticalScrollContainer.scrollTo(rowIndex);
+        this.grid.verticalScrollContainer.onChunkLoad
+        .pipe(first()).subscribe(() => {
+            cb();
+        });
+    }
+
+    public performHorizontalScrollToCell(rowIndex, visibleColumnIndex, isSummary = false, cb?) {
```
ChronosSF: more functions to add types for parameters, especially on cb

projects/igniteui-angular/src/lib/grids/grid/groupby-row.component.ts
```diff
@@ -244,14 +253,18 @@ export class IgxGridGroupByRowComponent {
         return this.grid.getColumnByName(this.groupRow.expression.fieldName).dataType;
     }
 
-    private handleTabKey(shift) {
+    private handleTabKey(shift, activeNode) {
```
ChronosSF: types

PavlovVasil:
@dkamburov @ChronosSF @MayaKirova @mpavlinov 
There are two issues I noticed while testing: 
1. The keyboard navigation scrolls the target cell into view in case it is not fully visible. However, if I scroll from the last cell of a row to the first of the next row, and there is a group row between the two rows, the navigation does not scroll horizontally, which results in navigating to partially visible cells. 
Steps to reproduce: 
    1. Run the dev demos and open http://localhost:4200/gridMRL
    2. Group by the ID column
    3. Select the last cell of the first row 
    4. Press "Tab"

Result: 

MRLSelectionNotScrolling

2. This reproduces in **Mozilla Firefox and MS Edge only** :
    1. Run the dev demos and open http://localhost:4200/gridMRL in Firefox/Edge
    2. Select the first cell of the first row - ALFKI
    3. Group by the ID column
    4. Scroll down using the mouse wheel 

Result - multiple simultaneously selected cells: 

 
MRLSelection

MayaKirova:
@PavlovVasil Please re-test with latest changes.

PavlovVasil:
@dkamburov @ChronosSF @MayaKirova @mpavlinov
The issue with "Tab" key not scrolling horizontally has been fixed. There are two more: 
1. When a cell from a pinned group gets focused, the bottom-border of the cells disappears. Running the samples repo and opening http://localhost:4200/samples/grid/grid-multi-row-layout would show you this if you select a cell in the first row: 

Sketch

The dev demos look like this if you click the button and pin the "Pin/Unpin Second Group" button at the bottom:
 
Sketch2

<br>
2. If Group By has been enabled: 
    1. Run the samples repo and open http://localhost:4200/samples/grid/grid-multi-row-layout
    2. Collapse all groups by clicking the button at the top left
    3. Expand the first three groups as well as the last one
    4. Click the middle cell of the last grid row - "London"
    5. Hold the up arrow key in order to scroll to the top
Result: 
When the navigation tries to scroll through the collapsed groups, it looses focus and stops working. The scroll position is not correct either, and scrolls to the bottom: 

MRLCollapsedGroups

MayaKirova:
@PavlovVasil Regarding 1). It is not an issue related to navigation, same can be reproduced in master if you select a cell range in MRL grid. Seems to be a styling issue due to an additional style that applies border-bottom: 0. Please log it separately.

MayaKirova:
@PavlovVasil Please re-test.

Issue #4678, opened 2019-05-03
Range error is thrown when you try to open MRL dev sample


ddincheva:
## Description  

 * igniteui-angular version: 7.3.x
 * browser: all

## Steps to reproduce  

1. Open the MRL dev sample
2. Observe the console

## Result  
What is the actual result after following the steps to reproduce?
![Screenshot_24](https://user-images.githubusercontent.com/34240583/57131961-baf30c00-6da6-11e9-96fa-b4eb95ca353e.png)

## Expected result  
What is the expected result after following the steps to reproduce?
No errors should be thrown and the column should be pinned

skrustev:
Currently in general the pinning with column widths set to percentages is not supported by the igxGrid, that is why I've changed the sample to have column widths in pixels.

zdrawku:
This issue aims to point out the returned error with that specific scenario (widths in percentages and enabled pinning feature). Either we should document this limitation as well, or at least handle the error because the scenario is not so uncommon.

The current limitation [states](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column_pinning.html#pinning-limitations): 

> Setting column widths in percentage (%) explicitly makes the Grid body and header content to be misaligned when there are pinned columns.

In this case, I can accept a misalignment, but not the "Maximum call stack size exceeded" error. Let me know if I am missing something here?

zdrawku:
Set the issue back in development, as we discussed with @skrustev we need to fix this issue.
Also, it looks like the Limitation is invalid, and actually the Grid set with percentages and using Column Pinning, is working just fine. Svetlio, you may consider this as well, and validate this behavior.

skrustev:
I've created a PR that should fix mainly the issue present in the Multi Row Layout that throws an error. 

With @rkaraivanov we noticed that while pinning works in the igxGrid without MRL, it does not work correctly when there are pinned columns and the unpinned area has horizontal scrollbar. I can look into that as a separate issue since it is not related to Multi Row Layout.

Because the sample has changed, in order to reproduce this issue the configuration would need to have a pinned block and width of at least 1 columns set to percentages.

PR #4778, opened 2019-05-14
Fix errors thrown when pinning blocks when there are columns with width in percentages. 

skrustev:
#4678

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -2886,21 +2886,6 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
     /**
      * @hidden
      */
-    get headerFixedWidth() {
-        let width = 0;
-        if (this.headerCheckboxContainer) {
-            width += this.headerCheckboxContainer.nativeElement.getBoundingClientRect().width;
-        }
-        if (this.headerDragContainer) {
-            width += this.headerDragContainer.nativeElement.getBoundingClientRect().width;
-        }
-
-        return width;
-    }
-
-    /**
-     * @hidden
-     */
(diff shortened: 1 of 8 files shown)


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -2990,11 +2975,19 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         return this._unpinnedWidth;
     }
 
+    /**
+     * @hidden
+     * Gets the combined width of the columns that are specific to the enabled grid features. They are fixed.
+     */
+    public get featureColumnsWidth() {
+        return this.getFeatureColumnsWidth();
```
kdinev: Why implement a method, which would be called inside a getter method? Also shouldn't some kind of caching be considered?

skrustev: This is because if I want to use in a child component that inherits the IgxGridBaseComponent `super.featureColumnsWidth` this is not supported for getters and in this case the child component only need to override the  'getFeatureColumnsWidth' method and it will be reflected in the getter as well.

Example is for the GroupBy in the IgxGridComponent that on top of the default calculations adds the expander column width. Also this is performed runtime and caching won't be viable here and the calculations are intensive anyway.

kdinev: This is available as a feature with the update to 8, but it's not in the current version.

kdinev: @skrustev Leave a TODO comment for this to be changed with 8.

PR #4800, opened 2019-05-15
refactor: move DOCUMENT imports from platform-browser to common

badre429:
upgrade to angular 8 does not work because DOCUMENT does not imports from platform-browser any more
 
 - [x] Bug fix


kdinev:
@badre429 I'll retarget this PR to the `ng-8-update` branch.

badre429:
this PR is not dedicated for angular 8 in this porject all instance of DOCUMENT are imported from @angular/common except for this file is importanted from @angular/platform-browser; 
then it still work with angular 7

kdinev:
@badre429 In this case, I'll retarget to `7.3.x` as `master` is currently frozen. The change will make it into master with #4783

kdinev:
@damyanpetev Can you verify that this file is even used anywhere because I have my doubts about that?

damyanpetev:
@kdinev Yeah, it is in use (though long overdue for a refactor 😃)

PR #4831, opened 2019-05-17
feat(column): caching calcWidth getter for optimized perf 7.2

### LINKED ISSUE #4575: Optimize parseInt requests from IgxForOf
## Description  
Chrome with IE polyfills include parseInt and the parseInt execution is a substantial part of the execution time during rendering.

Ensuring numeric input for column widths should provide with rendering speed increase.

 * browser: Chrome

ChronosSF:
Closes #4575   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/column.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/column.component.ts
@@ -29,7 +29,8 @@ import {
     IgxNumberFilteringOperand,
     IgxDateFilteringOperand,
     IgxStringFilteringOperand,
-    IgxFilteringOperand } from '../data-operations/filtering-condition';
+    IgxFilteringOperand
+} from '../data-operations/filtering-condition';
 import { IgxGridBaseComponent, IGridDataBindable } from './grid-base.component';
 import { FilteringExpressionsTree } from '../data-operations/filtering-expressions-tree';
 import { IgxGridFilteringCellComponent } from './filtering/grid-filtering-cell.component';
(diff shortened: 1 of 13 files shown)


PavlovVasil:
@ChronosSF 
Using the Column Resizing dev sample:

- I tried setting a width of '900px' to the ContactName column by using the API and the horizontal scrollbar does not get updated. This does not reproduce in 7.2.x. Here is what it looks like: 

resizeCalcWidth

- Resizing the grid also has a problem: 
1. The initial grid width is '900px': change it to '600px' with the API at runtime
2. Scroll horizontally to the middle
3. Resize the ContactName column: its initial width is 200px, toggle it to 900px with the API

Result: All the cell content flickers or gets hidden when a specific horizontal offset gets reached: 

resizeGridWidth

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -286,24 +287,23 @@ export class IgxColumnComponent implements AfterContentInit {
...
-            return this.width;
+        if (this._calcWidth !== null && !isNaN(this.calcPixelWidth)) {
+            return this._calcWidth;
         }
+        this.cacheCalcWidth();
+        return this._calcWidth;
     }
 
+    private _calcWidth = null;
+    public calcPixelWidth: number;
```
MayaKirova: Is there a reason to introduce another public prop (calcPixelWidth), which has the same value as calcWidth but parsed to number? Having both seems a bit redundant since the igxForOf parses the values to number anyway.

ChronosSF: I want it cached, because the parseInt operation which is done on changes when calculating sizes is expensive.

ChronosSF:
@PavlovVasil , please re-test after my changes.

The last issue
- the horizontal scrollbar does not appear/disappear after resizing a column is not caused by this PR but by the pinned/totalWidth caching. You can reproduce it in 7.2.x as well. Please, log it separately.

PR #4846, opened 2019-05-20
Fix shift+tab not scrolled in view with pinning

dkamburov:


projects/igniteui-angular/src/lib/grids/grid/grid-mrl-keyboard-nav.spec.ts
```diff
@@ -2478,6 +2479,55 @@ describe('IgxGrid Multi Row Layout - Keyboard navigation', () => {
             expect(cell.editMode).toBe(true);
         }
     });
+
+    fit('Shift + tab navigation should scroll to the approprate cell', async () => {
```
MayaKirova: Remove fit.

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
```diff
@@ -426,9 +426,15 @@ export class IgxGridMRLNavigationService extends IgxGridNavigationService {
...
+    public getParentUnpinnedIndex(visibleColumnIndex: number) {
+        const column = this.grid.unpinnedColumns.find((col) => col.columnLayout && col.visibleIndex === visibleColumnIndex);
+        return this.grid.pinnedColumns.length ? this.grid.unpinnedColumns.filter((c) => c.columnLayout).indexOf(column) :
+            visibleColumnIndex;
+    }
+
     protected performHorizontalScrollToCell(rowIndex, visibleColumnIndex, isSummary = false) {
         const col = this.grid.columns.find(x => !x.columnGroup && x.visibleIndex === visibleColumnIndex);
-        const unpinnedIndex = this.getColumnUnpinnedIndex(col.parent.visibleIndex);
+        const unpinnedIndex = this.getParentUnpinnedIndex(col.parent.visibleIndex);
```
MayaKirova: Instead of exposing another function that is very similar to the base one, can't we get the index directly from the igxFor, like it is done in performTab/performShiftTabKey functions, for example:

  const hScroll = this.horizontalScroll(rowIndex);
  const scrIndex = hScroll.igxForOf.indexOf(col.parent);
  hScroll.scrollTo(scrIndex);

projects/igniteui-angular/src/lib/grids/grid-mrl-navigation.service.ts
```diff
@@ -426,15 +426,21 @@ export class IgxGridMRLNavigationService extends IgxGridNavigationService {
         return '.igx-grid__mrl-block';
     }
 
+    public getParentUnpinnedIndex(visibleColumnIndex: number) {
```
MayaKirova: Since this function is no longer used it should be removed.

PR #4847, opened 2019-05-20
Fix shift + tab  when navigating though group row in multi row layout.

skrustev:
 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid/grid-mrl-keyboard-nav.spec.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid-mrl-keyboard-nav.spec.ts
@@ -1,4 +1,4 @@
-import { Component, ViewChild, TemplateRef } from '@angular/core';
+﻿import { Component, ViewChild, TemplateRef } from '@angular/core';
 import { async, TestBed, fakeAsync, tick } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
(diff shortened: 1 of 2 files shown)


MayaKirova:
@skrustev Some other scenarios with grouping are now broken, for example:
1. Open Grid MRL Dev sample.
2. Group by column ID
3. Click on first ID cell.
4. Hit up arrow - focus moves to group row correctly.
5. Hit down arrow - focus is lost.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -505,6 +505,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
 
     protected get selectionNode(): ISelectionNode {
         return {
+            cell: this,
```
MayaKirova: The cell instance's data may change when scrolling occurs. We should either  store the mrl specific coordinates of the cell or when navigation starts from group row:
- on ArrowUp - always move focus to the fist cell with the last rowStart in the current columnLayout
- on Arrow Down - always move focus to the fist cell with first rowStart in the current columnLayout

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -859,6 +867,12 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
             return;
         }
 
+        const cellData: LayoutCellData = {
```
MayaKirova: Is this needed? You should be able to get the cell using the information in the selectionNode.

skrustev: I wanted it to be consistent with other cases like in the group by since there you should not get the cell since the cell might not be rendered.

PR #4892, opened 2019-05-27
refactor(grid-mrl): Store starting selected cell and use it for keyboard nav

dkamburov:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -707,7 +707,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
      */
     pointerup = () => {
         if (this.grid.hasColumnLayouts) {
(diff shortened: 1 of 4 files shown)


MayaKirova:
@dkamburov Previous startNavigationCell should probably be cleared on Home/End keys (Ctrl + ArrowLeft/ArrowRight) to be more consistent with Excel like navigation. For example, for the following scenario:

1) Run grid Mrl dev sample.
2) Focus Maria Anders cell. Hit arrow left so that ID cell 'ALFKI' takes focus.
3) Hit End or Ctrl + Arrow Right.

The last cell from rowStart 2 is focused - Fax.

In Excel the previous rowStart/colStart is cleared on any large jump. 
To be consistent in that scenario, the last cell from rowStart 1 should be focused - Phone cell.

dkamburov:
@MayaKirova after discussing it, we've decided to leave as it is - persisting the navigation cell for this scenario. I believe if there is a demand on this acting exactly as Excel for ctrl + left/right we can easily clear the navigation cell.

Issue #4897, opened 2019-05-28
When no condition is provided, filter() method of grid throws undescriptive error


tachojelev:
## Description  
When no condition is provided, **filter()** method of grid throws undescriptive error.
When no condition is provided, the filterGlobal() on the other hand does nothing and does not throw error.

 * igniteui-angular version: 7.2.x, 7.3.x
 * browser: any

## Steps to reproduce  

1. Add the following lines to the **grid-search.sample.html** and the **grid-search.sample.ts** respectively:
```html
<button (click)="test()">test</button>
```
```typescript
test() {        
    this.grid1.filter('CompanyName', 'al');
}
```

2. Run the "Grid Search" dev demo.
3. Click the "test" button.

## Result  
Undescriptive error is thrown.

## Expected result  
Descriptive error is thrown for both the **filter** and **filterGlobal** methods.

## Attachments  

![image](https://user-images.githubusercontent.com/12784974/58472019-fc7c9a00-814d-11e9-9190-80ed6956e042.png)

PR #4900, opened 2019-05-28
 Throw descriptive error when filtering without condition - 7.3.x

tachojelev:
#4897

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -326,6 +326,9 @@ export class GridBaseAPIService <T extends IgxGridBaseComponent & IGridDataBinda
         }
 
         grid.filteringExpressionsTree = filteringTree;
+        if (!condition) {
```
DiyanDimitrov: You could move this check at the beginning of the method. However I think that for filterGlobal, we should make the condition argument required. This method doesn't make sense to be called without condition. Both changes are breaking, but I think that compilation error is better than a runtime error in this case.

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -310,6 +310,10 @@ export class GridBaseAPIService <T extends IgxGridBaseComponent & IGridDataBinda
     }
 
     public filter_global(term, condition, ignoreCase) {
+        if (!condition) {
+            throw Error('Invalid condition!');
```
bkulov: You should create an instance of the error here - ` throw new Error`.

projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts
```diff
@@ -150,7 +150,9 @@ export class IgxFilteringService implements OnDestroy {
             this.gridAPI.filter(field, value, conditionOrExpressionTree, filteringIgnoreCase);
         } else {
             const expressionsTreeForColumn = this.grid.filteringExpressionsTree.find(field);
-            if (expressionsTreeForColumn instanceof FilteringExpressionsTree) {
+            if (!expressionsTreeForColumn) {
+                throw Error('Invalid condition or Expression Tree!');
```
bkulov: Same as above - ` throw new Error`.

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -310,6 +310,10 @@ export class GridBaseAPIService <T extends IgxGridBaseComponent & IGridDataBinda
     }
 
     public filter_global(term, condition, ignoreCase) {
+        if (!condition) {
```
kdinev: It doesn't seem to me that this is the best solution. A few lined down the same method, the condition is checked for being defined, and if not, then no new filtering expression is added. I think the issue comes with the filtering operand reset which happens before the check is performed. If you move it inside the check, then there should be no error.

PR #4916, opened 2019-05-29
Implement required and  state validation 

### LINKED ISSUE #4353: igx-select does not honor [required]
Using the [required] attribute on an igx-select doesn't work.  It's very common to have a form item dynamically swap between required or not based on other selections.

The "actual" required status is of course set via my FormGroup, but the HTML element has to be able to be updated so that the asterisk appears and hides properly.

IvayloG:
Closes #4353

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -312,9 +377,51 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
+        if (this.ngControl) {
+            this._statusChanges$ = this.ngControl.statusChanges.subscribe(this.onStatusChanged.bind(this));
+        }
+    }
+
+    /**
+     * @hidden @internal
+     */
+    public ngOnDestroy() {
+        this.destroy$.complete();
```
ViktorSlavov: `this.destroy$` isn't needed - it's only initialized with the component and then completed in `onDestroy`

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -312,9 +377,51 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
+            this.valid = this.ngControl.valid ? IgxSelectState.VALID : IgxSelectState.INVALID;
+        }
+    }
+    /**
+     * @hidden @internal
+     */
+    public ngOnInit() {
+        try {
+            this.ngControl = this._injector.get(NgControl);
+        } catch (e) { }
```
ViktorSlavov: Is wrapping this in a `try - catch` necessary? I think `this._injector` will always be provided in an Angular app

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -312,9 +377,51 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
 
     /** @hidden @internal */
     public onBlur(): void {
+        if (this.ngControl && !this.ngControl.valid) {
```
ViktorSlavov: Could `onBlur` call `this.onStatusChanged()`? The logic seems to be the same

IvayloG: Actually I prefer to separate the initial+update validation logic from the one updating the states onBlur

projects/igniteui-angular/src/lib/select/select.component.ts
Lipata: The try - catch is not because of the injector, but because of the exception, thrown from the [get()](https://angular.io/api/core/Injector#get) method, when ngControl cannot be found.

projects/igniteui-angular/src/lib/select/select.component.ts
ViktorSlavov: Makes sense, I missed the `STATE.INITIAL` part - it'll behave like our components

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -57,12 +75,28 @@ const noop = () => { };
...
+export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelectBase, ControlValueAccessor,
+    AfterContentInit, OnInit, AfterViewInit, OnDestroy, EditorProvider {
 
     /** @hidden @internal do not use the drop-down container class */
     public cssClass = false;
 
+    /**
+     * @hidden @internal
+     */
+    @HostBinding('class.igx-input-group--valid')
```
damyanpetev: Not super sure the input group classes should be bound to the select. While this currently works, because the rule is not restrictive for the structure, it may very well be in the future or the class name can change.
Plus, it doesn't make sense semantically to have it on the host.

As for alternatives - you _could_ move this to the group or better yet - I see a `valid` setter on the input.

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -82,8 +116,10 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
     /** @hidden @internal */
     public height: string;
 
+    /** @hidden @internal */
+    private ngControl: NgControl = null;
```
damyanpetev: No need for hidden/internal on private, also move them before public props.

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -119,6 +122,28 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
      *
      */
     @Input() public placeholder;
+
+    private _required = false;
+    /**
+     * Sets/gets whether item selection is required.
+     * If not set, `required` will have value `false`.
+     * ```html
+     * <igx-select [required] = "true"></igx-select>
```
damyanpetev:
```suggestion
     * <igx-select [required]="true"></igx-select>
```

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -119,6 +122,28 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
+     * ```
+     * ```typescript
+     * let isRequired =  this.select.required;
```
damyanpetev:
```suggestion
     * let isRequired = this.select.required;
```

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -119,6 +122,28 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
+     */
```
damyanpetev: Missing closing backticks `` ``` ``

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -119,6 +122,28 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
      *
      */
     @Input() public placeholder;
+
+    private _required = false;
+    /**
+     * Sets/gets whether item selection is required.
+     * If not set, `required` will have value `false`.
```
damyanpetev: So just "Default is `false`"?

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -119,6 +122,28 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
+     * <igx-select [required] = "true"></igx-select>
+     * ```
+     * ```typescript
+     * let isRequired =  this.select.required;
+     */
+    public get required() {
+        return this._required;
+    }
+    @Input()
+    public set required(value: boolean) {
```
damyanpetev: Actually.. Do we even need to expose `[required]` on the select?
It is technically already handled by the [`RequiredValidator`](https://angular.io/api/forms/RequiredValidator) directive from @angular/forms.

Might be easier to not have the Input, than explaining the validation requires an ngControl anyway. Or it might be less discoverable that way?

// CC @rkaraivanov Thoughts?

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -312,9 +338,53 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
+    }
+
+    /**
+     * @hidden @internal
+     */
+    public ngAfterViewInit() {
+        if (this.ngControl) {
+            this._statusChanges$ = this.ngControl.statusChanges.subscribe(this.onStatusChanged.bind(this));
+        }
+        this.inputGroup.isRequired = this.required;
```
damyanpetev: Add a comment that this is set after view init on purpose, otheriwse the group itself will override it.

projects/igniteui-angular/src/lib/select/select.component.ts
rkaraivanov: @damyanpetev 

> Actually.. Do we even need to expose [required] on the select?
We don't

damyanpetev: This wouldn't work with reactive forms anyway. And here's a [sample showing the `ngModel` validator directive ](https://stackblitz.com/edit/angular-newbha?file=src%2Fapp%2Fselect-sample-3%2Fselect-sample-3.component.html) work just fine without this.
@IvayloG You can just drop the property.
Here's the related bit from the input checking the validators:
https://github.com/IgniteUI/igniteui-angular/blob/790dd199c9ccefdb323b8275ca42616d18844425/projects/igniteui-angular/src/lib/directives/input/input.directive.ts#L197-L201

The only thing that bugs me is detecting the change of the validator itself (i.e. runtime required change for the (*) indicator). Probably some validatorsChange subscription somewhere.

projects/igniteui-angular/src/lib/select/select.component.spec.ts
```diff
@@ -446,34 +447,42 @@ describe('igxSelect', () => {
...
             const fix = TestBed.createComponent(IgxSelectReactiveFormComponent);
+            let inputGroupWithRequiredAsterisk = fix.debugElement.query(By.css('.' + CSS_CLASS_INPUT_GROUP_REQUIRED));
             fix.detectChanges();
             const selectComp = fix.componentInstance.select;
             const selectFormReference = fix.componentInstance.reactiveForm.controls.optionsSelect;
             expect(selectFormReference).toBeDefined();
             expect(selectComp).toBeDefined();
             expect(selectComp.selectedItem).toBeUndefined();
             expect(selectComp.value).toEqual('');
+            expect(inputGroupWithRequiredAsterisk).toBeDefined();
```
damyanpetev: That check won't refresh for every expect, since the variable is assigned once and thus the query is ran once.

damyanpetev: Should probably add a test where the state is not constant (required changes)

PR #4924, opened 2019-05-30
feat(print-styles): improve Printing Experience

### LINKED ISSUE #1995: Feature-request: IgxGrid improve Printing Experience
### Description  
The grid does not seem to work very well for printing. The grid overflows and content is hidden. I'm not sure whether this is a bug, or there just needs to be some clarification in the documentation on how to make it work with printing.

 * igniteui-angular version: 6.0.1 (but issue seems to also be present on current stackblitz sample)
 * browser: Chrome

### Steps to reproduce  

1. Open https://lrqakgovrdo.angular.stackblitz.io/
2. Press Ctrl + p

### Result  
The grid overflows the page

### Expected result  
Grid should fit on page for printing, at least when it is defined in % and not px.

image

desig9stein:
Closes #1995 

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/themes/_core.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/themes/_core.scss
@@ -56,6 +56,7 @@
 @import '../components/toast/toast-component';
 @import '../components/tooltip/tooltip-component';
 @import '../components/time-picker/time-picker-component';
+@import '../print/index';
 
 /// Includes the base for each theme.
 
@@ -70,4 +71,5 @@
 
 @mixin igx-core() {
     @include igx-typography();
+    @include igx-print-layout;
 }
--- a/projects/igniteui-angular/src/lib/core/styles/themes/_core.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/themes/_core.scss
@@ -69,7 +69,10 @@
     }
 }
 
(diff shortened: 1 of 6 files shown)


projects/igniteui-angular/src/lib/core/styles/themes/_core.scss
```diff
@@ -70,4 +71,5 @@
 
 @mixin igx-core() {
     @include igx-typography();
+    @include igx-print-layout;
```
kdinev: This should be a function call. Also should it always get included?

simeonoff: @kdinev This is not a function, it's a mixin. 
@desig9stein However, the linter will throw because of this rule "scss/at-mixin-argumentless-call-parentheses": "always". So include it with parentheses.

desig9stein: Since this is a mixin and we don't expect any params it's not necessary to be invoked like a function, however, we have a Stylelint rule that forces us to add one, so I will add it :). I spoke with @simeonoff and we decide to make it optional but it will be ON by default.

PR #4974, opened 2019-06-04
Ensure there is no initial placeholder value

### LINKED ISSUE #4236: Select's label is positioned incorrectly
## Description  
The label of the IgxSelect is positioned in the upper right corner of the component.

 * igniteui-angular version:  7.2.0-rc4
 * browser: all

## Steps to reproduce  

1. Open the [StackBlitz](https://stackblitz.com/angular/earalgkryaav) select sample
2. Observe select's label

## Result  
The label is in the upper right corner of the component.

## Expected result  
Label should be placed in the input in the same way as placeholder is shown

IvayloG:
Closes #4236

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/select/select.component.ts
+++ b/projects/igniteui-angular/src/lib/select/select.component.ts
@@ -118,7 +118,7 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
      * An @Input property that sets input placeholder.
      *
      */
-    @Input() public placeholder = '';
+    @Input() public placeholder;
     /**
      * An @Input property that disables the `IgxSelectComponent`.
      * ```html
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -118,7 +118,7 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
      * An @Input property that sets input placeholder.
      *
      */
-    @Input() public placeholder = '';
+    @Input() public placeholder;
```
kdinev: Why don't you just host bind the attribute placeholder?

IvayloG: For the placeholder to work as expected, I need to pass it to the actual input element. So I am using an igxSelect   **@input** named as well "**placeholder**" to pass the user defined one to the **input's** placeholder attribute. 

Using the  **attr.placeholder** is needed, because otherwise the **@input** is evaluated as string and thus from the component's perspective there is always a placeholder (and the corresponding styles applied) even if it is an **empty string**, **null** or **undefined**.

PR #4977, opened 2019-06-04
Always reinit pinned columns state - 7.2.x #4905

### LINKED ISSUE #4905: Pinned area is not updated when columns collection is changed
## Description  
When a column, that is pinned, is removed from the columns collection it is not removed from the pinned columns area.

 * igniteui-angular version: 7.2.x, 7.3.x
 * browser: N/A

## Steps to reproduce  

1. Run [this ](https://stackblitz.com/edit/angular-xfci94)sample.
2. Click "Remove column" button in order to remove "Fax" column,
3. Observe pinned columns.

## Result  
"Fax" column is not removed from the pinned area
## Expected result  
"Fax" column to be removed from the pinned area

VMihalkov:
Closes #4905   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -4353,9 +4353,8 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
      * @hidden
      */
     protected reinitPinStates() {
-        if (this.hasColumnGroups) {
-            this._pinnedColumns = this.columnList.filter((c) => c.pinned);
-        }
+        this._pinnedColumns = this.columnList.filter((c) => c.pinned);
+        
         this._unpinnedColumns = this.columnList.filter((c) => !c.pinned);
     }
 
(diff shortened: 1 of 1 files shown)


PavlovVasil:
@rkaraivanov @ChronosSF  The sample from the linked issue uses Array.splice to mutate the original columns array. However, if the grid has MCH, splicing the columns actually breaks the layout. Add this to the Grid Column Groups template: 
```
 <span igxButton (click)="removeCol()">Remove "CompanyName"</span>
```
And this to the component: 
```
removeCol() {
    const element = this.grid.columns.splice(2, 1);
    console.debug('REMOVED', element[0].field);
}
```
Click the "Remove Company Name" button, then the "Pin first group" and "Hide first group" buttons. The result looks like this: 

Sketch

Is it supported for the user to be able to mutate the columns like this by splicing the array at runtime? The issue is not caused by this PR and reproduces is 7.2.x as well. Should I log it separately?

MayaKirova:
@PavlovVasil Actually the original sample does not splice the grid's columns array but the collection which is used to generate them in the template with ngFor. So the scenario you're describing is not the same as in the related issue. 
The grid has handling for scenarios when the content children (in this case the columns) are changed which in turn triggers the re-initialization of the pin states. However it does not have any special handling if someone directly changes the grids columns array. This would not actually remove a column in a normal grid nor in one with multi-column headers.

PavlovVasil:
I will log this as a separate issue, as it is not caused by the PR per se.

PR #5018, opened 2019-06-07
Fix calculating of the row height in 7.2.x

### LINKED ISSUE #5012: There is an empty space in the treeGrid when hide a column with summaries
## Description  
There is an empty space in the treeGrid when hide a column with summaries

 * browser: n/a 

## Steps to reproduce  

1. Open dev sample treeGridFlatData
2. Click button rootAndChildLevels
3. Hide column Salary
4. Scroll to button

Note: The issue appears only in 7.2.x, in the 7.3.x this is fixed with the PR #4730

## Result  
There is an empty space in the treeGrid

## Expected result  
There should be no empty space

## Attachments  

nrobakova:
Closes #5012  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
@@ -1195,8 +1195,8 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
     }
 
     private _getItemSize(item, dimension: string): number {
-        const hasDimension = (item[dimension] !== null && item[dimension] !== undefined);
-        return hasDimension ? parseInt(item[dimension], 10) : this.igxForItemSize;
+        const dim = item[dimension];
+        return typeof dim === 'number' ? dim : this.igxForItemSize;
     }
 }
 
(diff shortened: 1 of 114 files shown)


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -1297,6 +1297,8 @@ export class IgxGridForOfDirective<T> extends IgxForOfDirective<T> implements On
             size = parseInt(this.igxForItemSize, 10) || 0;
             if (item && item.summaries) {
                 size = item.max;
+            } else if (item && item.groups && item.height) {
+                size = item.height;
```
ChronosSF: this requires that the height is assigned to the ```IGroupByRecord``` by the ```groupDataRecursive<T>``` method, which is not the case in 7.2.x. You should get that code as well from the 7.3.x PR

nrobakova: @ChronosSF I updated the groupDataRecursive<T> to get the height as in the other PR

skrustev: In general for this related issue only the fix on line 670 would be enough.
The rest is for the group rows sizing when the regular rows have different height and so far it only happens for Multi Row Layout and it is not related to the IgxTreeGrid. Unless you want to bring the whole fix to 7.2.x as well.

ddincheva:
the issue is resolved, but it is necessary tests to be updated

PR #5035, opened 2019-06-10
style(drop-down): Adding classes for different density types

### LINKED ISSUE #2960: Extend DropDown and Combo to support DisplayDensity
Extend DropDown and Combo to support DisplayDensity similar to the input group and the grid.

desig9stein:
Closes #2960

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -452,7 +455,7 @@ export class IgxDropDownComponent extends IgxDropDownBase implements IDropDownBa
 
     ngAfterViewInit() {
         if (this.virtDir) {
-            this.virtDir.igxForItemSize = 32;
+            this.virtDir.igxForItemSize = 28;
```
damyanpetev: You can move the ItemHeight enum from the combo to the item base itself and use it here as well.
On a second thought, this item size set probably doesn't do much at all (since it won't trigger change detection for the igxFor directive). Basically the size still comes from the app configuration right now.
I'll dig a bit more into it and might log a separate issue.

Lipata:
This will not be part of 8.0.0, so please select a new section in the change log for 8.0.1 and add it there. Thank you!

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
ViktorSlavov: I've removed the DisplayDensity part from the drop-down virtualization sample. We'll need a way to cause re-calculation of the virutal directive's chunkSize - I've logged [an issue](https://github.com/IgniteUI/igniteui-angular/issues/5128).
I think it would be better to fix that in a separate PR - with the density feature, we do not impact any existing functionality.

PR #5149, opened 2019-06-21
Resize column to a smaller width than defaultMinWidth

### LINKED ISSUE #4608: Narrower column width than 88px
Current limitation of 88px as the minimum width is too wide. There are situations that I only want to allocate column width just enough to render a single character.

This feature is helpful when I have a column which renders an abbreviation of a word. For example, the column represents score of record item. In the screenshot, ‘高’ means ‘High’, ‘中’ means ‘Middle’, and ‘低’ means ‘Low’. I don't want to consume 88px width only for a single character. Rendering abbreviated word to save width on grid is a common use case in Japanese.

image
http://jsfiddle.net/rohs312n/

ddincheva:
Closes #4608

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


zdrawku:
@ddincheva as we discussed, you should make additional PR for the Changelog and update the topic in docfx repository related to the minWidth changes.

projects/igniteui-angular/src/lib/grids/grid-column-resizing.service.ts
```diff
@@ -54,11 +54,8 @@ export class IgxColumnResizingService {
      */
     get restrictResizeMin(): number {
         const columnLayoutMultiplier = this.column.grid.hasColumnLayouts ? this.column.gridColumnSpan : 1;
-        const actualMinWidth = parseFloat(this.column.minWidth);
-        const defaultMinWidth = parseFloat(this.column.defaultMinWidth) * columnLayoutMultiplier;
-
-        let minWidth = Number.isNaN(actualMinWidth) || actualMinWidth < defaultMinWidth ? defaultMinWidth : actualMinWidth;
-        minWidth = minWidth < parseFloat(this.column.width) ? minWidth : parseFloat(this.column.width);
+        const actualMinWidth = parseFloat(this.column.minWidth) * columnLayoutMultiplier;
```
skrustev: The columnLayoutMultiplier should be omitted in my opinion when used directly on user set minWidth because it is multiplied by the span of the column. For example if column has span 2 and user sets minWidth to 200px, the min width would be actually 400px.
It was used only on the default value since it is per single span and when a column spans 2 columns it should be multiplied by 2 but when users sets its value it should take exactly his value.

@dkamburov @StefanIvanov What do you think?

StefanIvanov: I'd rather delegate the comment on this to @simeonoff

StefanIvanov: I'd rather have @simeonoff share his opinion

zdrawku: After further discussion with @simeonoff, @rkaraivanov and @skrustev we've decided to remove the `columnLayoutMultiplier` in general. The spec will be updated accordingly. Remove ` a columns has a span 2 for example, its default min width is 2 * '80px'. Similarly for span 3 it is 3 * '80px`

PR #5164, opened 2019-06-24
fix(igxForOf): Recalculate chunk size when changing itemSize #5128

### LINKED ISSUE #5128: [igxForOf] Automatically recalc chunkSize on [itemSize] change or provide method to do so
When changing the `[itemSize]` input of the `igxForOf` directive, the chunk size is not automatically recalculated (unless the `[igxForOf]` collection changes as well)
This is illustrated in the [following sample](https://stackblitz.com/edit/virtual-directive-height-change).
Change the item height and observe the number of items rendered: 
 - When changing the item size from `Medium` to `Small`, the list height does not correspond to the scroll and at 2/3 scroll, the items disappear.
 - When setting the size to `Large`, less than the needed amount of items is needed and you cannot reach the last item by scrolling.

Changing the `[itemSize]` input should force a call to `_applyChanges` (`applyChunkSizeChange()`, resp.).
As this will impact performance and will likely be discarded.
Exposing a way to call `_applyChanges` manually would do the job.

You can cast the instance of `IgxForOfDirective` to `any` and call the `_applyChanges` method and it renders properly.
Here's the [very hacky example](https://stackblitz.com/edit/virtual-directive-height-change-hack).

## Additional context
Screenshot of the list with mismatching height:
image

dkamburov:
Closes #5128

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
@@ -101,18 +101,7 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
      * ```
      */
     @Input()
-    get igxForItemSize(): any {
-        return this._igxForItemSize;
-    }
-
-    set igxForItemSize(value) {
-        this._igxForItemSize = value;
-        if (this.dc) {
-            this.initSizesCache(this.igxForOf);
-            this._applyChanges();
-            this.onDataChanged.emit();
-        }
-    }
+    public igxForItemSize: any;
 
     /**
      * @hidden
@@ -196,7 +185,6 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
     protected heightCache = [];
     private _adjustToIndex;
     private MAX_PERF_SCROLL_DIFF = 4;
-    private _igxForItemSize;
 
     protected get sizesCache(): number[] {
         return this._sizesCache;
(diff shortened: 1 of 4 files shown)


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -101,7 +101,18 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
      * ```
      */
     @Input()
-    public igxForItemSize: any;
+    get igxForItemSize(): any {
```
MayaKirova: Input prop changes are handled in ngOnChanges. initSizesCache is already called there in case igxForItemSize changes. You can move the additional code there.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -101,7 +101,18 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
+        return this._igxForItemSize;
+    }
+
+    set igxForItemSize(value) {
+        this._igxForItemSize = value;
+        if (this.dc) {
+            this.initSizesCache(this.igxForOf);
+            this._applyChanges();
+            this.onDataChanged.emit();
```
MayaKirova: Is there a reason onDataChanged needs to be emitted here? It should be emitted only when the actual data changes.

dkamburov: Yes, I'll remove it.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
dkamburov: Actually I did it this way initially, but this will require having input binding for the directive to change the property. For example [this](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts#L458) won't trigger the change detection and won't change the item size.

dkamburov: Ok, ignore my previous comment, I discussed it with @ViktorSlavov  and @damyanpetev and I'm reverting it back to ngOnChanges

PR #5174, opened 2019-06-25
fix(forof): preventing reflows on scroll - 7.2

### LINKED ISSUE #5078: Multiple reflow per rendered row
### Description  
When the grid renders a row it causes reflow multiple times which impacts initial rendering performance. The grid queries the `igx-horizontal-virtual-helper` element's `scrollLeft` twice for each row rendered; which causes reflow.

Additionally, `style.left` is set instead of using GPU accelerated `translate: transformX(...)` which likely also worsens performance.

 * igniteui-angular version: 7.3.4
 * browser: Chrome 74

### Steps to reproduce  

1. Create a new Angular app with `igniteui-angular@7.3.4` as a dependency.
2. Add a basic grid
3. Use Chrome to performance profile the initial grid render.
4. Notice the purple blocks on Chrome Dev Tools Performance Tabs's flame chart.

### Result  
Flame Chart:
NonCachedHorizontalScrollFlame

Source View:
NonCachedHorizontalScroll

### Expected result  
No reflow per row rendered. The horizontal scroll position should be cached.

ChronosSF:
Closes #5078

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/for-of/for_of.sync.service.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/for_of.sync.service.ts
@@ -45,11 +45,4 @@ export class IgxForOfSyncService {
     public chunkSize(dir: string): number {
         return this._master.get(dir).state.chunkSize;
     }
-
-    /**
-     * @hidden
-     */
-    public scrollPosition(dir: string): number {
-        return this._master.get(dir).scrollPosition;
-    }
 }
--- a/projects/igniteui-angular/src/lib/directives/for-of/for_of.sync.service.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/for_of.sync.service.ts
@@ -45,4 +45,11 @@ export class IgxForOfSyncService {
     public chunkSize(dir: string): number {
         return this._master.get(dir).state.chunkSize;
     }
+
+    /**
+     * @hidden
+     */
+    public scrollPosition(dir: string): number {
+        return this._master.get(dir).scrollPosition;
+    }
 }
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/directives/for-of/for_of.sync.service.ts
```diff
@@ -45,4 +45,11 @@ export class IgxForOfSyncService {
     public chunkSize(dir: string): number {
         return this._master.get(dir).state.chunkSize;
     }
+
+    /**
+     * @hidden
+     */
+    public scrollPosition(dir: string): number {
+        return this._master.get(dir).scrollPosition;
```
MayaKirova: @ChronosSF 
Storing the scroll position in the master directive may be a problem for scenarios where the rows are cached, since it’s possible for the master directive that caches the position to be detached while the scrolling occurs, hence it will not detect any changes until it is re-added in the view.
As a result data cells may become de-synched from the scrollbar hence the headers and content will be misaligned.

Here’s an example scenario where can occurs:
1.	Run the gridGroupBy sample.
For the rows the first one is the one holding the “master” igxForOf that stores the position.
2.	Group By a column.
First row becomes group row, hence the data row is now cached (is in detached state) hence the “master” igxForOf directive is also detached.
3.	Scroll right.
4.	Ungroup
The data row is back from the cache, so is the igxForOf “master”. However it has retained its old scroll position and has forced all other rows to match it.

Result is that the header/footer cells and the data row cells are misaligned now, since data row cells are not synched with the scrollbar.

PR #5185, opened 2019-06-25
Fix - IgxCombo - Change combo default height to 40, itemsMaxHeight scales w/ item Height - 8.0.x

### LINKED ISSUE #5184: [igx-combo] Combo has incorrect default itemHeight and itemMaxHeight
## Description  
Since `displayDensity` is now defined for combo items, the default item height should be brought down from `48` (current default) to `40` (comfortable). The `itemsMaxHeight` default value should also be brought down from `480` (current) to `400` (10 * newHeight).

Furthermore, the combo's `itemsMaxHeight` should always default to 10 * itemHeight.  

 * igniteui-angular version: 8.0.x, 7.3.x
 * browser: N/A

## Steps to reproduce  
1. `<igx-combo [displayDensity]="'comfortable'">` w/o specifying anything in `[itemHeight]` input
2. Observe the items

## Result  
Combo items are larger than `40px` even when no `itemHeight` input is specified

## Expected result  
Combo items are `40px`, the list contains `10` items.

ViktorSlavov:
Closes #5184 

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/combo/combo.component.ts
+++ b/projects/igniteui-angular/src/lib/combo/combo.component.ts
@@ -72,7 +72,7 @@ enum DataTypes {
  * @hidden
  */
 const ItemHeights = {
-    'comfortable': 48,
+    'comfortable': 40,
     'cosy': 32,
     'compact': 28,
 };
@@ -140,6 +140,7 @@ export class IgxComboComponent extends DisplayDensityBase implements IgxComboBas
     private _data = [];
     private _filteredData = [];
     private _itemHeight = null;
+    private _itemsMaxHeight = null;
     private _positionCallback: () => void;
     private _onChangeCallback: (_: any) => void = noop;
     private overlaySettings: OverlaySettings = {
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -578,7 +579,16 @@ export class IgxComboComponent extends DisplayDensityBase implements IgxComboBas
      * ```
     */
     @Input()
-    public itemsMaxHeight = 480;
+    public get itemsMaxHeight(): number {
+        if (this._itemsMaxHeight === null || this._itemsMaxHeight === undefined) {
+            return this.itemHeight * 10;
```
wnvko: I would change this to:
this._itemsMaxHeight = this.itemHeight * 10;
this way we have single return.
10 could be const too.

ViktorSlavov: I agree that 10 should be a const - thanks for pointing it out!
We cannot do `this._itemsMaxHeight = this.itemHeight * 10`, as it would break the functionality - the next time we change item height, `this._itemsMaxHeight !== null` and we will not enter the `if` statement - the value won't change and the list will not scale to contain 10 items.

PR #5228, opened 2019-06-27
fix(igxHierarchicalGrid): Hierarchical Grid should update its childre…

### LINKED ISSUE #5102: Hierarchical Grid does not update its children when its [data] input changes
## Description  
When the igxHierarchicalGrid's [data] input changes, its children do not get updated. 

 * igniteui-angular version: 7.2.x, 7.3.x, 8.0.x
 * browser: n/a

## Steps to reproduce  

1. Run the attached sample
2. Expand the first row
3. Click the second button below the grid - it would replace the dataset with another one.
4. Expand the first row again

## Result  
The parent grid gets updated, but the child grid still shows the old data.

## Expected result  
The child grid should update when the dataset gets changed. 

## Attachments  

[TestApp.zip](https://github.com/IgniteUI/igniteui-angular/files/3301708/TestApp.zip)

MayaKirova:
…n when its [data] input changes.

Closes #5102   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/hierarchical-grid/child-grid-row.component.html
```diff
@@ -1,3 +1,3 @@
 <div class="igx-grid__hierarchical-indent" [ngClass]="{'igx-grid__hierarchical-indent--scroll': parentHasScroll}">
-    <igx-hierarchical-grid #hgrid></igx-hierarchical-grid>
+    <igx-hierarchical-grid #hgrid [data]='rowData.childGridsData[layout.key]'></igx-hierarchical-grid>
```
skrustev: When the row is expanded and the parent row reference to object is not updated and only the child data is changed this will not trigger change detection. It changes only when the row is collapsed/expanded again.
I am not sure how this should be detected though.

MayaKirova: @skrustev Modifying values in the data array will not trigger the hierarchical grid pipes that build the hierarchy, so the change will not be reflected. This is consistent with other features that rely on pipes to process the data. For example, in a normal grid with grouping, changing a value in the array will not in itself re-evaluate the groups. Some other action that triggers the grouping pipe needs to take place for the change to be applied to the groups (expand/collapse, sort etc.).
The behavior is consistent with the normal grid so I don't think we should cover such scenarios.

skrustev: I'm ok with keeping it as it is right now if there are no objections.

MayaKirova:
When testing please test with a Hierarchical Grid with Load On Demand and virtualization.

PR #5308, opened 2019-07-04
Focusing on parent grid using keyboard navigation in IE.

### LINKED ISSUE #4488: In IE parent grid scrolls when using ctr + right navigation.

MayaKirova:
Closes #4488
Internet explorer doesn't support [preventScroll](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) option. Try this sample [here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#Focus_prevent_scroll) (Click on the "Click me to focus on the button without scrolling!" in IE).
This is why we should not focus on a child grid but rather on the parent grid in some cases when using keyboard navigation in IE.
Please test the solution for IE with ctr + right, ctr + left, home, end, arrowLeft and arrowRight navigation keys.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
@@ -159,7 +159,6 @@ export class IgxGridNavigationService {
         if (!element.previousElementSibling && this.grid.pinnedColumns.length && index === - 1) {
             element.parentNode.previousElementSibling.focus({ preventScroll: true });
         } else if (!this.isColumnLeftEdgeVisible(visibleColumnIndex - 1)) {
-        } else if (!this.isColumnLeftFullyVisible(visibleColumnIndex - 1)) {
             this.getFocusableGrid().nativeElement.focus({ preventScroll: true });
             this.performHorizontalScrollToCell(rowIndex, visibleColumnIndex - 1, isSummary);
         } else {
--- a/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
@@ -131,7 +131,7 @@ export class IgxGridNavigationService {
                 if (this.isColumnLeftEdgeVisible(visibleColumnIndex + 1)) {
                     element.nextElementSibling.firstElementChild.focus({ preventScroll: true });
                 } else {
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -143,6 +143,7 @@ export class IgxGridNavigationService {
                 element.nextElementSibling.focus({ preventScroll: true });
             }
         } else {
+            this.getFocusableGrid().nativeElement.focus({ preventScroll: true });
```
skrustev: I don't think this is needed since in `performHorizontalScrollToCell()` method the grid is focused there as well.

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -158,6 +159,7 @@ export class IgxGridNavigationService {
         if (!element.previousElementSibling && this.grid.pinnedColumns.length && index === - 1) {
             element.parentNode.previousElementSibling.focus({ preventScroll: true });
         } else if (!this.isColumnLeftEdgeVisible(visibleColumnIndex - 1)) {
+            this.getFocusableGrid().nativeElement.focus({ preventScroll: true });
```
skrustev: Again I think this is not needed.

VMihalkov:
Ctrl + Down seems to be fixed in the first child grid, but trying it in the second child grid causes the following behavior: the last cell gets selected and scrolled into view, then it disappears for a moment, and shows again. Here are the steps to reproduce:

1. Open hierarchical grid dev sample on IE.
2. Select the first cell of the second child grid.
3. Use ctr + down for navigation.
CtrlDownInSecondChildGrid

PR #5358, opened 2019-07-08
Calculate column editable depending on rowEditable

### LINKED ISSUE #5022: When change input rowEditable from false to true, the non primary key columns are not editable
## Description  
When change input rowEditable from false to true, the non primary key columns are not editable

 * browser: n/a

## Steps to reproduce  

1. Open the [sample](https://stackblitz.com/edit/angular-fmql4v-v1varw?file=src%2Fapp%2Fgrid%2Fgrid-filtering-sample%2Fgrid-filtering-sample.component.ts)
2. Click on button "Enable Row Editng"
3. Double click on a cell

## Result  
Cell in not opened in edit mode

## Expected result  
When set rowEditable the columns property editable should be set to true

## Attachments  

wnvko:
Default value of `column.editable` was null. This was not correct. Now default value of inner field is `undefined`. The getter now checks if `editable` is set. If so returns what is set by the developer. If not returns what is set for `rowEditable` (if the column is primary column returns false in this case).

Closes #5022 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


damyanpetev:
I assume you got the tests in first so they can fail before the fix? Make the PR as draft for that with in development status 😉

projects/igniteui-angular/src/lib/grids/column.component.ts
```diff
@@ -107,18 +107,31 @@ export class IgxColumnComponent implements AfterContentInit {
...
+     * Gets whether the column is editable.
      * Default value is `false`.
      * ```typescript
      * let isEditable = this.column.editable;
      * ```
+     * @memberof IgxColumnComponent
+     */
+    @Input()
+    get editable(): boolean {
+        const editableInRowEdit = this.grid && this.grid.rowEditable && this.field !== this.grid.primaryKey;
```
kdinev: Why calculate the value of `editableInRowEdit` at all, if `this._editable` is defined?

wnvko: fixed.

PR #5367, opened 2019-07-09
fix(combo): enable combo checkboxes

### LINKED ISSUE #5366: All checkboxes in combo component have a disabled class applied to them
## Description  
In order to remove the focus form the checkbox, all checkboxes in combo component have a disabled class applied to them

 * igniteui-angular version: All
 * browser: All

## Steps to reproduce  
Inspect the checkbox component inside the combo component.

## Result  
All checkboxes have "igx-checkbox--disabled" class

## Expected result  
Checkboxes should not have "igx-checkbox--disabled" class

## Attachments  

ViktorSlavov:
Closes #5366

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/combo/combo-item.component.ts
```diff
@@ -99,6 +99,15 @@ export class IgxComboItemComponent extends IgxDropDownItemComponent implements D
         this.comboAPI.set_selected_item(this.itemID, event);
     }
 
+    /**
+     * @hidden
+     * @internal
+     */
+    handleFocus(event: MouseEvent) {
+        event.preventDefault();
+        this.dropDown.element.focus();
```
damyanpetev: Fairly certain you don't need this as the combo drop down is not focusable anyway. The prevent above does the heavy lifting

projects/igniteui-angular/src/lib/combo/combo-item.component.ts
```diff
@@ -99,6 +99,15 @@ export class IgxComboItemComponent extends IgxDropDownItemComponent implements D
         this.comboAPI.set_selected_item(this.itemID, event);
     }
 
+    /**
+     * @hidden
+     * @internal
+     */
+    handleFocus(event: MouseEvent) {
```
damyanpetev: Might need a better name and explanation comment for this - prevent/disable check or default or something.

_Details:_
_There's no focus (yet) on the initial click because of the way the checkbox is built - the visible portion is actually the internal input's label and the click event here originates from it._
_What you are preventing here is the default behavior of the HTML label, which would be to perform the default action on the target input - there's where the igx-checkbox handlers are._

_In this case, trigger the default action is click on the input and attempt to focus._
_First bit being worse actually, as it happens on the next cycle (after the combo item has already assigned state) and it toggles, messing up the checkbox state._

PR #5379, opened 2019-07-10
feat(IgxForOf): Adds local variables even, odd, first, last #5310

### LINKED ISSUE #5310: [igxForOf] Expose `even` and `odd` local variables in `igxForOf` template
I want to be able to use `igxForOf` as a drop-in replacement for `ngFor` when I want my content to be virtualized.
Currently, `ngFor` exposes some additional variables compared to `igxForOf`

Expose the following boolean properties in the `igxForOf` context:
 - `first`: True when the item is the first item in the iterable.
 - `last`: True when the item is the last item in the iterable.
 - `even`: True when the item has an even index in the iterable.
 - `odd`: True when the item has an odd index in the iterable.

I can just check the boolean value for each of the variables - they all depend on `index` and/or `igxForOf.length` and both are currently exposed.

## Additional context
This is primarily a nice-to-have enhancement. It would just make it easier for people who are using `even`/`odd`/`first`/`last` variables in their template to switch from using `ngFor` to `igxForOf`.

3phase:
Closes #5310   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -129,6 +129,23 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
     public totalItemCount: number = null;
 
+    /**
+     * A function that returns whether the element is the first or not
+     */
+    public first = (index) => { return index === 0};
+    /**
+     * A function that returns whether the element is the last or not
+     */
+    public last = (index) => { return index === this.igxForOf.length};
```
ChronosSF: arrays are 0-based so the last index is length - 1

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -323,7 +340,14 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
                 const input = this.igxForOf[i];
                 const embeddedView = this.dc.instance._vcr.createEmbeddedView(
                     this._template,
-                    { $implicit: input, index: this.igxForOf.indexOf(input) }
+                    { 
+                        $implicit: input, 
+                        index: this.igxForOf.indexOf(input), 
+                        first: this.first(this.igxForOf.indexOf(input)),
```
ChronosSF: don't indexOf for each asignment, also this should just be *i* (not sure why it was indexOf(input) before but just try with i and see if the tests succeed

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -129,6 +129,23 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
      */
     public totalItemCount: number = null;
 
+    /**
+     * A function that returns whether the element is the first or not
+     */
+    public first = (index) => { return index === 0};
```
ChronosSF: if these are in the context of forOf they should be affixed so that it's clear they return something regarding a forOf item, e.g.:

```typescript
private isFirstItem(index: number): boolean {
     return index === 0;
}
```

Notice that the function is also private, the parameters and return type is set. There is no reason for it to be public

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -129,6 +129,23 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
      */
     public totalItemCount: number = null;
 
+    /**
```
MayaKirova: While using functions is fine I like how ngFor defines the context better - https://github.com/angular/angular/blob/master/packages/common/src/directives/ng_for_of.ts#L14
It's defined in a separate class with getters for first/last/even/odd based on the index/count. This will allow for less code repetition and will be easier to maintain.

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
3phase: Done

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
3phase: Done

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
3phase: Done

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
3phase: Done

PR #5395, opened 2019-07-11
fix(igxGrid): Grid should render all columns when grid width is set to null.

### LINKED ISSUE #5288: Grid width set to `null` is not rendering grid columns at all.
## Description  
Setting grid with to `null` does not render columns at all making the grid empty and that is why rows render as 1px in height.
The same is for Tree Grid and Hierarchical Grid.

 * browser: any

## Steps to reproduce  

1. Open the "gridColumnResizing" demo sample
2. Set the width of the grid to `null`.
3. Observe the rendered grid.

## Result  
The grid is not rendering any columns.

## Expected result  
The grid should render all columns in the grid without horizontal virtualization. If there is vertical virtualization it should not be affected.

## Attachments  
width-null

MayaKirova:
Closes #5288  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/touch.ts
+++ b/projects/igniteui-angular/src/lib/core/touch.ts
@@ -19,11 +19,11 @@ export class HammerGesturesManager {
         inputClass: Hammer.TouchInput,
         recognizers: [
             [ Hammer.Pan, { threshold: 0 } ],
-            [ Hammer.Pinch, { enable: true } ],
-            [ Hammer.Rotate, { enable: true } ],
             [ Hammer.Swipe, {
                 direction: Hammer.DIRECTION_HORIZONTAL
-            }]
+            }],
+            [Hammer.Tap],
+            [Hammer.Tap, { event: 'doubletap', taps: 2 }, ['tap']]
         ]
     };
 
@@ -43,14 +43,14 @@ export class HammerGesturesManager {
     public addEventListener(element: HTMLElement,
                             eventName: string,
                             eventHandler: (eventObj) => void,
-                            options: object = null): () => void {
+                            options: HammerOptions = null): () => void {
 
         // Creating the manager bind events, must be done outside of angular
         return this._zone.runOutsideAngular(() => {
             let mc: HammerManager = this.getManagerForElement(element);
             if (mc === null) {
                 // new Hammer is a shortcut for Manager with defaults
-                mc = new Hammer(element, this.hammerOptions);
+                mc = new Hammer(element, Object.assign(this.hammerOptions, options));
                 this.addManagerForElement(element, mc);
             }
             const handler = (eventObj) => { this._zone.run(() => { eventHandler(eventObj); }); };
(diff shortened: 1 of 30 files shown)


ChronosSF:
check the comments in https://github.com/IgniteUI/igniteui-angular/pull/5396

MayaKirova:
Issues as described by @ChronosSF:

- [x] most (maybe all but I noticed it half-way through so I am not 100% sure) throw similar to the following error:
> ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'width: 38843px'. Current value: 'width: 38859px'.
    at viewDebugError (vendor.js:61388)
    at expressionChangedAfterItHasBeenCheckedError (vendor.js:61376)
    at checkBindingNoChanges (vendor.js:61576)
    at checkNoChangesNodeInline (vendor.js:71156)
    at checkNoChangesNode (vendor.js:71143)
    at debugCheckNoChangesNode (vendor.js:71747)
    at debugCheckRenderNodeFn (vendor.js:71701)
    at Object.eval [as updateRenderer] (GridPerformanceSampleComponent.ngfactory.js:237)
    at Object.debugUpdateRenderer [as updateRenderer] (vendor.js:71690)
    at checkNoChangesView (vendor.js:71044)
- [x] hidden columns are **incorrectly** included in the total width
- [x] gridCellEditing - the only column (the first one that's templated with buttons) which has width unset assigns a very weird 414px width. 
- [x] the column groups sample looks quite broken - **Note:** Please make sure grid is not in a flex container since flex containers don't show scrollbars so you won't see the whole grid. 
- [ ] moving sample - window doesn't scroll with the dragged column header (that's expected, I am not arguing that but weirdly enough the multi-cell selection works, maybe we should support it here as well)
- [ ]  **general navigation** - any navigated element out of view is just not followed by the scroll - if we are to support width=null we'll need new navigation code for this
- [ ] percent width columns sample - with % columns the columns are rendered in a very weird (small) size but not 0 (now I don't expect this to work generally speaking, but it will be nice to document it)
- [ ] remote sample - grid renders without width initially which makes sense but we might need to have a min width to render the Empty Data message first? or at least document this
- [ ] search sample - searching on the column that's out of view doesn't scroll to it (similar to navigation)
- [x] hierarchical grid sample
    - first grid - null width in child layout super breaks it
    - third sample - you can't navigate to the cells to the right in child layouts - this is similar to other navigation issues but I feel this might be harder to fix for hgrid
- [x] hierarchical grid updating sample - the third layout is broken. This seems to happen when there is no data even if columns are assigned (but not with width) - the content then gets squashed into a 1px width or so.

Note that the following: 
- moving sample
- **general navigation**
- search sample

Are related to features that do not support external scrollbars to begin with, so may potentially not be considered as part of the scope of this issue. 

Note that the following may need additional discussion:
- percent width columns sample - Since % of null width is nothing we could set the column widths to be the min allowed (136px). Let me know if that's ok or if we should just document is as not supported since it's a combination that does not make much sense.
- remote sample - Set 100% as default until data is received and columns can be generated or something else?

ChronosSF:
Check the new issues in #5397

PR #5396, opened 2019-07-11
fix(igxGrid): Grid should render all columns when grid width is set to null. 

### LINKED ISSUE #5288: Grid width set to `null` is not rendering grid columns at all.

MayaKirova:
Closes #5288   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -3999,7 +3999,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
      */
     protected _derivePossibleWidth() {
         if (!this.columnWidthSetByUser) {
-            this._columnWidth = this.getPossibleColumnWidth();
+            this._columnWidth = this.width ? this.getPossibleColumnWidth() : MINIMUM_COLUMN_WIDTH + 'px';
             this.columnList.forEach((column: IgxColumnComponent) => {
                 if (this.hasColumnLayouts && parseInt(this._columnWidth, 10)) {
                     const columnWidthCombined = parseInt(this._columnWidth, 10) * (column.colEnd ? column.colEnd - column.colStart : 1);
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -649,10 +649,12 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
     @HostBinding('style.width')
     get hostWidth() {
         if (this.width === null && this.columnList && this.columnList.length > 0) {
-            let width = this.columnList.reduce((sum, item) => sum + parseInt((item.width || item.defaultWidth), 10), 0);
+            let width = this.columnList.filter(x => !x.columnGroup)
+            .reduce((sum, item) => sum + parseInt((item.width || item.defaultWidth), 10), 0);
             if (this.hasVerticalSroll()) {
                 width += this.scrollWidth;
             }
+            width += this.getFeatureColumnsWidth();
             return width + 'px';
         } else {
             return this.width;
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -4244,6 +4244,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         if (!colSum) {
             return null;
         }
+        this.cdr.detectChanges();
         colSum += this.getFeatureColumnsWidth();
         return colSum;
     }
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -643,8 +643,11 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         }
     }
 
+    /**
+     * @hidden
+    */
     @HostBinding('style.width')
-    protected get hostWidth() {
+    get hostWidth() {
         if (this.width === null && this.columnList && this.columnList.length > 0) {
             let width = this.columnList.reduce((sum, item) => sum + parseInt((item.width || item.defaultWidth), 10), 0);
             if (this.hasVerticalSroll()) {
--- a/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.component.ts
@@ -4124,9 +4124,7 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
                 parseInt(this.document.defaultView.getComputedStyle(this.nativeElement).getPropertyValue('width'), 10);
         }
 
-        if (this.showRowCheckboxes) {
-            computedWidth -= this.headerCheckboxContainer ? this.headerCheckboxContainer.nativeElement.offsetWidth : 0;
-        }
+        computedWidth -= this.getFeatureColumnsWidth();
 
         const visibleChildColumns = this.visibleColumns.filter(c => !c.columnGroup);
 
@@ -4202,8 +4200,8 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
 
     private getColumnWidthSum(): number {
         let colSum = 0;
-        const cols = this.visibleColumns
-        .filter(x => !x.columnGroup);
+        const  cols = this.hasColumnLayouts ?
+         this.visibleColumns.filter(x => x.columnLayout) : this.visibleColumns.filter(x => !x.columnGroup);
         cols.forEach((item) => {
             colSum +=  parseInt((item.width || item.defaultWidth), 10) || MINIMUM_COLUMN_WIDTH;
         });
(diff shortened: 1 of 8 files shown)


ChronosSF:
I tested each sample in the demos repo with width null. These are my findings:

- most (maybe all but I noticed it half-way through so I am not 100% sure) throw similar to the following error:
> ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'width: 38843px'. Current value: 'width: 38859px'.
    at viewDebugError (vendor.js:61388)
    at expressionChangedAfterItHasBeenCheckedError (vendor.js:61376)
    at checkBindingNoChanges (vendor.js:61576)
    at checkNoChangesNodeInline (vendor.js:71156)
    at checkNoChangesNode (vendor.js:71143)
    at debugCheckNoChangesNode (vendor.js:71747)
    at debugCheckRenderNodeFn (vendor.js:71701)
    at Object.eval [as updateRenderer] (GridPerformanceSampleComponent.ngfactory.js:237)
    at Object.debugUpdateRenderer [as updateRenderer] (vendor.js:71690)
    at checkNoChangesView (vendor.js:71044)
- hidden columns are **incorrectly** included in the total width
- gridCellEditing - the only column (the first one that's templated with buttons) which has width unset assigns a very weird 414px width.
- the column groups sample looks quite broken
- moving sample - window doesn't scroll with the dragged column header (that's expected, I am not arguing that but weirdly enough the multi-cell selection works, maybe we should support it here as well)
- mrl sample - it seems the mrl grid cuts off with the window, not with the columns
- mrl generator sample - think the column don't have width assings which just destroys the grid
- **general navigation** - any navigated element out of view is just not followed by the scroll - if we are to supported we'll need new navigation code for this
- percent width columns sample - with % columns the columns are rendered in a very weird (small) size but not 0 (now I don't expect this to work generally speaking, but it will be nice to document it)
- remote sample - grid renders without width initially which makes sense but we might need to have a min width to render the Empty Data message first? or at least document this
- row dragging sample - broken (extra space)
- search sample - searching on the column that's out of view doesn't scroll to it (similar to navigation)
- hierarchical grid sample
    - first grid - null width in child layout super breaks it
    - third sample - you can't navigate to the cells to the right in child layouts - this is similar to other navigation issues but I feel this might be harder to fix for hgrid
- hierarchical grid updating sample - the third layout is broke. This seems to happen when there is no data even if columns are assigned (but not with width) - the content then gets squashed into a 1px width or so.

ChronosSF:
@kdinev @mpavlinov @rkaraivanov , because of the issues with navigation, search, having to change outer scroll position so that drag operations work, as well as, having to think of some default width scenarios when width would otherwise not be available because of the content box (we don't want to render grids looking broken right?) I suggest we keep width=null out of support. No one has asked for it, we found it while doing the sizing topic. This will require quite some investment and code bloat to resolve all of the above.

MayaKirova:
@ChronosSF Well, neither the search nor the navigation nor the drag operator would work with an external scrollbar in any scenario (even if we drop the width=null use case). For example, if you set a large grid width when it is in smaller container so that it generates scrollbars for that container you would have the same issues with those features as they are all specifically target the grid's virtual scrollbar.

ChronosSF:
> @ChronosSF Well, neither the search nor the navigation nor the drag operator would work with an external scrollbar in any scenario (even if we drop the width=null use case). For example, if you set a large grid width when it is in smaller container so that it generates scrollbars for that container you would have the same issues with those features as they all specifically target the grid's virtual scrollbar.

In either case we need to decide what to do in this case. I would assume people will hit out of view scenarios more often if they use width null but I might be wrong. You could argue that, because websites are designed to never render a horizontal scroll bar as it's generally horrible UX, they won't keep width=null if it's too large (and would instead use it for when the total width of columns is smaller than parent containers) but it's still something to consider.

MayaKirova:
Issues as described by @ChronosSF:

- [x] most (maybe all but I noticed it half-way through so I am not 100% sure) throw similar to the following error:
> ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'width: 38843px'. Current value: 'width: 38859px'.
    at viewDebugError (vendor.js:61388)
    at expressionChangedAfterItHasBeenCheckedError (vendor.js:61376)
    at checkBindingNoChanges (vendor.js:61576)
    at checkNoChangesNodeInline (vendor.js:71156)
    at checkNoChangesNode (vendor.js:71143)
    at debugCheckNoChangesNode (vendor.js:71747)
    at debugCheckRenderNodeFn (vendor.js:71701)
    at Object.eval [as updateRenderer] (GridPerformanceSampleComponent.ngfactory.js:237)
    at Object.debugUpdateRenderer [as updateRenderer] (vendor.js:71690)
    at checkNoChangesView (vendor.js:71044)
- [x] hidden columns are **incorrectly** included in the total width
- [x] gridCellEditing - the only column (the first one that's templated with buttons) which has width unset assigns a very weird 414px width. 
- [x] the column groups sample looks quite broken - **Note:** Please make sure grid is not in a flex container since flex containers don't show scrollbars so you won't see the whole grid. 
- [ ] moving sample - window doesn't scroll with the dragged column header (that's expected, I am not arguing that but weirdly enough the multi-cell selection works, maybe we should support it here as well)
- [ ]  **general navigation** - any navigated element out of view is just not followed by the scroll - if we are to support width=null we'll need new navigation code for this
- [ ] percent width columns sample - with % columns the columns are rendered in a very weird (small) size but not 0 (now I don't expect this to work generally speaking, but it will be nice to document it)
- [ ] remote sample - grid renders without width initially which makes sense but we might need to have a min width to render the Empty Data message first? or at least document this
- [ ] search sample - searching on the column that's out of view doesn't scroll to it (similar to navigation)
- [x] hierarchical grid sample
    - first grid - null width in child layout super breaks it
    - third sample - you can't navigate to the cells to the right in child layouts - this is similar to other navigation issues but I feel this might be harder to fix for hgrid
- [x] hierarchical grid updating sample - the third layout is broken. This seems to happen when there is no data even if columns are assigned (but not with width) - the content then gets squashed into a 1px width or so.
- [x] row dragging sample - broken (extra space)
- [x] mrl sample - it seems the mrl grid cuts off with the window, not with the columns. **Note:** Make sure container is not flex so that it can generate scrollbars as needed. Otherwise it will cut off the overflowing content.

Note that the following: 
- moving sample
- **general navigation**
- search sample

Are related to features that do not support external scrollbars to begin with, so may potentially not be considered as part of the scope of this issue. 

Note that the following may need additional discussion:
- percent width columns sample - Since % of null width is nothing we could set the column widths to be the min allowed (136px). Let me know if that's ok or if we should just document is as not supported since it's a combination that does not make much sense.
- remote sample - Set 100% as default until data is received and columns can be generated or something else?

MayaKirova:
@mpavlinov @kdinev @rkaraivanov @ChronosSF 

The only questions that remain:
1) Should the following features - navigation, moving and searching, work with external scrollbars (scrollbars that are not generated by the grid itself but from an external scrollable container)? For example scrolling the navigated/search matching cell in view, scrolling when dragging the movable header outside the view etc.
Here's a sample with external scroll so that you can get a better idea: https://stackblitz.com/edit/angular-dpvaad
If this should work, as it is not directly related to the width = null scenario, I suggest we log it separately for the related features.

2) If there are no visible columns - for example they should be autogenerated but there is no data yet (remote data scenario), or all columns are hidden, and width = null, should we set some default width for the grid until columns are generated/shown? If so what would a good default be?

kdinev:
@MayaKirova Here's the answers:

1. No, it's not required to handle container scrolling out of the box. This scenario is mostly for grids that have only a few columns, and in general wouldn't produce horizontal scrollbars.
2. The default width of the grid should be the width of the data loading/empty grid template, which means it's not something you should handle.

MayaKirova:
@kdinev @ChronosSF @mpavlinov 
Regarding 2)
In case we are showing just the loading indicator or the empty grid text, since they are rather small they don’t look very good IMO: 
image 
image


Also if we hide all columns in the grid (users can do that rather easily via the toolbar) and the width is null that would completely hide the grid since in that case there will be no default content to show.

Perhaps in that case it would be better to set 100% width or some default width. For example the min width for 1 or 2 columns, if we follow the same logic as the height for the loader as described here: https://github.com/IgniteUI/igniteui-angular/wiki/Hierarchical-Grid-Specification#6-loader
-> A grid that needs to load before displaying records, has a fixed height of 100px (based on the height of two rows on comfortable display density). 

Let me know what you think.

mpavlinov:
When there is no data I'm for having a minimal width of the sum for a 2 columns i.e. 2x136px = 272px

skrustev:
For a grid with width `null` and columns width set to percentages, sizes the columns afterwards based on the initially calculated grid size. The initial grid size is calculated on the basis that the columns have width `136px`

Steps to reproduce:
 1. Open the 'Grid percentage' demo
 2. Set all columns width to '30%', unpinned and the grid width to 'null'.
 3. Observe the size of the columns.

Their combined width is bigger than the calculated grid size and they go out of the grid view area.

ChronosSF:
Check the new issues in #5397

PR #5505, opened 2019-07-24
Add pointerEvents: none to use tag of icon's svg

### LINKED ISSUE #4901: In IE11 the excel style filter dialog is closed when change a filter condition
## Description  
In IE11 the excel style filter dialog is closed when change a filter condition

 * browser: IE11

## Steps to reproduce  

1. Open dev sample gridColumnMoving in IE11
2. Open text filter for column ConpanyName and condition contains
3. In the excel style filter dialog click to change the condition

## Result  
The excel filter dialog is closed

## Expected result  
The dialog should not be closed

wnvko:
When use `use` tag in `svg` most browsers creates clone of the targeted `svg`. When one clicks on such clone it correctly shows its parent. In same case in IE when one clicks on cloned `svg` in click event handler the `svg` has no parent. Additionally if try to get its bounding rectangle it returns the rectangle of the original `svg` from the clicked one was cloned.
All this breaks `IgxOverlay`'s `closeOnOutsideClick` as it cannot determine correctly if the click was over the shown element or not. Adding `pointer-events: none` to `use` tag of `svg` fixes this issue.

Closes #4901 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/src/assets/svg/filtering/does_not_contain.svg
+++ b/src/assets/svg/filtering/does_not_contain.svg
@@ -9,7 +9,7 @@
             <g id="Icon-24px">
                 <polygon id="Shape" points="0 0 24 0 24 24 0 24"></polygon>
                 <mask id="mask-2" fill="white">
-                    <use xlink:href="#path-1"></use>
+                    <use xlink:href="#path-1" style="pointer-events: none;"></use>
                 </mask>
                 <g id="Combined-Shape"></g>
                 <g id="Colors/grays.600" mask="url(#mask-2)" fill="#000000" fill-opacity="0.54">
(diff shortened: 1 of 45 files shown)


src/assets/svg/filtering/does_not_contain.svg
```diff
@@ -9,7 +9,7 @@
             <g id="Icon-24px">
                 <polygon id="Shape" points="0 0 24 0 24 24 0 24"></polygon>
                 <mask id="mask-2" fill="white">
-                    <use xlink:href="#path-1"></use>
+                    <use xlink:href="#path-1" style="pointer-events: none;"></use>
```
simeonoff: We we need to specify those inline styles? You're already applying this in the styles of the igx-icon?

wnvko: yes, this is correct. Will remove inline styles. Thank you.

PR #5546, opened 2019-07-30
Fix IgxButton has incorrect CSS class when setting new type - 7.3.x

### LINKED ISSUE #5404: IgxButton: The type property setter does not work correctly
## Description
The IgxButtonDirective type property setter does not set a new CSS class but it appends the new CSS class.

 * browser: All

## Steps to reproduce  

1. Open dev demos

2. Edit the "button.sample.html" file, by adding the following to the first button:
"#firstButton (click)="buttonHandler($event)""

3. Edit the "button.sample.ts" file, by adding the following after the last field definition:
@ViewChild('firstButton', { read: IgxButtonDirective, static: true})
public aButton: IgxButtonDirective;

public buttonHandler(e) {
    this.aButton.type = 'outlined';
    // type can be one of these 'flat', 'raised', 'outlined', 'fab', 'icon'
}

4. Start dev demos and load the Button directive sample
5. Press the first button in the sample.
6. Inspect the same button in browser's dev tools

## Result  
After setting the type property to outlined the button's class is the following:
"igx-button--flat igx-button--outlined"
Also it does not looks like the other outlined buttons.

## Expected result  
The other buttons on the page with default type "outlined" have the following CSS class, which is the correct:
"igx-button--outlined"

rmkrmk:
Closes #5404

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/button/button.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/button/button.directive.ts
@@ -17,22 +17,32 @@ import { DisplayDensityBase, DisplayDensityToken, IDisplayDensityOptions, Displa
     selector: '[igxButton]'
 })
 export class IgxButtonDirective extends DisplayDensityBase {
+
     /**
      *@hidden
      */
-    private _type = 'flat';
+    private _type: string;
+
     /**
      *@hidden
      */
-    private _cssClass = 'igx-button';
+    private _defaultType = 'flat';
+
+    /**
+     *@hidden
+     */
+    private _cssClassPrefix = 'igx-button';
+
     /**
      *@hidden
      */
     private _color: string;
+
     /**
      *@hidden
      */
     private _label: string;
+
     /**
      *@hidden
      */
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/directives/button/button.directive.ts
```diff
@@ -66,18 +76,26 @@ export class IgxButtonDirective extends DisplayDensityBase {
...
      * @memberof IgxButtonDirective
      */
-    @Input('igxButton') set type(value: string) {
-        this._type = value || this._type;
-        this._renderer.addClass(this.nativeElement, `${this._cssClass}--${this._type}`);
+    @Input('igxButton')
+    set type(value: string) {
+        const newValue = value ? value : this._defaultType;
+        if (this._type !== newValue) {
+            this._renderer.removeClass(this.nativeElement, `${this._cssClassPrefix}--${this._type}`);
```
DiyanDimitrov: We should not remove the class if the old type is empty.

rmkrmk: This will just try to remove it. If there is no such class it will just do nothing.

Issue #5557, opened 2019-07-31
In column groups when column cannot be pinned, the pinned button should be disabled in excel style filter


nrobakova:
## Description  
In column groups when column cannot be pinned, the pinned button should be disabled in excel style filter

 * igniteui-angular version: 8.1
 * browser: n/a

## Steps to reproduce  

1. Edit dev sample gridColumnGroups and add [filterMode]="'excelStyleFilter'"
2. Open the sample 
3. Pin second group "Address Information"
4. Open excel style filter for column in group General Information

## Result  
Button Pin is enabled

## Expected result  
Button pin should be disabled, because the group cannot be pinned

## Attachments  
![image](https://user-images.githubusercontent.com/11463104/62620194-51444a00-b921-11e9-8c40-031ada5c71f0.png)

PR #5609, opened 2019-08-07
Column cannot be pinned through ESF if top parent cannot be pinned - 8.1.x

tachojelev:
#5557

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid/grid-filtering-ui.spec.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid-filtering-ui.spec.ts
@@ -3504,7 +3504,8 @@ describe('IgxGrid - Filtering actions - Excel style filtering', () => {
             declarations: [
                 IgxGridFilteringComponent,
                 IgxTestExcelFilteringDatePickerComponent,
-                IgxGridFilteringESFTemplatesComponent
+                IgxGridFilteringESFTemplatesComponent,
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/grids/grid/grid-filtering-ui.spec.ts
```diff
@@ -5614,6 +5615,37 @@ describe('IgxGrid - Filtering actions - Excel style filtering', () => {
...
+
+        it('Should not pin column when its parent group cannot be pinned.', fakeAsync(() => {
+            // Test prerequisites
+            grid.width = '1000px';
+            fix.detectChanges();
+            tick(100);
+
+            // Pin the 'AnotherField' column.
+            GridFunctions.clickExcelFilterIcon(fix, 'AnotherField');
+            fix.detectChanges();
```
nrobakova: You can use here grid.pinColumn('AnotherField'); instead of opening the excel style filter for the column AnotherField.

tachojelev: I believe we should use the ESF for pinning, since the actual scenario for reproducing the issue is by using the UI as an end user.

PR #5663, opened 2019-08-19
Re-emitting events from child grid to related Row Island.

### LINKED ISSUE #5611: Hierarchical Grid child (row island) events
From what I gather the row islands are not actual grids and thus don't emit anything beyond the `onGridCreated` that can be seen in the [Data Binding](https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/hierarchical_grid.html#data-binding) section. But since the `IgxRowIslandComponent` extends the [H]Grid base it also inherits all the event emitters:
https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/igxrowislandcomponent.html

There's however no explicit mention that events can be handled by manually subscribing to the emitters trough `onGridCreated` and if that's the only way it can/should be done.
Docs improvement opportunity right here, regardless of the answer.

Either way, the listed output events on the Row island can cause confusion and lead to people wondering why they aren't working - case in point https://github.com/IgniteUI/igniteui-angular/issues/4953
They should either work (which might not be technically feasible, but sounds nice) or not be part of the API listing.

MayaKirova:
Closes #5611   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@mpavlinov @ChronosSF @damyanpetev Renaming "owner" to "context" in the base event args interface  since some events already use "owner" as an event arg (RowDragStartEventArgs, IRowDragEndEventArgs etc.) and their type is different than the owner component.

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -262,6 +262,13 @@ export interface CancelableEventArgs {
     cancel: boolean;
 }
 
+export interface IBaseEventArgs {
+    /**
+     * Provides reference to the context component.
+     */
+    context?: any;
```
mpavlinov: @MayaKirova: We discussed this and will want to use "owner" instead. For the event arguments that already have "owner" we will make a breaking change and document it.

ChronosSF:
@damyanpetev @Lipata , please review the new names for event argument properties related to the row dragging feature.

PR #5672, opened 2019-08-21
Adding templates for grid expansion indicator icons. Changing default icons.

### LINKED ISSUE #4530: Hide expansion indicator in hierarchical grid if no data found
## Question  
Run stackblitz example below, the first row doesn't have any data for Albums or Tours.
```
"Tours": [],
"Albums": []
```
In that case, how can I hide the expansion indicator to have the row just a normal grid row so the user knows without expanding the row that data is empty?

### The example here:
https://stackblitz.com/edit/angular-oq8q3z

### Result
Expansion indicators are shown for empty row-island grid

### Expected result
No expansion indicator should be displayed

### LINKED ISSUE #5540: Allow to expand all rows in hierach grid
It is always annoying if we want to see the full content of all rows that we need to click each one to expand it.

We have a button which allows to collapse all rows. It would be good if we could reuse this button to expand all rows if we hve no expanded rows. Since this can be quite a performance hit (if we lazy load a lot of rows) this should be configurable. Allowing this feature could be enabling by adding a new property <igx-hierarchical-grid [allowExpandAll]="true">
If this flag is true the behaviour of the button will be the same as the igx-grid in grouped mode

### LINKED ISSUE #5621: Expose inputs for grid expansion indicator icons
I should be able to replace the expand/collapse icons with icons that I want to use for my application.

Inputs for expand and collapse icon.

Another approach would be for templates to be exposed.

## Additional context
Also change the default icons to right and down chevrons.

MayaKirova:
Closes #5621  
Closes #5540
Closes #4530

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@mpavlinov Setting in awaiting test state so that #4530 can also be tested.

mpavlinov:
@MayaKirova Here is the feedback:
1. Toggle action logic stays above the template - template will be used only for visual changes
2. Expose an input for field name which will indicate whether to hide/show the expand/collapse

MayaKirova:
> @MayaKirova Here is the feedback:
> 
> 1. Toggle action logic stays above the template - template will be used only for visual changes
> 2. Expose an input for field name which will indicate whether to hide/show the expand/collapse

Regarding 1) -> Does this include the header expand/collapse template or just the ones for the data rows?
Regarding 2) -> Should this be an input on the hierarchical grid (one for all levels of the hierarchy) or should it also be settable per rowIsland?

mpavlinov:
> > @MayaKirova Here is the feedback:
> > 
> > 1. Toggle action logic stays above the template - template will be used only for visual changes
> > 2. Expose an input for field name which will indicate whether to hide/show the expand/collapse
> 
> @mpavlinov
> Regarding 1) -> Does this include the header expand/collapse template or just the ones for the data rows?
> Regarding 2) -> Should this be an input on the hierarchical grid (one for all levels of the hierarchy) or should it also be settable per rowIsland? Additionally, should this prevent expansion triggered from outside the template (for example, expandAll Api on the grid, toggle api called on the row instance etc.)
1. Only for the data rows
2. For hierarchical grid and for row islands. Row island will override the hierarchical grid one. Yes. It should prevent expansion from outside the template. After all the developer says that there are not children.

MayaKirova:
@kdinev Please confirm that the spec complies with your expectations so that we can continue with the implementation based on the new requirements.

kdinev:
@MayaKirova I'm wondering why in the example the hgrid instance has to be passed to the function call from that hgrid instance? 

```
<igx-icon role="button" fontSet="material" (click)="hgrid.expandAll(hgrid)">add</igx-icon>
```

I'm also starting to lead towards exposing both expand all and collapse all templates out of the box, with an hgrid input for whether the expandAll template should be shown, defaulting to false. This way the user would be able to show expandAll with our default template by simply setting `[showExpandAll]="true"`. Then, if they want to change the icon, they would pass an expand all template, but wouldn't have to call the expand all grid logic from the template element.

MayaKirova:
@kdinev 

hgrid doesn’t need to be passed as the expandAll api takes no params. Seems to be a problem in the spec snippet.

Reg 2) Yes. It’s currently a bit inconsistent since the header template allows changing the behavior (since api calls are done inside the template) while the ones for the rows do not. 
I’m ok with adding an input and documenting as a limitation that expand all operation will have a negative performance impact as it will create a large number of child hierarchical grids.

Let me know if that’s what we’ll go with and I’ll update the spec accordingly.

MayaKirova:
@kdinev 
I have updated the [Expand All Header Indicator section](https://github.com/IgniteUI/igniteui-angular/wiki/Hierarchical-Grid-Specification#expand-all-header-indicator) of the spec and added information on the new [hierarchical grid properties](https://github.com/IgniteUI/igniteui-angular/wiki/Hierarchical-Grid-Specification#properties).
Please review the spec and let me know if this will be the final approach we should implement. 
In short:
-	Templates (both row and header) will serve only visualization purpose ( cannot change default behavior via the templates.)
-	To achieve feature request #5540  (Allow to expand all rows in hierach. Grid) we will add an additional input – showExpandAll, which when set will render the expand all template and users will be able to expand all rows in the current grid via that template.
-	To achieve feature request #4530 (Hide expansion indicator in hierarchical grid if no data found) we will add an additional input – hasChildrenKey, which when set will check for a property(boolean) in the data source record, that indicates whether there are children for the particular record. If record does not have children (according to the prop value in the record) it will not show expand icon and will not be expandable (cannot be toggled via expand/collapse all api or other api calls).

kdinev:
@MayaKirova I like it! Go ahead and make the necessary changes.

MayaKirova:
@3phase Please re-retest according to the agreed on spec below:

> @kdinev
> I have updated the [Expand All Header Indicator section](https://github.com/IgniteUI/igniteui-angular/wiki/Hierarchical-Grid-Specification#expand-all-header-indicator) of the spec and added information on the new [hierarchical grid properties](https://github.com/IgniteUI/igniteui-angular/wiki/Hierarchical-Grid-Specification#properties).
> Please review the spec and let me know if this will be the final approach we should implement.
> In short:
> 
> * Templates (both row and header) will serve only visualization purpose ( cannot change default behavior via the templates.)
> * To achieve feature request #5540  (Allow to expand all rows in hierach. Grid) we will add an additional input – showExpandAll, which when set will render the expand all template and users will be able to expand all rows in the current grid via that template.
> * To achieve feature request #4530 (Hide expansion indicator in hierarchical grid if no data found) we will add an additional input – hasChildrenKey, which when set will check for a property(boolean) in the data source record, that indicates whether there are children for the particular record. If record does not have children (according to the prop value in the record) it will not show expand icon and will not be expandable (cannot be toggled via expand/collapse all api or other api calls).

PR #5716, opened 2019-08-29
feat(checkbox): add readonly prop #5675

### LINKED ISSUE #5675: Checkbox readonly property support
When a checkbox is used as a representational part of a larger complex object it's sometimes hard to control the checked state. We seem to hit case within a list/row selection scenario, where a listener would b active above the check to manage the state. This usually calls for preventing the default:
```html
<igx-checkbox (click)="$event.preventDefault()" [checked]="someProp"></igx-checkbox>
```

Perhaps a `readonly` property could make the usage less awkward:
```html
<igx-checkbox readonly [checked]="someProp"></igx-checkbox>
```

## Additional Context
I realize that the base HTML checkbox doesn't support `readonly` as it really doesn't make that much sense for a standalone check.
> It's actually read only by design, as the UI interaction toggles the `checked` state, which is separate from the usual input `value` that remains unchanged so the property does nothing.

However, there's already a precedent with our Ignite UI for JavaScript checkbox and this is more of a convenience thing. And again - this is really intended for use when the checkbox is in a more complex component that'll still respond to the same/similar interactions.

PS: This may very well be true for the Switch.

Lipata:
Closes #5675 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ For more information about the theming please read our [documentation](https://w
 - `IgxDrop`
     - Linking of drag and drop elements. This can be achieved by using the new provided `dropChannel` input, specifying each drop area to which channel it corresponds.
     - Drop strategies. Three new drop strategies have been provided - Append, Prepend and Insert.  Also an input `dropStrategy` to the `igxDrop` which specify which strategy should be used when dropping an element inside the drop area. Custom one can be specified as well.
+- `IgxCheckbox`
+    - introduced new property `readonly` that doesn't allow changing component state, but preserves default checkbox style.
```
damyanpetev:
```suggestion
    - introduced a new `readonly` property that doesn't allow user interaction to change the state, but keeps the default active style. Intended for integration in complex controls that handle the interaction and control the checkbox instead through binding.
```

projects/igniteui-angular/src/lib/checkbox/checkbox.component.ts
```diff
@@ -258,12 +258,24 @@ export class IgxCheckboxComponent implements ControlValueAccessor, EditorProvide
...
      * ```
      * @memberof IgxCheckboxComponent
      */
     @HostBinding('class.igx-checkbox--disabled')
     @Input() public disabled = false;
+    /**
+     * Sets/gets whether the checkbox is readonly.
+     * Default value is `false`.
+     * ```html
+     * <igx-checkbox [readonly] = "true"></igx-checkbox>
```
damyanpetev:
```suggestion
     * <igx-checkbox [readonly]="true"></igx-checkbox>
```

damyanpetev:
@Lipata Actually, might need a bit more for this:
While the toggle is correctly prevented, the native input is still focusable through tab navigation **and** you can toggle it with a space key. This doesn't change the component state indeed, but the native checkbox does change (not good if someone ever decides to use actual form submit). At least the aria-checked attribute is bound to the component, so readers report it correctly.

A prevent in the _onCheckboxClick should do it as both label click and space trigger the click.

PR #5721, opened 2019-08-30
Add seconds support for time-picker component

Tian-Hun:
Closes #  

 - [ ] Bug fix
 - [X] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [X] This PR includes `feature/README.MD` updates for the feature docs


kdinev:
@Lipata Please assign somebody to test this PR.

jackofdiamond5:
@Tian-Hun There seems to be a problem with the parsing of the input value's seconds - when you edit the seconds and expand the time picker, the changes are not reflected. Another thing that happens is when you expand the time picker and try to select a different value for the seconds, it updates the minutes' value instead.

Tian-Hun:
Ok, I understand and fix it.

damyanpetev:
@Tian-Hun Thanks!
While at it, do you mind renaming all the "second-" names to "seconds-" just so it won't be confused with secondary/another (e.g. `secondList` -> `secondsList`, `selectedSecond*s*`, etc)?

Ask away if you need help.

Tian-Hun:
@damyanpetev Thank you for your reminder, I will modify it.

jackofdiamond5:
@Tian-Hun 
The reason why no changes are reflected in the time-picker once you update the input’s value is because when you have templating you have to apply the changes by yourself in the `onBlur` event handler. The `timePickerValue` argument is what can be used to set the seconds.

Also, I see that you have updated the time-picker’s tests with the new functionality for seconds. And the reason why most of them fail is because you also have to update the testing components' formats, as well as their initial dates. Some tests will continue to fail, however, due to the new functionality, you should update them accordingly. For example, the `TimePicker UP Down Keyboard navigation` test fails because the time-picker's header is not set up to have seconds. So you may want to add that in the `time-picker.component` template and then update the test as well. Another thing is that the parsing of the mask seems to be incorrect. For example if you set the format to "h tt" and you select any value above 9, from the time-picker, it trims it to 1 or 2 (depending on format) and selecting any value below 10 is turned into a 0. This happens in the `time-picker.pipes`. In it the values for hours and minutes are trimmed out of the final result, trimming the seconds, however, while also trimming hours and/or minutes results in the entire string being removed, except for the AM/PM part. For example, you could use the `lastIndexOf(':')` function to first trim the seconds part and then the minutes part if needed.

Also, deleting the input's value while having the format set to "h tt" causes the mask to look like "--:-- tt" instead of "-- tt". You could fix it by adding an additional check in the format setter. At the moment you check `if (!this.showHoursList || !this.showMinutesList || !this.showSecondsList)` - and if this passes, you trim the first part of the mask. With the addition of seconds the mask looks like this - "00:00:00 LL" - trimming the first two zeros will cause the mask to be incorrect - you have to trim the first two zeros again if you have `!this.showSecondsList === true`.

And again if you need help with anything, just ask. :)

DSergii:
Hello everyone! 
Wich igniteui-angular version support seconds? I need it for angular 8.2+
I've tried with "igniteui-angular": "8.2.31", but it doesn't support seconds.
Higher versions of "igniteui-angular", like 9+, doesn't work with angular 8.2+

Lipata:
Hi, @DSergii.

When we develop features, we develop them for the upcoming versions - which in the case with the seconds list means that it was developed for version 9.0.0 and above. You can see that it was initially released with [9.0.0-beta.7](https://github.com/IgniteUI/igniteui-angular/commit/b6468e4debb0816e52d18f58234d8dd89034fd05).

Lipata:
@DSergii, you can following our [Update Guide](https://www.infragistics.com/products/ignite-ui-angular/angular/components/general/update_guide.html) and ask if you have some problems.

DSergii:
Thank you @Lipata, but unfortunately, I can't update angular version and seems 9.0.0-beta.7 will not work with angular 8.2+

PR #5802, opened 2019-09-13
fix(igxGrid): Fixing sizing in flexible layout so that grid content c…

### LINKED ISSUE #4244: Grid sizing in flexible layout
## Description  
The grid control doesn't properly adjust its size when inside a flexible layout

 * igniteui-angular version: 7.1.5
 * browser: chrome

## Steps to reproduce  

1. Go to https://stackblitz.com/edit/grid-in-flex-layout
2. Scale the right hand preview window by dragging it left and then right. The grid will grow bigger but not smaller.
3. Use the "open in new window" top right link on StackBlitz to see the sample in desktop modus with pinned side menu
4. Press the hamburger button to hide and show the menu. The grid will grow bigger or smaller but there will be empty white space at the rigth side of the grid
5. Scale the browser window down a little while still staying in desktop resolution. The grid will not scale with the page.

## Result  
See step 2, step 4 and step 5

## Expected result  
After step 2: Grid should grow smaller when the page width is reduced
After step 4: Grid should adjust its layout over the full width rather than getting some empty white space
After step 5: Grid should grow smaller when the page width is reduced

## Attachments  
Sample: https://stackblitz.com/edit/grid-in-flex-layout

### LINKED ISSUE #4952: [igx-grid][Chrome] the calculation of grid width is not correct when grid is in a flexbox.
## Description  
On Chrome, the calculation of grid width is not correct when the grid is in a flexbox.

 * igniteui-angular version: 7.3.2
 * browser: Chrome

## Steps to reproduce  

1. Run the attached sample on Chrome
2. Click "load data" button.
3. Enlarge browser width.
4. Shrink browser width.

## Result  
On step1(after initialization), the box that includes `igx-grid` is a little bigger than the lower box.
On step2(after loading data), the upper box becomes bigger.
On step4(after shrinking browser width), the upper box won't shrink.

Interestingly, the behavior on IE11 is perfect.
igx-grid-width-issue-in-flexbox

## Expected result  
The calculation of `igx-grid` width should be as same as the one on IE11.

## Attachments  
[igx-grid-sample.zip](https://github.com/IgniteUI/igniteui-angular/files/3241080/igx-grid-sample.zip)

MayaKirova:
…an be sized down after being rendered once and then window is resized.

Additional Notes:

In case a flex layout has some child element with width/min-width it cannot be sized down to a width smaller than that the width of those child element hence the observed behavior in the grid. Once rendered it has child elements that have width, so the flex container cannot be scaled down below that width.

Using flex-basis instead of width on the related direct child elements and overflow: hidden for other container elements seems to resolve this. 

Should be tested in different browsers and with the different grid features for all grids (grid, tree grid, hierarchical grid).

Adding this to 8.2.x only as the automatic detection of the size changes depends on the Resize Observer implementation (which is a new feature available only in 8.2.x and up).

Closes #4244 
Closes #4952

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -2749,7 +2749,11 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
         });
 
         this.resizeNotify.pipe(destructor, filter(() => !this._init), throttleTime(40))
-            .subscribe(() => this.notifyChanges(true));
+            .subscribe(() => {
+                this.zone.run(() => {
```
rkaraivanov: Not a fan of triggering a change detection cycle with a chatty observer

MayaKirova: In some cases, the resize observer is fired, but no change detection is triggered until you do some action on the grid (click a cell for example).  In those cases the grid appears broken (shows white spaces) until the next change detection cycle which updates the sizes. You can view such a scenario for example in the sample attached in issue #4952.

rkaraivanov: Then I suggest we increase the throttle time as having a spammable observer inside zone run is taxing on the UI thread

PR #5822, opened 2019-09-17
feat(IgxChip): Add new event when select animation is ended. Update the advanced filtering to use it instead.

### LINKED ISSUE #5821: The chip selection icon animation cannot be tracked and forces advance filtering to use setTimeout.
The advanced filtering currently uses setTimeout to wait for the chip select icon show/hide transition to end in order to reposition the filtering contextual menu.

The chip should emit an event when any transition animation ends when selection is changed.

skrustev:
Closes #5821

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid/grid-filtering-advanced.spec.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid-filtering-advanced.spec.ts
@@ -22,7 +22,7 @@ const ADVANCED_FILTERING_OPERATOR_LINE_OR_CSS_CLASS = 'igx-filter-tree__line--or
 const ADVANCED_FILTERING_OPERATOR_LINE_SELECTED_CSS_CLASS = 'igx-filter-tree__line--selected';
 const ADVANCED_FILTERING_TOOLBAR_ICON_BUTTON_FILTERED_CSS_CLASS = 'igx-grid-toolbar__adv-filter--filtered';
 
-fdescribe('IgxGrid - Advanced Filtering', () => {
+describe('IgxGrid - Advanced Filtering', () => {
     configureTestSuite();
     beforeEach(async(() => {
         TestBed.configureTestingModule({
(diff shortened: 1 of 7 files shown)


projects/igniteui-angular/src/lib/grids/grid/grid-filtering-advanced.spec.ts
```diff
@@ -2659,18 +2684,26 @@ describe('IgxGrid - Advanced Filtering', () => {
...
+}
+
+function selectColumnInEditModeExpression(fix, dropdownItemIndex: number) {
+    selectColumnInEditModeExpressionBase(fix, dropdownItemIndex);
     tick();
     fix.detectChanges();
 }
 
-function selectOperatorInEditModeExpression(fix, dropdownItemIndex: number) {
+function selectOperatorInEditModeExpressionBase(fix, dropdownItemIndex: number) {
```
tachojelev: It seems the only difference between the **selectOperatorInEditModeExpression** and the **selectOperatorInEditModeExpressionBase** methods is the type of waiting (tick or await wait()).
I believe we can simply create a new additional method called **selectOperatorInEditModeExpressionAsync** (without changing the existing non-async one), which is absolutely analogical to the **selectOperatorInEditModeExpression**, but uses await wait(), rather than tick(). This way it would become obvious pretty fast the difference between the two and tests will not need to call await wait() on their own.

projects/igniteui-angular/src/lib/grids/grid/grid-filtering-advanced.spec.ts
```diff
@@ -2659,18 +2684,26 @@ describe('IgxGrid - Advanced Filtering', () => {
     });
 });
 
-function selectColumnInEditModeExpression(fix, dropdownItemIndex: number) {
+function selectColumnInEditModeExpressionBase(fix, dropdownItemIndex: number) {
```
tachojelev: Same comments as for the **selectOperatorInEditModeExpressionBase** method.

MayaKirova:
@skrustev Event is triggered twice on select/deselect chip on Edge and IE.

PR #5829, opened 2019-09-19
fix(igxGrid): If set data is null/undefined update views accordingly.

### LINKED ISSUE #5751: IgxGrid binding issues
## Description  
Using `igx-grid` and binding it to `Observable` with `async` pipe. Observable emits an object that contains a `list` property of type `Array`. That property is used as data source for grid:
```
[data]="(data$ | async)?.list"
```
Everything works fine while `list` array has at least one item in it. When observable emits an object where `list` property is empty array or `null` or `undefined`, grid displays previously bound data and a message 'Grid has no data' underneath.

 * igniteui-angular version: 8.1.6
 * browser: Google Chrome (latest)

## Steps to reproduce  
I have created a small app to reproduce the issue:
https://stackblitz.com/edit/igniteui-angular-grid-no-data-issue?embed=1&view=preview
It has a filter input on top and two grids side by side. 
Left grid is bound to a object's property with `async` pipe (expression shown above). Right grid is bound to array.
To make the problem obvious enter `0` into a filter input to get three rows of data in both grids. Then enter something like `zzz` into filter input so that filtering would produce no data (data consists on numbers only).

Result is that left grid still shows same three rows from previous search and a message 'Grid has no data'.

Expected result would be that grid would show no data is data source has no data.

MayaKirova:
Closes #5751  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
@@ -411,7 +411,7 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
             if (changes) {
                 //  re-init cache.
                 if (!this.igxForOf) {
-                    return;
+                    this.igxForOf = [];
                 }
                 this._updateSizeCache();
                 this._zone.run(() => {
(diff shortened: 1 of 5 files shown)


kdinev:
@MayaKirova Nice coverage! 👍

PR #5908, opened 2019-10-02
Bypass all calls to browsers' object when not on browser platform

### LINKED ISSUE #4426: NavDrawer breaks server-side rendering
## Description  
I tried to build and app-shell that uses the nav-drawer in the app.component, and the build fails.

 * igniteui-angular version: 7.2.2
 * browser: N/A

## Steps to reproduce  

1. Create a default cli project with side-nav.
2. Add an app-shell and build.

## Result  
Build fails.

## Expected result  
Build should pass.

## Attachments  

> Hammer is not defined
> ReferenceError: Hammer is not defined
>     at new HammerGesturesManager (D:\GitHub\angular-training\node_modules\igniteui-angular\bundles\igniteui-angular.umd.js:67038:29)
>

wnvko:
There were some places in `touch.ts` as well as in `IgxNavDrawer` where we were calling browser related things, like `requestAnimationFrame`. Now we are checking if there platform where build runs is browser platform. If not, e.g. server side rendering, we skip calling browsers' methods.

Closes #4426   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/utils.ts
+++ b/projects/igniteui-angular/src/lib/core/utils.ts
@@ -231,7 +231,6 @@ export function isFirefox(): boolean {
 
 /**
  * @hidden
- * TODO: make injectable, check isPlatformBrowser()
  */
 @Injectable({ providedIn: 'root' })
 export class PlatformUtil {
@@ -241,12 +240,12 @@ export class PlatformUtil {
         this.platformBrowser = isPlatformBrowser(this.platformId);
     }
 
-    public isIOS(): boolean {
+    public get isIOS(): boolean {
         const iosBrowser = this.platformBrowser && /iPad|iPhone|iPod/.test(navigator.userAgent) && !('MSStream' in window);
         return iosBrowser;
     }
 
-    public isPlatformBrowser(): boolean {
+    public get isPlatformBrowser(): boolean {
         return this.platformBrowser;
     }
 }
--- a/projects/igniteui-angular/src/lib/core/utils.ts
+++ b/projects/igniteui-angular/src/lib/core/utils.ts
@@ -1,3 +1,6 @@
+import { Injectable, PLATFORM_ID, Inject } from '@angular/core';
+import { isPlatformBrowser } from '@angular/common';
+
 /**
  *@hidden
  */
@@ -230,10 +233,21 @@ export function isFirefox(): boolean {
  * @hidden
  * TODO: make injectable, check isPlatformBrowser()
  */
+@Injectable({ providedIn: 'root' })
 export class PlatformUtil {
-    static isIOS(): boolean {
-        const iosBrowser = typeof window !== 'undefined' && /iPad|iPhone|iPod/.test(navigator.userAgent) && !('MSStream' in window);
+    private platformBrowser: boolean;
+
+    constructor(@Inject(PLATFORM_ID) private platformId: Object) {
+        this.platformBrowser = isPlatformBrowser(this.platformId);
+    }
+
+    public isIOS(): boolean {
+        const iosBrowser = this.platformBrowser && /iPad|iPhone|iPod/.test(navigator.userAgent) && !('MSStream' in window);
         return iosBrowser;
+    }
+
+    public isPlatformBrowser(): boolean {
+        return this.platformBrowser;
     }
 }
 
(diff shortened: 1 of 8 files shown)


projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.ts
```diff
@@ -612,18 +614,26 @@ export class IgxNavigationDrawerComponent implements
     }
 
     private getWindowWidth() {
+        if (!this.platformUtil.isPlatformBrowser()) {
```
damyanpetev: This can likely stay (though I'd prefer a return value of something at least (0)).
Frankly, this only affects `checkPinThreshold` and while the current logic there will is fin like that, the whole check is pointless if there's no window to measure so it should be prevented instead.

projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.ts
```diff
@@ -805,19 +815,27 @@ export class IgxNavigationDrawerComponent implements
...
-            } else {
-                this.renderer.setElementStyle(this.drawer, 'transform', x ? 'translate3d(' + x + 'px,0,0)' : '');
-                this.renderer.setElementStyle(this.drawer, '-webkit-transform',
-                    x ? 'translate3d(' + x + 'px,0,0)' : '');
-            }
-            if (opacity !== undefined) {
-                this.renderer.setElementStyle(this.overlay, 'opacity', opacity);
-            }
-        });
+        if (this.platformUtil.isPlatformBrowser()) {
```
damyanpetev: This one shouldn't need a fallback, in fact this is only called during pan interaction, but if you want to make it safe just for future uses - just interrupt the method at start.

projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.spec.ts
```diff
@@ -532,8 +525,8 @@ describe('Navigation Drawer', () => {
         window.dispatchEvent(new Event('resize'));
 
         // wait for debounce
-        await wait(200);
-        expect(fixture.componentInstance.viewChild.pin).toBe(false, `Shouldn't change state on resize if window width is the same`);
+        await wait(2000);
```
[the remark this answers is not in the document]
wnvko: yes it can

projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.ts
wnvko: Actually in PWA with app shell, while building the app `ngAfterContentInit` is called. This calls `checkPinThreshold`. I am moving now the check to `checkPinThreshold` as it seams to be the only place where `getWindowWidth` is called with no browser.

projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.ts
wnvko: yeah, will remove that check

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -228,13 +231,23 @@ export function isFirefox(): boolean {
...
-    static isIOS(): boolean {
-        const iosBrowser = typeof window !== 'undefined' && /iPad|iPhone|iPod/.test(navigator.userAgent) && !('MSStream' in window);
+    private platformBrowser: boolean;
+
+    constructor(@Inject(PLATFORM_ID) private platformId: Object) {
+        this.platformBrowser = isPlatformBrowser(this.platformId);
+    }
+
+    public get isIOS(): boolean {
+        const iosBrowser = this.platformBrowser && /iPad|iPhone|iPod/.test(navigator.userAgent) && !('MSStream' in window);
```
rkaraivanov: Can't this just be a static property on the class itself?
Do we expect the this check to return a different value after script evaluation in the browser?

damyanpetev: Need instance to get the platform through DI before we can inspect user agent

wnvko: we finish with simple public fields

PR #5909, opened 2019-10-03
Improve caching by storing the igxForOf related scroll component

### LINKED ISSUE #5422: Reflow per rendered row
## Description  
Reflow occurs once per row render, rather than **once** per grid render. Since a grid tends to have many rows, reflow per row is impacting performance significantly.

There's various ways this could be achieved.
1. Considering that the horizontal scroll position is a property of the viewport of a grid, rather than an individual row. The scroll position should be externalized and provided to all rows, rather than each row individually querying the horizontal scroll position.
2. A momoized function, that is module (webpack module) scoped, could be used for obtaining the horizontal scroll position when rendering a row.

Related is #5078 which resolved the issue of reflow being caused multiple times when a single row is rendered.

 * igniteui-angular version: 7.3.9
 * browser: Chrome 75

## Steps to reproduce  

1. Create a new Angular app with igniteui-angular@7.3.9 as a dependency.
2. Add a basic grid
3. Use Chrome to performance profile the initial grid render.
4. Notice the purple blocks on Chrome Dev Tools Performance Tabs's flame chart.

## Result  
NonCachedHorizontalScrollAcrossRows

You can observe how reflow gets worse per row rendered, because the DOM node count and event handler count is increasing for each rendered row.

By the time the 12th row is rendered reflow is more costly than the DOM insertions for the row.

## Expected result  
No reflow per row rendered. At most reflow should occur once per grid render.

MayaKirova:
Closes #5422  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/for-of/base.helper.component.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/base.helper.component.ts
@@ -8,9 +8,6 @@ import {
 } from '@angular/core';
 
 export abstract class VirtualHelperBase implements OnDestroy {
-    @Output()
-    public onScrollEvent = new EventEmitter<Event>();
-
     public scrollAmount = 0;
 
     public _size = 0;
@@ -20,9 +17,6 @@ export abstract class VirtualHelperBase implements OnDestroy {
     @HostListener('scroll', ['$event'])
     onScroll(event) {
         this.scrollAmount = event.target.scrollTop || event.target.scrollLeft;
-        // TODO - Instead of every single igxForOf instance attaching a DOM event listener
-        // we could instaed subscribe to this onScrollEvent and remove related workaround logic in igxGridForOf.
-        this.onScrollEvent.emit(event);
     }
     constructor(public elementRef: ElementRef, public cdr: ChangeDetectorRef) { }
 
--- a/projects/igniteui-angular/src/lib/directives/for-of/base.helper.component.ts
+++ b/projects/igniteui-angular/src/lib/directives/for-of/base.helper.component.ts
@@ -1,0 +1,46 @@
+import {
+    EventEmitter,
+    Output,
+    HostListener,
+    ElementRef,
+    ChangeDetectorRef,
+    OnDestroy
+} from '@angular/core';
+
+export abstract class VirtualHelperBase implements OnDestroy {
+    @Output()
+    public onScrollEvent = new EventEmitter<Event>();
+
+    public scrollAmount = 0;
+
+    public _size = 0;
+
+    public destroyed;
+
+    @HostListener('scroll', ['$event'])
+    onScroll(event) {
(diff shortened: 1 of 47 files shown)


nrobakova:
Hi @MayaKirova,
I looked at this PR and found an issue:
When grid is scrolled down and change density/resize column/hide column/ reset data or do something that changes the grid size the grid layout is broken.
Steps to reproduce:
1. Open dev sample gridSummary
2. Scroll vertical to bottom
3. Click button Hide or enable paging or show toolbar or disable summary
image

projects/igniteui-angular/src/lib/directives/for-of/base.helper.component.ts
```diff
@@ -0,0 +1,42 @@
...
+    HostListener,
+    ElementRef,
+    ChangeDetectorRef,
+    OnDestroy
+} from '@angular/core';
+
+export abstract class VirtualHelperBase implements OnDestroy {
+    public scrollAmount = 0;
+
+    public _size = 0;
```
rkaraivanov: Probably better to make it a concrete class with a directive decorator mainly because of loss of metadata in
v9 with Ivy

vanslly:
I've compared 8.1.x with this PR and the difference looks promising:

From
image

To
image

No longer is the grid causing a reflow per row rendered which will be particularly good for customer's who have the grid occupy most of the viewport.

It seems like there's a lot of reflow instances that now happen before the grid starts rendering rows, and a few after the rows have been rendered; both are negatively impacting performance which I would like the team to backlog for future investigation at least:
1. I was expecting at least one reflow as the grid's viewport size is determined before rendering rows, however there are multiple reflows.
2. I wasn't expecting any reflows after the grid has rendered rows. One reason might be undesirable grid viewport resizing perhaps?

The reported issue is fixed by this PR - thank you.

ChronosSF:
@vanslly , we are investigating the remaining reflows. Our goal is to reduce them to their absolute minimum especially if the grid sizes are strictly set by the user (in which case there should be very few measurements required).

The result of this investigation will be additional issues that will appear in the repo in the coming week or so. I'll include you in the conversation about hose that may affect your app's performance so you can be in the loop.

skrustev:
@MayaKirova There's one small issue I found when pinning/unpinning columns, the content scrolls to left 0, while the scrollbar remains at the old position. It is reproduced only in this version.

Steps to reproduce:
1. Open the "Grid Column Pinning" demo
2. Scroll the unpinned area to the end.
3. Unpin the "Contact Title" column.
4. Observe the unpinned area and the scrollbar position. The unpinned area shouldn't be scrolled to the start.

MayaKirova:
@skrustev Please re-test.

PR #5919, opened 2019-10-04
Fixes issue with grouping by a column with a blank field

### LINKED ISSUE #5643: When group a template column it gets the name of column group before it
## Description  
When group a template column it gets the name of column group before it 

 * igniteui-angular version: 8.2
 * browser: n/a

## Steps to reproduce  

2. Move column group at the first position
3. Group by column for delete button

## Result  
In the grouping area the name of the group column is 'Column Group 1' and it cannot be sorted or removed

## Expected result  
The name for the grouped column should be empty or it should not be possible to group by a column which does not have defined field

## Attachments  
image

3phase:
Closes #5643   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
@@ -593,6 +593,9 @@ export class IgxGridComponent extends IgxGridBaseComponent implements IGridDataB
 	 * @memberof IgxGridComponent
      */
     public groupBy(expression: IGroupingExpression | Array<IGroupingExpression>): void {
+        if (this.checkIfNoField(expression)) {
+            return;
+        }
         this.endEdit(true);
         if (expression instanceof Array) {
             this._gridAPI.groupBy_multiple(expression);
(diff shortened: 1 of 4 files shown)


ChronosSF:
If you want a public function for the user to check if the grouping expressions he created and wants to use are valid, you should call it something like:

`isGroupingExpressionValid` and then have another , this time private method that calls it for each expression in the array for example.

But if you do that, you should also check if the field can be found in the columns array, because the expressions is hardly valid otherwise.

3phase:
> Field could also be `null` which would probably result in the same issue. I'd check for `!column.field`.
> 
> Also why is `checkIfNoField` public? It doesn't look like something the user would need to call for any reason.
> 
> I am not certain we need to sanitize the public groupBy function anyway. Are you sure you can literally send anything with this check in place and that will be enough for the grid to not get to an invalid state?

Thanks, requested changes are resolved now, reference the last two commits.

3phase:
> If you want a public function for the user to check if the grouping expressions he created and wants to use are valid, you should call it something like:
> 
> `isGroupingExpressionValid` and then have another , this time private method that calls it for each expression in the array for example.
> 
> But if you do that, you should also check if the field can be found in the columns array, because the expression is hardly valid otherwise.

I decided to make the newly introduced function private

Issue #5763, opened 2019-09-10
Dependant cell updates will not be refreshed in edit mode


RolandTa:
## Description  
The updated cell value will not be displayed in edit mode (cell.editValue).

 * igniteui-angular version: 8.1.4
 * browser: Google Chrome 76.0.3809.132  

## Steps to reproduce  
See https://stackblitz.com/edit/angular-1uuep6

Update the StockValue Column and Tab to the next cell OrderDate. The Order Date should display todays value. The value is properly updated within the rowData, but will not displayed if your directly jump to the edit mode of the cell. If you Tab back you see the update happen.

## Result  
A cell value update is not updated via runtime when directly switched into editmode of this cell. This happens for all update methods (via IgxGridCellComponent.update or IgxGridComponent.updateCell) or when directly changing the rowdata. Additionally it doesnt matter if I updated the value via a custom edit template change event (as in the stackblitz) or via the IgxGridComponent.onCellEdit event.

## Expected result  
Cell edit value should be properly updated for the current edit template

PR #6027, opened 2019-10-22
Call notifyChanges in updateRow instead of detectChanges

### LINKED ISSUE #5934: 8.2.* IgxGridComponent.updateRow poor performance
## Description  
After updating from 8.1.5 to 8.2.2, the performance of updateRow has degraded significantly , even if newer features are disabled such as advanced filtering.

 * igniteui-angular version: 8.2.2
 * browser: Chrome Version 73.0.3683.103 (Official Build) (32-bit)

## Steps to reproduce  
Click on the below links to view performance difference.

1. Create igx-grid with realtime updating rows via the updateRow function call

## Result  
The grid updates very slowly, with greatly increased time needed for each updateRow call.

## Expected result  
The grid should perform as it did in version 8.1.5

## Attachments  
igniteui-angular@8.1.5:  (grid updates normally)
    https://stackblitz.com/edit/angular-wkbmwm

igniteui-angular@8.2.2: (grid updates slowly)
    https://stackblitz.com/edit/angular-wkbmwm-vqzsup

wnvko:
We were calling `detectChanges` on each call to grid's `updateRow`. This made the grid unresponsive if `updateRow` was called very often. With this fix the grid will update when necessary without make the grid unresponsive.

Closes #5934 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.component.ts
```diff
@@ -3947,7 +3947,13 @@ export abstract class IgxGridBaseComponent extends DisplayDensityBase implements
...
             const row = new IgxRow(rowSelector, -1, this.gridAPI.getRowData(rowSelector));
             this.gridAPI.update_row(row, value);
-            this.cdr.detectChanges();
+
+            // TODO: fix for #5934 and probably break for #5763
+            // consider adding of third optional boolean parameter in updateRow.
+            // If developer set this parameter to true we should call notifyChanges(true), and
+            // vise-versa if developer set it to false we should call notifyChanges(false).
+            // The parameter should default to false
+            this.notifyChanges();
```
demingjs: Would UpdateCell benefit from this change as well?  I haven't tested specifically with that but see the detectChanges is being used there as well.

wnvko: @demingjs 
No, definitely no. This will break issue #5763.
There is a case when you may need to update a cell after another cell is changed, e.g. you have row with price, quantity and total. If user changes the value of let say quantity you would want to change the value of total by calling `updateCell`. If we do not call `detectChanges` at the end of `updateCell` total cell will not display updated value.
To summarize - if you need to update quickly the values of many cell, as in your case, you should call `updateRow`. If you need to change single cell you should call `updateCell`.

PR #3916, opened 2019-02-14
feat(igxGrid): Multi cell selection

rkaraivanov:
Closes #3915  
Related to #2903 

### Additional information (check all that apply):
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [x] All relevant tags have been applied to this PR
 - [ ] This PR includes unit tests covering all the new code
 - [ ] This PR includes API docs for newly added methods/properties
 - [ ] This PR includes `feature/README.MD` updates for the feature docs
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [ ] This PR includes `CHANGELOG.MD` updates for newly added functionality
 - [ ] This PR contains breaking changes
 - [ ] This PR includes `ng update` migrations for the breaking changes
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

PR #6056, opened 2019-10-25
perf(igx-grid): Reduce detection cycles in grid sizing logic

### LINKED ISSUE #6031: Increase rendering performance by circumventing the calcGridSizes last detectChanges call
## Description  
During Ivy development @rkaraivanov noticed that removing the `cdr.detectChanges()` call before the scroll checker increases rendering performance by a significant margin. We should invest time to determine if this breaks anything and possibly fix it. At a first glance the grid functions correctly but this detection cycle is most likely indirectly taken advanged of during tests and this is why a lot of them fail

rkaraivanov:
Cleaned up most of the DOM handlers to use `fromEvent` thus avoiding
manual removing of event handlers.
Use the resize observer as an actual observable, removing the extra grid
subject for proxying the calls.
Finally removed the `ngOnChanges` hook in the cell as it was obsolete
since #3767 and #3916.

Closes #6031

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


rkaraivanov:
@MayaKirova @ChronosSF 

I've removed the width = null behavior and the scrollbar checks in the sizing call. Let's see what can or will break and we can revert the changes if needed

zdrawku:
This PR will remain open, part of the suggestions were added with [this PR](https://github.com/IgniteUI/igniteui-angular/pull/6562) although the width null handling of the Grid is still not added.

PR #6106, opened 2019-11-01
fix(schematics): check polyfill in default project #6017

### LINKED ISSUE #6017: ng-add schematic fails to update polyfills in non-default workspaces
## Description  
In workspaces that do not conform to the default layout, `ng add` will fail due to not looking at the workspace file for the location of `polyfills.ts`.

 * igniteui-angular version: 8.2.4

## Steps to reproduce  

The simplest way to do this is create a workspace with Nx : 

`npx create-nx-workspace@latest ignite-repro`
choose Angular only, call the angular app whatever you like
`cd ignite-repro`
`ng add igniteui-angular`

## Result  

``` bash
? Enable polyfills for IE, Edge and Safari? Yes
? Add CSS library to reset HTML element styles across browsers? Yes
Cannot read property 'toString' of null
```

Due to this line which only expects there to be one file in a predictable place (this may not be true!) : 

https://github.com/IgniteUI/igniteui-angular/blob/07fb4881e3b487eae6bddacffa40dc340beef38c/projects/igniteui-angular/schematics/ng-add/index.ts#L53

Could be fixed with something like : 

```ts
      const workspace = getWorkspace(tree);
      Object.keys(workspace.projects).forEach(project => {
        const polyfillsFile = workspace.projects[project].sourceRoot + 'polyfills.ts';
        ...
      });
```

## Expected result  

The ng-add schematic should look at the `angular.json` file to determine where the applications are and should use that to figure out which `polyfills.ts` files to update (there may be more than one)

Lipata:
Closes #6017

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


Lipata:
You need to have this [fix ](https://github.com/IgniteUI/igniteui-cli/pull/591) in the CLI.

PR #6381, opened 2019-12-20
Remove additional initialization in the igxDrag to increase performance.

### LINKED ISSUE #6280: Improve the Drag directive initial rendering to be more performant.
## Is your request related to a problem? Please describe.
Initially when the igxDrag directive is initialized on AfterContentInit there are multiple force reflows that can be omitted without affecting the current behavior. Because of that when initialized on elements inside the igxGrid like columns, when they are scrolled the igxDrag is initialized for every column coming into view, which potentially slows down the scrolling.

The calculations related to the `_baseMarginLeft` and `_baseMarginTop` and other properties in the 
`ngAfterContentInit` method could possible be avoided without any side effects.

Other alternatives would be to use other types of methods that don't force reflows, but currently doesn't seem to be any.

skrustev:
Closes #6280

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


skrustev:
@rkaraivanov It seems like this will increase the performance only initially now, I took a look at it and I think Ivy no longer recreates templates on changes completely and now when the grid is scrolled the columns are not recreated as well.

PR #6383, opened 2019-12-20
fix(igxGrid): Making sure navigateTop/navigateBottom/goToFirstCell/go…

### LINKED ISSUE #4990: The igxGrid  keyboard navigation does not work when there are many collapsed groups at the bottom
## Description  
Navigating with Ctrl+ Arrow Up and then Ctrl + Arrow down does not work correctly when there is a large number of collapsed groups at the bottom of the grid.

 * browser: n/a

## Steps to reproduce  

1. Run the dev demos and open the GroupBy sample:  http://localhost:4200/gridGroupBy
2. Group by the "Address" column
3. Collapse the last 15 groups: everything below "Berkeley Gardens 12 Brewery" (it should be the last expanded group)
4. Select the "Consolidated Holdings" cell in that group
5. Press Ctrl + Arrow Up
6. Press Ctrl + Arrow Down

## Result  
Ctrl + Down does not scroll to the last visible cell - "Consolidated Holdings" in "Berkeley Gardens 12 Brewery". It scrolls to the bottom of the grid instead. Also, the focus is lost and the keyboard navigation stops working.

## Expected result  
Ctrl + Down should scroll to the last visible cell

## Attachments  

GroupByKeyboardNav

MayaKirova:
…ToLastCell target the first/last data row when scrolling.

Closes #4990   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -257,17 +259,31 @@ export class IgxGridNavigationService {
         }
     }
 
+    private findFirstDataRowIndex() {
+        const dv = this.grid.dataView;
+        return dv.findIndex(rec => !this.grid.isGroupByRecord(rec) && !this.grid.isDetailRecord(rec));
+    }
+
+    private findLastDataRowIndex() {
```
ChronosSF: While this function would not be called by a continuous key press making its performance less important, don't you think it would be a lot better implementing it like this:

```typescript
private findLastDataRowIndex() {
    let i = this.grid.dataView.length;
    while (i--) {
        const rec = this.grid.dataView[i];
        if (!this.grid.isGroupByRecord(rec) && !this.grid.isDetailRecord(rec)) {
             return i;
        }
    }
}
```
In this way we avoid the reverse call and the additional index search in the original array, which in the most common case will have to iterate it again fully.

Unfortunately, ES still doesn't define reverse find functions but you could also create one if you want this particular code to be clean.

PR #6444, opened 2020-01-10
fix(igxHGrid): Make sure visibleColumnIndex is not greater than columns l…

### LINKED ISSUE #5126: Keyboard navigation is not working when going from child to parent where parent has less columns.
## Description  
Navigating from child grid to parent grid where the child has more columns than the parent results in error thrown.

 * browser: any

## Steps to reproduce  

1. Open the "Hierarchical Grid" demo and use second grid
2. Expand the first row
3. Click on the second child's last cell to the right
4. Hold Arrow Down and try to navigate to the parent grid.

## Result  

## Expected result  
`ERROR TypeError: Cannot read property 'pinned' of undefined`

## Attachments  
Issue#5126

mddragnev:
…ength #5126

Closes #5126   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
@@ -609,7 +609,10 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
             if (nextIsSiblingChild) {
                 this.focusNextChild(next, visibleColumnIndex, nextParentGrid);
             } else {
-                this.focusNextRow(next, visibleColumnIndex, grid || nextParentGrid);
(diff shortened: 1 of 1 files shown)


ChronosSF:
The issue still reproduces if navigating **up** instead of down. Check if you can apply code similar to the one that prevents the bug from occuring when navigating between children of the same level:

```typescript
private focusNextChild(elem, visibleColumnIndex, grid) {
...
        // Update column index since the next child can have in general less columns than visibleColumnIndex value.
        const lastCellIndex = childGrid.unpinnedColumns[childGrid.unpinnedColumns.length - 1].visibleIndex;
        visibleColumnIndex = Math.min(lastCellIndex, visibleColumnIndex);
    }
```

PR #6481, opened 2020-01-15
Combo - properly update FormControl touched

### LINKED ISSUE #6475: Combo doesn't update FormControl.touched appropriately
## Description  
Similar as #6370, FormControl.touched not updated on focus/blur per `ControlValueAccessor` implementation. Since the FormControl manages value updates internally, setting `updateOn: 'blur'` without updating `touched`, the control never updates and value is never committed to the bound model.

 * igniteui-angular version: 8.2.x

## Steps to reproduce  

1. Run https://stackblitz.com/edit/angular-t7rmew
4. Focus the combo and observe the `touched` state
2. Select something in the combo
3. Focus out and observe the model `touched` state and value

## Result  
Form control touched should be updated and when updating on blur the bound model value should update and the model validity should reflect that.

## Expected result  
Touched is never set and the model value is never updated, causing the combo to be out of sync with the model - combo with selection and empty invalid model:
image

damyanpetev:
Closes #6475

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/combo/combo.component.html
+++ b/projects/igniteui-angular/src/lib/combo/combo.component.html
@@ -9,11 +9,11 @@
         <ng-content select="igx-hint, [igxHint]"></ng-content>
     </ng-container>
     <input igxInput #comboInput name="comboInput" type="text" [value]="value" readonly [attr.placeholder]="placeholder"
-        [disabled]="disabled" (blur)="onBlur()" />
+        [disabled]="disabled" (blur)="onBlur()" (focus)="onFocus()"/>
     <ng-container ngProjectAs="igx-suffix">
         <ng-content select="igx-suffix"></ng-content>
     </ng-container>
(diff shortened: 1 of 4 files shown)


projects/igniteui-angular/src/lib/combo/combo-dropdown.component.ts
```diff
@@ -87,6 +87,7 @@ export class IgxComboDropDownComponent extends IgxDropDownComponent implements I
     public navigatePrev() {
         if (this._focusedItem && this._focusedItem.index === 0 && this.virtDir.state.startIndex === 0) {
             this.combo.focusSearchInput(false);
+            this.toggle();
```
ViktorSlavov: This will lead to the drop-down closing when navigating from the 1st list item w/ `ArrowUp`.
The toggle removed from the `focusSearchInput` is wrapped in a second `if`, so moving it here would change the combo's behavior.

PR #6496, opened 2020-01-20
Expose onAppended event

### LINKED ISSUE #6472: Dropdown is not scrolled to the selected item.
## Description  
When select an item and open the drop down again, the drop down list is not scrolled to the selected item. 

In more detail, the dropdown scrolling is executed ok and the selected item is scrolled to. The actual issue is after attaching the dropdown in the overlay hence the scrolling is not persisted.

 * igniteui-angular version:  9.0
 * browser: n/a

## Steps to reproduce  

1. Open a dropdown with enough items so there is a scroll existing.
2. Scroll down to an initially non-visible item and select it.
3. Close & Reopen 

## Result  
The scroll is at the top. 

## Expected result  
The selected item should be visible and the items container scrolled down.

##
Related https://github.com/IgniteUI/igniteui-angular/issues/6459

IvayloG:
`onAppended` event is added to `igxOverlay` and `igxToggle`. This event will fire after the content, to be shown in the overlay, is moved in the DOM from its original position to the overlay.

This event allows developer to perform scrolling and other actions on the content before animation starts.

Closes #6472 

 - [x] Bug fix
 - [ ] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.ts
```diff
@@ -164,6 +165,25 @@ export class IgxToggleDirective implements IToggleView, OnInit, OnDestroy {
         @Optional() private navigationService: IgxNavigationService) {
     }
 
+    /**
+     * Emits an event after the toggle container content is positioned.
+     *
+     * ```typescript
+     * onPositioned(event) {
```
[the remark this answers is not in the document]
IvayloG: Yes it does. How do you feel about removing the same for both (onOpened)='onToggleOpened($event)' / onToggleOpened(event)  AND 
(onClosed)='onToggleClosed($event)'    /  onToggleClosed(event)
in this PR?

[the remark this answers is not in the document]
ViktorSlavov: I think it would be better to do those in a separate PR, as they *might* affect something else.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -304,6 +314,8 @@ export class IgxOverlayService implements OnDestroy {
                 true);
             info.settings.scrollStrategy.initialize(this._document, this, info.id);
             info.settings.scrollStrategy.attach();
+
+            this.onPositioned.emit({ id: info.id, componentRef: info.componentRef });
```
damyanpetev: Umm if this is `onPositioned` it would also make sense for it to emit on `reposition()`?
That might defeat the goal of the event to use as 'before show' but after 'opening/move' though.

IvayloG: Yes it does!
Alternativelly, we can rename this one to onInitiallyPositioned. However this may suggest the need of an non-initial event as well.

damyanpetev: Whelp it feels like it overlaps way too much with the positioning strategies. `onInitial` or emit argument also seem odd, we can just name after what it's used for - content move or something

projects/igniteui-angular/src/lib/directives/toggle/toggle.directive.ts
```diff
@@ -164,6 +165,25 @@ export class IgxToggleDirective implements IToggleView, OnInit, OnDestroy {
         @Optional() private navigationService: IgxNavigationService) {
     }
 
+    /**
+     * Emits an event after the toggle container content is appended.
```
damyanpetev:
```suggestion
     * Emits an event after the toggle element is appended to the overlay container.
```

PR #6503, opened 2020-01-21
refactor(*): make type enumerators more consistent across components

### LINKED ISSUE #854: Consistency in naming component "types"
### Description  
To have a consistent way of defining type for the components/directives, based on the way we create themes.

### Current state
**igx-badge API:**

type | string | Set the badge type to either **default, info, success, warning, or error**. Depending on the type, specific background color is set as declared in the default theme.
-- | -- | --

**igx-linear-bar API:**

type | string | Set type of the linear bar. Possible options - **default, success, info, warning, and danger**.
-- | -- | --

**igx-circular-bar** does not have a type in the API.

**igxButton** does not have type in the API, and can only be modified though the theming engine.

### Expectations

Our palettes accept arguments for **primary**, **secondary**, **info**, **success**, **warn**, and **error** colors. 
I would expect to get a "default" without explicitly setting it, so a "default" option seems redundant. I'd expect to receive the theming colors as options - namely: primary, secondary, info, success, and warn for each of the already mentioned components/directives.

- [x] remove "default" options.
- [x] rename (possibly add, to avoid breaking changes) the 'danger' type to 'error'
- [x] ~~add 'primary' and 'secondary' options to igx-linear-bar~~
- [x] ~~add 'primary' and 'secondary' options to igx-badge~~
- [x] ~~add type to igxButton~~
- [x] ~~add type to igx-circular-bar~~

simeonoff:
Closes #854

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [x] This PR includes `ng update` migrations for the breaking changes


projects/igniteui-angular/migrations/update-9_0_0/index.spec.ts
```diff
@@ -43,4 +43,32 @@ describe('Update 9.0.0', () => {
             `);
         done();
     });
+
+    it('should update Enum names.', done => {
+        appTree.create(
+            '/testSrc/appPrefix/component/enum.component.ts',
+            `import { AvatarType, Size, Type, SliderType } from 'igniteui-angular';
```
damyanpetev: Slight concern with the default migration for identifiers (class names, enums, etc).
Never assumed we'd be renaming something that can collide with the user's or 3rd party naming (unless the user manually created a shadowing `Igx-` class.
What I'm getting to is these might end up replacing an unrelated `Size` class for example as the migrations look for the identifiers globally.

Just FYI. If we think this may be too risky, let me know and I can log an issue and see if I can filter identifiers with import origin as well.

simeonoff: Now that you say that, it will be risky. Type and Size in particular could very well be defined in an app. Let's make sure that only imports from igniteui-angular get renamed.

simeonoff:
> Fairly certain there should be a sizable CHANGELOG entry for the changes, no?
> Also some component README-s list input properties with the enum type (saw `TabsType` for sure) and the slider has an API section for the `SliderType`

Thanks, introduced all those changes with the latest commits.

projects/igniteui-angular/src/lib/avatar/avatar.component.ts
```diff
@@ -272,20 +271,18 @@ export class IgxAvatarComponent implements OnInit, AfterViewInit {
      *
      * @memberof IgxAvatarComponent
      */
-    get type(): AvatarType {
+    get type(): IgxAvatarType {
```
damyanpetev: Missing the forth option in the enum will return `undefined` here

projects/igniteui-angular/src/lib/badge/badge.component.ts
```diff
@@ -60,7 +59,7 @@ export class IgxBadgeComponent {
     * ```
     */
     @Input()
-    public type: string | Type = 'default';
+    public type: string | IgxBadgeType = '';
```
damyanpetev: Just noticed the API doc on top (can't comment there) - but it does say "Allowed values are `default`...."

simeonoff: yep, updated.

projects/igniteui-angular/src/lib/avatar/avatar.component.ts
simeonoff: Added custom to remedy for that.

projects/igniteui-angular/migrations/update-9_0_0/index.spec.ts
MayaKirova: @damyanpetev If we don't filter based on the import origin  the "Type" migration will replace one of angular core's classes, which we use in the angular-samples repo:
import { Type } from "@angular/core";
So with the current migration the samples would be broken.
Perhaps the migration logic should be updated first (if possible) before we merge this PR.

damyanpetev: Ah so it's actually very likely these to clash with something. Will log an addition for the base migrations to filter out just our imports before this can be merged. Thanks.

PR #6547, opened 2020-01-27
Updates docs for the igxAvatarComponent API

3phase:
 - [ ] Bug fix
 - [ ] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/avatar/avatar.component.ts
```diff
@@ -164,8 +182,9 @@ export class IgxAvatarComponent implements OnInit, AfterViewInit {
     public roundShape = false;
 
     /**
-     * Sets the avatar's `initials`/`icon` color.
+     * Sets the color of the avatar's initials' icon.
```
damyanpetev: initials and icon are 2 separate types of avatar

projects/igniteui-angular/src/lib/avatar/avatar.component.ts
```diff
@@ -82,8 +94,9 @@ export class IgxAvatarComponent implements OnInit, AfterViewInit {
     protected iconTemplate: TemplateRef<any>;
 
     /**
-     * Returns the `aria-label` of the avatar.
+     * Returns the aria-label attribute of the avatar.
```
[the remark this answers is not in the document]
3phase: Not really. There was too many of the code blocks and I was following the pattern of the changes, applied beyond. I'd probably encircle all the attributes with `s again though.

3phase:
> @3phase Also (new addition to the wiki) - can you drop all `@memberof IgxAvatarComponent` from the property descriptions.

Done, thanks

projects/igniteui-angular/src/lib/avatar/avatar.component.ts
```diff
@@ -227,6 +213,7 @@ export class IgxAvatarComponent implements OnInit, AfterViewInit {
     @Input()
     public src: string;
 
+    private _size: string | Size = 'small';
```
rkaraivanov: Maybe we can collapse the union just to `Size` type

3phase: It's done

damyanpetev:
@3phase @rkaraivanov Restored the removed size test asset as it does test a valid code branch. Plus, there's no type safety for `@Input`.

Also added a friendly name to the `src` attribute per @SAndreeva's suggestion. Should we also add a friendly name for the component itself?

PS: [Wiki section on meta tags ](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines#metadata-tags )has been expanded

damyanpetev:
@rkaraivanov Slight issue with the simplification of the enum `@Input`-s. In TypeScript, unlike numeric values 🤷‍♂, string enum values are intentionally opaque (see https://github.com/microsoft/TypeScript/issues/17690#issuecomment-321319291), so this could break existing code that assigned strings to those props.

Also forgot there is also type check option ([strictTemplates](https://angular.io/guide/angular-compiler-options#stricttemplates)), and since assigning an actual enum in templates is rather cumbersome, the common way of using the value (` size="medium" `) will also fail type checks. Related: https://github.com/angular/angular/issues/35363

---
_We might have a misalignment of what we're trying to get out of enums and what they're intended to do. The point is to always refer to the public API (the enum members) so we avoid repeating strings (that are supposed to be free to change), yet we are very much stuck referring to the actual values for templates - kind of like the cake and eating it deal._

ChronosSF:
@damyanpetev , @rkaraivanov , could you please reach an agreement as to what to do with this change?

damyanpetev:
@ChronosSF I'd air on the side of caution and not break the current type.
Will continue to monitor the feature suggestion in Angular, though I doubt there's much that can be done with enums that are not meant to be interchangeable with string values. Might be better served with a string union type, but we'll need to look into that and definitely not the job of this PR :)

3phase:
@damyanpetev @ChronosSF @rkaraivanov the type of the avatar size is returned to its initial state now. Are there any other things to do?

PR #6554, opened 2020-01-28
Determine initial min-max value

### LINKED ISSUE #6494: igx-timepicker  In dropdown mode if we click cancel button without selecting any value null values are displaying
## Description  
In Dropdown mode if we click cancel button without selecting any value null values are displaying and after that can't select any time.

 * igniteui-angular version: 4.3.3
 * browser: Chrome

## Steps to reproduce  
Active dropdown mode and give min and max values
1. open the timepicker in dropdown mode
2.  click on cancel button and notice the input field value, null will display
3. again clear field and try to select the time again still it will display null.

## Result  
on cancel, time picker taking null values

## Expected result  
on cancel, time picker shouldn't take any values.

## Attachments  

jackofdiamond5:
Closes #6494

When the time picker has `minValue` and `maxValue` properties set and is bound to a date object that does not satisfy them, it does not set the internal value and that later on causes the wrong text to be displayed in the editor.

~~So now we determine whether we should use the min or max value for the editor, based on the value that is received through the model and based on whether `minValue` and `maxValue` are set.~~

The same behaviour can be seen with a `null` value passed in to the model. Now the editor will not throw any errors if you have no value or if the value is not within the valid range. This also prevents it from displaying incorrect text.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -139,6 +139,14 @@ export class IgxTimePickerComponent implements
             };
             this.onValueChanged.emit(args);
         } else {
+            const minTime = this.minValue ? this._convertMinMaxValue(this.minValue) : null;
```
damyanpetev: This looks like similar checks like `_isValueValid`, no? The issue with this is not in the value setter, outside/clear click doesn't fail like this going trough here as well.

Where it goes wrong is a combination of the `_onDropDownClosed` logic attempting to convert empty value (resulting in an Invalid Date) that passes as valid as is set to the editor, and likely the cancel action creating that empty value

projects/igniteui-angular/src/lib/time-picker/time-picker.component.spec.ts
```diff
@@ -183,6 +183,28 @@ describe('IgxTimePicker', () => {
...
+        UIInteractions.clickElement(timePickerTarget);
+        tick();
+        fixture.detectChanges();
+
+        const cancelButton = dom.query(By.css('.igx-button--flat'));
+        UIInteractions.clickElement(cancelButton);
+        tick();
+        fixture.detectChanges();
+
+        expect(timePicker.onValidationFailed.emit).not.toHaveBeenCalled();
```
damyanpetev: Shouldn't this also test what the editor is displaying since that's the main issue?

projects/igniteui-angular/src/lib/time-picker/time-picker.component.ts
```diff
@@ -1792,7 +1795,7 @@ export class IgxTimePickerComponent implements
      * ```
      */
     public cancelButtonClick(): void {
-        if (this.mode === InteractionMode.DropDown) {
+        if (this.mode === InteractionMode.DropDown && this.value) {
```
damyanpetev: Yeah, this doesn't quite do it - if there's no initial value(null, empty, etc), opening the drop-down will auto-populate the display with the currently selected time and this won't allow it to be cleared on cancel and still trigger invalid event for 4AM?

PR #6571, opened 2020-01-30
Move the label position in border-type input groups

### LINKED ISSUE #4535: Wrong label position for input with type "Border"
## Description  
According material.io the labels for input types "border" must not be inside the input, they should be at the middle of the border. 

You may see material-component implementation [here](https://material-components.github.io/material-components-web-catalog/#/component/select) (see the outlined demo).

 * igniteui-angular version: ^7.2.4
 * browser: Chrome

## Steps to reproduce  

1. Open your demo [here](https://stackblitz.com/angular/keropxgornk)
2. Go to the input named with "Border Type" and write some value...

## Result  
The label is moved at position -50% and scaled to 75%, like this:
image

## Expected result  
Label should be positioned at the middle of the border like this:
image

simeonoff:
Closes #4535 

Positions the label in border-type input groups to match the material design spec more closely. 

The label now requires background to be set when preparing custom themes. The default label background in border-type input groups is set to the surface color from the used palette. If the background in your app is set to the surface color from the palette the label background and the app background will not collide.

 - [ ] Bug fix
 - [x] New functionality
 - [x] Documentation
 - [x] Demos
 - [ ] CI/CD

### Checklist:


SisIvanova:
The background property for disabled input doesn't work.

simeonoff:
> The background property for disabled input doesn't work.

This has been fixed with the latest commit.

PR #6605, opened 2020-02-04
refactor(migrations): limit identifier rename from package imports

### LINKED ISSUE #6561: Migrations for renamed identifiers (class/enum) should only apply to imports from 'igniteui-angular'
Currently the [migrations for renamed Identifiers(Class)](https://github.com/IgniteUI/igniteui-angular/wiki/Update-Migrations#using-the-updatechanges-class) look for [identifiers](https://github.com/IgniteUI/igniteui-angular/wiki/Update-Migrations#class-rename) globally in the TypeScript source files. With the assumption we'd be replacing `Igx<Name>` to `Igx<NewName>`, the matches are not filtered based on context info (such as import package).

However, per the discussion in https://github.com/IgniteUI/igniteui-angular/pull/6503#discussion_r371847746, that will not be enough.

The built-in migrations functionality should consider only matches imported from 'igniteui-angular' to avoid potential naming conflicts.

damyanpetev:
Resolve initial identifier match only in our import statements -
was previously using a general identifier search w/o context.
Move to TypeScript's Language service to search for rename positions.

Closes #6561

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -57,3 +60,99 @@ export function getImportModulePositions(sourceText: string, startsWith: string)
...
+            return filePaths.indexOf(fileName) !== -1;
+        }
+    };
+    return ts.createLanguageService(servicesHost, ts.createDocumentRegistry());
+}
+
+function patchHostOverwrite(host: Tree, fileVersions: Map<string, number>) {
+    const original = host.overwrite;
+    host.overwrite = (path: string, content: Buffer | string) => {
+        if (!fileVersions.has(path)) {
```
ViktorSlavov: Very nit picky, especially after the black magic you're doing above:
Since `fileVersions` are entirely handled here, you can do something like:
```
const version = fileVersions.get(path) || 0;
if (!version) {
...
} else {
fileVersions.set(path, version + 1);
}
```
Just to avoid having to do two operations on the `Map`

PR #6769, opened 2020-02-18
Enhances pinning sample and fixes right pinning offset

3phase:
The cell into the column template component relies on the `rightPinnedOffset` method, which can cause trouble while calculating the offset of the column while a left pinning is in action, because it is dedicated to correctly displaying the right pinning.
Furthermore, I made a slight enhancement to the sample showing pinning, by adding an ability to switch between left and right column pinning at runtime.

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [x] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/columns/column.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/columns/column.component.ts
@@ -1003,9 +1003,9 @@ export class IgxColumnComponent implements AfterContentInit {
     }
 
     get rightPinnedOffset(): string {
-        return this.pinned ?
+        return (this.pinned && !this.grid.isPinningToStart) ?
             - this.grid.pinnedWidth + 'px' :
-            null;
+            this.grid.isPinningToStart ? 0 + 'px' : null;
     }
 
     get gridRowSpan(): number {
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1003,9 +1003,9 @@ export class IgxColumnComponent implements AfterContentInit {
     }
 
     get rightPinnedOffset(): string {
-        return this.pinned ?
+        return (this.pinned && !this.grid.isPinningToStart) ?
             - this.grid.pinnedWidth + 'px' :
-            null;
+            this.grid.isPinningToStart ? 0 + 'px' : null;
```
dkamburov: @3phase I think if pinning is to start, it would be better to leave it null, it is not necessary to apply the style left 0px for the cells that don't need it

dkamburov:
@3phase  and @ChronosSF should we close this as https://github.com/IgniteUI/igniteui-angular/pull/6775#pullrequestreview-361125678 is already merged?

Issue #6640, opened 2020-02-10
Row Pinning IgxGrid


radomirchev:
## Describe the solution you'd like
I would like to be able to pin/freeze rows on top/bottom of IgxGrid.

Sub-Tasks:

- [x] Basic implementation and API -> https://github.com/IgniteUI/igniteui-angular/pull/6848
- [x] Default UI – https://github.com/IgniteUI/igniteui-angular/pull/6984
- [x] TreeGrid - PB
- [x] HierarhicalGrid - SK

Ghost row changes tasks:
-	[x] Flat Grid Ghost rows - #7104
- 	[x] Excel exporter (change according to spec) (all grids) -> МК (https://github.com/IgniteUI/igniteui-angular/pull/7069)
-	[x] GroupBy (flat grid)-> MD
-	[x] Paging (all grids) -> SS
-	[x] `IgxPinnedIndicatorTemplateDirective` implementation - #7104
-	[x] getSelectedData test/fix for all grid types - #6979
-	[x] Default 'Pinned' chip to appear regardless of first column template - #7104
-	[x] Default 'Pinned' label in chip localization - #7104
-	[x] Search integration (all grids) 
-	[x] Add aria-disabled prop for ghost rows -> SK #7104
- 	[x] Ghost row selection on space, shift selection?

Feature Integrations:

-	[x] Updating -> https://github.com/IgniteUI/igniteui-angular/pull/6907
-	[x] Hiding -> https://github.com/IgniteUI/igniteui-angular/pull/6967
-	[x] Filtering -> https://github.com/IgniteUI/igniteui-angular/pull/6884
-	[x] Sorting -> https://github.com/IgniteUI/igniteui-angular/pull/6878
-	[x] Summaries -> #6942
-	[x] Master-Details -> #6909 
-	[x] MRL -> #6924
-	[x] Search -> #6919
- 	[x] State persistence directive -> https://github.com/IgniteUI/igniteui-angular/pull/6968
-  	[x] Navigation - should be resolved by https://github.com/IgniteUI/igniteui-angular/pull/6910

Docs and Samples
-	[x] Update Changelog -> https://github.com/IgniteUI/igniteui-angular/pull/6930
-	[x] Documentation – new topic. - https://github.com/IgniteUI/igniteui-docfx/issues/1791
-	[x] Samples 
    - [x] Potential sample with drag & drop of rows between pin/unpin area. - #1870
[...]

PR #6848, opened 2020-02-28
Row pinning base

MayaKirova:
Related to #6640 
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid/grid.component.html
```diff
@@ -101,16 +101,28 @@
...
         (onDragScroll)="dragScroll($event)" [igxGridDragSelect]="selectionService.dragMode"
-        [style.height.px]='calcHeight' [style.width.px]='calcWidth || null' #tbody>
+        [style.height.px]='totalHeight' [style.width.px]='calcWidth || null' #tbody>
         <span *ngIf="hasMovableColumns && draggedColumn && pinnedColumns.length <= 0"
             [igxColumnMovingDrop]="headerContainer" [attr.droppable]="true" id="left"
             class="igx-grid__scroll-on-drag-left"></span>
         <span *ngIf="hasMovableColumns && draggedColumn && pinnedColumns.length > 0"
             [igxColumnMovingDrop]="headerContainer" [attr.droppable]="true" id="left"
             class="igx-grid__scroll-on-drag-pinned" [style.left.px]="pinnedWidth"></span>
+        <div #pinContainer *ngIf='pinnedRecords.length > 0 && isRowPinningToTop' class='igx-grid__tr--pinned igx-grid__tr--pinned-top'>
```
ChronosSF: Is it possible to have both the top and bottom pinning use the same template so that there is less code repetition?

3phase:
When one attempts to pin a row with a non-existing row ID into the `pinnedRecords` collection, instead of nothing happening, an `undefined` is being pushed into the collection, which might result in errors while rendering the template.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -1375,6 +1377,16 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
...
 
+    /**
+     * Emitted when the pinned state of a row is changed.
+     * @example
+     * ```html
+     * <igx-grid [data]="employeeData" (onRowPinning)="rowPin($event)" [autoGenerate]="true"></igx-grid>
+     * ```
+     */
+    @Output()
+    public onRowPinning = new EventEmitter<IPinRowEventArgs>();
```
damyanpetev: If it's an `-ing` event, is it supposed to be cancellable?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3990,6 +4022,88 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
...
+     * @example
+     * ```typescript
+     * this.grid.pinRow(rowID);
+     * ```
+     * @param rowID The row id - primaryKey value or the data record instance.
+     * @param index The index at which to insert the row in the pinned collection.
+     */
+    public pinRow(rowID, index?): boolean {
+        const rec = this.gridAPI.get_rec_by_id(rowID);
+        if (this.pinnedRecords.indexOf(rec) !== -1 || !rec || this.data.indexOf(rec) === -1) {
```
damyanpetev: What's the point behind the `this.data.indexOf(rec) === -1` check? The `gridAPI.get_rec_by_id()` call should've already checked that and it handles transactions as well.
`if (!rec || this.pinnedRecords.indexOf(rec) !== -1)` would make more sense as a check.

MayaKirova: @damyanpetev The rowID can be either the primary key value (if primaryKey exists) or the record instance itself.
In case it is a record  get_rec_by_id does not do any additional checks to see if the passed record exists in the data or not - it simply returns the record instance itself. Hence the additional check.

At this time we don't want to allow pinning data that does not exist in the grid. We will mark this as a limitation for scenario where the user may want to pin something that is not in the data (like in remote virtualization scenarios for example).

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
MayaKirova: @damyanpetev Wasn't planned to be cancellable according to the spec, however I see no reason why it can't be. @ChronosSF what do you think?

skrustev:
There seems to be an issue when grid has `height` set to `null` and the user tries to pin a row. The rowData disappears after pinning.

Steps to reproduce:
1. Run the attached sample.
2. Pin a row.

[igxGrid-RowPinning.zip](https://github.com/IgniteUI/igniteui-angular/files/4319073/igxGrid-RowPinning.zip)

projects/igniteui-angular/src/lib/grids/grid/grid.component.html
```diff
@@ -101,16 +101,32 @@
...
+        [style.height.px]='totalHeight' [style.width.px]='calcWidth || null' #tbody>
         <span *ngIf="hasMovableColumns && draggedColumn && pinnedColumns.length <= 0"
             [igxColumnMovingDrop]="headerContainer" [attr.droppable]="true" id="left"
             class="igx-grid__scroll-on-drag-left"></span>
         <span *ngIf="hasMovableColumns && draggedColumn && pinnedColumns.length > 0"
             [igxColumnMovingDrop]="headerContainer" [attr.droppable]="true" id="left"
             class="igx-grid__scroll-on-drag-pinned" [style.left.px]="pinnedWidth"></span>
+        <ng-template #pinnedRecordsTemplate>
+            <ng-container *ngFor="let rowData of pinnedRecords
+                | gridTransaction:id:pipeTrigger
```
damyanpetev: Umm... fairly sure this will not play well with pinned rows since the transaction pipe will blindly add all the new records here.

If those are the records directly stored (even if it was a record created by transaction) it'd be in the `pinnedRecords` anyway, no?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -2435,6 +2461,12 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
     protected _columnPinning = false;
 
 
+    /**
+     * @hidden
+    */
+    public pinnedRecords = [];
```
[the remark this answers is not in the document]
MayaKirova: To make it consistent we'll add a pinnedRows getter, which will return the pinned row instances.

MayaKirova:
@skrustev please re-test.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
ChronosSF: If the corresponding column event is not cancellable this shouldn't be either.

In this case, create a feature-request issue to have them both cancellable separately from this PR

projects/igniteui-angular/src/lib/grids/row.directive.ts
```diff
@@ -58,6 +58,30 @@ export class IgxRowDirective<T extends IgxGridBaseDirective & GridType> implemen
...
+        return this.grid.pinnedRecords.indexOf(this.rowData) !== -1;
+    }
+    /**
+     * Sets whether the row is pinned.
+     * Default value is `false`.
+     * ```typescript
+     * this.grid.selectedRows[0].pinned = true;
+     * ```
+     */
+    public set pinned(value: boolean) {
```
damyanpetev: Not sure this is needed, considering `row.grid` (and thus its API) is public + you have the shorthand pin methods.
Not a big deal though.

MayaKirova: I assume it was added in the specification so that it's the same as the column, although its functionality overlaps with the pin/unpin methods. @ChronosSF let me know if you still think it is necessary.

damyanpetev: I was specifically looking at the spec for this, but I can't see anything for the row API additions.

MayaKirova: While there's no dedicated section for Row API methods, there is a mention under 3.3. Developer Experience:

"Similar to the column counterparts, the row directive has pin and unpin methods that are used under the hood, as well as pinned getter/setter for maximum flexibility to the user. "

damyanpetev: Fair enough, still feels a bit redundant since the column is declarative and the row isn't, but one more shorthand won't hurt much.

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
```diff
@@ -826,7 +826,11 @@ export class IgxGridComponent extends IgxGridBaseDirective implements GridType,
     /**
     * @hidden @internal
     */
-    public getContext(rowData, rowIndex): any {
+    public getContext(rowData: any, rowIndex: number, pinned?: boolean): any {
+        if (pinned && !this.isRowPinningToTop) {
```
damyanpetev: @MayaKirova @ChronosSF This should definitely be noted in the feature documentation, that pinned rows will affect the start index of normal rows or are affected by the length of the data view when on the bottom.

projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
```diff
@@ -152,3 +152,26 @@ export class IgxGridFilteringPipe implements PipeTransform {
         return result;
     }
 }
+
+
+@Pipe({
```
damyanpetev: This should be hidden to avoid it showing up in the API docs

PR #6919, opened 2020-03-17
Row pinning and search integration

VMihalkov:
Related to #6640

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@ChronosSF @VMihalkov  filteredSortedData is not correct when row is pinned initially (for example on afterViewInit) since the pinnedRows collection is not available yet. This can be amended by checking the _pinnedRecordIDs collection instead or using the hasPinnedRecords getter). 
Order is also incorrect if row position is set to RowPinningPosition.Bottom as the pinned rows are always added first in the collection.

VMihalkov:
> @ChronosSF @VMihalkov filteredSortedData is not correct when row is pinned initially (for example on afterViewInit) since the pinnedRows collection is not available yet. This can be amended by checking the _pinnedRecordIDs collection instead or using the hasPinnedRecords getter).
> Order is also incorrect if row position is set to RowPinningPosition.Bottom as the pinned rows are always added first in the collection.

@MayaKirova I tested with RowPinningPosition.Bottom and again the unpinned collection is always going first through the sort pipe, so the order seems to be correct. It works with both Top and Bottom row pinning. I only replaced pinnedRows with _pinnedRecordIDs.

MayaKirova:
@VMihalkov When row is pinned to bottom it actually becomes the last row in the grid, not the first. This is reflected in the row's index. So when findNext is used activation should move to through the unpinned row first, before reaching the pinned row. Currently it moves first to the pinned row, which is incorrect.

For example, if findNext('a') is called and RowPinningPosition.Bottom is used:
image

Should activate "Ana", not 'Alfred'.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -2935,6 +2937,18 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
         this.setupColumns();
     }
 
+    public setFilteredSortedData(data, pinned: boolean) {
```
ChronosSF: this should have @hidden and @Internal otherwise it'll show in the API Docs

PR #6935, opened 2020-03-19
Focusless kb navigation hierarchical-grid

MayaKirova:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


dkamburov:
Alt + ArrowLeft/ArrowRight seems to be not prevented, because the browser is navigated to previous or next page:
alt+arrows
Ctrl + Home is not working as expected when it needs to scroll horizontally:
ctrl+home
First active cell when navigating to a child is not being selected(ArrowDown is used for navigation):
noselection
Not sure if the next one is related to kb nav, but clicking on expand indicator seems to activate the first cell of the grid:
firstactive

MayaKirova:
1) Issue with Alt + ArrowLeft/ArrowRight should be fixed now.
2) Issue with Ctrl+Home
Cannot reproduce. Please share more detailed steps or if it reproduces on a particular browser.
I have tried with/without expanded children but both work:
ctrl+home

MayaKirova:
3. Issue with ArrowDown. 
Cannot reproduce. Please share more detailed steps or if it reproduces on a particular browser.
arrowdown-test

MayaKirova:
4. Issue with first cell being activated on expander click should be fixed now.

dkamburov:
2. I use the sample sample with Chrome without expanding the children 
firstactive
3. You are actually reproducing it only for the first cell in the child

MayaKirova:
Regarding 2) It appears to reproduce only when browser is zoomed. Happens in base grid as well. Should be logged and resolved in base grid implementation.

MayaKirova:
3) Issue with selection not moving with focus should be fixed now.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -836,6 +836,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
     public activate(event: FocusEvent) {
         const node = this.selectionNode;
         const shouldEmitSelection = !this.selectionService.isActiveNode(node);
+        this.grid.navigation.activeNode = this.selectionNode;
```
ddincheva: Due to some differences in selection node and active node in the multi row layout can omit this assignment or move to to the hierarchical cell activate() method?!

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -2,630 +2,339 @@ import { IgxGridNavigationService } from '../grid-navigation.service';
 import { IgxHierarchicalGridComponent } from './hierarchical-grid.component';
 import { first } from 'rxjs/operators';
 import { ISelectionNode } from '../selection/selection.service';
-import { isIE } from '../../core/utils';
+import { isIE, SUPPORTED_KEYS, NAVIGATION_KEYS } from '../../core/utils';
 import { FilterMode } from '../common/enums';
 import { IgxColumnComponent } from '../columns/column.component';
 import { Injectable } from '@angular/core';
```
ddincheva: There are unused imports .. would be nice to remove them

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -2,630 +2,339 @@ import { IgxGridNavigationService } from '../grid-navigation.service';
...
+import { IgxHierarchicalRowComponent } from './hierarchical-row.component';
+import { IgxChildGridRowComponent } from './child-grid-row.component';
+import { IgxRowDirective, IgxGridBaseDirective } from '../grid';
+import { GridType } from '../common/grid.interface';
+import { IPathSegment } from './hierarchical-grid-base.directive';
+import { isNumber } from 'util';
```
ddincheva: this method is deprecated .. can you replace it..

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -2,630 +2,339 @@ import { IgxGridNavigationService } from '../grid-navigation.service';
...
+                this.grid.gridAPI.set_row_expansion_state(row.rowID, true, event);
             }
-        } else if (currentRowIndex !== this.grid.dataView.length - 1) {
-            // scroll next in view
-            super.navigateDown(rowElement, selectedNode);
-        } else if (this.grid.parent !== null &&
-            currentRowIndex === this.grid.dataView.length - 1) {
-            // move to next row in sibling layout or in parent
-            this.focusNext(visibleColumnIndex);
+            return;
```
ddincheva: this logic seems quite similar to the logic implemented in the handleAlt() method inside the base navigation service .. can you please check to see if it is possible to use it instead :)

ddincheva:
Also I have noticed that when I expand the last row from the hierarchical grid and scroll to the bottom of the child grid can not navigate with the KB after that in any direction.
scroll

ddincheva:
When parent and child grid has scroll and try to navigate to child an error is thrown:
expand row
click over random cell and pres "end" button
then press "arrowdown" 
scrollEnd

dkamburov:
Navigation between children on the same level is not working:
childnav

MayaKirova:
@dkamburov Issue with moving between siblings should be fixed now.

dkamburov:
The navigation stops if there is a child with empty data. Note that I had to modify a row island as in that sample all have them have data.
emptychildnav

dkamburov:
The issue with empty child still reproduces. And now I noticed that there is an error thrown:
hgridempty

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.navigation.spec.ts
```diff
@@ -1081,105 +901,106 @@ describe('IgxHierarchicalGrid Smaller Child Navigation #hGrid', () => {
...
         expect(childLastCell.selected).toBe(true);
-        expect(childLastCell.focused).toBe(true);
+        expect(childLastCell.active).toBe(true);
     }));
 
     it('should navigate to last cell in next child using Arrow Down from last cell of previous child with more columns', (async () => {
-        const childGrids =  fixture.debugElement.queryAll(By.directive(IgxChildGridRowComponent));
-        const firstChildGrid = childGrids[0].query(By.directive(IgxHierarchicalGridComponent)).componentInstance;
-        const secondChildGrid = childGrids[1].query(By.directive(IgxHierarchicalGridComponent)).componentInstance;
+        const childGrids =  fixture.debugElement.queryAll(By.css('igx-child-grid-row'));
```
nrobakova, line 953: Can you revert the original code to get the Grids by directive or create constants for the:
igx-child-grid-row 
igx-hierarchical-grid

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.navigation.spec.ts
```diff
@@ -1081,105 +901,106 @@ describe('IgxHierarchicalGrid Smaller Child Navigation #hGrid', () => {
...
-
-        GridFunctions.simulateCellKeydown(parentCell, 'Tab', false, true);
-        await wait(DEBOUNCE_TIME);
-        fixture.detectChanges();
-
-        const childGrids =  fixture.debugElement.queryAll(By.directive(IgxChildGridRowComponent));
-        const childGrid = childGrids[1].query(By.directive(IgxHierarchicalGridComponent)).componentInstance;
+        // last cell in child should be focused
+        const childGrids =  fixture.debugElement.queryAll(By.css('igx-child-grid-row'));
+        const childGrid = childGrids[1].query(By.css('igx-hierarchical-grid')).componentInstance;
```
nrobakova, line 945: Can you revert the original code to get the Grids by directive or create constants for the:
igx-child-grid-row
igx-hierarchical-grid

nrobakova:
Hi @MayaKirova, I tested this PR and have some questions;
1. When click on expand/collapse indicator should the active cell will be created:
Example:
1.1 Open dev sample hierarchicalGridUpdating
1.2 Expand first row and click on a cell in the child grid.
1.3 Expand second row 
1.4. cell in the child is still activated, but when press arrow key Left/Right/Up/Down the cell is not activated

2. Pressing tab from the root to the children. 
exapmle:
2.1 Open dev sample hierarchicalGridUpdating
2.2 Expand first row and click on a cell in it
2.3 press tab 2 times, the fist cell in the child is not activated, it is activated when press tab another again (on the 3th tab key)

3. When navigate with Shift+Tab to a child grid with no data and press arrow key an error is returned:
image

MayaKirova:
Hi @nrobakova @ddincheva ,

Regarding 1) Since the expand/collapse indicator is not a cell I don't think clicking it should activate it or some other cell in the parent. However since clicking it does change the focused grid (focus moves to parent) the child activation should be cleared in order to not cause confusion that you can continue to navigate in child, when in fact you are now focusing the parent grid.

I think that in general when the tbody is blurred and the focus has moved from the container then any active cell in that tbody should become inactive (should remove active styles, but remain selected) not just in the hierarchical grid but in the normal grid as well.

Currently when the tbody is blurred (for example when you click outside the normal igx-grid) the cell remains active, creating the impression that you can continue navigation, when in fact you cannot. This would also make the behavior more similar to how focusing used to work in previous versions.

Regarding 2) and 3) The cause seems to be a HostBiding in the base grid directive:
    /**
     * @hidden @internal
     */
    @HostBinding('attr.tabindex')
    public tabindex = 0;

I assume it was just forgotten and needs to be removed. Let me know if it is there for some other purpose.

mddragnev:
@MayaKirova 
1) As Deyan mentioned above using Alt + ArrowLeft still seems not to be prevented and redirects you to the previous page.
2) Expand the last row
In the child grid scroll down to the last record
Press ArrowDown (once the focus is lost) and when clicked twice the first row cell of the same child grid is focused
HgridNavigation

MayaKirova:
@mddragnev 
1) I cannot reproduce this. Handling is done in base navigation service (handleAlt) and event is prevented properly.
2) Should be fixed now.

mddragnev:
@MayaKirova  It all seems okay now

MayaKirova:
Adding back tabIndex on root element. It was decided that root element will also be a tab stop, hence when tabbing from parent to child the first element to take focus will be the child grid root element.

PR #6979, opened 2020-03-25
Implement Row Pinning for igxHierarchicalGrid.

skrustev:
Related #6640

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-cell.component.html
```diff
@@ -0,0 +1,31 @@
+<ng-template #defaultCell>
```
MayaKirova: Is there a need for a new hierarchical-cell template? As the same checks for the displayPinnedChip and the same template for the additional chip will probably need to be added to tree grid cell, perhaps it will be best to have this in the base cell so it can be re-used.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
```diff
@@ -102,7 +124,7 @@
             [igxForContainerSize]='calcHeight' [igxForItemSize]="renderedRowHeight" [igxForTrackBy]='trackChanges'
             #verticalScrollContainer (onChunkPreload)="dataLoading($event)">
             <ng-template #hierarchical_record_template>
-                <igx-hierarchical-grid-row [gridID]="id" [index]="rowIndex" [rowData]="rowData" #row>
+                <igx-hierarchical-grid-row [gridID]="id" [index]="getRowIndex(rowIndex, false)" [ghostRow]="isGhostRecord(rowData)" [rowData]="!isGhostRecord(rowData) ? rowData : rowData.recordData" #row>
```
MayaKirova: Is there need to calculate the index again here ( [index]="getRowIndex(rowIndex, false)" ) if it is already populated correctly in the context passed to the row ( in getContext for the template outlet)? This does not seem to be necessary for the base grid.

MayaKirova: Regarding the name of the property - ghostRow. I'm not sure whether we should call it something generic like that as it shows pinning specific UI (pinned chip in the first cell) and is something unique to the pinning feature.   Perhaps something like 'pinPlaceholder' would make its purpose more obvious. Same for related apis - isGhostRecord, ghostRec etc.

dkamburov:
I encountered a bit weird behavior. I'm I missing something here? The second pinned row disappeared when scrolling and then the children seems to have bigger z-index as they are showing on top of pinned area:
hgridpin

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-cell.component.html
skrustev: It is supported by both but the regular cells do not support it, so it will only pollute their template. Also the tree grid cell has a separate cell template so it will need to be added there anyway.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
skrustev: Yep, it seems that when the context is passed to the igxTemplateOutlet, the index is not applied. It seem to be working with the ngTemplateOutlet, so it could be a bug in our template. As you can see in the grid the index is actually set:

https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid/grid.component.html#L161

skrustev: Well I named it like that because the ghost thing on its own, the way I implemented does not know about the pinning. The pinned chip should be shown only when the row returns true if it is pinned as well, not only being a ghost row. I am ok to name it something else.

MayaKirova: After discussing it further with the team we have agreed that we'll have a `disabled` input in the base grid row, which will apply disabled styles and prevent some user interactions when set to true (selection and editing). Pinning will utilize the `disabled` input and duplicated row in the unpinned area will have it set to true . If a row is both `disabled` and `pinned` it will also render the "Pinned" chip in the first cell.

MayaKirova:
@skrustev If the first column is templates, then the pinned chip is not rendered anywhere in the pinned row, hence it gives no indication that the row is pinned. The rows looks just disabled.
The pinned chip should behave like the tree grid expansion indicator - should be rendered in the first visible cell in the row, regardless of whether it is templated or not.

ChronosSF:
@simeonoff , please let us know if you are ok with the VD implementation of the ghost records based on the spec https://github.com/IgniteUI/igniteui-angular/wiki/Row-Pinning

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.virtualization.spec.ts
```diff
@@ -388,21 +388,21 @@ describe('IgxHierarchicalGrid Virtualization Custom Scenarios #hGrid', () => {
         const hierarchicalGrid = fixture.componentInstance.hgrid;
 
         const initialBodyWidth = hierarchicalGrid.tbody.nativeElement.offsetWidth;
-        expect(hierarchicalGrid.verticalScrollContainer.getScroll().parentElement.hidden).toBeTruthy();
+        expect(hierarchicalGrid.verticalScrollContainer.getScroll().parentElement.parentElement.hidden).toBeTruthy();
```
ChronosSF, line 391: You should probably shorthand this to something readable in the HierarchicalGridFunctions static methods

ChronosSF:
@skrustev , editing is not prevented on ghost rows if you enter edit mode with the keyboard (enter, f2)

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -1371,6 +1381,9 @@
         }
     }
 
+    %grid-cell--pinned-chip {
+        margin-right: rem(4px);
```
simeonoff, line 1385:
```suggestion
        margin-#{$right}: rem(4px);
```
Where did you get the 4px from?

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
```diff
@@ -45,6 +45,7 @@
 /// @prop {Map} edit-mode-color [igx-color: ('secondary', 500)] - The text color in edit mode.
 /// @prop {Map} edited-row-indicator [igx-color: ('grays', 400)] - The indicator's color of edited row.
 /// @prop {Map} cell-edited-value-color [igx-color: ('grays', 600)] - The color of cell edited value.
+/// @prop {Map} cell-disabled-color [igx-color: ('grays', 600)] - The text color of a disabled cell.
```
simeonoff, line 48: The assignment below states `igx-color: ('grays', 400)`, not 600;

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
```diff
@@ -221,6 +222,10 @@ $_light-grid: extend(
             igx-color: ('grays', 600)
         ),
 
+        cell-disabled-color: (
+            igx-color: ('grays', 400)
```
simeonoff, line 226: Where did you get this color from?

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
```diff
@@ -558,6 +566,11 @@
         @extend %igx-grid__hierarchical-expander !optional;
     }
 
+    @include e(hierarchical-expander, $m: empty) {
```
simeonoff, line 569: Is this modifier class additive? I.e. are you adding it to the class list, or do you replace igx-grid__hierarchical-expander with igx-grid__hierarchical-expander--empty?

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
```diff
@@ -673,6 +686,11 @@
             @extend %igx-grid__hierarchical-expander--cosy !optional;
         }
 
+        @include e(hierarchical-expander, $m: empty) {
```
simeonoff, line 689: If this is an additive class, you don't have to extend the placeholder selector for each density.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
```diff
@@ -772,6 +790,11 @@
         @include e(hierarchical-expander) {
             @extend %igx-grid__hierarchical-expander--compact !optional;
         }
+        
+        @include e(hierarchical-expander, $m: empty) {
```
simeonoff, line 794: Same as above.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
```diff
@@ -558,6 +566,11 @@
         @extend %igx-grid__hierarchical-expander !optional;
     }
 
+    @include e(hierarchical-expander, $m: empty) {
+        @extend %igx-grid__hierarchical-expander !optional;
+        @extend %igx-grid__hierarchical-expander--empty !optional;
```
simeonoff, line 571: I can't find any declaration for this placeholder selector? What's the purpose of this?

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
skrustev: In the current implementation `igx-grid__hierarchical-expander--empty` replaces the default `igx-grid__hierarchical-expander` class.

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
skrustev: This looked closest to the current design in the row pinning spec. You are free to correct me if this is not right.

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
skrustev: This is a number I seem to not have updated.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
skrustev: Based on the spec screenshots we have, this can easily be changed if you think pixels should not be used or anything.
Maybe @StefanIvanov could also say what should be the correct value.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
skrustev: The purpose is disable any pointer events and return the cursor to the default appearance for the expander cell in the pinned rows which doesn't have expander icon.

simeonoff: It doesn't change the fact that there is no declaration of `%igx-grid__hierarchical-expander--empty` anywhere in the codebase. Hence, this does nothing.

projects/igniteui-angular/src/lib/grids/cell.component.html
```diff
@@ -1,4 +1,5 @@
 <ng-template #defaultCell>
+    <igx-chip *ngIf="displayPinnedChip" class="igx-grid__td--pinned-chip" [disabled]="true" [displayDensity]="'compact'">Pinned</igx-chip>
```
StefanIvanov, line 2: I am thinking if the chip should respect the grid density rather than always being "compact" because it appears in body rows that change the row/cell height, what do you think.

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
StefanIvanov: The gray variant that we use for disabled is 500.

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
StefanIvanov: And the true nuance is in the middle, we use grays 500 for disabled.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
StefanIvanov: Yes we keep 4px between the chip and the cell text, not sure if the value should depend on display density but maybe we can use a fraction of the cell left/right padding.

StefanIvanov: We updated the spec with the margins for the different densities, thanks for noting this.

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
StefanIvanov: should be 500

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_grid.scss
StefanIvanov: should be 500

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
skrustev: Here it is, I can use the full name if we have some conventions I missed:
https://github.com/IgniteUI/igniteui-angular/pull/6979/files#diff-e4bf11c573bc2e6e6f2423c2a1e70cb4R35

PR #6984, opened 2020-03-26
Action Strip Component

### LINKED ISSUE #6941: Action Strip component
Create a new component that can be templated to show a set of actions contextual to the component over which it is shown.

Spec: https://github.com/IgniteUI/igniteui-angular/wiki/Action-Strip-Component

Sub-tasks:
- [x] README for the feature
- [x] Changelog
- [x] Implement base grid action component which should be inherited by all pinning and editing grid action components
- [x] Restrict grid action usage outside of a grid context. Those actions should not appear if they don't have defined grid
- [x] Styling ActionStrip and grid action components - #7117
- [x] have my action strip layout respect display densities i.e. for comfortable 24px till action strip border; for cosy 16px till action strip border; for compact 12px till action strip border. The spacing between the actions is not affected by the display density and is 4px between actions
- [x] Remove `id` and `isMenu` @Inputs
- [x] Implement navigate to disabled row
- [x] Inject Action Strip to grid aciton child and use its context to retrieve the current context and grid, also remove the @Inputs for grid and context

- [x] Topic - https://github.com/igniteui/igniteui-docfx/issues/1841
- [x] Sample - https://github.com/igniteui/igniteui-angular-samples/issues/1872

dkamburov:
Closes #6941, #7117 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    public context;
+
+    @ContentChildren(forwardRef(() => IgxGridActionsBaseDirective), { descendants: true })
+    public gridActions: QueryList<IgxGridActionsBaseDirective>;
+
+    @ContentChildren(IgxActionStripMenuItemDirective)
+    public menuItems: QueryList<IgxActionStripMenuItemDirective>;
+
+    show(context): void {
```
ChronosSF, line 126: Show and hide should also have API comments

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.renderer.appendChild(context.element.nativeElement, this._viewContainer.element.nativeElement);
+        this.sendContext();
+    }
+
+    hide(): void {
+        this.hidden = true;
+        this.renderer.removeChild(this.context.element.nativeElement, this._viewContainer.element.nativeElement);
+    }
+
+    private sendContext() {
```
ChronosSF, line 138: Can't we request the grid context to be specified by users of the action strip so that we don't have this coupling that's grid specific?

Also this wouldn't be a problem if we went for the default UI right from the get go :/

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-pinning-actions.component.ts
```diff
@@ -0,0 +1,31 @@
...
+    providers: [{provide: IgxGridActionsBaseDirective, useExisting: IgxGridPinningActionsComponent}]
+})
+
+export class IgxGridPinningActionsComponent extends IgxGridActionsBaseDirective implements AfterViewInit {
+    get pinned() {
+        return this.context &&
+            (this.context.pinned || this.context.row && this.context.row.pinned);
+    }
+
+    pin() {
```
ChronosSF, line 16: Comments

projects/igniteui-angular/src/lib/core/styles/themes/schemas/light/_action-strip.scss
```diff
@@ -0,0 +1,57 @@
...
+/// @author <a href="https://github.com/desig9stein" target="_blank">Marin Popov</a>
+////
+
+/// Generates a light action-strip schema.
+/// @type {Map}
+/// @prop {Color} $color [currentColor] - The color used for the action strip component text and icons.
+/// @prop {Map} $background [igx-color: ('grays', 100] - The color used for the action strip component content background.
+/// @prop {Map} $delete-action [igx-color: ('error')] - The color used for the delete icon in action strip component.
+/// @prop {Number} border-radius [1] - The border radius fraction, between 0-1 to be used for action strip actions container.
+
```
simeonoff, line 14: Check if the produced documentation includes this item. Not sure sassdoc plays nicely with empty lines.

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
```diff
@@ -0,0 +1,165 @@
...
+    NgModule,
+    QueryList,
+    Renderer2,
+    TemplateRef,
+    ViewContainerRef,
+    Optional,
+    Inject
+} from '@angular/core';
+import {CommonModule} from '@angular/common';
+import {IgxDropDownModule} from '../drop-down/index';
```
ChronosSF, line 17: VS Code -> right click -> format document is your friend :wink:

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
damyanpetev: And explicitly marked as public :)

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-actions-base.directive.ts
```diff
@@ -0,0 +1,17 @@
...
+    /**
+     * An @Input property that set an instance of the grid for which to display the actions.
+     * ```html
+     *  <igx-grid-pinning-actions [grid]="grid1"></igx-grid-pinning-actions>
+     *  <igx-grid-editing-actions [grid]="grid1"></igx-grid-editing-actions>
+     * ```
+     */
+    @Input() grid;
+
+    @Input() context;
```
MayaKirova, line 16: Assuming we will have a service that will pass the context, and the context for grid actions can be either row or cell, both of which have reference to the grid they belong to, these Inputs would probably not be necessary. Both can be getters that read from the service.

That will also make it easier to implement in a hierarchical grid scenario where you will not have a direct reference for child grid in your template that you can pass to the grid action `grid` input. It makes more sense for the action strip service to extract the grid from the cell/row context, rather than to require the user to explicitly set it.

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
ChronosSF: To other potential reviewers - we decided to move the context sharing logic in a service.

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Enter row or cell edit mode depending the grid rowEditable option
+     * @example
+     * ```typescript
+     * this.gridEditingActions.startEdit();
+     * ```
+     */
+    public startEdit(): void {
+        const context = this.strip.context;
```
damyanpetev, line 27: Do type the context here - this will ensure all the calls below use traceable API :)
Can even use a check since it's public API:
```
if (!(this.strip.context instanceof IgxRowDirective)) {
            return;
        }
```
which will actually type the code afterwards. You can make your base class check a custom type guard too - see https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
```diff
@@ -0,0 +1,48 @@
...
+        const firstEditable = context.cells ?
+            context.cells.filter(cell => cell.editable)[0] : context;
+        const grid = firstEditable.row.grid;
+        grid.crudService.begin(firstEditable);
+        firstEditable.focused = true;
```
damyanpetev, line 32: Is that really needed?

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
damyanpetev: This applies for all action components

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
```diff
@@ -91,8 +101,8 @@ export class IgxActionStripComponent extends DisplayDensityBase {
      * @hidden
      * @internal
      */
-    @HostBinding('class.igx-action-strip')
-    public cssClass = 'igx-action-strip';
+    @Input()
+    public class = 'igx-action-strip';
```
damyanpetev, line 105: Wouldn't that override the `'igx-action-strip'` class if something is set on the element? I assumed that's required for the component to work. The whole point was more toward not binding the entire `attr.class` if we can avoid it, so again - can the density classes live on the internal container?

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
```diff
@@ -4,16 +4,25 @@ import {
...
+    ElementRef,
+    TemplateRef
 } from '@angular/core';
 import { DisplayDensityBase, DisplayDensityToken, IDisplayDensityOptions } from '../core/density';
 import { IgxDropDownComponent } from '../drop-down';
+import { IgxGridBaseDirective, IgxRowDirective } from '../grids';
+
+export interface IActionStripContext {
+    element: ElementRef<any>;
+    grid?: IgxGridBaseDirective;
```
damyanpetev, line 22: Erm, shouldn't this be generic for the strip as a whole? I.e. Not related to the grid in any way?

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
dkamburov: It won't as their is additional `@HostBinding('attr.class')`. I tried applying 'igx-action-strip' without success. The styles are expecting to be applied on the host element.
And I prefer not to re-write them.
But if that is not OK, I'll restructure the strip and see if we can change the styles.

MayaKirova:
@dkamburov 

- [ ] ActionStip does not render correctly when a different display density is set (any density different than 'comfortable').
    Note: Appears that the default 'igx-action-strip' class is overwritten due to the @HostBinding('attr.class') binding.

ChronosSF:
> * [ ]  Menu Buttons are not visible in Edge:
>   image

I've noticed this in some other samples as well, It seems there is something weird going on on Edge with some icons in the grid (not just the pinning one).

Annotation 2020-04-27 183256

Both the filtering icon in the chip and the sorting icons are white and barely visible. @simeonoff , is this something dev samples specific due to theme applied or a bug? It doesn't reproduce in the live demos.

skrustev:
Issues currently found as well:

1. The `context` is not specified as an input according to spec and describtion:
Context_input

2. The overlay is not visible when there are right pinned columns:
AS_pinning_after

projects/igniteui-angular/src/lib/action-strip/action-strip.module.ts
```diff
@@ -0,0 +1,35 @@
...
+        IgxGridActionsBaseDirective
+    ],
+    entryComponents: [
+    ],
+    exports: [
+        IgxActionStripComponent,
+        IgxActionStripMenuItemDirective,
+        IgxGridPinningActionsComponent,
+        IgxGridEditingActionsComponent,
+        IgxGridActionsBaseDirective
```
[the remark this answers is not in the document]
dkamburov: Yes, it can be used to create a custom grid actions components. By extending it, the developer receives the context from action strip.

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
dkamburov: Decided to scrap the `IActionStripContext` and use `any` for the context, as the context can be anything.

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
dkamburov: @damyanpetev Is the current state of the classes acceptable? If not this will require re-writing some styles.

projects/igniteui-angular/src/lib/action-strip/index.ts
```diff
@@ -0,0 +1,5 @@
+export { IgxGridActionsBaseDirective } from './grid-actions/grid-actions-base.directive';
+export { IgxGridEditingActionsComponent } from './grid-actions/grid-editing-actions.component';
+export { IgxGridPinningActionsComponent } from './grid-actions/grid-pinning-actions.component';
+export { IgxActionStripComponent } from './action-strip.component';
```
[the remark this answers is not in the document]
dkamburov: @MayaKirova I'll need to revert that one. Without those exports I'm unable to do that:
`import { IgxActionStripComponent } from 'igniteui-angular';`

Issue #6639, opened 2020-02-10
Splitter


radomirchev:
## Describe the solution you'd like
I would like to have a Splitter layout UI that provides resizable, expandable, collapsible, nestable panes and control of the page layout

Spec: https://github.com/IgniteUI/igniteui-angular/wiki/Splitter-Specification

- [x] Basic implementation - based on igxDrag directive -> https://github.com/IgniteUI/igniteui-angular/pull/6985
- [x] Expand/Collapse panes. -> #7036
- [x] Keyboard interactions -> #7052 
- [x] RTL support
- [x] Automation
- [x] Styling - https://github.com/igniteui/igniteui-angular/issues/6949
- [x] Documentation - https://github.com/igniteui/igniteui-docfx/issues/1842
- [x] Samples - https://github.com/igniteui/igniteui-angular-samples/issues/1880

PR #7036, opened 2020-04-01
feat(IgxSplitter): Adding expand/collapse functionality

mddragnev:
Related to #6639 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.html
```diff
@@ -1,15 +1,15 @@
 <div class="igx-splitter__bar" [style.flex-direction]="direction" igxDrag [ghost]="false"
 (dragStart)='onDragStart($event)'
 (dragMove)="onDragMove($event)">
-    <div class="igx-splitter-bar-expander">
-        <igx-icon fontSet="material">{{collapsePrevIcon}}</igx-icon>
-    </div>    
+    <div class="igx-splitter-bar-expander" (click)='onCollapsing($event)'>
```
ChronosSF, line 4: you can just send false/true here so you don't have to extract the clicked button in the handler

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -122,4 +137,24 @@ export class IgxSplitBarComponent {
         return !!relatedTabs.find(x => x.resizable === false);
     }
 
+    public onCollapsing(event: any) {
+        const arrowElement = event.srcElement;
```
ChronosSF, line 141: if you send false for left/top and true for right/bottom this can be shortened to something like:
```typescript
public onCollapsing(direction: boolean) {
   const sibling = this.siblings[0];
   this.collapse.emit(sibling.hidden && !direction || !this.pane.hidden && direction ? sibling : this.pane);
}
```

Try it out and see if it works

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -92,6 +98,15 @@ export class IgxSplitBarComponent {
      */
     private startPoint!: number;
 
+    get prevButtonHidden() {
+        const isPaneHidden = this.pane.hidden;
+        return isPaneHidden ? true : false;
+    }
+
+    get nextButtonHidden() {
+        const isSiblingHidden = this.siblings[0].hidden;
```
ChronosSF, line 107: Why not 
```typescript
return this.siblings[0].hidden;
```

It seems hidden is something you created and it's always boolean

projects/igniteui-angular/src/lib/splitter/splitter.component.ts
```diff
@@ -151,6 +151,18 @@ export class IgxSplitterComponent implements AfterContentInit {
...
 
+    /**
+     * This method performs the collapsing of the pane
+     * @param pane
+     */
+    public onCollapsing(pane: IgxSplitterPaneComponent) {
+        if (!pane) {
+            return;
+        }
+        this.pane = pane;
```
ChronosSF, line 162: I have to admit I am not too certain on the semantics but should we really reset the main pane? What happens to this.sibling then? They may become the same.

ChronosSF: Also, don't we need a public event for collapsing? Actually it should probably be something like paneToggle with an argument of the pane that changed its hidden state.

MayaKirova:
There are two minor issues:
1) When two adjacent splitter panes are collapsed, there is no expand button displayed for the first pane. For example if "Pane 1" "Pane 2" are collapsed you cannot show Pane 1, before showing Pane 2:

image

There should be an expand button (right arrow) in the first splitter bar, which when clicked can show "Pane 1" and reach the following state:
image

2) Resizing is not disabled when a pane is collapsed. It should be disabled as the pane is hidden.

PR #7052, opened 2020-04-06
Splitter keyboard interactions

VMihalkov:
Related to #6639

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


3phase:
I am not sure whether the issues below concern the keyboard navigation and not the base splitter PR itself, but there's a weird behavior to observe when moving the right border of the central grid pane a little bit to the left and then using `Ctrl+Left Arrow` to hide the whole central pane. The second misbehavior is when focusing the left border of the central pane via `Shift+Tab` and clicking `Ctrl+Right Arrow`. The rightmost grid won't get expanded at its maximal width, instead would stay cut. My explanations are described better by the gif below.

splitter_bug_1

3phase:
A lag occurs when focusing the left border of the central pane and trying to move it to the left by holding the Right arrow key for a while.

splitter_2

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -113,6 +202,58 @@ export class IgxSplitBarComponent {
         return !!relatedTabs.find(x => x.resizable === false);
     }
 
+    private moveUpOrLeft() {
+        this.panesInitialization();
```
MayaKirova, line 206: The code (moveUpOrLeft/moveDownOrRight) seems very similar to what is done in the Splitter component onMoveStart/onMoving handlers, which are bound to the moveStart/moving events of the igxSlitterBar.
 Instead of duplicating the same logic is it not possible to emit the events with the related pane this.moveStart.emit(this.pane) and the related delta -  
this.moving.emit(delta) where the delta is +10/-10 depending on the direction?

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -77,9 +77,98 @@ export class IgxSplitBarComponent {
         return this.siblings[0].hidden && !this.siblings[1].hidden;
     }
 
+    get collapseNextIcon() {
```
MayaKirova, line 80: These getters - collapseNextIcon/collapsePrevIcon are no longer used anywhere and should be removed.

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -77,9 +77,98 @@ export class IgxSplitBarComponent {
...
+     * @internal
+     */
+    @HostListener('keydown', ['$event'])
+    keyEvent(event: KeyboardEvent) {
+        const key = event.key.toLowerCase();
+        const ctrl = event.ctrlKey;
+        event.stopPropagation();
+            switch (key) {
+                case 'arrowup':
+                    if (this.type === 1) {
```
MayaKirova, line 120: Should compare with the SplitterType enum for clarity.

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="igx-splitter-bar"
+<div tabindex="0" class="igx-splitter-bar"
```
MayaKirova, line 1: The main igx-splitter-bar element should be focusable, not the first child. TabIndex can be set via a HostBinding for the root element.

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -77,9 +77,98 @@ export class IgxSplitBarComponent {
...
+                        if (ctrl) {
+                            this.pane.hidden ? this.onCollapsing(true) : this.onCollapsing(false);
```
MayaKirova, line 122: Isn't Ctrl+ArrowUp/Ctrl+ArrowLeft  equivalent to clicking the previous button? So shouldn't it call onCollapsing(false) in both cases?

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -77,9 +77,98 @@ export class IgxSplitBarComponent {
...
+                        if (this.pane.resizable && this.siblings[1].resizable) {
+                            event.preventDefault();
+                            this.moveUpOrLeft();
+                        }
+                    }
+                    break;
+                case 'arrowdown':
+                    if (this.type === 1) {
+                        if (ctrl) {
+                            this.siblings[1].hidden ? this.onCollapsing(false) : this.onCollapsing(true);
```
MayaKirova, line 134: Isn't Ctrl+ArrowDown/Ctrl+ArrowRight equivalent to clicking the next button? So shouldn't it call onCollapsing(true) in both cases?

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -77,9 +77,98 @@ export class IgxSplitBarComponent {
...
+                        }
+                    }
+                    break;
+                case 'arrowleft':
+                    if (this.type === 0) {
+                        if (ctrl) {
+                            this.pane.hidden ? this.onCollapsing(true) : this.onCollapsing(false);
+                            break;
+                        }
+                        if (this.pane.resizable && this.siblings[1].resizable) {
```
MayaKirova, line 149: There an exposed getter -> resizeDisallowed, which makes the same check.

projects/igniteui-angular/src/lib/splitter/splitter-bar/splitter-bar.component.ts
```diff
@@ -77,9 +77,98 @@ export class IgxSplitBarComponent {
...
+                        }
+                    }
+                    break;
+                case 'arrowright':
+                    if (this.type === 0) {
+                        if (ctrl) {
+                            this.siblings[1].hidden ? this.onCollapsing(false) : this.onCollapsing(true);
+                            break;
+                        }
+                        if (this.pane.resizable && this.siblings[1].resizable) {
```
MayaKirova, line 161: There an exposed getter -> resizeDisallowed, which makes the same check.

dkamburov:
Ctrl + arrow keys doesn't seem to be prevented, which causes Edge and Firefox to scroll the page.
Open the sample on Edge, focus a vertical splitter. Press ctrl + arrow up/arrow down:
splitter-scrolling

PR #7057, opened 2020-04-06
Input group to retain focus on prefix/suffix click

### LINKED ISSUE #6531: Input Group consistent focus
As the Input Group combines a bunch of separate elements, the actual `<input>` only takes up a portion of the space and thus clicking outside - on the suffix/prefix/label - causes the browser to blur the input by default.
Currently, the Input Group handles a click event to ensure focus is restored to the input so interaction with the built-in elements doesn't disrupt, but that looks a bit weird:
suffix-click

label-click
Naturally that also means `focus` and `blur` events are re-triggered constantly on the input itself, which can be a bit annoying for users that try to handle those and also affects the integration with `FormControl` as it may cause a value to be updated sooner than expected (came up from #6370)

Looking at the material design spec https://material.io/components/text-fields/ - there's no mention of this, but the reference Web components behave as a single unit (focus doesn't shift around). It seems very reasonable to do that.

More demos of those: https://material-components.github.io/material-components-web-catalog/#/component/text-field

## Additional context
The current click handle can `preventDefault` on pointer down to make sure the focus remains in the input. The material reference components also use a clever trick with the input padding to make it span over other elements to prevent the focus, though I'm not sure this is feasible with our group (diff structure, we use ripples on some icons).
This will affect the Grid's filtering that uses the focus shift to it's befit https://github.com/IgniteUI/igniteui-angular/issues/542 and the existing `supressInputAutofocus` API that was introduced for it. Manual focus will likely be required.

wnvko:
Clicking on prefix/suffix of the input group will call now `preventDefault` of the `pointerDown` event. This will prevent focusing of the prefix/suffix and effectively will retain the focus on the input.

Closes #6531 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


wnvko:
Please when verifying this new feature check thoroughly grid row filtering.

damyanpetev:
@wnvko  Can you look into `supressInputAutofocus` which was added for the Grid filtering use. After this change it should be fairly pointless and might be up for deprecation

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.ts
```diff
@@ -498,6 +498,14 @@ export class IgxGridFilteringRowComponent implements AfterViewInit {
         this.value = value;
     }
 
+    /** @hidden @internal */
+    public inputGroupPrefixClick(event: MouseEvent, target: any) {
+        event.stopPropagation();
+        // TODO: try to get the target from the event
```
DiyanDimitrov, line 504: There is a leftover TODO.

wnvko: I was not sure, in the scope of filter row, which element should be taken. This is why I added this TODO for guys implementing the filtering to fix or remove it.

CHANGELOG.md
```diff
@@ -50,6 +50,9 @@ All notable changes for each version of this project will be documented in this
 - `IgxToast`:
     - Added functionality for displaying various content into the toast component. It also allows users to access toast styles through its host element.
 
+- `IgxInputGroup`:
+    - clicking on the `prefix` or `suffix` of input group will not trigger `blur` and `focus` event. Also if the input group is focused clicking on it will not trigger `blur` and `focus` events.
```
damyanpetev, line 54:
```suggestion
    - Clicking on prefix, suffix or label elements in the Input Group will not longer blur and re-focus the input.
```
(clicking on already focused input never did anything anyway)

CHANGELOG.md
```diff
@@ -50,6 +50,9 @@ All notable changes for each version of this project will be documented in this
...
+
```
damyanpetev, line 55: Add a note that `supressInputAutofocus` is deprecated, though depending on resolution of comments below, the note might be a behavior change as well.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -975,6 +975,11 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
         }
     }
 
+    public onInputGroupClick(event: MouseEvent, target?: HTMLElement) {
+        // TODO: should we call stopPropagation here?
```
damyanpetev, line 979: Don't see a reason to stop propagation of the click TBH

projects/igniteui-angular/src/lib/input-group/input-group.component.ts
```diff
@@ -174,12 +207,23 @@ export class IgxInputGroupComponent extends DisplayDensityBase implements IgxInp
...
             this.input.focus();
         }
     }
 
+    /** @hidden */
+    @HostListener('pointerdown', ['$event'])
+    public onPointerDown(event: PointerEvent) {
+        // if the pointer is not on the input, e.g. pointer is on the prefix, call preventDefault.
+        // This will prevent blur of the input and focusing of the element under the pointer.
+        if (event.target !== this.input.nativeElement || this._supressInputAutofocus) {
```
damyanpetev, line 222: So if the target is not the input, this will always prevent (which is the goal of the change)? Don't see how `supressInputAutofocus` fits in here just for clicks on the input itself

projects/igniteui-angular/src/lib/input-group/input-group.component.ts
```diff
@@ -174,12 +207,23 @@ export class IgxInputGroupComponent extends DisplayDensityBase implements IgxInp
      * @hidden
      */
     @HostListener('click', ['$event'])
-    public onClick(event) {
-        if (!this._supressInputAutofocus) {
+    public onClick(event: MouseEvent) {
+        // if the click target is not the input, e.g. click on prefix, focus the input.
+        if (event.target !== this.input.nativeElement || !this._supressInputAutofocus) {
```
damyanpetev, line 212: This isn't compatible with the current `onPointerDown` logic that will always prevent on prefix/suffix/label clicks as this will not trigger and when it does - it's on the input and it seems like it'll do nothing

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -975,6 +975,11 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
         }
     }
 
+    public onInputGroupClick(event: MouseEvent, target?: HTMLElement) {
```
damyanpetev, line 978: This can be `event.currentTarget` so you don't need two params

projects/igniteui-angular/src/lib/input-group/input-group.component.ts
damyanpetev: I was going to suggest dropping the entire `supressInputAutofocus` functionality, but I came up with at least one good reason to keep it - namely to maintain an option to use buttons/icons in the input group that can perform action without focusing the input. That's particularly useful for mobile devices to prevent the software keyboard popping up - similar to what we've done for the combo search and before.

@wnvko Ignore the deprecation note for the changelog and revert it instead, logic on the handlers should be updated. I'd make the pointer prevent only active when already focused.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
wnvko: resolved

projects/igniteui-angular/src/lib/input-group/input-group.component.ts
wnvko: fixed

projects/igniteui-angular/src/lib/input-group/input-group.component.ts
wnvko: done

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
wnvko: we actually do not need this anymore

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.ts
wnvko: done

nrobakova:
Hi @wnvko,
I tested thid PR and found one issue related to it:
1. Open dev sample datePicker
2. In the DropDown Date Picker bind with ngModel click on the prefix to open the calendar
The focus is moved to the input and then it is blurred:

kdinev:
@damyanpetev Please review again!

damyanpetev:
@kdinev waiting on [related PR ](https://github.com/IgniteUI/igniteui-angular/pull/7093)that will likely require minor conflicts to be resolved here

projects/igniteui-angular/src/lib/input-group/input-group.component.spec.ts
```diff
@@ -178,26 +178,50 @@ describe('IgxInputGroup', () => {
         expect(inputGroupElement.classList.contains(INPUT_GROUP_COMPACT_DENSITY_CSS_CLASS)).toBeTruthy();
     });
 
-    it('suppress focus on input when clicked', async () => {
-        const fixture = TestBed.createComponent(InputGroupSupressInputFocusComponent);
```
damyanpetev, line 182: If there's no other test using this options.. might be a good idea to keep it

projects/igniteui-angular/src/lib/input-group/input-group.component.spec.ts
```diff
@@ -178,26 +178,50 @@ describe('IgxInputGroup', () => {
...
-        fixture.detectChanges();
+        const prefix = fixture.componentInstance.prefix;
+        const input = fixture.componentInstance.igxInput;
+
+        const pointOnPrefix = UIInteractions.getPointFromElement(prefix.nativeElement);
+        const pointerEvent = UIInteractions.createPointerEvent('pointerdown', pointOnPrefix);
+        const preventDefaultSpy = spyOn(pointerEvent, 'preventDefault');
+
+        Object.defineProperty(pointerEvent, 'target', { value: input.nativeElement });
+        inputGroup.onPointerDown(pointerEvent);
```
damyanpetev, line 194: I actually prefer if you're using fixture to go ahead and use `triggerHandler` instead - that also verifies the correct host binding is active. You should be able to pass the event argument with spy function for `preventDefault` just the same.

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.ts
```diff
@@ -482,6 +482,14 @@ export class IgxGridFilteringRowComponent implements AfterViewInit {
         this.value = value;
     }
 
+    /** @hidden @internal */
+    public inputGroupPrefixClick(event: MouseEvent) {
+        event.preventDefault();
+        event.stopPropagation();
+        (event.currentTarget as any).focus();
```
damyanpetev, line 489: Also, not any, `HTMLElement` should do :)

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -1061,6 +1064,11 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
         this.onSelection.emit(date);
     }
 
+    public onPrefixClick(event: MouseEvent) {
```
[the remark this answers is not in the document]
wnvko: done

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.ts
```diff
@@ -482,6 +482,14 @@ export class IgxGridFilteringRowComponent implements AfterViewInit {
         this.value = value;
     }
 
+    /** @hidden @internal */
+    public inputGroupPrefixClick(event: MouseEvent) {
+        event.preventDefault();
```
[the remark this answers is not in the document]
wnvko: I was sure we needed this - test shown we did not
done

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.ts
wnvko: fixed

projects/igniteui-angular/src/lib/input-group/input-group.component.spec.ts
wnvko: new test added to test this

projects/igniteui-angular/src/lib/input-group/input-group.component.spec.ts
wnvko: yeah much better in this way - done

Issue #6926, opened 2020-03-18
Make Shadow Row Disable when Pinning


3phase:
## Is your feature request related to a problem? Please describe.
Add a generic property to disable a shadow row when pinning.

PR #7104, opened 2020-04-10
IgxGrid disabled row implementation.

skrustev:
Closes: #6926  
#6640

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 
Integration scenarios with ghost rows:

- [x] Updating
- [x] Hiding
- [x] Filtering
- [x] Sorting
- [x] Summaries
- [x] Master-Details view
- [x] Multi-row layout
- [x] State persistence directive
- [x] Group by (with/without group summary).
- [x] Export to excel - test with this: https://github.com/IgniteUI/igniteui-angular/pull/7069
- [x] Row Selection
- [x] Cell Selection
- [x] Paging

These should be tested.


projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -5137,7 +5156,7 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
      * ```
      */
     get dataView(): any[] {
-        return this.verticalScrollContainer.igxForOf;
+        return this.verticalScrollContainer.igxForOf.map(rec => this.isGhostRecord(rec) ? rec.recordRef : rec) ;
```
MayaKirova, line 5159: Is this mapping still necessary? We don't extract the original data from meta data records for other features (like group records, hierarchical records etc.) so unless there is some specific integration scenario that depends on the value to be extracted this is probably not needed.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.pipes.ts
```diff
@@ -89,10 +89,14 @@ export class IgxGridHierarchicalRowPinning implements PipeTransform {
 
     constructor(private gridAPI: GridBaseAPIService<IgxHierarchicalGridComponent>) { }
 
-    public transform(collection: any[], pinnedArea: boolean, pipeTrigger: number): any[] {
+    public transform(collection: any[], isPinned: boolean, pipeTrigger: number): any[] {
```
MayaKirova, line 92: As base grid pinning pipe and the one for the hierarchical grid are the same, we can remove the one for hierarchical grid pinning and re-use the base grid one.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
rkaraivanov: Additionally I suspect this getter is called quite a lot around the grid code/templates. With a sufficiently big data, this is basically creating and discarding a new array on each call.

PR #7192, opened 2020-04-23
Fix Search with Row Pinning

### LINKED ISSUE #7161: IgxHierarchicalGrid + RowPinning Issues
-  [x] When row is pinned and search is applied that matches value in the pinned row then both pinned and duplicate(disabled) row are highlighted as active and you cannot navigate to the row in the unpinned area. Same happens in base grid.
image

- [x] When row pinning mode is bottom, it takes an extra change detection cycle for the pinned chip to appear and disappear from the pinned/unpinned cell - https://github.com/IgniteUI/igniteui-angular/pull/7244

3phase:
Related issue: #7161   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
```diff
@@ -333,9 +357,18 @@ export class IgxTextHighlightDirective implements AfterViewInit, AfterViewChecke
...
-        const row = group.rowIndex === undefined ? group.row : group.rowIndex;
+        const row = group.rowIndex ? group.rowIndex : group.row;
+
+        let metadataMatch = true;
+
+        if (group.metadata) {
+            const metadata = group.metadata;
+            group.metadata.forEach((value, key) => {
+                metadataMatch = this.metadata.get(key) === value;
+            });
```
DiyanDimitrov, line 368: We need to check that the map sizes are equal, also we need to break if the metadataMatch gets false. You can use the approach described here https://stackoverflow.com/questions/35948335/how-can-i-check-if-two-map-objects-are-equal.

projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
```diff
@@ -333,9 +357,18 @@ export class IgxTextHighlightDirective implements AfterViewInit, AfterViewChecke
     public activateIfNecessary(): void {
         const group = IgxTextHighlightDirective.highlightGroupsMap.get(this.groupName);
         const column = group.columnIndex === undefined ? group.column : group.columnIndex;
-        const row = group.rowIndex === undefined ? group.row : group.rowIndex;
+        const row = group.rowIndex ? group.rowIndex : group.row;
```
DiyanDimitrov, line 360: This change should be reverted since rowIndex=0 is a valid value for a row index but is is a falsy value. Anyway the rowIndex should be deprecated in another PR.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
```diff
@@ -99,7 +99,7 @@
             | visibleColumns:hasVisibleColumns
             | gridRowPinning:id:true:pipeTrigger
             | gridFiltering:filteringExpressionsTree:filterStrategy:advancedFilteringExpressionsTree:id:pipeTrigger:filteringPipeTrigger:true
-            | gridSort:sortingExpressions:sortStrategy:id:pipeTrigger as pinnedData">
+            | gridSort:sortingExpressions:sortStrategy:id:pipeTrigger:true as pinnedData">
```
DiyanDimitrov, line 102: I am not sure if this change is related to this PR.

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.html
```diff
@@ -79,9 +79,9 @@
             | treeGridTransaction:id:pipeTrigger
             | visibleColumns:hasVisibleColumns
             | treeGridNormalizeRecord:pipeTrigger
+            | gridRowPinning:id:true:pipeTrigger
             | treeGridFiltering:filteringExpressionsTree:filterStrategy:advancedFilteringExpressionsTree:id:pipeTrigger:filteringPipeTrigger:true
-            | treeGridSorting:sortingExpressions:sortStrategy:id:pipeTrigger:true
-            | gridRowPinning:id:true:pipeTrigger as pinnedData'>
+            | treeGridSorting:sortingExpressions:sortStrategy:id:pipeTrigger:true as pinnedData'>
```
DiyanDimitrov, line 84: I am not sure if this change is related to this PR.

projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
```diff
@@ -179,6 +183,26 @@ export class IgxTextHighlightDirective implements AfterViewInit, AfterViewChecke
...
+     * @example
+     * ```typescript
+     * // Set a property, which would disable the highlight for a given element on a cetain condition
+     * if (dontActivateCondition) {
+     *     cell.highlight.metadata.set('highlightElement', false);
+     * }
+     *
+     * const metadataMap = new Map<string, any>();
+     * metadataMap.set('highlightElement') = true;
+     * IgxTextHighlightDirective.setActiveHighlight(id, {index: index, metadata: metadataMap});
```
DiyanDimitrov, line 200: I don't think this code snippet is appropriate since it is cell specific. We should have something similar to the row and column snippets.

projects/igniteui-angular/src/lib/grids/cell.component.html
```diff
@@ -4,7 +4,7 @@
 <ng-template #defaultCell>
     <div igxTextHighlight style="pointer-events: none" [cssClass]="highlightClass" [activeCssClass]="activeHighlightClass" [groupName]="gridID"
         [value]="formatter ? formatter(value) : column.dataType === 'number' ? (value | igxdecimal: grid.locale) : column.dataType === 'date' ? (value | igxdate: grid.locale) : value"
-        [row]="rowData" [column]="this.column.field" [containerClass]="'igx-grid__td-text'"
+        [row]="rowData" [column]="this.column.field" [containerClass]="'igx-grid__td-text'" [metadata]="searchMetadata"
```
DiyanDimitrov, line 7: This change should be made to the tree-cell as well.

MayaKirova: Also to the expandable cell.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -919,4 +919,14 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
...
             .map((child) => getNodeSizeViaRange(range, child)));
     }
+
+    /**
+     * @hidden
+     * @internal
+     */
+    public get searchMetadata() {
+        const meta = new Map<string, any>();
+        meta.set('dataIndex', (this.row as any).viewIndex);
```
MayaKirova, line 929: Rather than cast to any it would be better to have the viewIndex defined in the RowType as it is required for search to work correctly with pinned rows in all grids.

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.html
```diff
@@ -79,9 +79,9 @@
             | treeGridTransaction:id:pipeTrigger
             | visibleColumns:hasVisibleColumns
             | treeGridNormalizeRecord:pipeTrigger
```
MayaKirova, line 81: Copying of objects can now be removed from the treeGridNormalizeRecord pipe.

MayaKirova:
@3phase Also the ChangeLog should be updated with the new Input for the highlight directive and a test for that new input should be added.

projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
3phase: I have updated it, although I'm not sure if it's as readable as it should be...

projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
```diff
@@ -179,6 +183,32 @@ export class IgxTextHighlightDirective implements AfterViewInit, AfterViewChecke
...
+     * @example
+     * ```typescript
+     *  // Set a property, which would disable the highlight for a given element on a cetain condition
+     *
+     *  const metadata = new Map<string, any>();
+     *  metadata.set('highlightElement') = false;
+     *
+     *  const metadataMap = new Map<string, any>();
+     *  metadataMap.set('highlightElement') = true;
+     *  IgxTextHighlightDirective.setActiveHighlight(id, {index: index, metadata: metadataMap});
```
DiyanDimitrov, line 200: You can remove this part of the snippet.

projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
```diff
@@ -179,6 +183,32 @@ export class IgxTextHighlightDirective implements AfterViewInit, AfterViewChecke
...
+     * A map that contains all aditional conditions, that you need to activate a highlighted
+     * element. To activate the condition, you will have to add a new metadata key to
+     * the `metadata` property of the IActiveHighlightInfo interface.
+     *
+     * @example
+     * ```typescript
+     *  // Set a property, which would disable the highlight for a given element on a cetain condition
+     *
+     *  const metadata = new Map<string, any>();
+     *  metadata.set('highlightElement') = false;
```
DiyanDimitrov, line 196: It should be ` metadata.set('highlightElement', false);`.

projects/igniteui-angular/src/lib/directives/text-highlight/text-highlight.directive.ts
```diff
@@ -335,7 +365,26 @@ export class IgxTextHighlightDirective implements AfterViewInit, AfterViewChecke
...
+                group.metadata.forEach((value, key) => {
+                    if (!metadataMatch) {
+                        return;
+                    }
+                    if (this.metadata.has(key)) {
+                        metadataMatch = this.metadata.get(key) === value;
+                    } else {
+                        metadataMatch = false;
+                    }
+                });
```
DiyanDimitrov, line 383: Returning in the forEach won't break the loop. I would suggest to use `for of` instead so you can break the loop when metadataMatch is false.

DiyanDimitrov: I would also suggest to extract this as a generic helper method for comparing maps in the utils file since almost the same code is repeated in the `compareMetadata` function.

3phase: The use of `for of` would result in a build error, which prompts to the need to add a `downlevelIteration: true` to the tsconfig.json file, while targeting es5.

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -2,6 +2,7 @@ import { Injectable, PLATFORM_ID, Inject } from '@angular/core';
 import { isPlatformBrowser } from '@angular/common';
 import { Observable } from 'rxjs';
 import ResizeObserver from 'resize-observer-polyfill';
+import { IActiveHighlightInfo } from '../directives/text-highlight/text-highlight.directive';
```
DiyanDimitrov, line 5: This import is no more necessary.

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -355,3 +356,29 @@ export function resizeObservable(target: HTMLElement): Observable<ResizeObserver
...
+
+/**
+ * @hidden
+ * @internal
+ *
+ * Compares two metadata maps.
+ */
+export function compareMetadata(map1: Map<any, any>, map2: Map<any, any>) {
+    let metadataMatch = true;
+    if (map1 && map2) {
```
DiyanDimitrov, line 368: If only one of the maps is falsy, the method will return true, which is not correct.

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -355,3 +356,29 @@ export function resizeObservable(target: HTMLElement): Observable<ResizeObserver
...
+        if (map1.size !== map2.size) {
+            return false;
+        } else {
+            const keys = Array.from(map2.keys());
+            for (const key of keys) {
+                if (map1.has(key)) {
```
DiyanDimitrov, line 374: If map1 doesn't have the key, metadataMatch should be set to false.

mddragnev:
Hello @3phase, I have found two minor issues.
a) 
1.Open row pinning dev sample and on the base grid pin record with company name Ana
2.Change the items per page to 5
3.Sort CompanyName column with descending strategy so that the ghost row is on page 6
4.Search for "an"
5.Observe that the grid is navigated to page 6 where is the ghost row but actually the correct record is highlighted on the first page.

b)
1.On the tree grid search for Andrew
2.Pin arbitary records so that the pinning section gets half of the whole grid size
3.Pin the record that contains Andrew
4.From the search navigation go to the next match
5.Observe that the scroll is broken and the ghost row that matches Andrew is not visible but highlighted

3phase:
> Hello @3phase, I have found two minor issues.
> a)
> 1.Open row pinning dev sample and on the base grid pin record with company name Ana
> 2.Change the items per page to 5
> 3.Sort CompanyName column with descending strategy so that the ghost row is on page 6
> 4.Search for "an"
> 5.Observe that the grid is navigated to page 6 where is the ghost row but actually the correct record is highlighted on the first page.
> 
> b)
> 1.On the tree grid search for Andrew
> 2.Pin arbitary records so that the pinning section gets half of the whole grid size
> 3.Pin the record that contains Andrew
> 4.From the search navigation go to the next match
> 5.Observe that the scroll is broken and the ghost row that matches Andrew is not visible but highlighted

Hi @mddragnev, thank you for your remarks. I think I have them resolved. You could re-test.

skrustev:
- [ ] When there is paging and pinning is set to bottom items do not highlight that are in the unpinned area and are not on page 1.
 1. Pin a row when pinning is to bottom
 2. Search for an item that is on page different from the first.
 Result: The search scrolls to the item but it is not highlighted

3phase:
> * [x]  When there is paging and pinning is set to bottom items do not highlight that are in the unpinned area and are not on page 1.
>   Steps:
>   
>   1. Pin a row when pinning is to bottom
>   2. Search for an item that is on page different from the first.
>      Result: The search scrolls to the item but it is not highlighted

@skrustev, thanks for remarking that issue, I think it's resolved now.

PR #7195, opened 2020-04-23
feat(igxChip): Add support for setting tabindex through attribute or input. #7147

### LINKED ISSUE #7147: Can not change the tabIndex to the igx-chip component

As a developer I would like to be able to set the tabIndex property for igx-chip component. Currently there is no way to prevent focus on chip component.

skrustev:
Closes #7147

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/chips/chip.component.ts
```diff
@@ -353,6 +372,16 @@ export class IgxChipComponent extends DisplayDensityBase {
     @Output()
     public onDragEnter = new EventEmitter<IChipEnterDragAreaEventArgs>();
 
+    /**
+     * @hidden
+     * @internal
+     */
+    @HostBinding('attr.tabIndex')
```
MayaKirova, line 379: Can't this HostBinding be on the tabIndex Input rather than on a getter that returns no value and instead sets the internal tab index prop? 
I assume it was set like this to handle scenarios where the user has set the element attribute directly and wants the same attribute to be set automatically for all focusable children, however I don't think this is something we should handle. IMO if the user wants to have the same tab index for all elements inside the chip he should use the exposed Input.

MayaKirova: After discussing this it seems that the code is added to prevent the user from setting tabindex attribute directly on the chip itself for example:  
<igx-chip  tabIndex="4" ...

With the current implementation this would remove the tabIndex from the root element and will instead set it on the child elements (chip-area and chip delete button) hence preventing the user from having tab index on the root element. Hence the first focusable element will always be the chip area.

@ChronosSF @damyanpetev @rkaraivanov As we could not reach an agreement whether it is ok to change attributes that the user has set himself I'd like to hear your thoughts on this.
PS. I also don't like the fact that a getter is settings things for child elements - looks like a hack.

damyanpetev: Correct me if I'm wrong here, but I seem to remember angular handing native attributes _only_ if a respective `@Input` is not found, kinda like a fallback. I.e. if the component is configured as `<igx-chip tabIndex="4"` then all that is needed is.
```ts
/** @hidden @internal */
@Input()
    public set tabindex(....
```
Note the exact match in casing. This should get the value set and not require an explicit host bind to clear out the native attr. This can also transfer the index to the internal container or later on be removed if we change which element is focusable. Also since it's the native attr I wouldn't necessarily put it on the public API explicitly.

Maya brought up a very valid point though - moving the index doesn't do much in terms of behavior (user defines a chip, the container (the same thing for all the user is concerned) gets focus in the specified order, but it does mess with anything supplementing the focus - titles, labels, and such. That on its own might be a separate issue though.

skrustev: Ok so after discussing we decided that with this PR the tabindex will be applied to the chip inner element only, because when tabindex is set as an attribute the angular applies it to the native element as well ass reflecting its value in the input. 

In a separate issue the focusable element will be moved to be the main igx-chip element to support titles, labels, aria and etc since it will take more time to implement.

MayaKirova: Link to issue: https://github.com/IgniteUI/igniteui-angular/issues/7248

MayaKirova:
@skrustev There is one consistently failing test that should probably be updated after the latest changes:
IgxChip Rendering Tests: should apply correct tab indexes when tabIndex and removeTabIndex are set as inputs

PR #7255, opened 2020-05-04
feat(RowPinning): Initial implementation of jump to disabled row grid action.

skrustev:
#6941

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@skrustev Build fails with error:
Property '_pinnedRecordIDs' does not exist on type 'IgxGridBaseDirective'

Additionally pinning throws error in treegrid: 
Cannot set property 'ghostIndex' of undefined

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -229,7 +229,7 @@ export class IgxGridRowPinningPipe implements PipeTransform {
 
         if (grid.hasPinnedRecords && isPinned) {
             const result = collection.filter(rec => grid.isRecordPinned(rec));
-            result.sort((rec1, rec2) => grid.pinRecordIndex(rec1) - grid.pinRecordIndex(rec2));
+            result.sort((rec1, rec2) => grid.getPinedRecordIndex(rec1) - grid.getPinedRecordIndex(rec2));
```
MayaKirova, line 232: getPinedRecordIndex and setPinnedRecordGhostIndex will probably have to be overwritten for the treegrid as its data records are of different type - ITreeGridRecord. Hence these methods will not work as expected for them.

MayaKirova:
Issues:
- [ ] Pinned rows order is not retained correctly in TreeGrid.
 To reproduce pin the second row in tree grid and then the first row. Order in pinned area should be second row and then first row since this is the order of pinning.

skrustev:
This PR should be merged after #7192, since it contains changes that are required for the Tree Grid Jump To on another page to work.

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-pinning-actions.component.html
```diff
@@ -1,4 +1,7 @@
 <ng-container *ngIf="isRowContext">
+    <button *ngIf="inPinnedArea" igxRipple igxButton="icon" (click)="scrollToRow($event)">
+        <igx-icon fontSet="filtering-icons" name="jump_down"></igx-icon>
```
dkamburov, line 3: There should be jump_up icon when the rows are pinned to bottom

MayaKirova:
Hierarchical Grid:
- [ ] In root grid if a row in the middle is pinned and all rows are then expanded, "Go To" cannot scroll to the correct row.

Steps to reproduce:
1. In the gridRowPinning sample in the hierarchical grid, pin a row in the middle (8).
2.  Expand all rows via the header button and use the "Go To" button.

- [ ] HierarchicalGrid + Paging - In HierarchicalGrid if you go from a page that has no scrollbar to a target row on a page with scrollbar the row is not scrolled in view.

Steps to reproduce:
1. In the gridRowPinning sample in the hierarchical grid set page size to 5.
2. Pin row 3 and expand a row above it (for example row 2).
3. Go to the next page and use the "Go to" button.
Page is changed but the row is not scrolled in view.

PR #7269, opened 2020-05-05
fix(for-of): Revert change to small scrolls

### LINKED ISSUE #7219: Revew for-of "Skip logic on small scrolls"
## Description  
This change fixes a memory leak that should be investigated, what is the cause of it, and why the detached elements are not cleared by the garbage collector.
With the removal of the improvement related to the "small scroll" we introduce the following problem: 
if we have a cell in edit mode and the igx-focus is set to true, we can not perform scroll in the hierarchical grid. See the gif below:
focusScroll

Related to [this change](https://github.com/IgniteUI/igniteui-angular/pull/7212)

### LINKED ISSUE #7230: The not correct row is opened in edit mode in edit mode when the grid is scrolled and click on a child
## Description  
The not correct row is opened in edit mode in edit mode when the grid is scrolled and click on a child

 * igniteui-angular version: 9.1
 * browser: n/a

## Steps to reproduce  

1. Open dev sample hierarchicalGridUpdating
2. Expand row with CACTU
3. Scroll down and double click on a row in the child grid
4. Click on another row

Note: This issue is related to the issue #7219

## Result  
The grid scroll is changed and not correct row is opened in edit mode.

## Expected result  
The grid should not be scrolled

## Attachments  
hGridRowEditing

### LINKED ISSUE #7290: [hGrid] arrow up/down does not work when you reach any inner grid via kb nav
## Description  
Once you reach with kb nav any inner grid while navigating up/down, the navigation goes to the document body and the grid becomes uncontrollable.

 * browser: all

## Steps to reproduce  

1. Open any hGrid dev demos
2. navigate to a row island which is closer to the end of the display view of the grid so that when you open it with alt + arrow right it is expanding out of it.
3. Try to continue navigating up/down

## Result  
The focus goes to the document of the body and the grid stops reacting.

## Expected result  
The focus should remain on the grid you are in and continually responding to the kb interactions.

### LINKED ISSUE #7295: When navigate with arrow down in the hgrid focus does not stays on the grid
## Description  
When navigate with arrow down in the hgrid focus does not stays on the grid

 * igniteui-angular version: 9.1
 * browser: n/a

## Steps to reproduce  

1. Open dev sample hierarchicalGridUpdating
2. Expand row with id Anton
3. In the child grid expand first row
4. Start navigation with arrow down

Note: This issue can be similar to the #7290

## Result  
After goes to the second row in the child rid the focus is losed

## Expected result  
It should be possible to continue with navigation

## Attachments  

rkaraivanov:
Handle focus/blur interaction of templates inside the for-of directive itself.

Closes #7219 
Closes #7230 
Closes #7290
Closes #7295

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -850,27 +849,98 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
...
+                const container = this.dc.instance._viewContainer.element.nativeElement as HTMLElement;
+                const activeElement = document.activeElement as HTMLElement;
+
+                // Remove focus in case the the active element is inside the view container.
+                // Otherwise we hit an exception while doing the 'small' scrolls swapping.
+                // For more information:
+                //
+                // https://developer.mozilla.org/en-US/docs/Web/API/Node/removeChild
+                // https://bugs.chromium.org/p/chromium/issues/detail?id=432392
+                if (container.contains(document.activeElement)) {
```
MayaKirova, line 873: Not sure how performant this check will be since the display container can be quite big (for vertical igxGridForOf it will contain all visible rows).
Perhaps it would be better to have this in moveApplyScrollNext/moveApplyScrollPrev and have it check  just the particular view that is about to be detached&inserted for focus. That way it will search only in that one view rather than in all views and it will also be executed only for small scrolls (as it is probably not needed for big scrolls).

nrobakova:
Hi @rkaraivanov,
I tested this issue and the bug #7230 is still reproducible

MayaKirova:
@rkaraivanov Since focus should not be lost in a HierarchicalGrid with child grids scenario (refer to issue #7290) where navigation with arrow keys goes through small scrolls, rather than checking and blurring in all views (whole display container) it would be better to blur just in the particular view node that is going to be detached/attached. Other views don't need to be blurred.
For example:
            const embView = this._embeddedViews.pop();
            **const node = embView.rootNodes.find(node => node.nodeType === Node.ELEMENT_NODE) || embView.rootNodes[0].nextElementSibling;
            this.scrollFocus(node);**
            const view = container.detach(container.length - 1);
            this.updateTemplateContext(embView.context, i);
            container.insert(view, 0);
  ...

 protected scrollFocus(container: HTMLElement): void {
        const activeElement = document.activeElement as HTMLElement;
...
        if (container.contains(document.activeElement)) {
            activeElement.blur();
        }
    }

This should retain the existing navigation within hierarchical grid (should resolve #7290) without losing the child grid's focus.

PR #7302, opened 2020-05-11
fix(HierarchicalGrid): Making behavior between base and hierarchical …

### LINKED ISSUE #7277: [hGrid] tab back does not activate the first cell of the parent grid tbody
## Description  
[hGrid] tab back(shift+tab) does not activate the first cell of the corresponding grid tbody 

 * browser: all

## Steps to reproduce  

1. Open any hGrid dev demo
2. Try to shift + tab from some inner grid to the parent tbody

## Result  
The upper-left cell is not activated

Also once we shift + tab to the parent grid and the first left cell is activated no scroll is performed and the focus indicator remains over the child grid. This behavior could be seen in the accompanying gif file below.

## Expected result  
The upper-left cell should be activated 
IMO the parent grid should be scrolled and the active indication of the child grid should be gone as we as the selection cell indicator.

## Attachments
image

hgrid-focus

MayaKirova:
…grid consistent - focus always makes first cell active even if the clicked element is not a cell (for example is expander, row selector etc.).

**Note:**  Removing previous handling that prevented first cell from becoming active when you click an element that is not focusable (for example the row expander) as it was not consistent with the base grid behavior. 
In base grid the first cell is always selected on focus, so for example if a row selector is clicked the first cell becomes active even if it is not clicked:
row_selection_activate

Since this appears to be the intended behavior it should behave the same in hierarchical grid.

Closes #7277  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
After discussing the base grid behavior with @ddincheva we have decided that the default interactable grid elements that are not in the cells that can be clicked (like expanders, row selectors etc.) should not trigger activation so we will prevent the related events. Will include the fix for hierarchical grid specific UI elements (expanders) with this PR. The others will be addressed in a separate issue.

ddincheva:
This is related to issue #7311

PR #7307, opened 2020-05-12
refactor(date-time-editor): fix model binding;

### LINKED ISSUE #7264: IgxDateRangePicker incorrect model validation
## Description  
The `IgxDateRangePicker` updates its validity based on either the selected value from the calendar and/or the projected inputs (if any). With the projected inputs, we want the picker to update its validity during typing and then again after blurring.

 * igniteui-angular version: n/a
 * browser: n/a

## Steps to reproduce  

1. Open the date-range dev samples.
2. Clear the start or end field of a picker that's ngModel bound.
3. Observe

## Result  
The picker does not update its validity.

## Expected result  
It should update its validity.

wnvko:
- date-range: use callback to update model;
- add separate validation for projected inputs;
- use separate date-editors' values when updating date-range's value

Closes #7264   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -312,7 +312,7 @@ export interface IBaseEventArgs {
 
 export interface CancelableBrowserEventArgs extends CancelableEventArgs {
     /** Browser event */
-    event?: Event;
+    event?: MouseEvent;
```
damyanpetev, line 315: That interface is generic/shared, surely it doesn't refer only to mouse events

projects/igniteui-angular/src/lib/date-picker/date-picker.utils.ts
```diff
@@ -246,55 +246,53 @@ export abstract class DatePickerUtil {
     }
 
     /**
-     * Determines whether the provided value is less than the provided min value.
+     * Determines whether the provided value is greater than the provided min value.
```
damyanpetev, line 249: while at it
```suggestion
     * Determines whether the provided value is greater than the provided max value.
```

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker-inputs.common.ts
```diff
@@ -45,8 +45,9 @@ export class DateRangePickerFormatPipe implements PipeTransform {
     providers: [{ provide: IgxInputGroupBase, useExisting: IgxDateRangeInputsBaseComponent }]
 })
 export class IgxDateRangeInputsBaseComponent extends IgxInputGroupComponent {
+    /** @hidden @internal */
     @ContentChild(NgControl)
-    protected ngControl: NgControl;
+    public ngControl: NgControl;
```
damyanpetev, line 50: Um, why did this become public as well?

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -404,7 +408,7 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
     private get required(): boolean {
         if (this._ngControl && this._ngControl.control && this._ngControl.control.validator) {
             const error = this._ngControl.control.validator({} as AbstractControl);
-            return (error && error.required) ? true : false;
+            return error && error.required;
```
damyanpetev, line 411: Technically, the getter is now `boolean | undefined` like this

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -569,12 +569,27 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
...
             return { 'minValue': true };
         }
-        if (this.maxValue && !this.valueInRange(control.value)) {
+        if (this.maxValue && value && value.start && DatePickerUtil.greaterThanMaxValue(value.start, this.maxValue, false)) {
+            return { 'maxValue': true };
+        }
+        if (this.maxValue && value && value.end && DatePickerUtil.greaterThanMaxValue(value.end, this.maxValue, false)) {
             return { 'maxValue': true };
         }
+        if (control.validator) {
```
damyanpetev, line 585: This should definitely not be needed. As `Validator` you are part of the form control validators that are all called anyway. You only provide your bit of the check and for sure don't call the entire validation again (that should've even entered an endless loop with the call below).

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -654,23 +669,68 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
...
+            this.value = { start: this.value.start, end: this.value.start };
         }
         if (this.value && !this.value.start && !this.value.end) {
-            this.updateValue(null);
+            this.value = null;
+        }
+
+        let outsideClick = true;
+        if (event.event) {
+            const elementRect: DOMRect = this.element.nativeElement.getBoundingClientRect();
```
damyanpetev, line 684: What's this? Whatever you are trying to avoid here, there must be a better way to check.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -654,23 +669,68 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
...
+            // input click
+            this.focusInput();
+        }
+    }
+
+    private updateValidityOnBlur() {
+        this.onTouchCallback();
+        if (this.hasProjectedInputs) {
+            this.projectedInputs.forEach(i => {
+                const ngControl = this._ngControl || (i as IgxDateRangeInputsBaseComponent).ngControl;
```
damyanpetev, line 703: Yeah, nope. Fairly certain this applies *only if* the range itself has `ngControl`, and that's still odd use for me

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -723,14 +783,15 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
...
-                this.projectedInputs.map(i => i.isRequired = this.required);
-            });
-        }
+        // workaround for igxInput setting required
+        Promise.resolve().then(() => {
+            const isRequired = this.required;
+            if (this.inputGroup && this.inputGroup.isRequired !== isRequired) {
+                this.inputGroup.isRequired = isRequired;
+            } else if (this.hasProjectedInputs && this._ngControl) {
+                this.projectedInputs.map(i => i.isRequired = isRequired);
```
damyanpetev, line 792: that looks like a `forEach` to me, rather than a map

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -768,7 +829,7 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
     private extractRange(selection: Date[]): DateRange {
         return {
             start: selection[0],
-            end: selection.length > 1 ? selection[selection.length - 1] : null
+            end: selection.length > 0 ? selection[selection.length - 1] : null
```
damyanpetev, line 832: Okay, so that ensures with one selection you get a range of the same start date? Does it affect anything else in terms of behavior?

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -785,23 +846,23 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
...
-                        this.updateValue({ start: this.value?.start, end: date as Date });
+                    .subscribe(() => {
+                        this.value = { start: start.dateTimeEditor.value, end: end.dateTimeEditor.value };
                     });
                 start.dateTimeEditor.validationFailed
                     .pipe(takeUntil(this.$destroy))
-                    .subscribe((event: IgxDateTimeEditorEventArgs) => {
-                        this.updateValue({ start: event.newValue as Date, end: this.value?.end });
+                    .subscribe(() => {
+                        this.value = { start: start.dateTimeEditor.value, end: end.dateTimeEditor.value };
```
damyanpetev, line 860: Why on earth do we update the range value on .. validationFailed?!

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -814,16 +875,21 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
                     .pipe(takeUntil(this.$destroy))
                     .subscribe(() => {
                         if (this.collapsed) {
-                            this.onTouchCallback();
+                            this.updateValidityOnBlur();
+                        }
+                        if (this.value && !this.value.start) {
+                            this.value = { start: null, end: null };
```
damyanpetev, line 881: Why is the value changing on the touch callback? Also if the goal is to clear the value.. do set a null TBH. That way a required validator can check it as well

projects/igniteui-angular/src/lib/core/utils.ts
wnvko: The source of truth is always the `overlay` service. It is used in `toggle` and `date-picker` where event is passed from the `overlay`. Then it is used in `time-picker`, `date-range-picker` and `drop-down` where event is provided from the `toggle`, and effectively from `overlay`.
In `overlay` service it is already [passed ](https://github.com/IgniteUI/igniteui-angular/blob/6b5cdc9bdf72e89378870ec87aa1580ea20bdda4/projects/igniteui-angular/src/lib/services/overlay/overlay.ts#L644)as `MouseEvent` in document handler. I think it is better to have it as `MouseEvent` as it is such, than more general `Event`.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker-inputs.common.ts
wnvko: In `updateValidityOnBlur` if there is `ngControl` we need to get if it is valid or not and update the input directive validity. However, when in two inputs mode `ngModel` could be applied on `date-range-component` or on each of the inputs. In second case we need access to input's `ngModel` from `date-range-component` and it became `public`

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
wnvko: reverted

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
wnvko: fixed - if it is in drop down mode and click in the input it will not blur. In all other cases outside click will force blur

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
wnvko: `extractRange` is called only in `handleSelection`. Last one is called at two places:
- in calendar as handler for `onSelection`.
- in public `selectRange` method. This one says _"Selects a range of dates. If no `endDate` is passed, range is 1 day (only `startDate`)"_
I think the fix is ok.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
wnvko: yeah with `null` is much better.
If one deletes start value and leave the component we should delete and the end value.
and this has nothing in common with touch callback, but the naming :)

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
wnvko: I've cleaned this a little

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
wnvko: when we have required on `ngModel` angular test if component's `value` is empty like this:
```
function isEmptyInputValue(value) {
    // we don't check for string here so it also works with arrays
    return value == null || value.length === 0;
}
```
If our `value` is `{ start: null, end: null }` angular is Ok with this `value` and does not set `required` to `true`.

And no this will not enter infinite loop as we are calling `validator` against empty object

wnvko: when we have required on `ngModel` angular test if component's `value` is empty like this:
```
function isEmptyInputValue(value) {
    // we don't check for string here so it also works with arrays
    return value == null || value.length === 0;
}
```
If our `value` is `{ start: null, end: null }` angular is Ok with this `value` and does not set `required` to `true`.

And no this will not enter infinite loop as we are calling `validator` against empty object

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker-inputs.common.ts
wnvko: in the end reverted, not public anymore

projects/igniteui-angular/src/lib/core/utils.ts
damyanpetev: Yeah, except that's in core/utils and is presumably available for use outside the overlay and related components, thus it doesn't make sense to limit it to mouse event.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
damyanpetev: Fairly certain you are calling validator function (which combines all validator functions, including this one) with an empty control (normally called with `this`), that will merely have `undefined` value. If the required validator runs, so should this one, so still a chance this will loop.

In either case - the required check is fine, there's very little reason to have an object with two null values. If there's no range and the value is cleared I'd just have it as `null`. Plus it's easier to check if `value` exist rather than going `value.start && value.end`

wnvko: what if we have value and start or end is null - it is still not valid. And yes we do hope one will set range to null if both start and end are null, but will they all do it?

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
damyanpetev: Yes, but you already handle the value change for those inputs, so I still don't see why the range adjusts values on blue as well.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
damyanpetev: First - the scenario is a bit odd, since that's only possible with projected inputs that are meant to validate on their own (i.e. have their own models). And ideally the control itself should not enter this state at all (the TODO you've left). At the end of the day if a user forcibly assigns a range with nulls, it should be acceptable to also add a custom validator to handle it.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -845,37 +911,23 @@ export class IgxDateRangePickerComponent extends DisplayDensityBase
         }
     }
 
+    private initialSetValue() {
+        // if there is no value, no ngControl but we have inputs we may have value set trough
+        // inputs' ngModels - we should generate our initial control value
+        if (!this.value && this.hasProjectedInputs && !this._ngControl) {
```
damyanpetev, line 917: Likely a candidate for refactor - this is also the condition on which you don't use `updateInputs` (since the value is coming from the projected inputs)

PR #7329, opened 2020-05-13
fix(select): items

### LINKED ISSUE #7190: igx-select long text items
## Description  
igx-select long text items display

 * igniteui-angular versions: 9.0.x, 8.2.x
 * browser: any

## Steps to reproduce  
Open the following [sample](https://stackblitz.com/edit/igx-select-long-text-list-items).

### igniteui-angular 8.2.x:
 * When igx-select item has a long text it's not stretching itself wide enough and the text gets cut off. There isn't any padding on the right side as well.

### igniteui-angular 9.0.x:
 * A horizontal scroll appears when igx-select item has a long text. Again there is no padding on the right side (see the attached image). The selection appears only on the part of the item which is placed below the toggle button. If the text is really long and we scroll further to the right, it seems like there isn't a selected item.

## Expected result  
Since there isn't any description how it should behave in the spec my assumption is: 
 * Either there should be a horizontal scroll, the select should be applied all the way to the end and the padding on the right sight should be similar to this on the left.
 * Or the text should get cut off and the padding on the right side should be again similar to this on the left. 

## Attachments  
* igniteui-angular 8.2.x:
image

* igniteui-angular 9.0.x:
image

desig9stein:
Closes #7190 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


ViktorSlavov:
This PR introduces a behavioral change.
The `igx-drop-down__item` has a display `flex`, but `igx-drop-down__inner` has display `block`, so this might lead to different experience in specific apps that would rely on that.
In particular, this will affect two of our select samples, I'll log an issue in the samples browser for those.

simeonoff:
> Behavioral change caused by missing/moved `flex` display needs to be added to the Changelog

@desig9stein could you update the changelog, please?

desig9stein:
Sorry about the message of the last commit it should be "**updating the changelog.md**"

CHANGELOG.md
```diff
@@ -129,9 +135,6 @@ All notable changes for each version of this project will be documented in this
     ```typescript
     public pinningConfiguration: IPinningConfig = { columns: ColumnPinningPosition.End };
     ```
-    - Added new properties for paging:
```
Lipata, line 132: unintentionally removed?

PR #7339, opened 2020-05-14
chore(*): Fix failing tests

HristoP96:


zdrawku:
Based on this [commit change](https://github.com/IgniteUI/igniteui-angular/pull/7339/commits/90a03f3a0006a15d1408d2e8b08043c09004892d), we should test where the package is added, within the consuming project (dependency, devDep, or nowhere). In our case, it shouldn't be added at all (in the package.json consuming project) when the end-user executes `ng add igniteui-angular`, as the watermark is a dependency of igniteui-angular and it will be installed in node_modules/igniteui-angular/node_modules.

Testing steps:
1. `npm run build:lib`, 
2. `npm pack`
3. `npm install` in external project
4. `ng g igniteui-angular:ng-add` - ng-add is the name of the schematic

projects/igniteui-angular/src/lib/core/i18n/grid-resources.ts
```diff
@@ -90,6 +90,7 @@ export interface IGridResourceStrings {
     igx_grid_advanced_filter_initial_text?: string;
     igx_grid_advanced_filter_column_placeholder?: string;
     igx_grid_advanced_filter_value_placeholder?: string;
+    igx_grid_watermark_placeholder?: string;
```
damyanpetev, line 93: That doesn't appear to be used anywhere 🤔

zdrawku: I hurried to assign you as a reviewer before discussing what's left to be done here with @HristoP96 😅 the status will be updated ones it is ready for review, извинявай Damyanski 🙏🏻

nrobakova:
Hi @HristoP96,
When open the page under Edge or IE11, an error is returned and the contend is not loaded.

projects/igniteui-angular/src/lib/core/i18n/grid-resources.ts
3phase: > извинявай Damyanski 🙏🏻

@zdrawku hahahha patron 🙇🏻‍♂️

HristoP96:
> Hi @HristoP96,
> When open the page under Edge or IE11, an error is returned and the contend is not loaded.

image
 I am not reproducing this issue

HristoP96:
We need to update the package the a new version, which is compatible with any versions of any browsers, so this PR should not be merged before that.

projects/igniteui-angular/src/lib/core/i18n/grid-resources.ts
HristoP96: > That doesn't appear to be used anywhere 🤔

@damyanpetev This is now used in the text attribute of the  [watermark ](https://github.com/IgniteUI/igniteui-angular/pull/7339/files#diff-aa19f92a91132fc5c20ebdb1bc687c91R134)

zdrawku:
@nrobakova could you test again with IE and Edge, now the trial watermark shouldn't break the application

zdrawku:
@kdinev @damyanpetev do you have any other comments or concerns? As per our last conversation, Konstantin said that its okay to leave the localization capabilities, so I assume there are no more questions there.

HristoP96:
@zdrawku @kdinev @damyanpetev  Tha latest requirements are applied.
P.S. @damyanpetev  I removed the changes in the package-lock, that are not coherent to the new package and I strongly agree with you on that the package-lock must not me touched if it is not necessary, I must have pushed it inadvertently. :)

PR #7340, opened 2020-05-14
Migrate deprecated rowSelectable property to rowSelection

### LINKED ISSUE #7325: Add migrations for rowSelectable prop
## Description  
`rowSelectable` was deprecated in 9.0.x and needs a migration.

https://github.com/IgniteUI/igniteui-docfx/issues/1930

dafo:
Closes #7325   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/migrations/common/UpdateChanges.ts
+++ b/projects/igniteui-angular/migrations/common/UpdateChanges.ts
@@ -8,6 +8,17 @@ import { getLanguageService, getRenamePositions } from './tsUtils';
 import { getProjectPaths, getWorkspace, getProjects, escapeRegExp } from './util';
 import { LanguageService } from 'typescript';
 
+
+export enum InputPropertyType {
+    EVAL = 'eval',
+    STRING = 'string'
+}
+declare type TransformFunction = (args: BoundPropertyObject) => void;
+export interface BoundPropertyObject {
+    value: string;
+    bindingType: InputPropertyType;
+}
+
 // tslint:disable:arrow-parens
 export class UpdateChanges {
     protected workspace: WorkspaceSchema;
@@ -19,7 +30,7 @@ export class UpdateChanges {
     protected themePropsChanges: ThemePropertyChanges;
     protected importsChanges: ImportsChanges;
     protected conditionFunctions: Map<string, Function> = new Map<string, Function>();
-    protected valueTransforms: Map<string, Function> = new Map<string, Function>();
+    protected valueTransforms: Map<string, TransformFunction> = new Map<string, TransformFunction>();
 
     private _templateFiles: string[] = [];
     public get templateFiles(): string[] {
@@ -131,7 +142,7 @@ export class UpdateChanges {
         this.conditionFunctions.set(conditionName, callback);
     }
 
-    public addValueTransform(functionName: string, callback: (oldValue: string) => string) {
+    public addValueTransform(functionName: string, callback: TransformFunction) {
         this.valueTransforms.set(functionName, callback);
     }
 
@@ -213,8 +224,8 @@ export class UpdateChanges {
             let searchPattern;
 
             if (type === BindingType.output) {
-                base = String.raw`\(${change.name}\)=(["'])`;
-                replace = `(${change.replaceWith})=$1`;
+                base = String.raw`\(${change.name}\)=(["'])(.*?)\1`;
+                replace = `(${change.replaceWith})=$1$2$1`;
             } else {
                 // Match both bound - [name] - and regular - name
                 base = String.raw`(\s\[?)${change.name}(\s*\]?=)(["'])(.*?)\3`;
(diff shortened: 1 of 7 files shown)


projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -202,39 +210,40 @@ export class UpdateChanges {
...
+                base = String.raw`(\s\[?)${change.name}(\s*\]?=)(["'])(.*?)\3`;
+                replace = String.raw`$1${change.replaceWith}$2$3$4$3`;
                 groups = 3;
             }
 
             let reg = new RegExp(base, 'g');
             if (change.remove || change.moveBetweenElementTags) {
                 // Group match (\1) as variable as it looks like octal escape (error in strict)
-                reg = new RegExp(String.raw`\s*${base}.*?${'\\' + groups}(?=\s|\>)`, 'g');
+                reg = new RegExp(String.raw`\s*${base}.*?(?=\s|\>)`, 'g');
```
damyanpetev, line 231: Okay this cannot be right, this add the "search till end of value" (`.*?`) based on the old base regex, but that's now added in the base as well. That should result in some pretty wonky expression that can match something after the closing quote mark or even still is doing a lot more pointless lookups.
This separate regex for the deletion shouldn't be needed anymore.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -202,39 +210,40 @@ export class UpdateChanges {
...
+                    break;
+                case 'directive':
+                    searchPattern = String.raw`\<[^\>]*[\s\[]${change.owner.selector}[^\>]*\>`;
+                    break;
             }
 
             const matches = fileContent.match(new RegExp(searchPattern, 'g'));
 
             for (const match of matches) {
+                let replaceStatement = replace;
```
damyanpetev, line 246: Just rename the actual variable TBH.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -244,9 +253,23 @@ export class UpdateChanges {
                     fileContent = this.copyPropertyValueBetweenElementTags(fileContent, match, moveMatch);
                 }
 
+                if (change.valueTransform) {
+                    // We need to remove 'g' flag to see capture groups
+                    // This will FAIL if there are multiple bindings in A SINGLE TAG to the SAME property
```
damyanpetev, line 258: It actually won't since the actual replace still uses the `reg` that is global. Bet you can put it in a test to make sure :)

projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -244,9 +253,23 @@ export class UpdateChanges {
...
+                    const regExpMatch = match.match(new RegExp(base));
+                    const boundProp = regExpMatch && regExpMatch[1] === ' [' ? InputPropertyType.EVAL : InputPropertyType.STRING;
```
damyanpetev, line 260: I'd take `endsWith('['` instead of hard equals in this case, just to proof this against future capture group changes.

projects/igniteui-angular/migrations/common/schema/index.ts
```diff
@@ -74,3 +76,8 @@ export enum ElementType {
     directive = 'directive',
     component = 'component'
 }
+
+export enum InputPropertyType {
```
damyanpetev, line 80: That's not part of the schema, thus shouldn't be in this file :)

projects/igniteui-angular/migrations/update-9_1_0/index.ts
```diff
@@ -4,14 +4,29 @@ import {
...
+                switch (oldValue) {
+                    case 'true':
+                        return '\'multiple\'';
+                    case 'false':
+                        return '\'none\'';
+                    default:
+                        return `${oldValue} ? 'multiple' : 'none' `;
+                }
+            } else {
+                return 'true';
```
damyanpetev, line 27: Eh, what now? How is `true` a valid value for the new `rowSelection` enum?

projects/igniteui-angular/migrations/update-9_1_0/index.ts
```diff
@@ -4,14 +4,29 @@ import {
...
 
 const version = '9.1.0';
 
-export default function(): Rule {
+export default function (): Rule {
     return (host: Tree, context: SchematicContext) => {
         context.logger.info(`Applying migration for Ignite UI for Angular to version ${version}`);
 
         const update = new UpdateChanges(__dirname, host, context);
+        update.addValueTransform('rowSelectable_is_deprecated', function (oldValue: string, bindingType?: InputPropertyType): string {
```
damyanpetev, line 16: Okay, actually looking at this - there's a little bit missing still. For this migration ideally we'd want to convert `[rowSelectable]="true"` to `rowSelection="multiple"` without the extra property binding. You could also land on the case where a old string property `prop="value"` should be updated to an object or something that requires an expression (e.g. `[newProp]="{}"`).

This will require the transform function to be able to return both the value and change the type, possibly best though an object that can also describe the input parameter. This will require more manipulation of the replace group references.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
ViktorSlavov: The `base` value is declared only once outside of the loop's scope, so I reassign it here as to not mutate the `base` value

projects/igniteui-angular/migrations/common/UpdateChanges.ts
ViktorSlavov: Will leave this as is for the moment, as it seems to capture some edge cases (leading/trailing spaces). Could investigate in the future.

PR #7465, opened 2020-06-02
fix(*): Add a check not to update groupingExpansionState when no expansion states are changed

### LINKED ISSUE #7218: when we should emit groupingExpansionStateChange
## Question  
When we should emit groupingExpansionStateChange? 
As far as I understood the event corresponds to the interaction of the expand/collapse rows which means that the event should not be triggered when we are clearing the grouped columns, as I saw that we have other events for that purpose **groupingExpressionChange** and **onGroupingDone**.

The other note is that the event is emitted only when the column group is cleared but not when it is added to an expression.

 * browser: all

mddragnev:
Closes #7218 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid/grid-api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid-api.service.ts
@@ -43,15 +43,15 @@ export class IgxGridAPIService extends GridBaseAPIService<IgxGridComponent> {
             names.forEach((colName) => {
                 const grExprIndex = groupingState.findIndex((exp) => exp.fieldName === colName);
                 const grpExpandState = this.grid.groupingExpansionState;
-                /* Do not set the new instance produced by filter
-                when there are not any expansion states */
-                if (grpExpandState.length) {
-                    /* remove expansion states related to the cleared group
+                 /* remove expansion states related to the cleared group
                     and all with deeper hierarchy than the cleared group */
-                    this.grid.groupingExpansionState = grpExpandState
-                        .filter((val) => {
-                            return val.hierarchy && val.hierarchy.length <= grExprIndex;
-                        });
+                const newExpandState = grpExpandState.filter((val) => {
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/grids/grid/grid-api.service.ts
```diff
@@ -43,12 +43,16 @@ export class IgxGridAPIService extends GridBaseAPIService<IgxGridComponent> {
...
                 const grpExpandState = this.grid.groupingExpansionState;
-                /* remove expansion states related to the cleared group
-                and all with deeper hierarchy than the cleared group */
-                this.grid.groupingExpansionState = grpExpandState
-                    .filter((val) => {
-                        return val.hierarchy && val.hierarchy.length <= grExprIndex;
-                    });
+                /* Do not set the new instance produced by filter
+                when there are not any expansion states */
+                if (grpExpandState.length) {
```
ChronosSF, line 48: As discussed today I think the correct logic would be:

```typescript
var newState = this.grid.groupingExpansionState = grpExpandState
    .filter((val) => {
         return val.hierarchy && val.hierarchy.length <= grExprIndex;
    });
if (newState.length !==this.grid.groupingExpansionState.length) {
    this.grid.groupingExpansionState = newState;
}
```

MayaKirova: @ChronosSF You are right as this would also prevent the event from emitting when a different column than the one with expansion state is unpinned so it is a better solution.

I noticed that the first line of the suggested code has a typo:
var newState = **this.grid.groupingExpansionState** = grpExpandState
    .filter((val) => { ...

Should be just:
 var newState = grpExpandState
                .filter((val) => {...

mddragnev:
It should be okay right now. @MayaKirova the pr is ready for testing.
Thanks!

PR #7479, opened 2020-06-03
fix(group-by): Expose the column on which the grouping is applied #6939

### LINKED ISSUE #6939: Use header instead of field in the default groupby row template
Fields are not always human-readable so column-related ui strings in the grid should use the header when available.

It's better to use the header instead of the field for the default template.

The row is templatable so this is not critical but a nice to have default

## Additional context
Related to #6933

dkamburov:
Related #6939  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@dkamburov Based on the description of the issue I assume that the default group row template should also be updated so that if the column exists and has a header it would by default show the header and not the field name.  Currently the default template still shows only the field name. For example, when grouping a column with header set to "Company Name Header" it shows the correct header text in the chip but in the group row shows the field name:

image

dkamburov:
@MayaKirova That's the plan for the next version. We don't want to introduce behavioral change in the current 9.1.x.

dkamburov:
The pull request was based to master - https://github.com/IgniteUI/igniteui-angular/pull/7532

PR #7498, opened 2020-06-05
fix(igxGrid): Include custom footer, which is rendered next to pager,…

### LINKED ISSUE #6065: Grid body height is not calculated correctly when there is a custom footer.
## Description  
When defining a custom footer for the IgxGrid, the body container where the grid rows should be rendered is not resized so it takes into account the footer. That is why the vertical scrollbar disappears behind the footer when the user have scrolled to the bottom.

 * igniteui-angular version: 8.2.x
 * browser: any

## Steps to reproduce  

1. Open the Column Moving demo and add a custom footer.
2. Scroll to the bottom of the grid.

## Result  
The vertical scrollbar gets partially hidden by the footer.

## Expected result  
The footer shouldn't appear on top of the vertical scrollbar..

## Attachments  
Custom_footer_grid

MayaKirova:
in height calculations.

Closes #6065 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/src/app/grid-column-moving/grid-column-moving.sample.html
+++ b/src/app/grid-column-moving/grid-column-moving.sample.html
@@ -36,6 +36,11 @@
                                                     [hasSummary]="false"
                                                     [dataType]="c.type">
                 </igx-column>
+                <igx-grid-footer>
+                    <div style='height:200px;'>
+                        Test
+                    </div>
+                </igx-grid-footer>
             </igx-grid>
             Drag mode: {{ grid1.selectionService.dragMode }}
             <div class="sample-buttons">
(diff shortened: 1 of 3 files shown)


src/app/grid-column-moving/grid-column-moving.sample.html
```diff
@@ -36,6 +36,11 @@
                                                     [hasSummary]="false"
                                                     [dataType]="c.type">
                 </igx-column>
+                <igx-grid-footer>
+                    <div style='height:200px;'>
```
kdinev, line 40: Can this be applied directly to the `igx-grid-footer` element?

MayaKirova: @kdinev You can but the igx-grid-footer has display:inline by default so it would have no affect. You could set a different display and a height, for example:
` <igx-grid-footer style='height:200px;display:block;'>
      Test
   </igx-grid-footer> 
`

PR #7525, opened 2020-06-09
Extend IGridEditEventArgs

### LINKED ISSUE #4965: [igx-grid] improve IGridEditEventArgs
No.

1. I'd like `IGridEditEventArgs` to have the reference to the cell instance like `IGridCellEventArgs` does.
```ts
onCellEdit(event: IGridEditEventArgs) {
    // reference to IgxGridCellComponent
    const cell = event.cell;

    // do something to the cell
}
```
2. I'd like `IGridEditEventArgs.cellID.columnID` is renamed to `IGridEditEventArgs.cellID.columnIndex`. Because, columnID is not id but index of column, which is confusing,
```ts
onCellEdit(event: IGridEditEventArgs) {
    // Bad
    const columnIndex = event.cellID.columnID;
    // Good
    const columnIndex = event.cellID.columnIndex;
    ...
}
```

No

## Additional context
I want to get the reference  to`IgxGridCellComponent` instance to do something to the cell when I hook the following events, 
- onCellEditEnter
- onCellEdit
- onCellEditCancel
- onRowEditEnter
- onRowEdit
- onRowEditCancel

Currently, we can't get the cell directly from event params. We have to code like below. It's a pain. 
```ts
onCellEdit(event: IGridEditEventArgs) {
    const columnIndex = event.cellID.columnID;
    const column = this.grid.columns[columnIndex];
    const cell = this.grid.getCellByKey(event.rowID, column.field);

    // do something to the cell
}
```
In addition, `event.cellID.columnID` property is confusing because it means not `ID` but `index` of column.

### LINKED ISSUE #7304: igx-grid onCellEdit/onRowEdit event issues
## Description  
I'm handling the `onCellEdit` event and I want to commit my transaction to my backend service. 
The `onRowEdit` and `onCellEdit` events are documented as events that fire after editing is committed, which is simply not true, since both events are cancellable. There's currently no events that can be subscribed to, which fire after editing has been committed.

> Note: Initially this issue was also describing the problems below, however they were resolved by #4965 
> - I cannot extract from the event which column the edit was performed for.
> - Also, the event args have no `owner` and I cannot extract the grid context.
> - Also, if hiding is performed, the column index is wrong, and the correct column cannot be retrieved.

 * igniteui-angular version: 9.0.13
 * browser: N/A

## Steps to reproduce  

1. Define a grid
2. Define editing for cells in it
3. Handle `onCellEdit` and inspect the event arguments

```html
<igx-grid [data]="orders" [height]="'300px'" primaryKey="id" (onCellEdit)="editDone($event)" [cellSelection]="'none'">
  <igx-column field="confirmed" [dataType]="'boolean'" [editable]="true" header="Потвърдена">
  </igx-column>
  <igx-column field="shipped" [dataType]="'boolean'" [editable]="true" header="Изпратена">
  </igx-column>
</igx-grid>
```

## Result  
`console.log(args)` gives me this:
image

## Expected result  
I want editing events after edits are committed (as documented).

## Attachments  

wnvko:
When emitted `IGridEditEventArgs` will commit `owner`, the grid where event was emitted, and `column` where is possible (e.g. will not pass `column` in `OnRowWhatever` events).

`onCellEdit` will now fire after the data in the cell is updated, as is `onRowEdit` now. Note, if one sets `cancel` to `true` in `On[Cell/Row]Edit` edit mode will not end, but the value will be submited.

Closes #4965 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/common/events.ts
```diff
@@ -29,6 +29,7 @@ export interface IGridEditEventArgs extends CancelableEventArgs, IBaseEventArgs
     oldValue: any;
     newValue?: any;
     event?: Event;
+    columnField?: string;
```
rkaraivanov, line 32: IMO it maybe better to actually return the column instance when emitting with cell editing?
@damyanpetev your thoughts about this?

ddincheva:
Hello @wnvko ,
In the event arguments we can not provide the actual IgxCellComponent, due to the fact when the cell in edit mode is out of the current view and try to submit it, the IgxCellComponent is undefined.
 
image
So you can define a new object that uses the newly added interface CellType  or you can extent the existing IgxCell class and return it instead the IgxCellComponent object.

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -143,6 +143,8 @@ export class GridBaseAPIService <T extends IgxGridBaseDirective & GridType> {
         cell.editValue = value;
 
         const args = cell.createEditEventArgs();
+        args.cell = this.grid.getCellByColumn(args.cellID.rowIndex, args.column.field);
+        args.owner = this.grid;
```
ddincheva, line 147: Maybe it is good idea this code snippet to be part of createEditEventArgs() method?

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -225,6 +227,7 @@ export class GridBaseAPIService <T extends IgxGridBaseDirective & GridType> {
         this._update_row(row, value);
 
         const args = row.createEditEventArgs();
+        args.owner = this.grid;
```
ddincheva, line 230: Same applies here .. this should be moved inside createEditEventArgs() method of IgxRow class.

projects/igniteui-angular/src/lib/grids/selection/selection.service.ts
```diff
@@ -139,11 +140,9 @@ export class IgxGridCRUDService {
             console.warn('The grid must have a `primaryKey` specified when using `rowEditable`!');
         }
         this.row = this.createRow(this.cell);
-        const args = {
-            rowID: this.row.id,
-            oldValue: this.row.data,
-            cancel: false
-        };
+        // TODO: should we emit newValue in onRowEditEnter?
```
ddincheva, line 143: Why Cell/Row edit enter events need to have a property newValue since no edit action is performed?!

projects/igniteui-angular/src/lib/grids/common/cell.interface.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface CellType {
+    gridID: string;
+    column: ColumnType;
+    columnIndex: number;
+    row: RowType;
+    rowData: any;
+    rowIndex: number;
+    value: any;
+    visibleColumnIndex: number;
+}
```
ddincheva, line 13: Don't you think it is good idea the column to provide all the information about the column like visibleColumnIndex, column index, field and etc. , same applies to the row?!

wnvko: We will stick to `IgxCell` and get rid of `CellType`

projects/igniteui-angular/src/lib/grids/selection/selection.service.ts
```diff
@@ -53,20 +54,24 @@ export class IgxRow {
...
 
-    constructor(public id: any, public index: number, public data: any) { }
+    constructor(public id: any, public index: number, public data: any, public grid: IgxGridBaseDirective & GridType) { }
 
-    createEditEventArgs(): IGridEditEventArgs {
-        return {
+    createEditEventArgs(includeNewValue = true): IGridEditEventArgs {
+        const args: IGridEditEventArgs = {
             rowID: this.id,
+            rowData:  { ... this.data },
```
damyanpetev, line 62: Might be good to leave the reference to the actual data record in case the event handler want to find in the data

PR #7574, opened 2020-06-11
Deselect Row on Transaction Redo

### LINKED ISSUE #5673: When redo transactions the deleted row stays selected
## Description  
When redo transactions the deleted row stays selected

 * igniteui-angular version: 8.2
 * browser: n/a

## Steps to reproduce  

1. Open dev sample treeGrid
2. Select first row and click button 'Delete selected row'
3. Click undo button
4. Select same row
5. Click redo button

## Result  
The row is deleted but it is still selected

## Expected result  
The row should not be selected

## Attachments  
image

3phase:
Closes #5673   

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/services/transaction/transaction.ts
```diff
@@ -26,6 +35,11 @@ export interface State {
     type: TransactionType;
 }
 
+export interface IStateUpdateEvent {
```
wnvko, line 38: we do not need `I` in front of the interface. In type script given class can `extend` other class or `implement` an interface. This is why when you read the class signature you know what class it extends and what interfaces it implements.

projects/igniteui-angular/src/lib/services/transaction/transaction.ts
```diff
@@ -26,6 +35,11 @@ export interface State {
...
+    origin: TransactionEventOrigin;
+    actions?: any;
```
wnvko, line 40: we can strongly type this. Please add interface `Action` like this:
```
export interface Action {
    transaction: Transaction;
    recordRef: any;
}
```
Note in `IgxTransactionService.redo` method there is `useInUndo` parameter in the action type - this is a leftover, and you can safely remove it.

projects/igniteui-angular/src/lib/services/transaction/igx-transaction.ts
```diff
@@ -127,7 +127,7 @@ export class IgxTransactionService<T extends Transaction, S extends State> exten
             this._undoStack.push(actions);
             this._redoStack = [];
 
-            this.onStateUpdate.emit();
+            this.onStateUpdate.emit({ origin: TransactionEventOrigin.END});
```
wnvko, line 130: you can add the actions to the event arguments here too

projects/igniteui-angular/src/lib/services/transaction/igx-transaction.ts
```diff
@@ -47,7 +47,7 @@ export class IgxTransactionService<T extends Transaction, S extends State> exten
         if (!this._isPending) {
             this._undoStack.push([{ transaction, recordRef }]);
             this._redoStack = [];
-            this.onStateUpdate.emit();
+            this.onStateUpdate.emit({ origin: TransactionEventOrigin.ADD });
```
wnvko, line 50: what we push in `_unodStack` is actually an `action`. Add this action to the event arguments.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 All notable changes for each version of this project will be documented in this file.
 
+## 10.0.0
+
+### General
+- `IgxTransaction` - The `onStateUpdate` now emits with information of its origin. The emitted value is of type `IStateUpdateEvent`, which has two properies:
+    - `origin` - a mandatory property. It can vary within the values of the `TransactionEventOrigin` interface; 
+    - `actions` - an optional property that contains information about the transactions, that caused the emission of the event. It is used by `undo` and `redo`.
```
wnvko, line 10: we will use the `actions` in every place where event emits - no need to clarify it is used in...

projects/igniteui-angular/src/lib/services/transaction/transaction.ts
wnvko, line 40: you can make this mandatory too and send empty array in `clear` method. For all other places you have the necessary actions.

3phase:
@wnvko thanks for your suggestions, I think I have resolved them. Could you take it a look again, please?

projects/igniteui-angular/src/lib/services/transaction/transaction.ts
```diff
@@ -1,11 +1,20 @@
 import { EventEmitter } from '@angular/core';
+import { ThrowStmt } from '@angular/compiler';
 
 export enum TransactionType {
     ADD = 'add',
     DELETE = 'delete',
     UPDATE = 'update'
 }
 
+export enum TransactionEventOrigin {
```
skrustev, line 10: @wnvko
Looking at the listed event origins for onStateUpdate I don't see when it is emitted from a commit. Looking at the IgxTransactionService I also didn't notice any place on commit that the onStateUpdate is called except on clear. Is this the expected behavior of the onStateUpdate event?

Also due to the IgxHierarchicalTransactionService commit calling clear() a second time when childDataKey is not provided, the onStateUpdate is emitted two times which I also am not sure if it's a valid scenario anyway in the igxTreeGrid but just asking.

wnvko: @skrustev 
Calling `onStateUpdate` in `clear` is OK from my point of view.
Calling `clear` in `IgxHierarchicalTransactionService.commit` is a bug. You can log this, as I do not thing it is OK to fix it in this feature PR.

skrustev: @wnvko 
Ok so if the user wants to perform action on `commit` he will have to check for the `clear` event origin ? It's kinda weird to me but if you think it's ok I am fine as well.

3phase: @skrustev @wnvko I have logged a separate issue (#7615) to address that problem.

Issue #5486, opened 2019-07-23
Add support for mixing px and % column widths


mpavlinov:
## Is your feature request related to a problem? Please describe.
Right now we have the following Known limitation [Currently we do not support mixing of column widths with % and px](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/grid.html#known-limitations)

- [x] Add support for Column Resizing and AutoSizing of % widths. - https://github.com/IgniteUI/igniteui-angular/pull/7662
- [x] Add support for Multi-Column Headers with % widths and mixed widths. - https://github.com/IgniteUI/igniteui-angular/pull/7625
    
1. Run sample at: https://stackblitz.com/edit/angular-butmkq
2. For the column with % width (ContactName), body cell has correct width, while header has much bigger width. This is resulting in misalignment between header and body cells.

- [x] Update related specifications for [Column Resizing](https://github.com/IgniteUI/igniteui-angular/wiki/Column-Resizing) and  [Multi-Column Headers](https://github.com/IgniteUI/igniteui-angular/wiki/Grid-multi-column-headers).
- [x] Remove known issue and document sizing with mixed widths. Section can be added [here](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/sizing.html#column-sizing). PR -> https://github.com/IgniteUI/igniteui-docfx/pull/2045

## Describe the solution you'd like
Introduce support for mixing % and px column widths

## Additional context
https://github.com/IgniteUI/igniteui-angular/issues/5201

PR #7625, opened 2020-06-18
feat(igxGrid): Handle % width for MCH.

MayaKirova:
Related to #5486  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1863,7 +1863,7 @@ export class IgxColumnComponent implements AfterContentInit {
         const colWidth = this.width;
         const isPercentageWidth = colWidth && typeof colWidth === 'string' && colWidth.indexOf('%') !== -1;
         if (isPercentageWidth) {
-            this._calcWidth = parseInt(colWidth, 10) / 100 * (grid.calcWidth - grid.featureColumnsWidth());
+            this._calcWidth = parseInt(colWidth, 10) / 100 * grid.calcWidth;
```
skrustev, line 1866: Why do you remove the feature columns width in this case? In this case the user will need to take into consideration the row selection column for example, because his defined 100% column widths will render a scrollbar.

MayaKirova: @skrustev When width is in % and the related flex styles are applied (flex-basis, min/max width etc.) they are resolved based on the whole container width. So the actual rendered element widths will be the column percentage width *  the grid container width.

The previous calculation was wrong as it did not reflect the size of the % width elements as rendered in the DOM. This was not as apparent before since the calcWidth was only used for internal horizontal virtualization calculations. However since it now applies the size for mch header elements the misalignment between the header and content cells is more obvious (you can refer this test -  'Should calculate column width when a column has width in % and row selectors are enabled' , which reflects this scenario and shows the misalignment).

This should also fix the following scenario:
https://stackblitz.com/edit/angular-dba7e7?file=src/app/grid/grid-sample-selection/grid-selection.component.html
Where all columns have widths in % and due to the row selector the virtualization calculations are wrong and the last column is cut off.

skrustev: I see, I guess in the context that columns are sized based on the whole grid size and not scrollable area it is fine, since I saw that other features also treat it that way.

PR #7627, opened 2020-06-18
fix(drop-down): remove shadow when list is empty #3977

### LINKED ISSUE #3977: There is a shadow, when autocomplete list is empty
## Description  
When autocomplete is filtered so that there are no items left, there is still a shadow.

## Attachments  
image

Lipata:
Closes #3977

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/drop-down/drop-down.component.html
+++ b/projects/igniteui-angular/src/lib/drop-down/drop-down.component.html
@@ -1,5 +1,10 @@
-<div [ngClass]="listClass()" igxToggle [style.width]="width"
-(onOpening)="onToggleOpening($event)" (onOpened)="onToggleOpened()" (onAppended)="onToggleContentAppended()"
+<div [ngClass]="{
+    'igx-drop-down__list': items.length > 0,
+    'igx-drop-down__list--empty': items.length === 0 }"
+[style.width]="width"
+igxToggle
+(onAppended)="onToggleContentAppended()"
+(onOpening)="onToggleOpening($event)" (onOpened)="onToggleOpened()"
 (onClosing)="onToggleClosing($event)" (onClosed)="onToggleClosed()">
     <div class="igx-drop-down__list-scroll" #scrollContainer [attr.id]="this.listId" role="listbox"
     [style.height]="height"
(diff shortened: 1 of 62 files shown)


projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -524,6 +524,13 @@ export class IgxDropDownComponent extends IgxDropDownBaseDirective implements ID
         }
     }
 
+    /**
+     * @hidden @internal
+     */
+    public listClass() {
```
kdinev, line 530: Getter?

projects/igniteui-angular/src/lib/drop-down/drop-down.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="igx-drop-down__list" igxToggle [style.width]="width"
+<div [ngClass]="listClass()" igxToggle [style.width]="width"
```
simeonoff, line 1: I would drop the ngClass directive in favor of host-binding the two classes.

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
simeonoff: I think I would either change the return signature of this method to be an object with the keys holding the class names where the values are expressions that evaluate to boolean based on whether the list has items in it or not.
The other option is to make a getter like @kdinev suggested, which returns the length of the list and to which the `--empty` class gets host-bound.

Lipata: @kdinev, @simeonoff I've now applied the comments from @rkaraivanov: https://github.com/IgniteUI/igniteui-angular/pull/7628#discussion_r444686832. @simeonoff, as you can see the value, is an object with keys holding the class names. However, in that case, where we have more like an `if-else` case, maybe it is clear to use that expression `[ngClass]="{ items.length > 0 ?  'igx-drop-down__list': 'igx-drop-down__list--empty'}"` ?

@simeonoff, as you've also said the other option is to have a getter with `@HostBinding('attr.class')`. It seems both options can be applied, and I prefer the first one only because the expression that is used is short and don't need to be in the .ts file.

PR #7629, opened 2020-06-18
fix(mask): set default placeholder properly

Lipata:
WARN: 'IgxMaskDirective.placeholder: "placeholder" is deprecated, use native placeholder instead.' is removed

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
@@ -153,8 +153,9 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
 
     /** @hidden */
     public ngOnInit(): void {
-        this.renderer.setAttribute(this.nativeElement, 'placeholder',
-            this.placeholder ? this.placeholder : this.maskOptions.format);
+        if (!this.nativeElement.placeholder) {
+          this.renderer.setAttribute(this.nativeElement, 'placeholder', this.maskOptions.format);
+        }
     }
 
     /**


projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -153,8 +153,9 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
 
     /** @hidden */
     public ngOnInit(): void {
-        this.renderer.setAttribute(this.nativeElement, 'placeholder',
-            this.placeholder ? this.placeholder : this.maskOptions.format);
+        if (!this.nativeElement.placeholder) {
+          this.renderer.setAttribute(this.nativeElement, 'placeholder', this.maskOptions.format);
```
kdinev, line 157: But the placeholder option is just deprecated and not removed yet, correct? Cause you're stop using it completely in this change, which means that you're effectively removing it.

damyanpetev: The mask's `placeholder` is merely a proxy to the native attribute - added specifically to intercept it as input and then manually assign the native one anyways.

The deprecation removed the input decorator (leaving the native assign to go through). This fixes the mask self-triggering the deprecation warning.

And yes, the idea is to remove the native public prop or at least make it private.

Lipata: @kdinev let's remove the warning with that fix, we will remove all deprecated properties for 10.0.0 in another PR?

PR #7663, opened 2020-06-23
Add hammerjs import in test.ts for ng add

### LINKED ISSUE #7515: hammerjs test import missing via ng add igniteui-angular
## Description      
When building a project using IgniteUI for Angular CLI, the hammerjs is imported into the two angular.json script arrays - for building and testing. When installing IgniteUI for Angular with "ng add igniteui-angular", hammerjs is imported into the main.ts file, causing depending tests to fail.
 
IvayloG:
Closes #7515  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/schematics/utils/dependency-handler.ts
```diff
@@ -59,16 +59,28 @@ function getTargetedProjectOptions(project: WorkspaceProject<ProjectType>, targe
...
 }
 
-export function getConfigFile(project: WorkspaceProject<ProjectType>, option: string): string {
-    const buildOptions = getTargetedProjectOptions(project, 'build');
-    if (!buildOptions[option]) {
-        throw new SchematicsException(`Could not find the project ${option} file inside of the ` +
-            `workspace config (${project.sourceRoot})`);
-    }
+export function getConfigFile(project: WorkspaceProject<ProjectType>, option: string, configSection: string): string {
+    switch (configSection) {
```
ViktorSlavov, line 63: Does this have to be a `switch` statement? Currently, the only thing it's serving is providing an "enum" for what we _except_ to be present in the project options (we're not even checking for it). You can change it to the following (I'd even get rid of the **section** error)
```suggestion
    const options = getTargetedProjectOptions(project, configSection);
    if (!options) {
        throw new SchematicsException(`Could not find matching ${configSection} section` +
            `inside of the workspace config ${project.sourceRoot} `);
    }
    if (!options[option]) {
        throw new SchematicsException(`Could not find the project ${option} file inside of the ` +
            `workspace config ${project.sourceRoot}`);
    }
    return options[option];
```

projects/igniteui-angular/schematics/utils/dependency-handler.ts
```diff
@@ -59,16 +59,28 @@ function getTargetedProjectOptions(project: WorkspaceProject<ProjectType>, targe
     throw new SchematicsException(`Cannot determine the project's configuration for: ${target}`);
 }
 
-export function getConfigFile(project: WorkspaceProject<ProjectType>, option: string): string {
-    const buildOptions = getTargetedProjectOptions(project, 'build');
-    if (!buildOptions[option]) {
-        throw new SchematicsException(`Could not find the project ${option} file inside of the ` +
-            `workspace config (${project.sourceRoot})`);
-    }
+export function getConfigFile(project: WorkspaceProject<ProjectType>, option: string, configSection: string): string {
```
ViktorSlavov, line 62: I'd add `configSection` w/ a default value, (since, up until now, it's always been used for the 'build' section), so that the changes to the `ng-add/index.ts` file won't be needed
```suggestion
export function getConfigFile(project: WorkspaceProject<ProjectType>, option: string, configSection: string = 'build'): string {
```

projects/igniteui-angular/schematics/utils/dependency-handler.ts
```diff
@@ -148,15 +160,25 @@ function includeDependencies(pkgJson: any, context: SchematicContext, tree: Tree
...
                 // if there are no elements in the architect.build.options.scripts array that contain hammerjs
                 // and main.ts does not contain an import with hammerjs
                 if (!projectOptions.scripts.some(el => el.includes('hammerjs')) && !mainTsContent.includes(hammerImport)) {
                     // import hammerjs in the main.ts file
-                    const contents = hammerImport + mainTsContent;
-                    tree.overwrite(mainTsPath, contents);
+                    const mainContents = hammerImport + mainTsContent;
+                    tree.overwrite(mainTsPath, mainContents);
+                }
+                // make sure test.ts has hammerjs import even for projects already containing it in main.ts
```
ViktorSlavov, line 176: Since there is a lot of repetition here, I'd extract this in a separate function, something like (name pending):
```suggestion
function addHammerToConfig(project: WorkspaceProject<ProjectType>, tree: Tree, config: string) {
    const projectOptions = getTargetedProjectOptions(project, config);
    const tsPath = getConfigFile(project, 'main', config);
    const hammerImport = 'import \'hammerjs\';\n';
    const tsContent = tree.read(tsPath).toString();
    // if there are no elements in the architect[config]options.scripts array that contain hammerjs
    // and the "main" file does not contain an import with hammerjs
    if (!projectOptions.scripts.some(el => el.includes('hammerjs')) && !tsContent.includes(hammerImport)) {
        // import hammerjs in the main file
        const mainContents = hammerImport + tsContent;
        tree.overwrite(tsPath, mainContents);
    }
}
```
Might be a bit too much, but it's there's more granularity this way. Thoughts?

projects/igniteui-angular/schematics/utils/dependency-handler.ts
IvayloG: Yes, this remove a bit of the duplication, and configSection could be further improved somewhat in the future, containing an enum with the meaningful sections in angular.json

projects/igniteui-angular/schematics/utils/dependency-handler.ts
IvayloG: It seems nice, although I like a bit passing the 'build'/'test'  param for readability;
Ref: getConfigFile(project, 'main', 'build');

ViktorSlavov: Yeah, you're right, without meaningful `enum`s they're just *some strings* in the moment - leaving them is fine too!

PR #7696, opened 2020-06-26
Mpopov/indigo navdrawer

desig9stein:
 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


didimmova:
Wouldn't it be a better approach for the light theme to use the igx-contrast-color() function or at least the surface color like in the dark theme instead of pure white?

StefanIvanov:
I think @didimmova makes a very good recommendation here, @desig9stein could you implement the styles via the recommended approach?

desig9stein:
> I think @didimmova makes a very good recommendation here, @desig9stein could you implement the styles via the recommended approach?

@StefanIvanov There are two approaches here, one with the contrast function and one with the surface color. I personally think that using the surface color for text color is wrong for both dark and light versions. I am ok with the contrast function approach.

StefanIvanov:
> > I think @didimmova makes a very good recommendation here, @desig9stein could you implement the styles via the recommended approach?
> 
> @StefanIvanov There are two approaches here, one with the contrast function and one with the surface color. I personally think that using the surface color for text color is wrong for both dark and light versions. I am ok with the contrast function approach.

The contrast color was created particularly for text on background scenarios and I also felt it to be more appropriate, let's use that one.

projects/igniteui-angular/src/lib/core/styles/components/navdrawer/_navdrawer-theme.scss
```diff
@@ -183,7 +185,9 @@
         transition: width, transform;
         transition-timing-function: $ease-out-quad, $ease-out-quad;
         box-shadow: --var($theme, 'shadow');
-        border-#{$right}: 1px solid --var($theme, 'border-color');
+        @if $variant != 'indigo-design'{
```
simeonoff, line 188:
```suggestion
        @if $variant != 'indigo-design' {
```

PR #7760, opened 2020-07-08
feat(grid): column-actions component implementation

### LINKED ISSUE #6549: Provide "Unfreeze All" option under Freeze button on Data Grid & Tree Grid
User is requesting to provide "Unfreeze All" option under Freeze button on Data Grid & Tree Grid. 

User is requesting to provide "Unfreeze All" option under Freeze button on Data Grid & Tree Grid.  it is similar to "Show All" & "Hide All" button provided on Hide Button on both grids.

ChronosSF:
Closes #6549   

### Remaining work
- [x] ARIA support - spec and implementation
- [x] 31 failing tests in grid 
- [x] 2 failing tests in hgrid
- [x] testing as part of the toolbar
- [x] testing as standalone component
- [x] testing navigation to work as before (and explaining it in the spec)
- [x] finalizing spec
- [ ] new document explaining how to use it standalone and with different directives
- [ ] changing the current hiding and pinning documents to replace the old ones
- [ ] creating a sample for the topic

- ~~[ ] possibly migrations?~~


MayaKirova:
Issues:

- [ ] After a filter is applied to the column list via the filter input the hide all/show all or pin all/unpin all actions are now applied to all columns, not just the ones that match the filter criteria. In previous version the selected action was applied only to the columns that match the filter criteria.
This is a behavioral change and show either be fixed so that it behaves as before or should be documented in the spec and changelog.

projects/igniteui-angular/src/lib/core/styles/themes/_core.scss
```diff
@@ -33,6 +33,7 @@
 @import '../components/carousel/carousel-component';
 @import '../components/checkbox/checkbox-component';
 @import '../components/chip/chip-component';
+@import '../components/column-actions/column-actions-component';
 @import '../components/column-hiding/column-hiding-component';
```
simeonoff, line 37: If you are to copy the column hiding theme, then we should also remove the column hiding source code as it's essentially just more of the same.

PR #7766, opened 2020-07-09
Make onSearchInput cancellable

### LINKED ISSUE #7282: igxCombo component has to include caseSensitive property in filter search
image

PlamenaMiteva:
#7282  

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes for each version of this project will be documented in this
 ### General
 - `igxCombo`
     - **Behavioral Change** - Change default positioning strategy from `ConnectedPositioningStrategy` to `AutoPositionStrategy`. The [`Auto`](https://www.infragistics.com/products/ignite-ui-angular/angular/components/overlay_position.html#auto) strategy will initially try to show the element like the Connected strategy does. If the element goes out of the viewport Auto will flip the starting point and the direction, i.e. if the direction is 'bottom', it will switch it to 'top' and so on. If after flipping direction the content goes out of the view, auto strategy will revert to initial start point and direction and will push the content into the view. Note after pushing the content it may hide the combo's input.
+    - Make `onSearchInput` event cancellable. The event args type has been changed to `IComboSearchInputEventArgs`, which have two properties: `change` - holds the change in the search input and `cancel` - indicates whether the event should be canceled.
```
Lipata, line 10: Please rename "change" to "value" or similar, which will not lead to the understanding that arguments hold the difference from the previous time the event was fired.

PlamenaMiteva: Renamed to 'searchTerm'

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -483,7 +489,7 @@ export class IgxComboComponent extends DisplayDensityBase implements IgxComboBas
      * ```
      */
     @Output()
-    public onSearchInput = new EventEmitter();
+    public onSearchInput = new EventEmitter<IComboSearchInputEventArgs>();
```
Lipata, line 492: Let's extend also the IBaseEventArgs (like the `onAddition` event). Also the `searchTerm` should be renamed to 'value'

PR #7772, opened 2020-07-10
Combo caseSensitive filtering

### LINKED ISSUE #7282: igxCombo component has to include caseSensitive property in filter search

PlamenaMiteva:
Closes #7282  

 - [ ] Bug fix
 - [X] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -144,6 +153,10 @@ export class IgxComboComponent extends DisplayDensityBase implements IgxComboBas
      * @hidden @internal
      */
     public defaultFallbackGroup = 'Other';
+    /**
+     * @hidden @internal
+     */
+    public caseSensitive = false;
```
Lipata, line 159: Please create an object `filtertingOptions` that for the moment will hold only `caseSensitive` option. For the moment we will not expose this as an input.

PlamenaMiteva: I have created the filteringOptions object an an interface IComboFilteringOptions

CHANGELOG.md
```diff
@@ -67,6 +67,8 @@ The following example shows how you can use the Indigo theme:
     - `activeViewChanged` event emitted after the active view (DEFAULT, YEAR, DECADE) is changed after user interaction.
     - `viewDate` day value is always 1.
     - `activeView` setter is now available as an input property.
+- `IgxCombo`
+    - Added `showCaseIcon` to display a case sensitive search icon in the search input. Icon click allows the user to easily toggle the search case sensitivity.
```
IvayloG, line 71: This may have already being discussed, yet I would prefer a bit longer yet more meaningful description for the property `showCaseIcon` --> `showMatchCaseIcon` OR `showCaseSensitiveIcon`

PlamenaMiteva: Renamed to showSearchCaseIcon

IvayloG, line 71: One last place to replace `showCaseIcon` with `showSearchCaseIcon`

PlamenaMiteva: Done

src/app/combo/combo.sample.html
```diff
@@ -39,7 +39,7 @@ <h5 class="sample-title">Reactive Form</h5>
                     </form>
                 </div>
                 <div class="input-row">
-                    <igx-combo #playgroundCombo class="input-container" [placeholder]="'Locations'"
+                    <igx-combo #playgroundCombo class="input-container" [placeholder]="'Locations'" [showSearchCaseIcon]="'true'"
```
IvayloG, line 42: [showSearchCaseIcon]="'true'" should be passed as a boolean like [showSearchCaseIcon]='true'

projects/igniteui-angular/src/lib/combo/combo.pipes.ts
```diff
@@ -13,18 +14,21 @@ import { DefaultSortingStrategy } from '../data-operations/sorting-strategy';
     name: 'comboFiltering'
 })
 export class IgxComboFilteringPipe implements PipeTransform {
-    public transform(collection: any[], searchValue: any, displayKey: any, shouldFilter: boolean) {
+    public transform(collection: any[], searchValue: any, displayKey: any,
```
IvayloG, line 17: Currently, when clicking the `showSearchCaseIcon` while there is already a search criteria typed, the icon toggles, yet the pipe transform method is not called and thus the dropdown list result items are not updated to respect the toggled caseSensitive value.  It is not the major use case, yet one would expect the items are updated in this case as well. Adding gif-s for comparison. 
showSearchCaseIcon-no-update-on-icon-click-igxCombo
showSearchCaseIcon-update-on-icon-click-igxCombo

I have played a bit and could force the pipe to work calling it's transform method manually and passing the required arguments:

```
<igx-icon fontSet="case-sensitive" name="case-sensitive" [isActive]="filteringOptions.caseSensitive"
    (click)="**toggleAndUpdate(searchValue, displayKey, filterable, filteringOptions**)">
</igx-icon>
```

```
    public toggleAndUpdate( searchValue, displayKey, filterable, filteringOptions) {
        this.filteringOptions.caseSensitive = !this.filteringOptions.caseSensitive;
        // If you mutate the array, no pipe is invoked and the display isn't updated;
        // if you replace the array, the pipe executes and the display is updated.
        this.data = cloneArray(this.data);
        this.filteringPipe.transform(this.data, searchValue, displayKey, filterable, filteringOptions);
    }
```
Notice, a new data collection should be passed, as change detection will not like the old one being mutated. This is required so the list items actually rerenders with the correct items returned from the pipe.

PR #7777, opened 2020-07-13
feat(grid): Support for binding to nested data source

### LINKED ISSUE #7083: Support for nested properties
Currently the data grid only supports flat data. 

You can currently use cell templates for more complicated data but then filtering and sorting are not supported so that isn't really viable where you just want to pull through 'normal' data. 

You could flatten the data first but if you are using the grid for editing data then this becomes a bit complicated as you then have to then try and transform it back to its original shape before submitting it back.

It would be good if instead it could support nested objects. So for example for an object like:
{teamName: 'team1', teamId: 1, reportingDetails: { area: 'TC', region: 'LDN'}, region: 'IND'}
You could use field in the column as reportingDetails.area
This should support filtering, sorting, cell edits etc

rkaraivanov:
Closes #7083   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


ddincheva:
The following problems are still reproducible :

- [ ] when editing a row in the grid the number of performed changes is not correctly counted;
- [ ] when editing a row the style of the modified cell is not applied correctly;
- [ ] when try to enter edit mode in the hierarchical grid the whole page freeze;/

       Open hierarchicalGrid dev sample;
       Enable row editing for the first or second grid for example;
       Try to enter edit mode/

PR #7782, opened 2020-07-13
feat(grid): Add Column Sorting Indicators

### LINKED ISSUE #7493: Sorting Indicator on all Grid components
When user apply sorting on multiple columns on grid components, it is difficult to know which is first, second column. User has to remember the order of sorting on columns. 

There should sorting indicator on grid column. When user perform first sorting column should show 1. it is rounded and highlighted. then if user do second sorting it should show 2 along with Arrow. This way it will be easy to understand how sorting is applied on multiple columns. 
i am looking for Index indicator on column header itself along with Arrow.

HristoP96:
Closes #7493


rkaraivanov:
@SisIvanova @simeonoff 
Can we make the background of the `:after` pseudo-element transparent?
Currently it looks like this:
image
and while sorting with 10+ columns seems unlikely, it will be nice to have in that one edge case.

StefanIvanov:
@rkaraivanov currently it is by design and follows our guidance taking care of the edge case with two-digit indices

projects/igniteui-angular/src/lib/grids/headers/sorting-index.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+
+@Pipe({
+  name: 'sortingIndex',
+  pure: true
+})
+export class SortingIndexPipe implements PipeTransform {
+
+  constructor(private igxHeader: IgxGridHeaderComponent) {}
+  transform(columnField: string, sortingExpressions: ISortingExpression[]): number {
+    const grid = this.igxHeader.grid;
```
rkaraivanov, line 13: If you need the grid instance in the pipe just pass it as a parameter in the pipe `transform` method.
Don't see the point of injecting the header component just to get a grid instance.

projects/igniteui-angular/src/lib/grids/headers/sorting-index.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+    let sortIndex = sortingExpressions.filter(expression => grid.getColumnByName(expression.fieldName) ?? false)
+                                      .findIndex(expression => expression.fieldName === columnField);
```
rkaraivanov, line 15:
```suggestion
    let sortIndex = sortingExpressions.findIndex(expression => expression.fieldName === columnField);
```

HristoP96: @rkaraivanov I added the filter function in order to clear out the expressions, which have wrong field names, otherwise the sorting indicators will have wrong indexes

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -1504,6 +1504,19 @@
             height: rem(15px);
             min-width: rem(15px); /* yeah IE, it really needs to be 15px wide... */
             font-size: rem(15px);
+            position: relative;
+
+            &::after {
+                content: attr(data-sortIndex);
+                position: absolute;
+                top: -5px;
+                right: -1px;
```
simeonoff, line 1513: @SisIvanova You should use the interpolated value of `#{$right}` here to ensure RLT compatibility is preserved.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -1504,6 +1504,19 @@
...
+                font-size: rem(10px);
+                text-align: right;
+                font-family: sans-serif;
+                line-height: .625rem;
```
simeonoff, line 1517: @SisIvanova Please use the `rem` function so the resulting value takes the global `$browser-context` value into account.

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -1504,6 +1504,19 @@
...
             font-size: rem(15px);
+            position: relative;
+
+            &::after {
+                content: attr(data-sortIndex);
+                position: absolute;
+                top: -5px;
+                right: -1px;
+                font-size: rem(10px);
+                text-align: right;
```
simeonoff, line 1515: @SisIvanova `right` here should also be based on the value of the `$right` variable.

projects/igniteui-angular/src/lib/grids/headers/sorting-index.pipe.ts
rkaraivanov: @HristoP96 
Our API calls generate the correct sorting expression objects. The only way the column field would be different in the expression is if the user misspells the field name in which case I don't think we should be a spell checker 😃

PR #7793, opened 2020-07-14
Consecutevly close overlays on escape keypress

### LINKED ISSUE #7697: Do not close modal overlay on ESC key press
There is no way to stop closing of `IgxDialog` on `ESC` key press. This is valid also for all components using `IgxOverlay` in modal mode.

Allow configuration of the overlay allowing to discard `ESC` key press. This could be added in `OverlaySettings`. We have now `closeOnOutsideClick` property. There should be and something like `closeOnEsc`.

jackofdiamond5:
Closes #7697

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -476,11 +477,22 @@ export class IgxOverlayService implements OnDestroy {
     }
 
     private setUpCloseOnEscape(info: OverlayInfo) {
+        if (!info.settings.modal) {
```
wnvko, line 480: this will not work!
If I have two overlays like this:
`Overlay1` no-modal closeOnEsc=true
`Overlay2` no-modal closeOnEsc=false
If I open `Overlay1` and then `Overlay2` I will end up with `kyedown` event attached to the `document`. If I click `ESC` key it will close `Overlay2`
Two things to fix here:
1. The check for `closeOnEsc` should be in the event handler. Each overlay should add handler with or without `closeOnEsc` set to true.
2. Check the logic in `addOutsideClickListener` and do the same.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -330,9 +331,7 @@ export class IgxOverlayService implements OnDestroy {
             wrapperElement.classList.add('igx-overlay__wrapper--modal');
         }
 
-        if (info.settings.closeOnEsc) {
-            this.setUpCloseOnEscape(info);
-        }
+        this.setUpCloseOnEscape();
```
wnvko, line 334: rename this to `addCloseOnEscapeListener` just to be consistent with the other two methods and move the call just after `addResizeHandler`
Let's make our code better :)

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -475,12 +474,23 @@ export class IgxOverlayService implements OnDestroy {
...
-        const wrapperElement = info.elementRef.nativeElement.parentElement.parentElement;
-        fromEvent(wrapperElement, 'keydown').pipe(
+    private setUpCloseOnEscape() {
+        if (this._overlayInfos.length - this._overlayInfos.filter(x => x.closeAnimationPlayer
+            && x.closeAnimationPlayer.hasStarted()).length === 1) {
+            this.subscribeOnEscape(this._document);
+        }
+    }
+
+    private subscribeOnEscape(target: FromEventTarget<unknown>) {
```
wnvko, line 484: we do not need this method anymore. Move this logic in `setUpCloseOnEscape` (or `addCloseOnEscapeListener`)

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -700,6 +710,14 @@ export class IgxOverlayService implements OnDestroy {
         }
     }
 
+    private hasRemainingOverlaysAfterClose(): boolean {
```
wnvko, line 713: leftover

projects/igniteui-angular/src/lib/services/overlay/overlay.spec.ts
```diff
@@ -2753,11 +2735,12 @@ describe('igxOverlay', () => {
             overlayWrapper.addEventListener('keydown', (event: KeyboardEvent) => {
                 if (event.key === targetButton) {
                     overlayWrapper = document.getElementsByClassName(CLASS_OVERLAY_WRAPPER_MODAL)[0];
+                    expect(overlayWrapper).toBeFalsy();
```
wnvko, line 2738: I really do not like this way of testing. You can check the count of the overlays attached to the document

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -735,6 +726,22 @@ export class IgxOverlayService implements OnDestroy {
         }
     }
 
+    private addCloseOnEscapeListener() {
+        //  if all overlays minus closing overlays equals one add the handler
+        if (this._overlayInfos.length - this._overlayInfos.filter(x => x.closeAnimationPlayer
+            && x.closeAnimationPlayer.hasStarted()).length === 1) {
+            fromEvent(this._document, 'keydown').pipe(
```
damyanpetev, line 733: Right, so once this subscription is active, there's no mechanism to remove it (outside of destroy, which is basically never for a root service)? I don't like a key handler to be permanently active TBH. Also the check is all wrong - length alone doesn't tell you much, where's the `closeOnEsc` before setting the listener? And won't this keep adding subscriptions if just one overlay is shown at a time (length === 1)?

Don't get me wrong, the old version with subscribing to the wrapper is also no good as it never unsubscribed and rxjs won't do that for removed elements either. If you want to use the event observable - do store subscription, check `closeOnEsc` before even setting it up and use similar logic to unsubscribe like outside click.

projects/igniteui-angular/src/lib/dialog/dialog.component.ts
```diff
@@ -83,6 +83,7 @@ export class IgxDialogComponent implements IToggleView, OnInit, OnDestroy, After
         this._isModal = val;
     }
 
+    @Input()
```
damyanpetev, line 86: Umm what is this? Where the API description comment? What's the default? Mark as public as well.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
wnvko: Unsubscribing is fixed now.
This will never subscribe more than once. When calling `show` for the first overlay it will subscribe. Then if you close this overlay and during its close animation open a new one two things will happen. First in `hide` call subscription will be removed. Then in `show` there will be still two overlays active, and one with closing animation playing - it will subscribe which is correct.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -735,6 +723,30 @@ export class IgxOverlayService implements OnDestroy {
...
+            });
+        }
+    }
+
+    private removeCloseOnEscapeListener() {
+        const closingOverlaysCount =
+        this._overlayInfos
+            .filter(o => o.closeAnimationPlayer && o.closeAnimationPlayer.hasStarted())
+            .length;
+        if (this._overlayInfos.length - closingOverlaysCount === 1 && this._keyPressEventListener && !this._keyPressEventListener.closed) {
```
damyanpetev, line 744: Don't think the `closed` check does much if you are de-ref the subscription property anyway. Is there a way for it to be still assigned yet be closed?

valadzhov:
Verified that the PR is working but logged https://github.com/IgniteUI/igniteui-angular/issues/7803 related to the overlay not closing while another overlay's closing animation is in progress.

PR #7794, opened 2020-07-14
Expose grid active node evt

### LINKED ISSUE #7601: exposing onActiveNodeChange output in the Grid
The event would be really helpful in a situation where the user would like to identify which item in the grid is currently active.
One example is the development of the [keyboard navigation sample](https://staging.infragistics.com/products/ignite-ui-angular/angular/components/grid/keyboard_navigation.html#demo), where we are literally using plenty of the grid API so we can take the active item as a reference to enable/disable particular navigation commands.

The event could contribute to a more centralized, tidied, and concise code.

Emitting the event every time the active node is changed, providing info about the container and the cell.

Aleksandyr:
Closes #7601 

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -428,34 +502,43 @@ export class IgxGridNavigationService {
...
-            .filter(col => col.visibleIndex < this.activeNode.mchCache.visibleIndex)[0];
-            this.activeNode.mchCache.level = nextCol.level;
+            .filter(col => col.visibleIndex < newHeaderNode.visibleIndex)[0];
+            newHeaderNode.level = nextCol.level;
         }
-        this.activeNode.column = nextCol.visibleIndex;
-        this.activeNode.level = nextCol.level;
-        this.performHorizontalScrollToCell(this.activeNode.column);
+
+        this.setActiveNode({
```
zdrawku, line 535: Also, can we use interfaces for these type of objects, whenever you create an object we should strive to always ensure type safety. As I see, we have a similar `IActiveNode` interface, where gridID and layout are optional.

You've added the logic of Object.assigned to the setActiveNode, which is nice.

Aleksandyr:
@zdrawku We are using string union types because of the strict template type checking. The compiler does not handle it well when we use enum types. 
Btw we should consider updating all our enums. It is a known issue, but I suppose nobody has logged it yet.

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
Aleksandyr: Actually setActiveNode method does the validation. As you can observe the method validates the arguments itself.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -23,9 +23,10 @@ import { DeprecateMethod } from '../core/deprecateDecorators';
 import { HammerGesturesManager } from '../core/touch';
 import { ColumnType } from './common/column.interface';
 import { RowType } from './common/row.interface';
-import { GridSelectionMode } from './common/enums';
+import { GridSelectionMode, GridKeydownTargetType } from './common/enums';
 import { GridType } from './common/grid.interface';
 import { ISearchInfo } from './grid/public_api';
+import { retry } from 'rxjs/operators';
```
ddincheva, line 29: This is unused import .. so you can delete it.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -885,11 +887,13 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
...
     private setActiveNode() {
+        const currRow = this.grid.dataView[this.rowIndex];
+        const type: GridKeydownTargetType = this.grid.isDetailRecord(currRow) ? 'masterDetailRow' : 'dataCell';
         if (this.grid.navigation.activeNode) {
-            Object.assign(this.grid.navigation.activeNode, {row: this.rowIndex, column: this.visibleColumnIndex});
+            this.grid.navigation.setActiveNode({ row: this.rowIndex, column: this.visibleColumnIndex}, type);
         } else {
             const layout = this.column.columnLayoutChild ? this.grid.navigation.layout(this.visibleColumnIndex) : null;
-            this.grid.navigation.activeNode = { row: this.rowIndex, column: this.visibleColumnIndex, layout: layout };
+            this.grid.navigation.setActiveNode({ row: this.rowIndex, column: this.visibleColumnIndex, layout: layout }, type);
```
ddincheva, line 896: is this if clause still necessary ... since in both statements we use setActiveNode method and if any additional logic is necessary is better to be moved inside setActiveNode method.

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -148,45 +151,53 @@ export class IgxGridNavigationService {
...
+            this.setActiveNode({ row: 0, column: 0 }, type);
             this.grid.navigateTo(0, 0, (obj) => {
                 this.grid.clearCellSelection();
                 obj.target.activate(event);
             });
         }
     }
 
-    focusFirstCell(header = true) {
+    focusFirstCell(header = true, footer = false) {
```
ddincheva, line 193: why do you need to introduce this new parameter... since in the context of this method if the header param return false means that you are in the footer context. Can you think to remove the footer param?

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -335,16 +351,74 @@ export class IgxGridNavigationService {
             && !curRow.childGridsData && (includeSummary || !curRow.summaries);
     }
 
+    public setActiveNode(activeNode: IActiveNode, tag: GridKeydownTargetType) {
```
ddincheva, line 354: isn't it a better idea to make the tag parameter as optional and if is not passed to make here the check what is the tag parameter for each element, because from I can see from the code above you make a check on a few other places and in this way will be much more clean and easy to maintain.

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -299,7 +310,12 @@ export class IgxGridNavigationService {
 
     protected navigateInBody(rowIndex, visibleColIndex, cb: Function = null): void {
         if (!this.isValidPosition(rowIndex, visibleColIndex) || this.isActiveNode(rowIndex, visibleColIndex)) { return; }
-        this.grid.navigateTo(this.activeNode.row = rowIndex, this.activeNode.column = visibleColIndex, cb);
+        const currRow = this.grid.dataView[rowIndex];
+        const type: GridKeydownTargetType = this.isDataRow(rowIndex) ? 'dataCell' :
+            this.grid.isGroupByRecord(currRow) ? 'groupRow' :
+            this.grid.isDetailRecord(currRow) ? 'masterDetailRow' : 'summaryCell';
+        this.setActiveNode({ row: rowIndex, column: visibleColIndex}, type);
```
ddincheva, line 317: this check maybe is better to be inside setActiveNode method.

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
Aleksandyr: Because this method is used for the tBody as well, so I need to differentiate where the activeNode is exactly, otherwise when the focus is over the grid or over the trial watermark the method tracks the navigation as it is over the summary section.

Aleksandyr: I've got what you mean. I will remove the footer parameter and I'm gonna use the header insted.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6393,6 +6411,7 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
      * @internal
      */
     public endRowEdit(commit = true, event?: Event) {
+        console.log('end row');
```
ddincheva, line 6414: You can remove this log :)

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -821,9 +821,11 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
      * @hidden
      * @internal
      */
-    public activate(event: FocusEvent | KeyboardEvent) {
+    public activate(event: FocusEvent | KeyboardEvent, position) {
```
ddincheva, line 824: activate method is called in the scope of the cell component. We do not have a case where we call this method on one cell in order to activate another. So why is necessary to pass position arguments since in the code below you can just pass this.rowIndex and this.visibleColumnIndex as params to the setActiveNode method.
 this.grid.navigation.setActiveNode({ row: this.rowIndex, column: this.visibleColumnIndex });

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
ddincheva, line 6414: Please remove this logger :)

PR #7843, opened 2020-07-22
fix(tooltip): Handle closeOnEsc via overlay #7638

### LINKED ISSUE #7638: [igxTooltipTarget] [performance] Permanently attached document listeners can cause the UI to become unresponsive with large number of instances
## Description  
The `igxTooltipTarget` attached a `HostListener` directly to the document to listen for Escape key:
https://github.com/IgniteUI/igniteui-angular/blob/473a31a63ff4f15c5f2cbbf25e6eb1ffcf3247fd/projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts#L255

Since that handlers cannot be prevented and are permanently attached, when they start piling up (such as a grid of small items - imagine calendar days with a few months visible, menus with lots of small items or grid cells), hitting escape to close a single visible tooltip will go though all handlers and can block the main thread for a bit. It's been noted that overriding the handler code with empty one doesn't alleviate the issue - i.e. the problem is with triggering the all.

While this was brought up for escape key, I suspect a similar behavior can be reproduced with touch, looking at:
https://github.com/IgniteUI/igniteui-angular/blob/473a31a63ff4f15c5f2cbbf25e6eb1ffcf3247fd/projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts#L359
And since that's touch start it's gonna be even more pronounced.

 * browser: Chromium

## Steps to reproduce  

1. Open https://stackblitz.com/edit/angular-qgmo68
2. Hover over an avatar and hit esc (make sure you have focused the sample iframe)
3. Observe the delay till the tooltip closes
4. Move the slider to above 100/200 and repeat 2,3

## Result  
At around 100 active instances I begin to notice some delay (less than a second), and after 200-300 the delay is very noticeable. A single instance keydown handler takes about 2ms (basically non-blocking). A total of 2 attached document listeners.
Instances | Keydown handler time | document events
---|---|---
25| 28ms|50
50| ⚠️ 76ms|100
100| ⚠️ 265ms|200
200| ⚠️ 964ms|400
300| ⚠️ 1.96s|600

As you can see after 25 we have dropped frames and growing exponentially at 300 the close of single tooltip takes a good 2 seconds
image
with a whole lot of handlers
image

## Expected result
Outside of the initialization of such large number of directives, there shouldn't really be much delay difference in interacting with a single tooltip.

I have no idea why the tooltip target requires such complicated logic to close, considering it's still using the overlay service that already has a built-in Esc handler (that's only active when showing something). So consider that perhaps or at the very least attach and detach the handlers as needed (perhaps [`renderer.listen`](https://angular.io/api/core/Renderer2#listen)?). The second bit also applies to the touch listener.

## Attachments   
[Profile-20200619T195209.zip](https://github.com/IgniteUI/igniteui-angular/files/4805839/Profile-20200619T195209.zip)

astaev:
Closes #7638

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts
```diff
@@ -196,6 +196,16 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
 
         this._overlayDefaults.positionStrategy = new AutoPositionStrategy(positionSettings);
         this._overlayDefaults.closeOnOutsideClick = false;
+        this._overlayDefaults.closeOnEscape = true;
+
+        this.target.onClosing.subscribe((event) => {
```
jackofdiamond5, line 201: We should remove all subscriptions in `ngOnDestroy` to prevent any potential memory leaks.
Maybe something like this:
```typescript
private destroy$ = new Subject();

this.target.onClosing.pipe(takeUntil(this.destroy$)).subscribe(/* ... */);

public ngOnDestroy(): void {
   this.destroy$.next();
   this.destroy$.complete();
}	
```

astaev: Done

projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts
```diff
@@ -196,6 +200,21 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
...
+            const hidingArgs = { target: this, tooltip: this.target, cancel: false };
+            this.onTooltipHide.emit(hidingArgs);
+
+            if (hidingArgs.cancel) {
+                event.cancel = true;
+            }
+        });
+    }
+
+    ngOnDestroy() {
```
jackofdiamond5, line 215:
```suggestion
     /** @hidden */
    public ngOnDestroy() {
```

astaev: Done

PR #7868, opened 2020-07-27
Make combo filtering pipe pure

### LINKED ISSUE #7282: igxCombo component has to include caseSensitive property in filter search

PlamenaMiteva:
#7282 

 - [ ] Bug fix
 - [X] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/combo/combo.component.html
+++ b/projects/igniteui-angular/src/lib/combo/combo.component.html
@@ -41,7 +41,7 @@
             aria-autocomplete="both" [attr.aria-owns]="dropdown.id" [attr.aria-labelledby]="ariaLabelledBy" />
         <igx-suffix *ngIf="showSearchCaseIcon">
             <igx-icon fontSet="case-sensitive" name="case-sensitive" [isActive]="filteringOptions.caseSensitive"
-                (click)="filteringOptions.caseSensitive=!filteringOptions.caseSensitive">
+                (click)="toggleCaseSensitive()">
             </igx-icon>
         </igx-suffix>
     </igx-input-group>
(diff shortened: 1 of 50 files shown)


projects/igniteui-angular/src/lib/combo/combo.pipes.ts
```diff
@@ -11,12 +11,11 @@ import { IComboFilteringOptions } from './combo.component';
  * @hidden
  */
 @Pipe({
-    name: 'comboFiltering',
-    pure: false
+    name: 'comboFiltering'
 })
 export class IgxComboFilteringPipe implements PipeTransform {
     public transform(collection: any[], searchValue: any, displayKey: any,
-                    shouldFilter: boolean, filteringOptions: IComboFilteringOptions) {
```
kdinev, line 19: You don't need to define the pipeTrigger parameter here. As long as you pass the parameter in the pipe calls, it would trigger.

projects/igniteui-angular/src/lib/combo/combo.pipes.ts
```diff
@@ -11,12 +11,11 @@ import { IComboFilteringOptions } from './combo.component';
...
+                    shouldFilter: boolean, filteringOptions: IComboFilteringOptions, pipeTrigger: number) {
```
kdinev, line 18: You don't need to define the pipeTrigger parameter here. As long as you pass the parameter in the pipe calls, it would trigger.

PlamenaMiteva: Removed!

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -1578,6 +1582,14 @@ export class IgxComboComponent extends DisplayDensityBase implements IgxComboBas
...
         this.onClosed.emit();
     }
+
+    /**
+     * @hidden @internal
+     */
+    public toggleCaseSensitive() {
+        this.filteringOptions.caseSensitive = !this.filteringOptions.caseSensitive;
+        this.filteringPipeTrigger++;
+    }
```
rkaraivanov, line 1592:
```suggestion
        this.filteringOptions = { caseSensitive: !this.filteringOptions.caseSensitive };
```
Changing the reference to the `filteringOptions` object will trigger the pipe. No need for an additional parameter in this case.

projects/igniteui-angular/src/lib/combo/combo.component.html
```diff
@@ -51,7 +51,7 @@
         [style.maxHeight.px]="itemsMaxHeight" [igxDropDownItemNavigation]="dropdown" (focus)="dropdown.onFocus()"
         [tabindex]="dropdown.collapsed ? -1 : 0" role="listbox" [attr.id]="dropdown.id">
         <igx-combo-item role="option" [itemHeight]='itemHeight' *igxFor="let item of data
-            | comboFiltering:searchValue:displayKey:filterable:filteringOptions
+            | comboFiltering:searchValue:displayKey:filterable:filteringOptions:filteringPipeTrigger
```
rkaraivanov, line 54: See below

projects/igniteui-angular/src/lib/combo/combo.component.ts
PlamenaMiteva: Done as suggested

PR #7932, opened 2020-08-05
fix(navDrawer): Set width on open change #6321

### LINKED ISSUE #6321: Nav-drawer doesn't apply width if initial state is closed.
## Description  
Describe the issue.

 * igniteui-angular version: 8.1
 * browser: any

## Steps to reproduce  

1. Create a nav-drawer
2. Set its width to 500px and its `isOpen` to false.
3. Set `isOpen` to `true` to open the drawer

## Result  
The width of the nav-drawer is not 500px.

## Expected result  
The width should be as set.

## Attachments  
https://stackblitz.com/edit/angular-etncmg?embed=1&file=src/app/navdrawer/navdrawer.component.html

VMihalkov:
Closes #6321   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.ts
```diff
@@ -495,6 +495,10 @@ export class IgxNavigationDrawerComponent implements
             this.setDrawerWidth(changes.width.currentValue);
         }
 
+        if (changes.isOpen) {
+            this.setDrawerWidth(this.isOpen ? this.width : '');
```
skrustev, line 499: I think it should take into account also if there is mini drawer templated. Then the width when closed is equal to the `miniWidth`.

VMihalkov: I tested it with mini drawer, which is set initially or in runtime and the width in closed state is correctly set to the miniWidth. I also tried this with setting the miniWidth to a custom value, like 100px, and it works correctly as well. I don't think that any changes are needed in this case.

damyanpetev: Okay, so the issue with the bug (which is not specified) is opening the drawer though the `isOpen` binding, rather than calling `open()` which applies the widths correctly. As it stands opening and closing the drawer though the `isOpen` input is not handled correctly and @skrustev has a point - the else case of the check is basically a close equivalent and this is what it sets:
https://github.com/IgniteUI/igniteui-angular/blob/c8bd808f20a6e20b5d608209d232d6294eabe58d/projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.ts#L566
The reason it currently works is the mini template setter - Angular runs a refresh on child queries _after_ setting component inputs. The setter fixes the width (because the template can be added/removed by angular at any moment), but it's kind of a gimmick to rely on that order.

I'd update that else clause (or you can refactor this into a tidier function) and you might want to ignore the `firstChange` there as `onInit` and the mini setter kind of handle that bit.

PR #7995, opened 2020-08-13
Add validation when disabledDates in date picker is set v.10.1

### LINKED ISSUE #7800: igx-date-picker: DisabledDates property is not used in a Reactive Form
## Description  
 DisabledDates property is not used in a Reactive Form

 * igniteui-angular version: 10.0.3
 * browser:  Chrome 83.0.4103.116

## Steps to reproduce  

1. Open sample [Sample StackBlitz](https://stackblitz.com/edit/github-4j77ke?file=package.jsonl)
2. The button "Book" is disabled as the field "Date" is required and does not have a value.
3. Click on "Change Range" button, the button "Book" is enabled.

## Result  
The disabled dates property is not used to updated the state of the form input.

## Expected result  
The state of the form input is invalid as the value is in the range of the disabled dates.

## Attachments  
https://stackblitz.com/edit/github-4j77ke?file=package.json

PlamenaMiteva:
Closes #7800  

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/date-picker/date-picker.component.spec.ts
+++ b/projects/igniteui-angular/src/lib/date-picker/date-picker.component.spec.ts
@@ -1,6 +1,6 @@
 import { Component, ViewChild, ElementRef, EventEmitter } from '@angular/core';
 import { async, fakeAsync, TestBed, tick, flush, ComponentFixture } from '@angular/core/testing';
-import { FormsModule, FormGroup, FormBuilder, ReactiveFormsModule, Validators } from '@angular/forms';
+import { FormsModule, FormGroup, FormBuilder, ReactiveFormsModule, Validators, NgControl } from '@angular/forms';
 import { By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { IgxDatePickerComponent, IgxDatePickerModule } from './date-picker.component';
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -308,6 +313,10 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
     }
     public set disabledDates(value: DateRangeDescriptor[]) {
         this._disabledDates = value;
+        if (this._ngControl && this._ngControl.control && this._ngControl.control.validator && this._value && this._disabledDates) {
```
jackofdiamond5, line 316: This is not how we should be doing validity checks. We do not need `onChangeCallback`  here as it needs to be called only if the value of the control has changed. Instead, we should implement the `Validator` interface from `@angular/forms` and then register a validator change callback, something like this:

```typescript
private _onValidatorChange = () => { };
public registerOnValidatorChange(fn: any) {
    this._onValidatorChange = fn;
}
```
Then call that `this._onValidatorChange` right after `this_disabledDates` is updated. Everything else can be removed.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -876,9 +891,9 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
         if ((this._ngControl.control.touched || this._ngControl.control.dirty) &&
```
jackofdiamond5, line 891: We should now check if the control is touched or dirty OR it has a validator or async validator. This is because `onStatusChanged` will be invoked right after we update `disabledDates` and the current implementation expects both for the control to be in a specific state AND to have validators. While this was working just fine, now that we've made our DatePicker into a validator, we have to make sure that its validity is updated no matter if it has been touched or not, since you can easily change the `disabledDates` to be outside of a specific range and the control should become invalid.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.spec.ts
```diff
@@ -1332,12 +1332,33 @@ describe('IgxDatePicker', () => {
...
             expect(inputDirective.valid).toEqual(IgxInputState.INITIAL);
 
-            inputDirectiveElement.triggerEventHandler('blur', { target: { value: ''}});
+            inputDirectiveElement.triggerEventHandler('blur', { target: { value: '' } });
             fixture.detectChanges();
 
             expect(inputDirective.valid).toEqual(IgxInputState.INVALID);
         }));
 
+        it('Should set date picker status to invalid when date is disabled', fakeAsync(() => {
```
jackofdiamond5, line 1341: We could add an additional test that checks for essentially the same thing but with a user's template.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -753,29 +765,34 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
...
 
-    /**
-     * @hidden @internal
-     */
-    public getInputGroupElement() {
-        if (this.inputGroup) {
-            return this.inputGroup.element.nativeElement;
+    /** @hidden @internal */
+    public getInputGroupElement(): HTMLElement {
+        if (this._inputGroup) {
```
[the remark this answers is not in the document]
IvayloG: done

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
IvayloG: done

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
IvayloG: Fixed, though a bit differently,
as this suggestion is not entirely true. It will force the component to set IgxInputState when there is no user interaction. For example, this will cause setting IgxInputState.**INVALID** and the corresponding RED-ish styles when there is error.required, and we do not want that.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.spec.ts
IvayloG: done

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
IvayloG: done

jackofdiamond5 (approved):

IvayloG (approved):

wnvko:
The fix does not work when templated `IgxDatePicker` is blurred. `onBlur` is not called for templated component and respectively `_updateValidityOnBlur` is not called too.
We should find a way to handle `onBlur` for the templated input.
Note: this should work also in case the `mode` gets changed at run-time

damyanpetev (approved):

PR #8051, opened 2020-08-27
Implement input group type token

### LINKED ISSUE #6751: Expose way to set the Input Group `type` globally since input components depend on it to apply themes correctly.

When <input-group> is used with a different theme a `type` input needs to be set to the corresponding theme to change the structure of the component in order to apply the theme correctly. It appears that there is no easy way to infer the `type` from the theme. In most cases the user can set the `type` himself on the component. Components however that use the <input-group> internally don’t have a way for the user to set the type, which means that there’s no way out of the box to style them according to the applied theme.

Some such components:
1)	Grid Editors – the editors for edit functionality are by default input-groups/date-picker. Unless the user templates each cell editor and sets the `type` input the editors cannot be styled according to the theme.
2)	Date Picker – uses input-group for both default templates. Does not appear to be a way to set type. 
3)	Time Picker – same as Date Picker.
4)	Combo – this one has its own `type` @Input that is then set to the main input, however there is also a search input in the drop-down (which is enabled when filterable)  that does not seem to be settable based on the theme.

Expose some way to set the Input Group `type` globally so that components that use it internally inherit it from the parent or use the globally set one (maybe something similar to the DisplayDensity implementation or a global service that sets the type).

## Additional context
Related to: https://github.com/IgniteUI/igniteui-angular-samples/issues/1757
Once we have a way to set this globally we can apply it in the Samples Browser.

IvayloG:
Closes #6751   

 - [ ] Bug fix
 - [x] New functionality
 - [x] Documentation
 - [x] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -182,13 +182,21 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
      * An @Input property that sets how the select will be styled.
-     * The allowed values are `line`, `box` and `border`. The default is `line`.
+     * The allowed values are `line`, `box` and `border`. The input-group default is `line`.
      * ```html
      * <igx-select [type]="'box'"></igx-select>
      * ```
      */
-    @Input()
-    public type = 'line';
+    @Input('type')
```
ViktorSlavov, line 190: The explicit naming of the input is not needed

projects/igniteui-angular/src/lib/input-group/input-group.component.ts
```diff
@@ -219,7 +208,7 @@ export class IgxInputGroupComponent extends DisplayDensityBase
      * ```
      */
     public get type() {
-        return this._type;
+        return this._type ? this._type : this._inputGroupType ? this._inputGroupType : this._type = 'line';
```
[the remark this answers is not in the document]
ViktorSlavov: The same change can be done to the `IgxCombo` as well - add the token to the constructor (w\ `@Optional()` decorator). If no token is provided, fall back to the default `'box'` value, to avoid any behavioral changes.

```
private _type = null;
...
public get type(): IgxInputGroupType {
        return this._type || this._inputGroupToken || 'box';
    }

public set type(val: IgxInputGroupType) {
    this._type = val;
}

```

ViktorSlavov (changes requested):

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -182,13 +182,21 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
...
      * An @Input property that sets how the select will be styled.
-     * The allowed values are `line`, `box` and `border`. The default is `line`.
+     * The allowed values are `line`, `box` and `border`. The input-group default is `line`.
      * ```html
      * <igx-select [type]="'box'"></igx-select>
      * ```
      */
     @Input()
-    public type = 'line';
+    public set type(value: IgxInputGroupType) {
```
ViktorSlavov, line 191: This implementation works fine, but I feel like it should be the same as in the `IgxInputGroup` component. If we expose the input to be templateable or if (for some reason) a user passes items that have and `input` in them, those should behave consistently w/ the input group from the view template. So the select component should get the `InputGroupToken` injected in it's template (optional) and `type` should return `this._type || this._inputGroupToken || 'line'`. As we are passing any specified type to the `input` in the template, there should be no inconsistencies when it is provided.
@damyanpetev , thoughts?

ViktorSlavov (approved):

projects/igniteui-angular/src/lib/select/select.component.ts
damyanpetev: I don't think might be an issue, but the input group child may one day not be static and interfere with this setter. So I'd still go that route anyway.

projects/igniteui-angular/src/lib/input-group/inputGroupType.ts
```diff
@@ -0,0 +1,19 @@
...
+    box,
+    border,
+    search,
+}
+
+/**
+ * Defines the InputGroupType DI token.
+ */
+ // Should this go trough Interface https://angular.io/api/core/InjectionToken
+ export const InputGroupToken = new InjectionToken<IgxInputGroupType>('InputGroupType');
```
damyanpetev, line 14: I'd take the more fitting token name of `IGX_INPUT_GROUP_TYPE`

projects/igniteui-angular/src/lib/select/select.component.ts
```diff
@@ -182,13 +183,19 @@ export class IgxSelectComponent extends IgxDropDownComponent implements IgxSelec
 
     /**
      * An @Input property that sets how the select will be styled.
-     * The allowed values are `line`, `box` and `border`. The default is `line`.
+     * The allowed values are `line`, `box` and `border`. The input-group default is `line`.
```
ViktorSlavov, line 186: Nit-pick, but this change isn't really needed

ViktorSlavov (approved):

Issue #5946, opened 2019-10-09
Native add row


Eralmidia:
## Is your feature request related to a problem? Please describe.
It would be nice to have native support for adding a new row. This can of course be accomplished today with the edit row functionality, but it is a bit cumbersome.

For instance in some cases a grid have read only columns, meaning that row edit for those columns are disabled. But when adding a new row, these columns may still need an input. In other cases, a grid might be fully read only, but still have the ability to add new rows, in which case we need to enable/disable row editing when row adding starts/ends.

## Describe the solution you'd like
A distinction between add new row and edit exiting row

Eralmidia:
Just to add the discussion from #5947 to this thread: it would be great if the add row used the same transaction based system as editing. Like many others, we use state managements libraries (ngrx, ngxs, akita ect). These systems stream immutable objects, which in turn typically means that the data source for the grid also is immutable, and edits should result in store updates, rather than manipulation of a local array of data.

Eralmidia:
@kdinev Any chance of including this in the not to distant future? 🙏

kdinev:
@Eralmidia It's actually a good time to ask this question. We're currently building the roadmap for the next iterration.

@radomirchev Can we consider this for the next milestone, or is the milestone full already?

Eralmidia:
@kdinev @radomirchev  Did this make it into the milestone?

radomirchev:
@Eralmidia what we have in the current milestone is Row Pinning #6640 which is prerequisite for a solution for your request.  The development has started this sprint. Once is ready we will start working on the sample. You could follow the sprints schedule and progress [here](https://github.com/orgs/IgniteUI/projects).

Eralmidia:
@radomirchev  I see this was moved to validation in the current sprint. Will it be in 10.2?

StefanIvanov:
Hello @Eralmidia we have a design that we are validating internally, so to be more exact the feature is still in the design/specification phase that we plan to finish very soon

ChronosSF:
@Eralmidia , I am happy to inform you that we added the requested functionality with the latest 10.2.0 release. Please, check the documentation for more information - https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/row_adding.html

Eralmidia:
[...]

PR #8093, opened 2020-09-04
Add row basic functionality

mddragnev:
Related to  #5946  and IgniteUI/trial-watermark#50

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -136,6 +137,24 @@ export class GridBaseAPIService <T extends IgxGridBaseDirective & GridType> {
...
 
+    public submit_add_value() {
+        const cell = this.grid.crudService.cell;
+        if (cell) {
+            this.update_add_cell(cell, cell.editValue);
+            this.escape_editMode();
+        }
+    }
+
+    update_add_cell(cell: IgxCell, value: any) {
```
ChronosSF, line 148: explicitly specify access

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -355,3 +355,39 @@ export class IgxColumnFormatterPipe implements PipeTransform {
...
+    name: 'gridAddRow',
+    pure: true
+})
+export class IgxGridAddRowPipe implements PipeTransform {
+
+    constructor(private gridAPI: GridBaseAPIService<IgxGridBaseDirective & GridType>) { }
+
+    transform(collection: any, parentID: number, enableAddRow: boolean, cancelAddMode: boolean, pipeTrigger: number) {
+        const grid = this.gridAPI.grid;
+        if (!enableAddRow) {
```
ChronosSF, line 369: don't pass grid props as pipe state, you can simply pass the grid. it would also be wrong as you can use the API to start adding a row without the need of a UI flag

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -640,6 +651,15 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
         this.notifyChanges();
     }
 
+    @Input()
+    get enableAddRow(): boolean {
```
ChronosSF, line 655: remove these as per the new spec

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -355,3 +355,39 @@ export class IgxColumnFormatterPipe implements PipeTransform {
...
+            return collection;
+        }
+        if (parentID === -1) {
+            const result = collection.map(record => {
+                return grid.isAddRowRecord(record) ? record.recordRef : record;
+            });
+            grid.data = result;
+            return result;
+        }
+        const parentRecordIndex = collection.findIndex(record => record[grid.primaryKey] === parentID);
```
ChronosSF, line 379: I am wondering, if we trigger this from code (the beginAddRow method) that knows the record index, do we have to search for it again.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -4010,6 +4042,26 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
         this.cdr.detectChanges();
     }
 
+
+    public beginAddRow(rowID?: any) {
```
ChronosSF, line 4046: missing public API comment

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -355,3 +355,39 @@ export class IgxColumnFormatterPipe implements PipeTransform {
...
+        if (cancelAddMode) {
```
ChronosSF, line 380: Really unsure about this. Is the add row (while being in edit mode) ever part of the collection before this pipe is triggered? Don't you want to add this row _only_ when the state is "add row UI present". Once the user presses esc, you say "add row UI no longer present" and escape the pipe never reaching a point that adds the row.

ChronosSF (changes requested):

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -355,3 +355,39 @@ export class IgxColumnFormatterPipe implements PipeTransform {
...
+            collection.splice(parentRecordIndex + 1, 1);
+            return collection;
+        }
+        const row = {...collection[parentRecordIndex]};
+        Object.keys(row).forEach(key => row[key] = undefined);
+        const rec = {
+            recordRef: row,
+            addRow: true
+        };
+        collection.splice(parentRecordIndex + 1, 0, rec);
```
MayaKirova: Since this pipe runs after all other pipes, there may be cases when the row should not be shown below the parent record index. For example, if you have a grid with master-detail views, if the new row is displayed under the parent (master row) it would visually break the hierarchy since the new row will appear above the details view:
image

In those cases it may be better to show the add row above it.
Also the add row should not allow expand/collapse operations

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -355,3 +355,39 @@ export class IgxColumnFormatterPipe implements PipeTransform {
...
+    transform(collection: any, parentID: number, enableAddRow: boolean, cancelAddMode: boolean, pipeTrigger: number) {
+        const grid = this.gridAPI.grid;
+        if (!enableAddRow) {
+            return collection;
+        }
+        if (parentID === -1) {
+            const result = collection.map(record => {
+                return grid.isAddRowRecord(record) ? record.recordRef : record;
+            });
+            grid.data = result;
```
MayaKirova: This will pollute the original grid data with non-data records! We should not be setting anything directly back to the grid's data.

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -355,3 +355,39 @@ export class IgxColumnFormatterPipe implements PipeTransform {
...
+            return result;
+        }
+        const parentRecordIndex = collection.findIndex(record => record[grid.primaryKey] === parentID);
+        if (cancelAddMode) {
+            collection.splice(parentRecordIndex + 1, 1);
```
MayaKirova: Directly slicing the collection may result in polluting the original grid data. You should ensure you work with a copy of the data so that the original data remains the same and does not have non-data records after it passes through the pipe.

MayaKirova (changes requested):

ChronosSF (dismissed):

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -5782,6 +5855,13 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
         event.clipboardData.setData('text/plain', result);
     }
 
+    /**
+     * @hidden @internal
+     */
+    public navigateToAddedRow(event) {
```
MayaKirova, line 5861: This should probably be implemented for the base version.

projects/igniteui-angular/src/lib/grids/grid/grid.component.html
```diff
@@ -229,6 +230,9 @@
         </ng-template>
     </div>
     <div class="igx-grid__scroll-end" [style.float]='"right"' [style.width.px]='pinnedWidth' [style.min-width.px]='pinnedWidth' [hidden]="pinnedWidth === 0 || isPinningToStart"></div>
+    <div class="igx-grid__grid-addrow-snackbar">
+        <igx-snackbar #addRowSnackbar [actionText]="snackbarActionText" (onAction)="navigateToAddedRow($event)" displayTime='2000'>{{snackbarLabel}}</igx-snackbar>
```
MayaKirova, line 234: Since this will always show the "Show" action even if the added row is visible, perhaps we can conditionally hide the action if the added row is in view.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6441,6 +6526,26 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
...
 
+    public endAdd(commit = true, event?: Event) {
+        const row = this.crudService.row;
+        const cell = this.crudService.cell;
+        if (!row && !cell) {
+            return;
+        }
+        if (commit) {
+            this.gridAPI.submit_add_value();
+            this.data.push(row.data);
```
MayaKirova, line 6537: This does not take into consideration the transaction service. Should probably use  this.gridAPI.addRowToData instead. Related method and transaction pipe should probably also be updated to handle null PK and use the row instance instead. Otherwise add row will not work with transaction service enabled.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6441,6 +6526,26 @@ export class IgxGridBaseDirective extends DisplayDensityBase implements
...
     }
 
+    public endAdd(commit = true, event?: Event) {
+        const row = this.crudService.row;
+        const cell = this.crudService.cell;
+        if (!row && !cell) {
+            return;
+        }
+        if (commit) {
+            this.gridAPI.submit_add_value();
```
MayaKirova, line 6536: This seems to generate an update transaction. There's no need to generate an update transaction before adding the row.

MayaKirova (changes requested):

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -5843,7 +5843,8 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
      * @hidden @internal
      */
     public navigateToAddedRow(event) {
-        // TO DO: add navigation logic
+        this.navigateTo(this.dataView.length - 1);
```
MayaKirova, line 5846: New record will not necessarily be added to the end of the data view. Some features (sorting, filtering, etc.) may change the target row location. You will need to check the added record index after the pipes have been executed to find the correct location.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6517,10 +6518,18 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
             return;
         }
         if (commit) {
+            this.verticalScrollContainer.onDataChanged.pipe(first()).subscribe(() => {
+            // A check whether the row is in the current view
+            const shouldScroll = this.navigation.shouldPerformVerticalScroll(this.dataLength - 1, 0);
```
MayaKirova, line 6522: New record will not necessarily be added to the end of the data view. Should check its resolved index after the pipes.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6517,10 +6518,18 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
+            if (shouldScroll) {
+                this.addRowSnackbar.show();
```
MayaKirova, line 6524: Instead of hiding the whole snackbar, maybe we can just show/hide the "SHOW" button inside. So that the grid will still show that row was added but will not give the option to scroll to the row as the row is already visible.

MayaKirova (changes requested):

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -5842,9 +5842,18 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
-        this.addRowSnackbar.hide();
+    public showSnackbarFor(id: number) {
+        if (id === -1) {
+            this.addRowSnackbar.actionText = '';
+            this.addRowSnackbar.show();
+            return;
+        }
+        this.addRowSnackbar.actionText = this.snackbarActionText;
+        this.addRowSnackbar.show();
+        this.addRowSnackbar.onAction.subscribe(() => {
```
MayaKirova: This will subscribe a new listener every time a new row is added so the event will fire multiple times when clicked after a few rows are added. Instead you can subscribe once, cache the last add row index in a private prop and use it  to navigate to the correct index.

MayaKirova (changes requested):

MayaKirova (approved):

3phase:
When adding the Add Row functionality to the MRL template, you cannot edit the newly added rows in the row that appears beneath the click target of the action strip. 
bug_add_new_row_mrl

3phase:
When adding a new row into a grid with enabled filtering, the grid is not added (the `onRowAdded` event won't be fired at all).

bug_add_new_row_filtering

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -2335,6 +2335,13 @@
         }
     }
 
+    %igx-grid__addrow-snackbar {
+        position: absolute;
+        z-index: 5;
+        left: 35%;
```
simeonoff, line 2341: This property should be the interpolated version of the `$left` variable for RTL compatibility purposes.
```suggestion
        #{$left}: 35%;
```

Also, are you sure you want to use percentages as the unit to set the position of the snackbar? If yes, how did you come up with those numbers?

projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-component.scss
```diff
@@ -544,6 +544,10 @@
         @extend %igx-grid__row-editing-outlet !optional;
     }
 
+    @include e(grid-addrow-snackbar) {
```
simeonoff, line 547: The produced selector will be `igx-grid__grid-addrow-snackbar`. This name is rather convoluted.
I would suggest the following change:
```suggestion
    @include e(addrow-snackbar) {
```

ChronosSF (approved):

PR #8197, opened 2020-09-24
refactor(toast): utilize Overlay Service to display in DOM

### LINKED ISSUE #6802: Toast does not show correctly in tabs
## Description  
I have a component which shows `igxToast` messages. When I put this component in `igxTab`'s tabs the toast message does not shows correctly in the first tab, and does not show at all in all other tabs. 

 * igniteui-angular version: 9.0.1
 * browser: all

## Steps to reproduce  

1. Start this [Stackblitz](https://stackblitz.com/edit/angular-okg5s9?file=src%2Fapp%2Ftabs-sample-1%2Ftabs-sample-1.component.html).
2. Click on `Show notification` button.
3. Click on second tab.
4. Click on `Show notification` button.

## Result  
On first tab notification is cut.
On all other tabs notification does not appear at all.

## Expected result  
Notification should be visible in all tabs

simeonoff:
Closes #6802 

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [x] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -277,17 +297,42 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
...
+        this._isVisible = true;
 
         if (this.autoHide) {
             this.timeoutId = window.setTimeout(() => {
                 this.hide();
             }, this.displayTime);
         }
 
+        this.overlayService.show(this.overlayId);
         this.onShown.emit(this);
```
wnvko, line 336: this will fire too early before toast is actually shown. You should subscribe to `IgxOverlay`'s `onOpened` event and emit `onShown` there. IMO it will be much more easier if you try to use `IgxToggleDirective` instead of `IgxOverlayService`.

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -277,17 +297,42 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
...
+
+        this.overlayId = this.overlayService.attach(
+            this.elementRef,
+            overlaySettings
+        );
+
         this.onShowing.emit(this);
-        this.isVisible = true;
-        this.animationState = 'visible';
+        this._isVisible = true;
```
wnvko, line 327: because of the animation toast is not actually 100% visible here. You could handle `onOpened` event of the `IgxOverlayService` and set _isVisible there.

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -300,10 +345,9 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
      */
     public hide(): void {
         this.onHiding.emit(this);
-        this.isVisible = false;
+        this._isVisible = false;
```
wnvko, line 348: same as above, it is better to set this in `onClosed` handler of `IgxOverlayService`.

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -300,10 +345,9 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
...
         this.onHidden.emit(this);
```
wnvko, line 349: same as above, this will fire too early because of the animation.

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -331,31 +375,7 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
...
-       return this.position === IgxToastPosition.Middle;
-    }
-
-    /**
-     * @hidden
-     */
-    @HostBinding('class.igx-toast--bottom')
-    public get cssClassBottom() {
-       return this.position === IgxToastPosition.Bottom;
+        this._isVisible ? this.close() : this.open();
```
wnvko, line 378: this will not work because of the animation. You can try to create a button and call `toggle` method by quickly clicking on the button several times. If you observe the console you will see a lot of warnings. All these is solved in `IgxToggleDirective` and you can use it for the toast.

wnvko (changes requested):

projects/igniteui-angular/src/lib/toast/toast.component.ts
simeonoff: This should be handled better now since the event is fired on the `onOpened` event of the toggle directive.

projects/igniteui-angular/src/lib/toast/toast.component.ts
simeonoff: Same here. The `isVisible` property is updated in the `onOpening` event of the toggle directive.

projects/igniteui-angular/src/lib/toast/toast.component.ts
simeonoff: Same here.

projects/igniteui-angular/src/lib/toast/toast.component.ts
simeonoff: And here.

projects/igniteui-angular/src/lib/toast/toast.component.ts
simeonoff: Should be handled better now, as it uses the `collapsed` property of the toggle directive to decide whether to call the `open` or `closed` methods.

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -249,7 +262,7 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
      * @memberof IgxToastComponent
      */
     public get element() {
-        return this.elementRef.nativeElement;
+        return this.element.nativeElement;
```
wnvko, line 265: this should throw stack overflow.
This should either
 - get the element from `this.elementRef.nativeElement` and make `elementRef` protected in toggle.
 - get the element from `this._element.nativeElement` and `_element` private field in the constructor.

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -277,18 +295,43 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
...
+            closeOnOutsideClick: false,
+            modal: false,
+            outlet: this.outlet,
+        };
+
+        if (message !== undefined) {
+            this.toastMessage = message;
+        }
+
+        this.onOpening.subscribe(() => {
```
wnvko, line 319: this may leak as on each call to `show`/`hide` a new subscriptions are created. You should subscribe once in `ngOnInit`. Then you should unsubscribe all subscriptions in `ngOnDestroy` with `takeUntil`.
Additionally there is no reason to subscribe to `onShowing` - you may call all the logic you need before you call `super.open()`.

wnvko (changes requested):
I have one general question about `igxToast` (not sure this is related to this PR). If I have an `IgxToast` which is not show yet, and if I set `isVisible` to `true`, should this show the toast?

simeonoff:
> I have one general question about `igxToast` (not sure this is related to this PR). If I have an `IgxToast` which is not show yet, and if I set `isVisible` to `true`, should this show the toast?

Yeah, this should be fixed in a separate PR.

SisIvanova (approved):

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -331,49 +371,28 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
...
 
     /**
      * @hidden
      */
-    public ngOnInit() {
-        if (this.navService && this.id) {
-            this.navService.add(this.id, this);
-        }
+    ngOnInit() {
+        this.onAppended.pipe(takeUntil(this.d$)).subscribe(() => {
```
wnvko, line 381: `onAppended` is emitted when the content is moved to the overlay's content `div`. At this point content is still not visible, neither it is shown. Why not subscribe to `onOpened`?

projects/igniteui-angular/src/lib/toast/toast.component.ts
```diff
@@ -299,12 +341,10 @@ export class IgxToastComponent implements IToggleView, OnInit, OnDestroy {
      * @memberof IgxToastComponent
      */
     public hide(): void {
-        this.onHiding.emit(this);
-        this.isVisible = false;
-        this.onHidden.emit(this);
-        this.animationState = 'invisible';
-
         clearInterval(this.timeoutId);
+        this.isVisible = false;
```
wnvko, line 345: at this point toast is still visible. After calling `supper.close()` the first thing that happens close animation start. After close animation ends the content is removed from the DOM and at this point it is not visible anymore. This is why IMO it is better to set `isVisible` in `onClosed` subscription just before we emit `onHidden`.

wnvko (changes requested):

wnvko (approved):

Issue #8195, opened 2020-09-24
IgxDatePipe is not transforming values of type ISO string and number


zdrawku:
## Description  
Currently the default column template for `date` dataType is not transforming data values of type ISO string.
[The input value for the DatePipe](https://angular.io/api/common/DatePipe) is of types - `Date object`, `number` and `ISO string` and our formatter should comply with the same types as well.

 * igniteui-angular version: any
 * browser: any

## Steps to reproduce  

1. Open [the sample](https://stackblitz.com/edit/github-ysxqf8?file=src/app/grid/grid-editing-sample/grid-editing-sample.component.html)
2. Check the first two dates in `Order Date` column

## Result  
The dates are shown as ISO string:
```
2012-02-12T00:00:00.000Z
```

## Expected result  
Dates to be formatted with the default column formatting 
```
Feb 12, 2012
```

PR #8239, opened 2020-09-29
refactor(grids): forward-compatibility for upcoming type improvements in Angular Pipes

AndrewKushnir:
There is an upcoming improvement in common pipes in Angular framework that would be released in the next major version (v11). The changes can be found in https://github.com/angular/angular/pull/37447.

This commit refactors `IgxDatePipeComponent` and `IgxDecimalPipeComponent` pipes to provide forward-compatibility for the next major release of Angular framework. The improved typings make `transform` function signatures of the mentioned pipes incompatible with the `DatePipe` and `DecimalPipe`. This commit updates the mentioned pipes to initialize `DatePipe` and `DecimalPipe` in constructors instead of extending them (to keep existing `transform` function signatures) and use created instances inside the `transform` function as needed.

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/common/pipes.ts
+++ b/projects/igniteui-angular/src/lib/grids/common/pipes.ts
@@ -171,20 +171,20 @@ export class IgxHasVisibleColumnsPipe implements PipeTransform {
 @Pipe({
     name: 'igxdate'
 })
-export class IgxDatePipeComponent extends DatePipe implements PipeTransform {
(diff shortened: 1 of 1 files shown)


zdrawku:
Thank you @AndrewKushnir for your contribution! We are currently working [on a problem](https://github.com/IgniteUI/igniteui-angular/issues/8195) related to both igx pipes that might result in a slight refactoring causing a removal of both pipes entirely.

hanastasov:
Resolved by #8195

PR #8281, opened 2020-10-05
Upgrade excludePositionTarget to a collection of HTML Elements

IvayloG:
Closes #8148 
Closes #7974   

### Additional information (check all that apply):
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [x] All relevant tags have been applied to this PR
 - [x] This PR includes unit tests covering all the new code ([test guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Test-implementation-guidelines-for-Ignite-UI-for-Angular))
 - [ ] This PR includes API docs for newly added methods/properties ([api docs guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines))
 - [ ] This PR includes `feature/README.MD` updates for the feature docs
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [ ] This PR includes `CHANGELOG.MD` updates for newly added functionality
 - [ ] This PR contains breaking changes
 - [ ] This PR includes `ng update` migrations for the breaking changes ([migrations guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Update-Migrations))
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

PR #8298, opened 2020-10-06
Check if position target is point

### LINKED ISSUE #8297: Error thrown when trying to close the overlay on outside click and target is set as Point
Set positioning setting target to a Point instead of an HTML Element.
The following error is thrown when trying to close the overlay: 

oerr

wnvko:
We are checking now it the target has `contains` method before call it.

Closes #8297

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


IvayloG (dismissed):

wnvko:
Do not merge before #8281

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -655,7 +655,7 @@ export class IgxOverlayService implements OnDestroy {
                     // if we should exclude position target check if the click is over it. If so do not close overlay
                     const positionTarget = info.settings.target as HTMLElement;
                     let clickOnPositionTarget = false;
-                    if (positionTarget) {
+                    if (positionTarget && positionTarget.contains) {
```
damyanpetev, line 658: Since this is pretty much the only use of the target here, you could also get away with a type check which will also let you not hard type assert `as HTMLElement` above too.

PS: I realize this is not cross-frame friendly, but then again there's a precedent of `instanceof HTMLElement` check in utils and none of the positioning or these handlers support targets in other frames atm, so those rare cases will have to resort to point use anyway.

damyanpetev (approved):

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -804,12 +804,12 @@ export class IgxOverlayService implements OnDestroy {
                 return;
             }
             if (info.settings.closeOnOutsideClick) {
-                const target = ev.target as any;
+                const target = ev.target as Node;
```
damyanpetev, line 807: That assert is kind of pointless if it's going to be done again below :)

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -804,12 +804,12 @@ export class IgxOverlayService implements OnDestroy {
...
                 const overlayElement = info.elementRef.nativeElement;
                 // check if the click is on the overlay element or on an element from the exclusion list, and if so do not close the overlay
-                const excludeElements = info.settings.excludeFromOutsideClick ?
+                const excludeElements: HTMLElement[] = info.settings.excludeFromOutsideClick ?
```
damyanpetev, line 810: `excludeFromOutsideClick` is already typed as `HTMLElement[]` so this shouldn't be needed. In fact I think with this version the issue is no longer reproducible anyway

damyanpetev (approved):

IvayloG (approved):

PR #8337, opened 2020-10-09
Fix animations when content is flipped

### LINKED ISSUE #8238: Auto position strategy use wrong animation when flipped
## Description  
In `IgxOverlay` when using `AutoPositionStrategy` in case the content is flipped, e.g. initially should show in down direction, but there is not space and it shows in upper direction, the animation is not changed.

 * igniteui-angular version: 9.1.x; 10.1.x; 10.2.x
 * browser: all

## Steps to reproduce  

Show overlay with `AutoPositionStrategy` forcing the content to be flipped.

## Result  
Animation starts from the top of the content.

## Expected result  
Overlay should use the opposite animation - from bottom to top.

wnvko:
New util function is introduced - `oppositeAnimationResolver`. Passing an animation to this function returns opposite animation if one exists. By the help of this function, when content shown in overlay is flipped by the auto position strategy, we are changing and provided animations.

Closes #8238 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -451,3 +471,125 @@ export function yieldingLoop(count: number, chunkSize: number, callback: (index:
 }
 
 export function mkenum<T extends { [index: string]: U }, U extends string>(x: T) { return x; }
+
+export function reverseAnimationResolver(animation: AnimationReferenceMetadata): AnimationReferenceMetadata {
+    return oppositeAnimation.get(animation) ?? animation;
+}
+
+const oppositeAnimation: Map<AnimationReferenceMetadata, AnimationReferenceMetadata> = new Map([
```
ViktorSlavov, line 487: I feel like the animation breakup could be expanded. Currently, we don't differentiate between animations regarding the X and Y axis. In a case where we're only flipping the container horizontally, but the animation is vertical, it would still be "flipped" (even though it doesn't need to, since the overlay would still appear, for example, on the bottom of the target)

ViktorSlavov: Also, some animations are both vertical and horizonal.
For example, to display a container to the `VerticalAlignment.Top` and `HorizontalAlignment.Left` side of a component, you'd use `scaleInBr`.
Depending on the viewport, the autoposition could force several different display flips:
`HorizontalAlignment` changes to `Left` - animation should change to `scaleInBl`.  
`VerticalAlignment` changes to `Bottom` - animation should change to `scaleInTr`.
If **both** `Horizontal` and `VerticalAlignment` change - animation should change to `scaleInTl`.

ViktorSlavov (changes requested):

simeonoff (dismissed):

ViktorSlavov (dismissed):

ViktorSlavov (approved):

PR #8357, opened 2020-10-13
fix(dialog): exposed opened property-10.2.x #8320

### LINKED ISSUE #8320: igx-dialog - isOpen is not bindable
## Description  

The property `isOpen` in `IgxDialog` is not bindable. It's declared as an `@Input()` but it only has a getter, which leads to errors.

```ts
export class IgxDialogComponent implements IToggleView, OnInit, OnDestroy, AfterContentInit {
    ...
    @Input()
    get isOpen() {
        return !this.toggleRef.collapsed;
    }
    ....
```

 * igniteui-angular version: "10.1.0"
 * browser: any

## Steps to reproduce

```html
<igx-dialog title="My Dialog"
    [isOpen]="this.myDialogShouldBeOpen"
>
    Content
</igx-dialog>
```

## Result  

In runtime the console prints this error:
> TypeError: Cannot set property isOpen of [object Object] which has only a getter

## Expected result  
The binding works and the dialog is open if and only if the binding is `true`.

## Why?

Right now Dialog can only be opened using imperative API, if this property is bindable I could use it in a fully declarative way. Furthermore, I could remove the dialog entirely when it's not needed (ngIf):

```html
<igx-dialog
    *ngIf="this.myDialogShouldBeOpen"
    title="My Dialog"
    [isOpen]="true"
>
    Content
</igx-dialog>
```

dobromirts:
Closes #8320

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


hanastasov (changes requested):
Rename the Input to `isOpened` and add a getter (you can use the old getter ;) )

dobromirts:
> Rename the Input to `isOpened` and add a getter (you can use the old getter ;) )

Done

hanastasov (approved):

hanastasov (changes requested):
Fix Tslint errors.

hanastasov (approved):

hanastasov (approved):

hanastasov (changes requested):
When setting the opened property to false, dialog should be closed.

hanastasov (dismissed):

projects/igniteui-angular/src/lib/dialog/dialog.component.ts
```diff
@@ -368,11 +368,35 @@ export class IgxDialogComponent implements IToggleView, OnInit, OnDestroy, After
...
     }
 
+
+    /**
+     * An @Input property that allows you to open the dialog declaratively.
+     * ```typescript
+     * @ViewChild("MyDialog")
+     * public dialog: IgxDialogComponent;
+     * ngAfterViewInit() {
+     *     let dialogOpen = this.dialog.opened=true;
```
HristoP96, line 383:
```suggestion
     *     this.dialog.opened=true;
```

HristoP96 (changes requested):

hanastasov (approved):

HristoP96 (approved):

damyanpetev:
Just, FYI on the property name choice - if we are to move away from `on-` prefixed events globally for the product, the current `onOpen` event won't have a clear transformation - will collide with the `open()` method as is. Alternatively, if we properly expose all the events from the underlying toggle directive/overlay service, those are `onOpening` and `onOpened` and the latter will directly collide with this input's name as both will have to be named `opened` after dropping the prefix :)

Also, as the open action can be prevented and the dialog can be closed with interactions (the buttons?) the value bound to the dialog can easily end up out of sync. Unless this is intended to be initial-only property (the feature request seems to be for totally declarative control), the input should come with a matching output `<name>Change` for two-way binding IMO.

Is there a sample with the proposed `*ngIf` usage?

// cc @rkaraivanov

pmoleri:
@damyanpetev @dobromirts 
About the `opened` naming, I just want to add one data point that you may want to consider, `nav-drawer` uses `isOpen` and is bindable.

rkaraivanov:
@pmoleri 

Are you interested in the state of the dialog (aka what the navdrawer `isOpen` property does) or do you want to actually show/hide the dialog declaratively?
Can you provide a reference sample?

pmoleri:
> @pmoleri
> 
> Are you interested in the state of the dialog (aka what the navdrawer `isOpen` property does) or do you want to actually show/hide the dialog declaratively?
> Can you provide a reference sample?

@rkaraivanov in short: show/hide declaratively.
There's an example in the issue [8320 ](https://github.com/IgniteUI/igniteui-angular/issues/8320#issuecomment-707193271)

hanastasov:
I guess keeping the input name to be `isOpen` will avoid any name collisions in the future, and will be consistent with the nav-drawer naming.

@damyanpetev

damyanpetev:
@hanastasov @dobromirts Yeah, with `open()` and `opened` for the method/event, don't really have ideas other than `isOpen` for the property :)

For the declarative control, just a reminder from my previous comment that we should consider also adding a `matching output <name>Change for two-way binding`. That way the first example @pmoleri provided will look more like:
```
<igx-dialog
    title="My Dialog"
    [(isOpen)]="this.myDialogShouldBeOpen"
>
    Content
</igx-dialog>
``` 
Still not sure on the ngIf case - likely to mess up the closing animation at the very least.

HristoP96 (dismissed):

hanastasov (changes requested):

hanastasov (approved):

HristoP96 (approved):

rkaraivanov (approved):

kdinev (approved):

pmoleri:
> @hanastasov @dobromirts Yeah, with `open()` and `opened` for the method/event, don't really have ideas other than `isOpen` for the property :)
> 
> For the declarative control, just a reminder from my previous comment that we should consider also adding a `matching output <name>Change for two-way binding`. That way the first example @pmoleri provided will look more like:
> 
> ```
> <igx-dialog
>     title="My Dialog"
>     [(isOpen)]="this.myDialogShouldBeOpen"
> >
>     Content
> </igx-dialog>
> ```
> 
> Still not sure on the ngIf in the second case - likely to mess up the closing animation at the very least.

Hi @damyanpetev @dobromirts 
I was trying to use this, under the impression that it would support 2 way binding after what was said in the quoted message, but it doesn't.

This works ok:
```
<button igxButton="raised" igxRipple="white" (click)="this.dialogOpen = true">Show Alert Dialog</button>

<igx-dialog
    [isOpen]="this.dialogOpen"
    (onClose)="this.dialogOpen = false"
    title="Notification"
    message="Your email has been sent successfully!"
    leftButtonLabel="OK"
    (onLeftButtonSelect)="this.dialogOpen = false">
</igx-dialog>
```

This doesn't:
```
<button igxButton="raised" igxRipple="white" (click)="this.dialogOpen = true">Show Alert Dialog</button>

<igx-dialog
    [(isOpen)]="this.dialogOpen"
    title="Notification"
    message="Your email has been sent successfully!"
    leftButtonLabel="OK"
    (onLeftButtonSelect)="this.dialogOpen = false">
</igx-dialog>
```

- Click to open
- Close with the button (works fine, it's still 1 way)
- Click to open
- Esc to close -> the binding is not updated (2 way failed)

Was it supposed to be 2 way?

hanastasov:
@dobromirts is working on this, it seemed that the component is missing a ControlValueAccessor. I think he is going to push to this    issue.

damyanpetev:
This is not related or dependent on `ControlValueAccessor` FYI :)

hanastasov:
@dobromirts will change the implementation to keep the `isOpen` state accordingly. Also, implementation should respect if the open/close events are cancelled.

pmoleri:
Thanks. BTW, I'm yet to try this kind of binding in real world. But at first glance it looks really promising, I think I no longer need to care about things like `@ViewChild`, checking it `AfterContentInit` or worrying if _Expression has changed after it was checked_.
It seems to me that regular declarative bindings are more robust than the corresponding `dialog.Open()` `.close()` alternatives.

pmoleri:
Hi, I just noticed another issue. When I initialize with `[isOpen]="true"` the dialog is not measured correctly:
image

This is my workaround in the parent component:
image

PR #8430, opened 2020-10-20
fix(grid): update transactionState rowEdit is canceled

### LINKED ISSUE #8381: Cell editing styling remains when rowEdit is canceled and hit cancel
## Description  
Editing styling remains when rowEdit is canceled and hit cancel

 * igniteui-angular version: 10.2.x
 * browser: All

## Steps to reproduce  

1. Open editing lifecycle sample
2. cancel rowEdit
3. Enter in edit and try to commit a new value
4. Cancel/esc the editing 

## Result  
The cell value remains italic as if the update process hasn't ended yet.

## Expected result  
The cell and whole row should apply the default non-editable style once you cancel/end the editing.

## Attachments  
image

Aleksandyr:
Closes #8381 

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid/grid-row.component.html
```diff
@@ -135,13 +135,13 @@
...
-        [class.igx-grid__td--edited]="rowID | transactionState:col.field:grid.rowEditable:grid.transactions:grid.pipeTrigger:grid.crudService.cell"
+        [class.igx-grid__td--edited]="rowID | transactionState:col.field:grid.rowEditable:grid.transactions:grid.pipeTrigger:grid.crudService.cell:grid.crudService.row"
         [class.igx-grid__td--number]="col.dataType === 'number' && col.visibleIndex !== 0"
         [ngClass]="col.cellClasses | igxCellStyleClasses:rowData[col.field]:rowData:col.field:viewIndex:grid.pipeTrigger"
         [ngStyle]="col.cellStyles | igxCellStyles:rowData[col.field]:rowData:col.field:viewIndex:grid.pipeTrigger"
         [expanded]="expanded"
-        [editMode]="col.editable && crudService.isCellInEditByPos(index, col.index)"
-        [addMode]="col.editable && crudService.isInAddMode(index,col.index)"
+        [editMode]="col.editable && crudService.targetInEdit(index, col.index)"
+        [addMode]="col.editable && crudService.targetInEdit(index, col.index)"
```
damyanpetev, line 144: If this is the exact same check as in `editMode`.. is it actually needed at all? @mddragnev ?

mddragnev: IMO addMode input may be omitted. It is used only in cell component when the template getter is called. This was implemented because long ago we wanted to have row adding without editing as we thought to expose additional grid input. Now, we cannot have adding without editing so this is unnecessary. @Aleksandyr Feel free to remove it and refactor cell template get method.

projects/igniteui-angular/src/lib/grids/grid/grid-row.component.html
```diff
@@ -98,15 +98,14 @@
 <ng-template #cellTemplate let-col>
         <igx-grid-cell
         class="igx-grid__td igx-grid__td--fw"
-        [class.igx-grid__td--edited]="rowID | transactionState:col.field:grid.rowEditable:grid.transactions:grid.pipeTrigger:grid.crudService.cell"
+        [class.igx-grid__td--edited]="rowID | transactionState:col.field:grid.rowEditable:grid.transactions:grid.pipeTrigger:grid.crudService.cell:grid.crudService.row"
```
damyanpetev, line 101: Those might be a bit overzealous for pipe triggers - each time _any_ edited cell changes the pipes on all cells run. That's preexisting though and adding the row on top will hardly make a difference I suppose.

damyanpetev (approved):

projects/igniteui-angular/src/lib/grids/grid/grid-row.component.html
Aleksandyr: @damyanpetev it makes a difference as if you imagine the moment we cancel **rowEdit**, non of the cells change is triggered and once you enter edit and apply any new changes and then exit the edit row the **td--edited** class remains as nothing triggers the pipe to change its state.

damyanpetev: Yeah, that's your fix. My comment was about performance - because all the cells in the grid view (on all rows) are using the static crud service state as pipe argument, when you enter/exit edit on one cell, the entire view will go calculating the transaction state. But that's already tied to `crudService.cell` from before, which will trigger a lot more than row and thus your addition will likely have minimal perf impact and it should be fine :)

Aleksandyr: o yeah, I get it. But probably at that point, there is no other way of catching that? 
Do you have smth in mind that could be beneficial for any performance improvement?

zdrawku (approved):

PR #8744, opened 2021-01-06
fix(radio-group): set group value to checked button

### LINKED ISSUE #8718: Radios in Radio Groups do not change state properly
* igniteui-angular version:  10.2.x, 11.0.x, 11.1.x
 * browser:  all

## Steps to reproduce  

1. Open the following sample - [link](https://stackblitz.com/edit/github-7y9wjs?file=src/app/radio-sample-1/radio-sample-1.component.html)
2. Click on  the `Option 2` radio button
3. Observe

## Result  
The `Option 1` radio button is not deselected.

## Expected result  
The `Option 1` radio button is deselected.

simeonoff:
Closes #8718  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.ts
@@ -366,13 +366,11 @@ export class IgxRadioGroupDirective implements AfterContentInit, ControlValueAcc
      * @internal
      */
     private _selectedRadioButtonChanged(args: IChangeRadioEventArgs) {
-        if (this._selected !== args.radio) {
-            if (this._selected) {
-                this._selected.checked = false;
-            }
-            this._selected = args.radio;
-        }
+        this.radioButtons.forEach(button => {
+            button.checked = button.id === args.radio.id;
+        });
 
+        this._selected = args.radio;
         this._value = args.value;
 
         if (this._isInitialized) {
--- a/projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.spec.ts
+++ b/projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.spec.ts
@@ -12,6 +12,7 @@ describe('IgxRadioGroupDirective', () => {
         TestBed.configureTestingModule({
             declarations: [
                 RadioGroupComponent,
+                RadioGroupSimpleComponent,
                 RadioGroupWithModelComponent,
                 RadioGroupReactiveFormsComponent,
                 RadioGroupDeepProjectionComponent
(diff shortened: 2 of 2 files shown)


projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.ts
```diff
@@ -351,6 +351,10 @@ export class IgxRadioGroupDirective implements AfterContentInit, ControlValueAcc
             this.radioButtons.forEach((button) => {
                 Object.assign(button, props);
 
+                if (!this._value && button.checked) {
```
jackofdiamond5, line 354: After running the samples with this branch I get the following:
image
This is the same application as the one from the issue.

I don't think the problem is in the initialization step of the directive.  Instead of changing the `value` here, we can simply go to the `button.change` handler - `_selectedRadioButtonChanged` and go through all of the buttons that we have and update their states based on which radio button was clicked. 

Maybe something like this:
```
 this.radioButtons.forEach(button => {
     button.checked = button.id === args.radio.id;
 });
 this._selected = args.radio;
```

The way I see it `this._selected` keeps a reference to the last button that was changed. So on lines `373` through `378` we update the state only if a different button was targeted and then we set its value to `false` explicitly. I don't think we need to do that since we have access to all buttons at once in the `this.radioButtons` member, thus we can simply loop them and update them all because, correct me if I'm wrong, the radio group should hold only radio buttons and only one of them can be selected at a time. I don't think looping said buttons would be a performance issue either since having hundreds or thousands of radio buttons does not seem like a viable scenario to me.

Also I think we should add at least on test just to be future-proof.

jackofdiamond5 (changes requested):

projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.ts
simeonoff: You are correct, the proposed solution will only work if `value` is defined on the `igx-radio` component instances in the radio group. If the user omits the `value` input on those radio buttons, this will not work. Your solution will work in all cases. Even though the selected option will always yield `undefined` as the value of the group. I don't know how much value that has in the context of a web form, regardless, it fixes the issue.

I have checked the native behavior of inputs of type radio when a part of a group. If the value attribute is omitted on any of the radio inputs, the value becomes 'on', instead of `undefined` as in our case. Perhaps this calls for a separate issue, it's just a discovery I made when inspecting the code and behavior.

jackofdiamond5 (commented):
Please add at least one test which covers the new code.

simeonoff:
> Please add at least one test which covers the new code.

Done deal.

jackofdiamond5 (approved):

jackofdiamond5 (approved):

PR #8747, opened 2021-01-06
Handle keydown and iterate items by first character, if no item matched typed input - master

### LINKED ISSUE #8436: When you type fast the same letter in the select, the items are not navigated
## Description  
When igxSelect has items in the list, with the same first letter, then when that same letter is typed fast, then those items are not navigated and the highlighted item is only the first one.

 * igniteui-angular version: 10.2.0
 * browser: any

## Steps to reproduce  

1. Open the following igxSelect [demo](https://stackblitz.com/github/IgniteUI/igniteui-live-editing-samples/tree/master/angular-demos/data-entries/select-sample-2).
2. Open the select component
3. Type fast and multiple times 'p'

## Result  
Notice that the only 'Potato' is selected.

## Expected result  
When typing 'p' multiple times, the navigation should move between 'Potato' to 'Pepper'.

IvayloG:
Closes #8436   

- Make the component feel more responsive now as if no exact match is found for the current input stream, a next item with the same start letter will be selected (if such an item).
- Behave more like a native HTML select in terms of fast searching through items by 1st character, and yet do not suffer from the negative downside of not matching same second character if holding down or typing fast same char. Native select basically ignores selecting items via the current input stream while the same char is continuously typed as it seems to prioritize fast items iteration by first char instead.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/select/select-navigation.directive.ts
+++ b/projects/igniteui-angular/src/lib/select/select-navigation.directive.ts
@@ -62,13 +62,14 @@ export class IgxSelectItemNavigationDirective extends IgxDropDownItemNavigationD
         }
 
         super.handleKeyDown(event);
+        this.captureKey(event);
     }
 
     // tslint:disable:member-ordering
     private inputStream = '';
     private clearStream$ = Subscription.EMPTY;
 
-    @HostListener('keydown', ['$event'])
+    // @HostListener('keydown', ['$event'])
     public captureKey(event: KeyboardEvent) {
         // relying only on key, available on all major browsers:
         // https://caniuse.com/#feat=keyboardevent-key (IE/Edge quirk doesn't affect letter typing)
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/select/select-navigation.directive.ts
```diff
@@ -112,6 +118,15 @@ export class IgxSelectItemNavigationDirective extends IgxDropDownItemNavigationD
         this.target.navigateItem(items.indexOf(nextItem));
     }
 
+    private findNextItem(items: IgxSelectItemComponent[],  text: string) {
+        const activeItemIndex = items.indexOf(this.target.focusedItem as IgxSelectItemComponent) || 0;
+
+        // Match next item in ddl items and wrap around if needed
+        const item = items.slice(activeItemIndex + 1).find(x => !x.disabled && (x.itemText.toLowerCase().startsWith(text.toLowerCase()))) ||
```
ViktorSlavov, line 125: No need to assign to a variable, you could just `return items.slice(...`

ViktorSlavov (changes requested):

ViktorSlavov (approved):

PR #8772, opened 2021-01-11
feat(tree-grid): add cascading row selection #8040

### LINKED ISSUE #8040: Auto select children in tree grid
It would be useful to have a mode for the grid, where selecting a parent would also select all children (recursively).

Essentially the same request as #7344, but for tree grid instead of grouping.

mmart1n:
Closes #8040   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


rkaraivanov (changes requested):

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,80 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
             this.loadChildrenOnRowExpansion(args);
         });
 
+        this.onRowAdded.subscribe(args => {
```
damyanpetev, line 365: Don't leave the args as `any`, the event emitter is typed but that's not picked up when you don't use a pipe. Then again, you might as well use one too :)

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,80 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
+            if (this.rowSelection === 'multipleCascade') {
+                let rec = (this.gridAPI as IgxTreeGridAPIService).get_rec_by_id(this.primaryKey ? args.data[this.primaryKey] : args.data);
```
damyanpetev, line 367: There's a typed `this._gridAPI` so you don't have to use type assertion.

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,80 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
+                if (rec && rec.parent) {
+                    // if batch editing is enabled
+                    (this.gridAPI as IgxTreeGridAPIService).handleCascadeSelectionByFilteringAndCRUD(
+                        new Set([rec.parent]), true, undefined, rec.parent.rowID);
+                } else {
+                    // if batch editin is disabled
+                    requestAnimationFrame(() => {
```
damyanpetev, line 374: Is this the same as handleCascadeSelection? Kind of looks like it. Also the batch comment might need elaborating.

damyanpetev: Also, is the animation frame because the event emits too soon (unlike filtering?) in that case leave a comment why it's there as this will likely need to be remove in the future. Also doesn't the event itself in the tree grid provide record parent information?!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -36,6 +37,7 @@ import { IgxGridNavigationService } from '../grid-navigation.service';
 import { GridType } from '../common/grid.interface';
 import { IgxColumnComponent } from '../columns/column.component';
 import { IgxTreeGridRowComponent } from './tree-grid-row.component';
+import { IgxTreeGridHierarchizingPipe } from './tree-grid.pipes';
```
damyanpetev, line 40: This is unused atm

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -679,6 +763,13 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
         }
         this.selectionService.clearHeaderCBState();
         this._pipeTrigger++;
+        if (this.rowSelection === 'multipleCascade') {
+            this.cdr.detectChanges();
```
damyanpetev, line 767: I don't get what's the change to detect here, plus, why run it before checking if the parent is selected anyway? And then considering children are available right here, could they not just be added to the selection directly?

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -696,4 +787,18 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
             }
         }
     }
+
+    private handleCascadeSelection(event: any, rec?: ITreeGridRecord) {
```
damyanpetev, line 791: Again, avoid `any` type :)

projects/igniteui-angular/src/lib/grids/common/enums.ts
```diff
@@ -34,6 +34,12 @@ export const GridSelectionMode = mkenum({
 });
 export type GridSelectionMode = (typeof GridSelectionMode)[keyof typeof GridSelectionMode];
 
+export const HierarchicalGridSelectionMode = mkenum({
```
kdinev, line 37: Don't introduce a new enum, just add the new selection mode to the `GridSelectionMode`. If anyone sets it on a flat or hierarchical grid, then selection would still work as multiple, because the `isMultiRowSelectionEnabled` would still return true, even though there's no cascading in flat and hiearchical grids.

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
```diff
@@ -260,11 +260,39 @@ export class IgxGridComponent extends IgxGridBaseDirective implements GridType,
...
+        if (!this._init) {
+            this.selectionService.clearAllSelectedRows();
+            this.notifyChanges(true);
+        }
+    }
+
+    /**
+     * @hidden @internal
+     */
+    public get isMultiRowSelectionEnabled(): boolean {
```
kdinev, line 292: You don't need this. Leave the base implementation take care of it. It will return true on `GridSelectionMode.multiple` and on `GridSelectionMode.multipleCascade`

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
```diff
@@ -282,6 +288,33 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
...
+        if (!this._init) {
+            this.selectionService.clearAllSelectedRows();
+            this.notifyChanges(true);
+        }
+    }
+
+    /**
+     * @hidden @internal
+     */
+    public get isMultiRowSelectionEnabled(): boolean {
```
kdinev, line 314: You don't need this either. Leave the base implementation take care of it. It will return true on GridSelectionMode.multiple and on GridSelectionMode.multipleCascade

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
```diff
@@ -122,6 +123,11 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
      */
     public parent = null;
 
+    /**
+     * @hidden
+     */
+    protected _rowSelectionMode: GridSelectionMode;
```
kdinev, line 129: You wouldn't need this.

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
```diff
@@ -260,11 +260,39 @@ export class IgxGridComponent extends IgxGridBaseDirective implements GridType,
      * @hidden
      */
     protected groupingDiffer;
+    protected _rowSelectionMode: GridSelectionMode;
```
kdinev, line 263: You wouldn't need this.

projects/igniteui-angular/src/lib/grids/row.directive.ts
```diff
@@ -403,7 +403,7 @@ export class IgxRowDirective<T extends IgxGridBaseDirective & GridType> implemen
         if (this.grid.rowSelection === 'none' || this.deleted || !this.grid.selectRowOnClick) {
             return;
         }
-        if (event.shiftKey && this.grid.rowSelection === 'multiple') {
+        if (event.shiftKey && this.grid.isMultiRowSelectionEnabled) {
```
kdinev, line 406: Good!

projects/igniteui-angular/src/lib/grids/row.directive.ts
```diff
@@ -436,7 +436,7 @@ export class IgxRowDirective<T extends IgxGridBaseDirective & GridType> implemen
      */
     public onRowSelectorClick(event) {
         event.stopPropagation();
-        if (event.shiftKey && this.grid.rowSelection === 'multiple') {
+        if (event.shiftKey && this.grid.isMultiRowSelectionEnabled) {
```
kdinev, line 439: Good!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-expanding.spec.ts
```diff
@@ -992,7 +992,7 @@ describe('IgxTreeGrid - Expanding / Collapsing #tGrid', () => {
             });
 
             it('check row selection when expand a row', async () => {
-                treeGrid.rowSelection = GridSelectionMode.multiple;
+                treeGrid.rowSelection = HierarchicalGridSelectionMode.multiple;
```
kdinev, line 995: You would no longer need this :)

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -679,6 +766,14 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
         }
         this.selectionService.clearHeaderCBState();
         this._pipeTrigger++;
+        if (this.rowSelection === 'multipleCascade') {
```
kdinev, line 769: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

src/app/tree-grid/tree-grid.sample.ts
```diff
@@ -18,14 +18,15 @@ export class TreeGridSampleComponent implements OnInit {
...
+    public selectionModes: HierarchicalGridSelectionMode[] = ['none', 'single', 'multiple', 'multipleCascade'];
 
     private nextRow = 1;
 
     constructor(private excelExporterService: IgxExcelExporterService,
         private csvExporterService: IgxCsvExporterService) {
 }
     public ngOnInit(): void {
-        this.selectionMode = GridSelectionMode.multiple;
+        this.selectionMode = HierarchicalGridSelectionMode.multiple;
```
kdinev, line 29: You wouldn't need this change!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -379,6 +450,21 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
     }
 
+    /**
+     * @hidden
+     */
+    public ngAfterViewInit() {
+        super.ngAfterViewInit();
+        // TODO: pipesExectured event
+        // run after change detection in super triggers pipes for records structure
+        if (this.rowSelection === 'multipleCascade' && this.selectedRows.length) {
```
kdinev, line 460: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,81 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
         this.transactions.onStateUpdate.pipe(takeUntil<any>(this.destroy$)).subscribe((event: StateUpdateEvent) => {
             let actions = [];
             if (event.origin === TransactionEventOrigin.REDO) {
                 actions = event.actions ? event.actions.filter(x => x.transaction.type === TransactionType.DELETE) : [];
+                if (this.rowSelection === 'multipleCascade') {
+                    this.handleCascadeSelection(event);
+                }
             } else if (event.origin === TransactionEventOrigin.UNDO) {
                 actions = event.actions ? event.actions.filter(x => x.transaction.type === TransactionType.ADD) : [];
+                if (this.rowSelection === 'multipleCascade') {
```
kdinev, line 432: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,81 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
+                this.gridAPI.grid.selectionService.updateCascadeSelectionOnFilterAndCRUD(leafRowsDirectParents);
+                this.notifyChanges();
+            }
+        });
+
         this.transactions.onStateUpdate.pipe(takeUntil<any>(this.destroy$)).subscribe((event: StateUpdateEvent) => {
             let actions = [];
             if (event.origin === TransactionEventOrigin.REDO) {
                 actions = event.actions ? event.actions.filter(x => x.transaction.type === TransactionType.DELETE) : [];
+                if (this.rowSelection === 'multipleCascade') {
```
kdinev, line 427: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,81 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
+                    requestAnimationFrame(() => {
+                        this.gridAPI.grid.selectionService.updateCascadeSelectionOnFilterAndCRUD(leafRowsDirectParents);
+                        this.notifyChanges();
+                    });
+                }
+            }
+        });
+
+        this.onFilteringDone.pipe(takeUntil(this.destroy$)).subscribe(() => {
+            if (this.rowSelection === 'multipleCascade') {
```
kdinev, line 411: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,81 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
+                                new Set([rec.parent]), rec.parent.rowID);
+                        }
+                        this.notifyChanges();
+                    });
+                }
+            }
+        });
+
+        this.onRowDeleted.pipe(takeUntil(this.destroy$)).subscribe(args => {
+            if (this.rowSelection === 'multipleCascade') {
```
kdinev, line 388: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -360,12 +362,81 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
             this.loadChildrenOnRowExpansion(args);
         });
 
+        this.onRowAdded.pipe(takeUntil(this.destroy$)).subscribe(args => {
+            if (this.rowSelection === 'multipleCascade') {
```
kdinev, line 366: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-selection.service.ts
```diff
@@ -0,0 +1,280 @@
...
+
+
+@Injectable()
+export class IgxTreeGridSelectionService extends IgxGridSelectionService {
+    private rowsToBeSelected: Set<any>;
+    private rowsToBeIndeterminate: Set<any>;
+
+    /** Select specified rows. No event is emitted. */
+    selectRowsWithNoEvent(rowIDs: any[], clearPrevSelection?): void {
+        if (this.grid && this.grid.rowSelection === 'multipleCascade') {
```
kdinev, line 13: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-selection.service.ts
```diff
@@ -0,0 +1,280 @@
...
+            this.cascadeSelectRowsWithNoEvent(rowIDs, clearPrevSelection);
+            return;
+        }
+        super.selectRowsWithNoEvent(rowIDs, clearPrevSelection);
+    }
+
+    /** Deselect specified rows. No event is emitted. */
+    deselectRowsWithNoEvent(rowIDs: any[]): void {
+        if (this.grid.rowSelection === 'multipleCascade') {
```
kdinev, line 22: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-selection.service.ts
```diff
@@ -0,0 +1,280 @@
...
+            this.cascadeDeselectRowsWithNoEvent(rowIDs);
+            return;
+        }
+        super.deselectRowsWithNoEvent(rowIDs);
+    }
+
+    public emitRowSelectionEvent(newSelection, added, removed, event?): boolean {
+        if (this.grid.rowSelection === 'multipleCascade') {
```
kdinev, line 30: Checks should be performed against the enum, rather than against the string representation. This way we can change the string without having to find and replace it everywhere!

kdinev (changes requested):

projects/igniteui-angular/src/lib/grids/common/enums.ts
damyanpetev: Technically, the 'enum' is only there for compatibility with older code when those options were actual enums and for this specifically, it's not even needed but likely better to have for consistency. The selection property (and others we've converted) are simple string union types now - their usage in no way requires the exported constants like this one as they are not even necessary. Just like in markup, this is perfectly valid:
```ts
this.grid1.rowSelection = 'multiple';
```
There's intellisense autocomplete, it's type checked (i.e. you can't just assign a random string) and will even validate conditions like this:
image

It costs us very little override the `'single' | 'multiple' | 'multipleCascade'` with `'single' | 'multiple'` for the flat grid and I'd rather do that to avoid the API exposing something it doesn't support even if it would technically still work. There's no impact on existing code, even for the tree grid you can still assign values from the old constant too.
Again, the exported constants are only there for backwards compatibility and/or [questionable] convenience and the selection pre-dated the migration, so the new one is there to be consistent, though for new 'enums' those constants are not even needed to be publicly exported and can be kept just for internal references.

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
damyanpetev: Fair enough, there's very little reason to override this check indeed

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
damyanpetev: This is is overridden just so the getter doesn't complain about the return type, though (if the types are kept) this is easily adjusted in the getter as well with a type assert.

kdinev: @damyanpetev I don't want a new enum type though, and so this won't be necessary.

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
damyanpetev: Agreed, while it's technically safe, it's better if we refer to those values through the exported const if we ever have to rename something.

projects/igniteui-angular/src/lib/grids/common/enums.ts
kdinev: @damyanpetev This is unnecessary code pollution, which adds no value to the actual features. All it does for us is increase the side of the bundle.

damyanpetev: Yes, no value added for the feature itself, the only benefit was keeping the public API of the unaffected grids clean. We were more concerned with that, rather than the duplicating one get/set (the H Grid override is supposed to be temporary). However, I see your point about pollution and duplication and I agree it's unnecessary debt and a small bump to the packed size.

Not sure if unreasonable, but the duplication can be avoided if the base is generic on the selection mode, and this I think can be implemented fairly gracefully with very little changes overall by making it optional as well. Basically, just convincing TypeScript to show the correct type, while keeping all the code the same.

If you think that's overkill, then a comment on the property the `multipleCascade` only applies to the Tree Grid (or doesn't apply to flat and H grids) will have to do.

damyanpetev:
@teodosiah Can you also add an entry for the new feature in the changelog :)

kdinev (dismissed):

projects/igniteui-angular/src/lib/grids/selection/selection.service.ts
```diff
@@ -4,6 +4,7 @@ import { FilteringExpressionsTree } from '../../data-operations/filtering-expres
 import { IGridEditEventArgs, IGridEditDoneEventArgs } from '../common/events';
 import { GridType } from '../common/grid.interface';
 import { IgxGridBaseDirective } from '../grid/public_api';
+import { IgxTreeGridAPIService } from '../tree-grid/public_api';
```
damyanpetev, line 7: This import is now unused, as are the selection modes in Grid/HGrid

kdinev (approved):

damyanpetev (approved):

rkaraivanov (approved):

PR #8796, opened 2021-01-14
Add missing "ing" or "ed" events for grid features

### LINKED ISSUE #8773: Intruduce missing "ing" or "ed" events for grid features.
Currently some features introduce only "ing" events, like `onColumnPinning`, others introduce only "ed" events, like `onColumnResized`. Also, different naming convetion is used in event names and in interfaces.

First table below shows current situation, next table shows implementation plan.

⚠ Behavioral changes
🆕 New stuff
❌ Missing

  | Ing event | Ed event
-- | -- | --
Sorting | ❌ | onSortingDone
Filtering | ❌ | onFilteringDone
Paging | onPagingDone | ❌
Pinning | onColumnPinning | ❌
Hiding | ❌ | onColumnVisibilityChanged
Resizing | ❌ | onColumnResized
Moving | onColumnMovingStart, onColumnMoving | onColumnMovingEnd

Missing events need to be introduced. Also, try to unify naming convention and interfaces (in terms of structure and naming) as much as possible, with as less as possible breaking or behavioral changes. See table below for what will be implemented.

Еvent | Interface
-- | --
sorting   🆕 | ISortingEventArgs   🆕
filtering 🆕 | IFilteringEventArgs   🆕
onPaging 🆕 | IPagingEventArgs🆕
onColumnPinning ⚠ | IPinColumnCancellableEventArgs   🆕
columnVisibilityChanging   🆕 | IColumnVisibilityChangingEventArgs 🆕
columnResizing   🆕 | IColumnResizingEventArgs 🆕
onColumnMovingStart | IColumnMovingStartEventArgs
onColumnMoving | IColumnMovingEventArgs
 |  
onSortingDone | ISortingExpression   \| Array<ISortingExpression>
onFilteringDone | IFilteringExpressionsTree
onPagingDone ⚠ | IPageEventArgs
columnPinned 🆕 | IPinColumnEventArgs
onColumnVisibilityChanged | IColumnVisibilityChangedEventArgs
onColumnResized | IColumnResizeEventArgs
onColumnMovingEnd | IColumnMovingEndEventArgs

📝 Notes:
- All '-ing' events interfaces will extend `IBaseEventArgs` and `CancelableEventArgs`

⚠ Behavioral changes:
- `pageChange` is now hidden, and exposes a message to use `onPagingDone` instead.
- `onPagingDone` will be emitted AFTER the operation has completed
- `onColumnPinning` to emit `IPinColumnCancellableEventArgs`, instead of `IPinColumnEventArgs`. `isPinned` in both "ing" and "ed" events will contain the current pin state of the column.

NOTE: Paging related stuff will be implemented with #8923

hanastasov:
Closes #8773 

 - [x] Bug fix
 - [x] New functionality
 - [x] Documentation
 - [x] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1741,25 +1747,34 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
...
+            this._unpinnedIndex = grid._unpinnedColumns.length;
+        }
+
         index = (index !== undefined ? index :
             this._unpinnedIndex !== undefined ? this._unpinnedIndex : this.index);
+
+        const args: IPinColumnCancellableEventArgs = { column: this, insertAtIndex: index, isPinned: false, cancel: false };
+        this.grid.onColumnPinning.emit(args);
+
+        if (args.cancel) { return; }
```
Aleksandyr, line 1774: we should change isPinned to false if the event is canceld in order to reciev the right state of the column

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1741,25 +1747,34 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
...
 
+        // correct unpinnedIndex, if it exceeds unpinned columns length
+        if (this._unpinnedIndex > grid._unpinnedColumns.length) {
+            this._unpinnedIndex = grid._unpinnedColumns.length;
+        }
+
         index = (index !== undefined ? index :
             this._unpinnedIndex !== undefined ? this._unpinnedIndex : this.index);
+
+        const args: IPinColumnCancellableEventArgs = { column: this, insertAtIndex: index, isPinned: false, cancel: false };
```
Aleksandyr, line 1771: The column is not unpinned yet, so we should change isPinned to false here.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1682,14 +1681,19 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
...
 
-        this._pinned = true;
-        this.pinnedChange.emit(this._pinned);
         this._unpinnedIndex = grid._unpinnedColumns.indexOf(this);
         const rootPinnedCols = grid._pinnedColumns.filter((c) => c.level === 0);
         index = index !== undefined ? index : rootPinnedCols.length;
-        const targetColumn = grid._pinnedColumns[index];
-        const args = { column: this, insertAtIndex: index, isPinned: true };
-        grid.onColumnPinning.emit(args);
+        const args: IPinColumnCancellableEventArgs = { column: this, insertAtIndex: index, isPinned: true, cancel: false };
```
Aleksandyr, line 1687: The column is not pinned yet, so we should change isPinned to false here.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1682,14 +1681,19 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
...
+        this.grid.onColumnPinning.emit(args);
+
+        if (args.cancel) { return; }
+
+        grid.endEdit(true);
+
+        this._pinned = true;
+        this.pinnedChange.emit(this._pinned);
```
Aleksandyr, line 1695: we should emit pinnedChange after the state of the column is visually changed.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1855,6 +1868,20 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
         return children;
     }
 
+    /**
+     * Toggles column vibisility and emits the respective event.
+     * @hidden
+     */
+    public toggleVisibility(value?: boolean) {
+        const newValue = value ?? !this.hidden;
+        const eventArgs: IColumnVisibilityChangingEventArgs = { column: this, newValue, cancel: false };
```
Aleksandyr, line 1877: IMO the value of the args should not be updated here. Because if we cancel the event the column remains in the same state as it were so as the value does.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/grid.excel-style-filtering.component.ts
```diff
@@ -411,6 +412,15 @@ export class IgxGridExcelStyleFilteringComponent implements OnDestroy {
      * @hidden @internal
      */
     public onHideToggle() {
+        const eventArgs: IColumnVisibilityChangingEventArgs = {
+            column: this.column,
+            newValue: !this.column.hidden,
```
Aleksandyr, line 417: the value should reamin as it were and change it only after cancelation is passed

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
hanastasov: `isPinned` as part of the event arguments is meant to indicate the new value that will be applied in the end. I know the name can be confusing, but I kept it in order to keep consistency between `IPinColumnEventArgs` and the new `IPinColumnCancellableEventArgs`.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
hanastasov: isPinned as part of the event arguments is meant to indicate the new value that will be applied in the end. I know the name can be confusing, but I kept it in order to keep consistency between IPinColumnEventArgs and the new IPinColumnCancellableEventArgs.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
hanastasov: It is expected that the event arguments provide the new value. Also, the name `newValue` suggests the same. Otherwise an event would be emitted that does not contain neither the current nor the old value.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/grid.excel-style-filtering.component.ts
hanastasov: Please note that `newValue` is only passed in the event arguments and the value of `this.column.hidden` is not changed here.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -514,9 +520,16 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
-        this.onPagingDone.emit({ previous: this._page, current: val });
+
+        const eventArgs = { previous: this._page, current: val };
+        const cancelableEventArgs = { ...eventArgs, cancel: false, owner: this };
+        this.onPaging.emit(cancelableEventArgs);
+
+        if (cancelableEventArgs.cancel) { return; }
+
         this._page = val;
         this.pageChange.emit(this._page);
```
Aleksandyr, line 531: Shall we annotate that event as deprecated?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -1391,9 +1443,21 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
+     * Emitted before paging is performed.
+     * @remarks
+     * Returns an object consisting of the previous and current pages.
+     * @example
+     * ```html
+     * <igx-grid #grid [data]="localData" [height]="'305px'" [autoGenerate]="true" (onPagingDone)="pagingDone($event)"></igx-grid>
+     * ```
+     */
+    @Output()
+    public onPaging = new EventEmitter<IPagingEventArgs>();
```
Aleksandyr, line 1455: The event is with on prefix because of the existing paging method?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
hanastasov: Actually, I have wrongly indicated `pageChange` to be deprecated. It is an event to serve the two way data binding of the `page` property. I will correct the table with changes in the issue description now. Good catch.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
hanastasov: Exactly. I think it is still a matter of change in the naming, if we decide to.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
Aleksandyr: shaw we document it as hidden so it is not presented into the API doc?

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
Aleksandyr: We can also use the direct refference of the hidden prop, without changing it's value and have:
{..., value: this.hidden...}
and then once the state of the column is changed pass the new value after it is changed.

It is more logical to me that ing events emits before the any of the operations is done, which means that the value or the state of the current item is still the same otherwise there is no point of canceling it.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/grid.excel-style-filtering.component.ts
Aleksandyr: I see that you are using ! operator which changes it's state before the emition of the ing evt as you are doing after the canceletion which changes it's state IMO

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
Aleksandyr: Which means that we should expose another prop of the arguments which is just the value accompanied with the newValue

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
Aleksandyr: It is more subjective the thing we are discussing. Again IMO ing events shows the state before it is applied. Because If I decide to cancel it I will have in the arguments the value saying that the column is pinned which is not true. 

I may be biased a little bit, but that is why in the crud service we are exposing the old or current value and the new one, so the user is not getting confused

Aleksandyr: But if we are seeing it from the perspective where we are saying that this is the newValue really, it may be the right implementation.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
hanastasov: Well now I see it is indeed public. I agree we should hide it, so I will address this into the changes.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
Aleksandyr: Probably it could happen in that mass event naming change

ddincheva:
After testing I found a few problems:

- [ ] when you try to hide a column (through the toolbar) but cancel **columnVisibilityChanging** event in order to prevent that the checkbox status is incorrect. The same applies and for the _column pinning_ through the toolbar:
hiddingCancel

- [ ] when the column pinning position is set to end and you try to pin any column the value of the property **insertAtIndex** of the events **onColumnPinning** and **columnPinned** is not correct:
pinnedEnd

- [ ] when you try to filter a certain column but cancel **filtering** event the state of ESF is not correct:
filteringCancel

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
Aleksandyr: The more I think about that, the more I see the need of exposing the current and the new value in the ing evt args, but for now let it be like that and if we see an urge about it we are going to expose it.

Aleksandyr (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 All notable changes for each version of this project will be documented in this file.
 ## 11.1.0
+### General
+- `IgxGrid`, `IgxTreeGrid`, `IgxHierarchicalGrid`
+    - The following new events are introduced: `sorting`, `filtering`, `onPaging`, `columnPinned`, `columnVisibilityChanging`.
+    - **Behavioral Change** -
+    - `onColumnPinning` to emit `IPinColumnCancellableEventArgs` instead of `IPinColumnEventArgs`.
```
kdinev, line 9: Just these events should no longer be on-prefixed.

kdinev (changes requested):

CHANGELOG.md
hanastasov: We should decide whether to rename `paging` property in order to use this name for the event.

hanastasov: I have prepared a branch for it (currently adding migrations). When we decide on the name, this branch can be used.

hanastasov: #8911 is ready to be tested and merged if `paging` needs to be freed to be used as name for the event.

hanastasov: Grid properties related paging should be deprecated. If any input/output handling needs to be done, a template with a paginator should be used and all handling to happen on paginator level.

ddincheva (approved):

zdrawku (approved):

PR #8800, opened 2021-01-15
feat(toolbar): expose toggle evts and select/deselect functionality

### LINKED ISSUE #8732: Re-expose the igxToggle instance for toolbar actions components
## Description  
In versions prior to 11.0 users could use the Toolbar hiding and pinning dropdowns to achieve custom functionality such as reset the filtering UI when the end user closes the dropdown. As neither the igxToggle nor the igxColumnActions associated with the IgxGridToolbarHiding|PinningComponent are exposed users can no longer easily achieve the same functionality using the default toolbar components and have to create custom ones for otherwise trivial modifications.

 * igniteui-angular version: 11.x
 * browser: n/a

## Expected result  
For example, you should be able to easily subscribe to the igxToggle directive on the igxColumnActions component onClosed event to reset the ColumnActions UI .

Both the igxToggle and the igxColumnActions instances can be exposed as public properties of the igxGridToolbarHiding/PinningComponent

Aleksandyr:
Closes #8732

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar.base.ts
+++ b/projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar.base.ts
@@ -88,7 +88,7 @@ export abstract class BaseToolbarDirective implements OnDestroy {
     /** @hidden @internal */
     public toggle(anchorElement: HTMLElement, toggleRef: IgxToggleDirective, actions?: IgxColumnActionsComponent): void {
         if (actions) {
-            this._setupListeners(toggleRef, actions);
+            this._setupListeners(toggleRef);
             const setHeight = () => actions.columnsAreaMaxHeight = this.columnListHeight ?? `${Math.max(this.grid.calcHeight, 200)}px`;
             toggleRef.onOpening.pipe(first()).subscribe(setHeight);
         }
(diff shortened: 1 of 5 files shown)


Aleksandyr:
@rkaraivanov @ChronosSF Please do elaborate more on whether we would need to expose additional functionalities?

ChronosSF:
@Aleksandyr , I'll test the customer's case with your code and let you know if it is achievable

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar.base.ts
```diff
@@ -79,3 +130,16 @@ function _makeOverlaySettings(): OverlaySettings {
         closeOnOutsideClick: true
     };
 }
+
+export abstract class BaseToolbarColumnActions extends BaseToolbarDirective {
+
+    protected columnActionsUI: IgxColumnActionsComponent;
```
ChronosSF, line 136: This should be public so that users can change all the column action components properties by a ref to its toolbar wrapper.

ChronosSF (changes requested):

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar.base.ts
Aleksandyr: That was the first intention, not exposing those references directly to the users but just the API they would need.
That was the approach we've discussed with @rkaraivanov, building a facade around the toggle and column actions functionalities.

ChronosSF: Sure. But then you have to expose more stuff. The customer wants to reset the `filterCriteria` on closing so you need to expose that property from the actions component at least. I'd probably expose all public properties that control its state just to be on the safe side.

Aleksandyr: Yep, I understand. That is why the PR is in a DRAFT state as it is not ready and it just serves as a starting point of discussion. I will definitely expose some more methods.

rkaraivanov (approved):

PR #8823, opened 2021-01-21
feat(overlay): add detach method to overlay service

### LINKED ISSUE #6063: Add detach method to IgxOverlay
Right now `IgxOverlay` has `attach` method. This method resolves the passed component and generates unique ID. Later on developer can show the overlay with provided ID.
However when developer calls `hide` method it resets the overlay for provided ID and this ID cannot be used anymore.
There should be `detach` method which should reset the overlay for provided ID. Hide method should only hide the overlay for provided ID without resetting the overlay.

Add `detach` method.

One should call `attach` after each time `hide` was called. This forces overlay to resolve same component several times.

### LINKED ISSUE #7088: Overlay in edit row template
## Description  
Clicking really fast over the row editable rows by constantly exiting and entering in edit mode, leads to confusion in overlay id's. 

After some investigation, it was determined that the **attach** method gets triggered before the animation has ended. So the callback in **onDone** of the **animationPlayer** lags.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open grid edit row samples
2. Hit a breakpoint in base-grid.directive.ts in the subscription of **onOpened** of the overlay
3. Compare the emitted **ids** and the ones registered in the overlay.

## Result  
There is no consistency in the registration and emit process of the overlay id's.

## Expected result  
Needs consistency in the cycle of registering and emitting correct overlay id's

wnvko:
Closes #6063 
Closes #7088 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


DiyanDimitrov:
We have been debugging the ESF and we didn't find any issue with populating the items. However we have found that if we replace the `igFor` with `ngFor` in the ESF search component, the tests are passing. We have no idea how the changes to the overlay cause the regression in the `igFor`.
@ChronosSF, Could someone from your team investigate this?

wnvko:
What I've found is while the failing test, `Should filter and clear the excel search component correctly` is running `IgxForOfDirective.addLastElem` method is called seven times, while the `displayedListData` (collection shown in `IgxExcelStyleSearchComponent`) has six items. Here is how `igx-for-of` is used in the search component:
```
<igx-list-item *igxFor="let item of displayedListData scrollOrientation : 'vertical'; containerSize: containerSize; itemSize: itemSize">
...
</igx-list-item>
```
The first and last item shown are the same, `Select All` in this case. What I've found too is if I log the `embeddedView` in `addLastElem` the test passes. Therefore this is some timing issue may be.

ChronosSF:
I'll check it out but it seems like a pretty big change and virtualization tests are sometimes quite difficult to unravel . It'll take a while before I can give you some idea what could be wrong.

ChronosSF:
Yeah, seems like a timing issue. It seems to get solved if your `displayedListData` collection is initialized with an empty array from the get-go. Having it undefined at start leads to some stuff getting fixed in forOf at a very late stage.

```typescript
public displayedListData: FilterListItem[] = [];
```

I still can't tell why this differs from before the detach method though. I'll keep looking into it.

wnvko:
@ChronosSF I have added an initialization for `displayedListData` and still get the same errors in test. Can you check this?

ChronosSF:
Well it wasn't a good fix anyway. I am glad it doesn't work. I'll take another look but it's timing related and infuriating to debug.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -997,7 +997,10 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
     /** @hidden @internal */
     public ngOnDestroy(): void {
         if (this._componentID) {
-            this._overlayService.hide(this._componentID);
+            this._overlayService.detach(this._componentID);
+        }
+        if (this._statusChanges$) {
```
ViktorSlavov, line 1002: This duplicate seems like a leftover

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3185,7 +3185,8 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
      */
     public hideOverlays() {
         this.overlayIDs.forEach(overlayID => {
-            this.overlayService.hide(overlayID);
+            // this.overlayService.hide(overlayID);
+            this.overlayService.detach(overlayID);
             this.overlayService.onClosed.pipe(
```
ViktorSlavov, line 3190: Since `detach` doesn't emit `onClosed`, could this not cause a memory leak?

ViktorSlavov (changes requested):

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
wnvko: `hide` is deleted and `detach` is added. I do not see a duplicate or I am missing something?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
wnvko: yeah, not sure if this will leak, but absolutely surely we do not need here this subscription
nice catch :)

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
wnvko: find it finally
leftover removed

ChronosSF:
By the way, having to move more ESF tests to async probably relates with the issues @simeonoff has with the calendar. @rkaraivanov and @nrobakova have been examining issues in the ESF tests greatly slowing down test execution. It is likely that all these issues have a common cause.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -933,4 +875,75 @@ export class IgxOverlayService implements OnDestroy {
...
+            //  This is logged in Angular here - https://github.com/angular/angular/issues/18891
+            //  As soon as this is resolved we can remove this hack
+            const innerRenderer = (info.openAnimationPlayer as any)._renderer;
+            info.openAnimationInnerPlayer = innerRenderer.engine.players[innerRenderer.engine.players.length - 1];
+            info.openAnimationPlayer.onDone(() => {
+                this.onOpened.emit({ id: info.id, componentRef: info.componentRef });
+                if (info.openAnimationPlayer) {
+                    info.openAnimationPlayer.reset();
+                    // calling reset does not change hasStarted to false. This is why we are doing it here via internal field
+                    (info.openAnimationPlayer as any)._started = false;
```
jackofdiamond5, line 904: I traced the call to `reset` in the renderer to a `WebAnimationsPlayer.reset` which actually sets `_started` to `false`, is there a situation where a different player is used which doesn't do it?

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -933,4 +875,75 @@ export class IgxOverlayService implements OnDestroy {
...
+        }
+        if (info.settings.positionStrategy.settings.closeAnimation) {
+            const animationBuilder = this.builder.build(info.settings.positionStrategy.settings.closeAnimation);
+            info.closeAnimationPlayer = animationBuilder.create(info.elementRef.nativeElement);
+
+            //  AnimationPlayer.getPosition returns always 0. To workaround this we are getting inner WebAnimationPlayer
+            //  and then getting the positions from it.
+            //  This is logged in Angular here - https://github.com/angular/angular/issues/18891
+            //  As soon as this is resolved we can remove this hack
+            const innerRenderer = (info.closeAnimationPlayer as any)._renderer;
```
jackofdiamond5, line 921: 👀

projects/igniteui-angular/src/lib/services/overlay/utilities.ts
```diff
@@ -174,9 +178,10 @@ export class Util {
      * @hidden
```
jackofdiamond5, line 178: If the whole class is marked as `@hidden / @internal`, its members do not need to be marked separately.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
wnvko: there are some updates in Angular animation library. However there are two things here:
- we can clean up our overlay animations implementation in separate task.
- from the other side we are now in a process of moving overlay animations in separate igxAnimation service when it is not even sure whether or not we will use Angular animations.
This is why I do not think we should fix animations in this PR

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -284,91 +299,127 @@ export class IgxOverlayService implements OnDestroy {
...
 
-        this._show(info);
-        return id;
+        const eventArgs: OverlayCancelableEventArgs = { id, componentRef: info.componentRef, cancel: false };
+        this.onOpening.emit(eventArgs);
+        if (eventArgs.cancel) {
+            return;
+        }
+        if (settings) {
+            // TODO: update attach
```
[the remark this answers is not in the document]
wnvko: This will not be added in this PR. This is a new feature. Now one can easily work around this by calling `detach ->attach(newSettings) -> show`.

jackofdiamond5 (approved):
I think we should clean up our overlay animations' implementation as soon as possible since we are accessing internal members of the animation players and Angular could change them without prior notice which will break our overlay.

jackofdiamond5 (approved):

ViktorSlavov (approved):

jackofdiamond5 (approved):

PR #8827, opened 2021-01-22
Apply formatter in summaries

### LINKED ISSUE #8786: Summaries does not respect column formatter
## Description  
Summaries does not respect column formatter. The values in the summary results respect the column pipeArgs, but not the column formatter.

 * igniteui-angular version: 10.2
 * browser: NA

## Steps to reproduce  

1. Open Grid Formatting dev demo
2. Set IgxDateSummaryOperand to the OrderDate column:
```ts
public dateSummary = IgxDateSummaryOperand;
```
```html
<igx-column field="OrderDate" width="200px" [dataType]="'date'" filterable="true" [hasSummary]="true" [summaries]="dateSummary"
```

3. Change the return value in the formatter function if you would like
4. Click the Use formatter function for OrderDate column.

## Result  
The summary results does not respect the formatter function.

## Expected result  
The summary results should respect the formatter function.

## Attachments  
image

mtsvyatkova:
Closes #8786

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -627,7 +628,13 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
     @notifyChanges()
     @WatchColumnChanges()
     @Input()
-    formatter: (value: any) => any;
+    set formatter (value: any) {
+        this._formatter = value;
+        this.grid.summaryService.clearSummaryCache();
```
DiyanDimitrov, line 633: Clearing the summary cache is overkill since we only need somehow to trigger the change detection of the summary cell.

projects/igniteui-angular/src/lib/grids/summaries/summary-cell.component.html
```diff
@@ -12,7 +12,11 @@
 
                 <span class="igx-grid-summary__label" [title]="summary.label">{{ translateSummary(summary) }}</span>
                 <span class="igx-grid-summary__result" [title]="summary.summaryResult">
-                    {{ summary.summaryResult }}
+                    {{
+                        (isDateSummary(summary.key) && column.formatter)
```
DiyanDimitrov, line 16: We need to call the formatter for each column not only the date columns.

DiyanDimitrov (changes requested):
I still believe that it is better to move the pipe formatting to the summary cell template instead of formatting the values in the summary operands. The operands should only calculate the value no matter how it is going to be visualized.

@hanastasov @rkaraivanov, What do you think about this? Is it too late to change this?

hanastasov:
> I still believe that it is better to move the pipe formatting to the summary cell template instead of formatting the values in the summary operands. The operands should only calculate the value no matter how it is going to be visualized.
> 
> @hanastasov @rkaraivanov, What do you think about this? Is it too late to change this?

For me it also makes more sense that the operand only calculate a value. The reason I put the formatting logic in the operands is that otherwise it would be impossible to put custom formatting (custom in terms that differs from the column pipe args) (1).

Thinking more about this I find this case (1) to fall amongst the 20% case, rather than in the 80% case, when a dev would format both the data cells and the summary cells using the same formatter/pipe args.

I dont think it is late for a change.

projects/igniteui-angular/src/lib/grids/summaries/summary-cell.component.html
hanastasov, line 16: Does `isDateSummary` covers the case of a custom summary key? If not, the column dataType could be considered (assuming we wont be covering another case - when summary data type differs from column data type).

DiyanDimitrov:
> > I still believe that it is better to move the pipe formatting to the summary cell template instead of formatting the values in the summary operands. The operands should only calculate the value no matter how it is going to be visualized.
> > @hanastasov @rkaraivanov, What do you think about this? Is it too late to change this?
> 
> For me it also makes more sense that the operand only calculates a value. The reason I put the formatting logic in the operands is that otherwise it would be impossible to put custom formatting (custom in terms that differs from the column pipe args) (1).
> 
> Thinking more about this I find this case (1) to fall amongst the 20% case, rather than in the 80% case, when a dev would format both the data cells and the summary cells using the same formatter/pipe args.
> 
> Also, I guess we have to understand `custom summaries` in terms of placing custom business logic in calculating, and not mix this concept with formatting.
> 
> Now it seems this was a mistake so I think it is fine to fix it.

On second thought the summary result might not be the same type as the column values like for example the `count` summary will return number for all column types. We need some further discussion on the matter.

hanastasov:
> > > I still believe that it is better to move the pipe formatting to the summary cell template instead of formatting the values in the summary operands. The operands should only calculate the value no matter how it is going to be visualized.
> > > @hanastasov @rkaraivanov, What do you think about this? Is it too late to change this?
> > 
> > 
> > For me it also makes more sense that the operand only calculates a value. The reason I put the formatting logic in the operands is that otherwise it would be impossible to put custom formatting (custom in terms that differs from the column pipe args) (1).
> > Thinking more about this I find this case (1) to fall amongst the 20% case, rather than in the 80% case, when a dev would format both the data cells and the summary cells using the same formatter/pipe args.
> > Also, I guess we have to understand `custom summaries` in terms of placing custom business logic in calculating, and not mix this concept with formatting.
> > Now it seems this was a mistake so I think it is fine to fix it.
> 
> On second thought the summary result might not be the same type as the column values like for example the `count` summary will return number for all column types. We need some further discussion on the matter.

True, I haven't thought about it too. But even if the column is numeric type, it does not make sense to format the `count` using the same formatter - formatter might output a label `"low"` or `"high"` for certain values, while this is not applicable for `count`.

hanastasov:
Following a discussion, we agreed to:

1. Intoduce a new input for the column, which is a formatter to format the summary result. Possible name is `summaryFormatter`. The `summaryFormatter` function should take as input the value and the summary operand that will calculate the summary result.
2. Keep operands to return a "clear", non-formatted value (this means to not take pipeArgs argument). summary-cell.component.html template will format the `summaryResult` through the `summaryFormatter` by default, or by the column `pipeArgs` when `summaryFormatter` is not available
3. Let's see how 2) would work for summaries of different types. For example: If the "ReleaseDate" column has `pipeArgs`, it does not make sense to format the `Count` summary using those `pipeArgs`.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -569,6 +570,11 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
     @WatchColumnChanges()
     @Input()
     formatter: (value: any) => any;
+
+    @notifyChanges()
+    @WatchColumnChanges()
+    @Input()
+    summaryFormatter: (summary: IgxSummaryResult, summaryOperand: IgxSummaryOperand) => string;
```
DiyanDimitrov, line 577: Let's make the return type `any` in case they want to format the values in other types.
```suggestion
    summaryFormatter: (summary: IgxSummaryResult, summaryOperand: IgxSummaryOperand) => any;
```

projects/igniteui-angular/src/lib/grids/summaries/summary-cell.component.html
```diff
@@ -12,7 +12,11 @@
 
                 <span class="igx-grid-summary__label" [title]="summary.label">{{ translateSummary(summary) }}</span>
                 <span class="igx-grid-summary__result" [title]="summary.summaryResult">
-                    {{ summary.summaryResult }}
+                    {{
+                        column.summaryFormatter
+                        ? (summary | summaryFormatter:column.summaries:column.summaryFormatter)
+                        : appySummaryPipe(summary)
```
DiyanDimitrov, line 18: It is better to apply the pipes in the template so that we don't create the pipes each time in the code and also benefit from the built-in cache of the pipes.

projects/igniteui-angular/src/lib/grids/summaries/summary-cell.component.ts
```diff
@@ -91,7 +96,46 @@ export class IgxSummaryCellComponent {
...
         return this.grid.resourceStrings[`igx_grid_summary_${summary.key}`] || summary.label;
     }
+
+    /**
+     * @hidden @internal
+     */
+    public appySummaryPipe(summary: IgxSummaryResult): any {
+        const pipeArgs = this.column.pipeArgs;
+        const locale = this.grid.locale;
+        if (summary.key === 'count') {
```
DiyanDimitrov, line 116: IMO we should transform with the pipe the `count` key only if it is from some of our summary operand types.

projects/igniteui-angular/src/lib/grids/summaries/grid-summary.ts
```diff
@@ -163,26 +162,25 @@ export class IgxNumberSummaryOperand extends IgxSummaryOperand {
     public operate(data: any[] = [], allData: any[] = [], fieldName?: string, locale: string = 'en-US',
         pipeArgs: IColumnPipeArgs = {}): IgxSummaryResult[] {
```
DiyanDimitrov, line 163: We should remove the `locale` and `pipeArgs` parameters from the `operate` methods.

projects/igniteui-angular/src/lib/grids/summaries/summary-cell.component.html
```diff
@@ -12,7 +12,21 @@
...
+                        : (isNumberOperand() && summary.key !== 'count')
+                        ? (summary.summaryResult | number:column.pipeArgs.digitsInfo:grid.locale)
+                        : (isDateOperand() && summary.key !== 'count')
+                        ? (summary.summaryResult | date:column.pipeArgs.format:column.pipeArgs.timezone:grid.locale)
+                        : (isCurrencyOperand() && summary.key !== 'count')
+                        ? (summary.summaryResult | currency:currencyCode:column.pipeArgs.display:column.pipeArgs.digitsInfo:grid.locale)
+                        : (isPercentOperand() && summary.key !== 'count')
+                        ? (summary.summaryResult | percent:column.pipeArgs.digitsInfo:grid.locale)
+                        : (summary.key === 'count' && (isNumberOperand() || isDateOperand() || isCurrencyOperand() || isPercentOperand()))
+                        ? (summary.summaryResult | number)
```
DiyanDimitrov, line 27: We should still use the `grid.locale` for the built-in `count` summaries.
```suggestion
                        ? (summary.summaryResult | number:undefined:grid.locale)
```

projects/igniteui-angular/src/lib/grids/summaries/summary-cell.component.ts
```diff
@@ -91,7 +103,59 @@ export class IgxSummaryCellComponent {
         return (this.column.grid as any);
     }
 
+    /**
+     * @hidden @internal
+     */
+    public get currencyCode(): string {
+        return this.column.pipeArgs.currencyCode ?
+            this.column.pipeArgs.currencyCode  : getLocaleCurrencyCode(this.grid.locale);
```
DiyanDimitrov, line 111:
```suggestion
            this.column.pipeArgs.currencyCode : getLocaleCurrencyCode(this.grid.locale);
```

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -569,6 +570,11 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy {
     @WatchColumnChanges()
     @Input()
     formatter: (value: any) => any;
+
```
DiyanDimitrov, line 573: Missing API comment.

projects/igniteui-angular/src/lib/grids/common/pipes.ts
```diff
@@ -307,6 +308,15 @@ export class IgxColumnFormatterPipe implements PipeTransform {
     }
 }
 
+@Pipe({ name: 'summaryFormatter' })
+export class IgxSummaryFormatterPipe implements PipeTransform {
+
+    transform(summaryResult: IgxSummaryResult, summaryOperand: IgxSummaryOperand,
+        summaryFormatter: (s: IgxSummaryResult, o: IgxSummaryOperand) => string) {
```
DiyanDimitrov, line 315:
```suggestion
        summaryFormatter: (s: IgxSummaryResult, o: IgxSummaryOperand) => any) {
```

DiyanDimitrov (changes requested):
We should remove the `clearSummaryCache` call from the `pipeArgs` setter.

DiyanDimitrov (dismissed):

projects/igniteui-angular/src/lib/grids/grid/grid-summary.spec.ts
```diff
@@ -27,6 +27,8 @@ import { GridSummaryCalculationMode } from '../common/enums';
 import { IgxNumberFilteringOperand, IgxStringFilteringOperand } from '../../data-operations/filtering-condition';
 import { SortingDirection } from '../../data-operations/sorting-expression.interface';
 import { DropPosition } from '../moving/moving.service';
+import { DatePipe, registerLocaleData } from '@angular/common';
+import localeFR from '@angular/common/locales/fr';
```
kdinev, line 31: In this PR: #9182 the `registerLocaleData` call is being moved to the `test.ts` file, instead of having to call it in each file individually. I would wait for that PR to get merged and then remove the specific locale registration from this test.

kdinev (changes requested):

projects/igniteui-angular/src/lib/grids/grid/grid-summary.spec.ts
kdinev: @mtsvyatkova I'm doing the merge now and I will resolve my own comment :)

kdinev (dismissed):

kdinev:
@mtsvyatkova The pipelines build will fail with one test failure, which is normal. The test should fail and it's a known issue with `ivy=false`.

hanastasov (approved):

DiyanDimitrov (approved):

kdinev (approved):

PR #8835, opened 2021-01-25
feat(themes): add scoped custom CSS properties for themes, typography, and elevations

### LINKED ISSUE #8824: Themes: Components should use global palette values (CSS variables)
Generated component themes specify their own custom CSS properties for the various colors they use. This is great as it allows for per-component custom themes, however, those custom CSS properties do not refer to colors from a common palette in the generated CSS, even though they use the same palette when generating the theme in Sass. All custom CSS color properties of a component theme use hard-coded hex/rgba values.

While there are somewhat valid reasons for that, like IE11 support and modifying some of the palette colors at build time for some of the components, there are huge drawbacks as well. For one, by not referring to common palette colors in the form of CSS variables, individual component colors are difficult to update at once as each component must update its individual color properties, which becomes unwieldy if one wanted to overhaul all common colors for many components. Conversely, this forces users to create custom component themes for all components in their application so that they themselves rewire the component colors to point to the same color reference on a higher level.

1. Themes produced by including `igx-theme` should add the colors of the passed palette to the scope of the theme in the document as CSS custom properties (variables) by default;
2. Component themes should refer to the produced custom CSS properties from the palette, instead of using the hard-coded color values from the passed $palette Sass map;
3. The generated color palette must include the contrast colors for each shade of the palette.
4. Modification of the colors should happen using pure CSS color functions. (This pertains to the default colors as defined in the component theme schema files);

#### Optional:
1. Generating the color shades of a palette may need to be revised so that the procedure is easily translatable to native CSS color functions, instead of being specific to Sass. This will greatly reduce the amount of colors that need to be modified at runtime and will allow primary, secondary, etc. colors to be replaced with color references from another theme, like Bootstrap.

### LINKED ISSUE #8830: Themes: igx-color should return references to CSS variables
In order for us to be able to implement the features listed in #8824, some changes need to be implemented in the `igx-color` function.

With legacy support off, the `igx-color` function should:

- If palette is null, the `igx-color` function should try to retrieve the color from the CSS vars palette
- If palette is not null, the `igx-color` function should retrieve the color from the passed palette

With legacy support on, the `igx-color` function should:

- If palette is null, the `igx-color` function should try to retrieve the color from the `$default-palette`
- If palette is not null, the `igx-color` function should retrieve the color from the passed palette

## Additional context
This should not affect the current API and no breaking changes should be introduced.

### LINKED ISSUE #8831: Themes: igx-contrast-color should return references to CSS variables
In order for us to be able to implement the features listed in #8824, some changes need to be implemented in the `igx-contrast-color` function.

With legacy support off, the `igx-color` function should:

  - If palette is null, the `igx-contrast-color` retrieves the color from the CSS vars palette
  - If palette is not null, the `igx-contrast-color` retrieves the color from the passed palette

With legacy support on, the `igx-color` function should:

  - If palette is null, the the `igx-contrast-color` retrieves the color from the `$default-palette` 
  - If palette is not null, the `igx-contrast-color` function should retrieve the color from the passed palette

## Additional context
This should not affect the current API and no breaking changes should be introduced.

### LINKED ISSUE #8832: Themes: to-opaque(hexrgba) should be able to work with palette var refs
In order for us to be able to implement the features listed in #8824, some changes need to be implemented in the `to-opaque(hexrgba)` function.

The to-opaque function should be able to receive a palette. Based off of whether a palette is present or not, the `to-opaque` function should be able to:

With legacy support:
 - act as it currently does, retrieving the colors from the current palette
 
Without legacy support:
- deconstruct the css var reference, if it's a valid palette color, try to resolve it to a color from the passed palette, or from the `$default-palette`.

### LINKED ISSUE #8833: Themes: Grid selected cell background not listed in theme props
## Description  
The `igx-grid-theme` should list a property for `$row-selected-cell-background` instead of trying to calculate it from the row-selected-background as it currently is. The property should be added to both the theme and the schemas of the grid.

 * browser: all

### LINKED ISSUE #8834: Themes: Do not pass palette to individual component themes
In order for us to be able to implement the features listed in #8824, some changes need to be implemented in the `igx-theme` mixin.

Theme should be omitted by default for individual component themes included in the `igx-theme` mixin. In addition to that, individual component theme functions should use the `$default-palette` if no `$palette` is provided.

### LINKED ISSUE #8848: Themes: Include contrast colors in the generated palette
Palettes generated with the `igx-palette` function do not include the contrast colors for each shade. The only way to retrieve a corresponding contrast color is to use the `igx-contrast-color` function, which will get the original color shade from the palette and calculate contrast for it.

Generated palettes should include contrast colors for each color shade/variant. In this way all contrast colors are calculated just once for all palette colors.

### LINKED ISSUE #8868: Themes: Scope custom CSS props to component selectors
Right now all component theme-related custom CSS props are scoped to the `:root` selector. This works, but to avoid name collision we have to prefix all props with `--igx-{component-name}` and it results in some very long prop names and pollutes the `:root` scope.

To simplify the names and ensure consistency between Sass theme prop names and custom CSS props, CSS props should be scoped to the component selector.

In this way we will go from this:
```css
:root {
  --igx-avatar-background: gray;
  --igx-avatar-color: white;
}
```

to this:

```css
igx-avatar {
  --background: gray;
  --color: white;
}
```

This should not introduce any breaking changes.

### LINKED ISSUE #8920: Themes: Expose elevations as custom CSS props
## Is your feature request related to a problem? Please describe
Expose all levels of elevation (0..24) as custom CSS props.

```css
  --igx-elevation-0: ...;
  --igx-elevation-1: ...;
```

1. Create a mixin called `igx-elevation-vars` that include all custom CSS elevation props in a given scope for a given Sass map as produced by the `igx-elevations` function.

2. Update the `igx-elevation` function to retrieve an elevation level as either :
  - list of box-shadows (when a concrete `$elevations` map is passed or `$igx-legacy-support` is set to `true`);
  - CSS variable ref (when a concrete `$elevations` map is omitted and `$igx-legacy-support` is set to `false`);
  - if `$igx-legacy-support` is set to `true` and the `$elevation` map is not passed, try to resolve the elevations from the default `$elevation` map;

## Additional context
This feature is a part of #8868

### LINKED ISSUE #8921: Themes: Cover functions and mixins with unit tests
Cover all functions and mixins with tests using True.
This includes functions and mixins declared in `utilities`, `typography`, and `elevations`.

### LINKED ISSUE #9119: Themes: extended palette variants should be based on the 500 variant
The palette variants exposed as CSS variables should be based on the 500 variant so they can be modified at runtime without the need to change every single color variant separately.

With the upcoming changes to the theming in 12.0.0, users will be able to modify palette colors using CSS variables:

```css
:root {
  --igx-primary-500: orange;
}
```

This will work but it will not automatically update all other variants of the primary color. Users will need to change each one individually.

We should try to come up with a CSS compatible solution that generates color variants from the base 500 variants for the primary, secondary, and grays palette at run-time. We can enforce setting the primary color as HSL values so that we can recalculate them;

For instance, `#09f` (Infragistics blue) can be represented like so:

```css
:root {
  --igx-primary-h: 204;
  --igx-primary-s: 100%;
  --igx-primary-l: 50%;

  /* later */
  --igx-primary-500: hsl(var(--igx-primary-h), var(--igx-primary-s), var(--igx-primary-l));

  /* the 600 variant of the primary color generated using calc */
  --igx-primary-600: hsl(var(--igx-primary-h), var(--igx-primary-s), calc(var(--igx-primary-l) * .9));
}
```
This requires a paradigm shift in how our users will define their main `primary`, `secondary` and so on colors, but it allows for a much more flexible and run-time possible modification of the entire color palette.

simeonoff:
Closes #8824 
Closes #8830
Closes #8831
Closes #8832
Closes #8833
Closes #8834
Closes #8848 
Closes #8868 
Closes #8920 
Closes #8921
Closes #9119

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


SisIvanova:
Samples that don't display correctly:
- http://localhost:4200/samples/interactions/drag-dialog-sample

SisIvanova:
- grid summary labels have wrong color
- input group inside the grid filtering row has a missing border
- slider track steps have wrong color

SisIvanova (approved):

PR #8864, opened 2021-01-28
igx-carousel accessibility improvements

### LINKED ISSUE #8202: igx-carousel accessibility improvements
The current `aria` roles applied on the `igx-carousel` widget fail in an accessibility scan. The carousel is marked as `role="list"`, but slides are not marked as list items. 

The carousel should be handled as a `tab` widget, where the buttons to choose a slide are `tabs` and the slides are `tabpanels`.

## Additional context
Put a carousel component in a document and run the lighthouse scan to check the a11y failures.

ddincheva:
Closes #8202  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


kdinev (dismissed):

projects/igniteui-angular/src/lib/carousel/carousel.component.html
```diff
@@ -21,30 +21,32 @@
     </a>
 </ng-template>
 
-
-<div *ngIf="showIndicators" [ngClass]="indicatorsOrientationClass">
+<div *ngIf="showIndicators" [ngClass]="indicatorsOrientationClass" [attr.role]="'tablist'">
     <div *ngFor="let slide of slides"
         class="igx-carousel-indicators__indicator"
         (click)="select(slide)"
+        [id]="setId(slide)"
```
kdinev, line 28: Use a getter instead of a method for the attributes. Same is applicable to `attr.aria-label`. You can host bind from within the component also with a getter.

projects/igniteui-angular/src/lib/carousel/carousel.component.ts
```diff
@@ -106,6 +106,14 @@ export class IgxCarouselComponent implements OnDestroy, AfterContentInit {
      */
     @HostBinding('attr.role') public role = 'region';
 
+    /** @hidden */
+    @HostBinding('attr.aria-roledescription')
+    public roleDescription = 'carousel';
+
+    /** @hidden */
+    @HostBinding('attr.aria-labelledby')
+    public labelId = `igxCarouselLbl`; // tab-${this.current + 1}-3
```
kdinev, line 115: Should this be static?

kdinev (changes requested):

projects/igniteui-angular/src/lib/carousel/carousel.component.ts
zdrawku: yep, this is something that we discussed while she was addressing some in-progress stuff, this will become:

```
public labelId = `${this.id}-label`;
```

the template will be updated accordingly

projects/igniteui-angular/src/lib/carousel/carousel.component.html
ddincheva: I agree with you that there is no need to use a method here, but I can not use a getter because the id and the labels that are constructed depends on the slide index and I can not pass a parameter to a getter, but maybe is better here to just construct the string literal directly in the template ... I don't see any benefit in using a function for that simple purpose. I can no use a host binding because tab elements are not in a separated component, they are part of the carousel component template.

projects/igniteui-angular/src/lib/carousel/carousel.component.ts
```diff
@@ -118,30 +109,23 @@ export class IgxCarouselComponent implements OnDestroy, AfterContentInit {
...
-        return 0;
-    }
+    @HostBinding('attr.role') public role = 'region';
+
+    /** @hidden */
+    @HostBinding('attr.aria-roledescription')
+    public roleDescription = 'carousel';
+
+    /** @hidden */
+    @HostBinding('attr.aria-labelledby')
```
zdrawku, line 127: Let's set this aria attr only when the label is visible - which is happening when the total slides are more than the `maximumIndicatorsCount` (5)

projects/igniteui-angular/src/lib/carousel/slide.component.ts
```diff
@@ -54,34 +57,40 @@ export class IgxSlideComponent implements OnDestroy {
...
      * ```typescript
-     * let slide = this.slide.ariaLive;
+     * let slide = this.slide.ariaSelected;
      * ```
      *
      */
     @HostBinding('attr.aria-selected')
-    public get ariaLive() {
-        return this.active ? 'polite' : null;
+    public get ariaSelected(): boolean {
```
zdrawku, line 92: As for the aria-selected error of the slide element, it appears that [when aria-selected is used](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role), its controlled `tabpanel` should have its `aria-expanded` attribute set to true and its hidden attribute set to false, otherwise the reverse.

image

ddincheva: aria-selected attribute should be added only to the element with role 'tab'. 'tabpanel' element should not have such an attribute /[aria tabpanel spec](https://www.w3.org/TR/wai-aria/#tabpanel)/.. so I just removed it

kdinev (approved):

zdrawku (approved):

PR #8939, opened 2021-02-09
IgxDatePicker refactoring

### LINKED ISSUE #6483: IgxDatePicker refactoring
Dependant on #6271

Related to #7199

[Specification](https://github.com/IgniteUI/igniteui-angular/wiki/Date-Picker-Specification)

With the new `IgxDateTimeEditorDirective` implemented, we should refactor the `IgxDatePicker` to use the new editor to handle mask parsing, spinning, date validation and any other action related to the date picker's input.

Tasks during refactoring:
- Inheritance of new base parent - introducing new members, renaming old ones (1 day) 
*  Integration of `IgxDateTimeEditorDirective`  (10 days)
     - Code clean up for already existing functionality in the `IgxDatePickerComponent`
     - Refactoring of model binding
     - Move `CalendarContainer` to global date directory
- Have `IgxDateRangePickerComponent` extend `PickersBaseDirective` (2 days)

-  Refactoring of `DatePickerUtils` - will be renamed to `DateTimeUtils` (4 days)
     - Unit tests for `DateTimeUtils`'s public methods
*  Simplify templating #8665
- New features that will be implemented
    - min and max values (will come from igxDateTimeEditor) #6286 
    - add type property #7196 
    - add ARIA (2 days) #6346
    - implementation of Validator interface (1 day)

* Test refactoring (2 days)
* Bugs that must be resolved (1 day) - #8551, #8567, #8475, #8065, #6749, #6925, #5525, #8613, #8913, #8919, #9154, #9189, #9215
* Migrations for `IgxDatePickerComponent` (2 days)
* Updating of `IgxDatePickerComponent`'s documentation and samples

Current estimations are purely for development time.

### LINKED ISSUE #7169: IgxDateTimeEditor should support spin delta
At the moment, the editor will only spin by `1` which may not cover some user scenarios.

I should be able to provide a number that will determine how the editor will increment or decrement its date parts.

n/a

## Additional context
n/a

### LINKED ISSUE #7199: Refactor igxDatePicker, igxTimePicker interfaces and define a common one
igxDatePicker implements the following [interface](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/date-picker/date-picker.common.ts#L5) and igxTimePikcer implements [this](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/time-picker/time-picker.common.ts#L8) one. After the `IgxDateTimeEditorDirective` is created, these interfaces should be revised. There are properties/methods/events (mode, calendar events) that are common for both, as well for the `IgxRangeDatePickerComponent`, so they should be extracted to a separate interface. Implements IToggleInterface (the latter should be extended and synced within the product)

- [x] Define a common abstract class for the `IgxDatePickerComponent`, `IgxTimePickerComponent` and `IgxDateRangePickerComponent` with base functionality (and possible additional extension for each component)
- [x] Update specification of `IgxDatePickerComponent`
- [ ] Update specification of `IgxTimePickerComponent`

<hr>

- [ ] Refactoring of `IgxDatePickerComponent`
- [ ] Refactoring of `IgxTimePickerComponent` 

`IgxDateTimeEditorDirective` needs to be used as the input for all pickers.

jackofdiamond5:
Closes #6483
Closes #7199
Closes #7169

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/date-picker/date-picker.common.ts
```diff
@@ -1,12 +1,20 @@
...
-    inputMask: string;
-    rawDateString: string;
-    dateFormatParts: any[];
-    invalidDate: string;
+/**
+ * Provides information about date picker reference and its current value
+ * when onDisabledDate event is fired.
+ */
+export interface IDatePickerDisabledDateEventArgs extends IBaseEventArgs {
+    datePicker: IgxDatePickerComponent;
```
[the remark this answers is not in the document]
jackofdiamond5: We have a deprecated `onDisabledDate` event which emits these args. As soon as we remove the event, we can remove the interface as well.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -508,990 +324,646 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
...
+    /**
+     * Selects today's date.
+     *
+     * @example
+     * ```typescript
+     * this.datePicker.selectToday();
+     * ```
+     */
+    public selectToday(): void {
+        this.select(new Date());
```
damyanpetev, line 560: Did we talk about removing this? Technically it should be new Date with cleared time portion, but that is prone to shift and I really would rather not go there if we can :)

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -508,990 +324,646 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
...
-        if (!this.disabled) {
-            this.isEmpty = true;
-            this.invalidDate = '';
-            this.deselectDate();
-            this._setCursorPosition(0);
+    //#region Control Value Accessor
+    /** @hidden @internal */
+    public writeValue(value: Date) {
+        this._value = value;
+        if (this.dateTimeEditor.value !== value) {
```
damyanpetev, line 697: Is this not bound already?

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -508,990 +324,646 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
...
+            Object.assign(errors, { value: true });
         }
+        if (value && this.disabledDates && isDateInRanges(value, this.disabledDates)) {
+            Object.assign(errors, { dateIsDisabled: true });
+        }
+        Object.assign(errors, DateTimeUtil.validateMinMax(value, this.minValue, this.maxValue));
 
-        if (this.collapsed) {
-            this._updateValidityOnBlur();
+        if (this.inputGroup && !this.inputGroup.isFocused) {
```
damyanpetev, line 736: Fairly certain you don't need to manipulate the input group state here, the whole point of validator integration is it will trigger the model state change we're hooked into already

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -508,990 +324,646 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
...
+    //#endregion
 
     /** @hidden @internal */
-    public onFocus(): void {
-        this._isInEditMode = true;
-        if (this.value && this.invalidDate === '') {
-            this._transformedDate = this._getEditorDate(this.value);
+    public ngOnChanges(changes: SimpleChanges): void {
+        if (changes['locale']) {
+            this.inputFormat = DateTimeUtil.getDefaultInputFormat(this.locale || 'en')
```
damyanpetev, line 749: if the directive handles this already, is this needed? Similar to the placeholder check

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -508,990 +324,646 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
...
+                || DateTimeUtil.DEFAULT_INPUT_FORMAT;
+        }
+        if (changes['displayFormat'] && this.dateTimeEditor) {
+            this.dateTimeEditor.displayFormat = this.displayFormat;
```
damyanpetev, line 753: If this is bound (it seems to be) than I don't see the point of it in ngOnChanges. Same for input format

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -508,990 +324,646 @@ export class IgxDatePickerComponent implements IDatePicker, ControlValueAccessor
...
-        this._onTouchedCallback();
-        const input = this.inputDirective;
-        if (input && this._ngControl && !this._ngControl.valid) {
-            input.valid = IgxInputState.INVALID;
-        } else {
-            input.valid = IgxInputState.INITIAL;
-        }
+    /** @hidden @internal */
+    public applyCustomFormat() {
+        return this.formatter ? this.formatter(this.value) : this.displayFormat;
```
damyanpetev, line 813: Yeah, those are very much _not_ the same thing. Display format is the template for formatting ('shortDate' or some pattern), while the formatter function straight up returns the formatted value, which is not compatible w/ the display format this is bound to on the directive.
Similar functionality may still be available on the mask/date-time directive, not sure

projects/igniteui-angular/src/lib/date-picker/date-picker.common.ts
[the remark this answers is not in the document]
damyanpetev: Oh in that case don't rename the arg - don't care to migrate a deprecated event before removing it :)

projects/igniteui-angular/src/lib/date-picker/date-picker.component.html
```diff
@@ -1,54 +1,30 @@
...
-            (blur)="onBlur($event)"
-            [tabindex]='editorTabIndex'
-            readonly
-        />
-    </igx-input-group>
-</ng-template>
+    <input class="igx-date-picker__input-date" igxInput [igxDateTimeEditor]="this.inputFormat"
+        [minValue]="this.minValue" [maxValue]="this.maxValue" [spinDelta]="this.spinDelta" [spinLoop]="this.spinLoop"
+        [displayFormat]="this.applyCustomFormat()" [disabled]="this.disabled" [placeholder]="this.placeholder"
+        [readonly]="!this.isDropdown" [igxTextSelection]="this.isDropdown" [required]="this.required"
```
damyanpetev, line 9: IIRC, we were avoiding setting the native `required` on inputs in other editors, because there's no ngControl on them and the `igxInput` directive will try and validate using the native call, which may collide with the parent component model state. Worth looking into

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
jackofdiamond5: No it's now bound, here we only update the internal `_value` and we also have to tell the directive to update its own value as well.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
jackofdiamond5: The validity is updated properly but the input group does not change its colors on its own if it has no focus.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -607,7 +607,7 @@ export class IgxDatePickerComponent extends PickerBaseDirective implements Contr
     public close(): void {
         if (!this.collapsed) {
             this._overlayService.hide(this._overlayId);
-            // TODO: detach()
+            this._overlayService.detach(this._overlayId);
```
damyanpetev, line 600: Umm wouldn't mess up the closing animation?

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.html
```diff
@@ -48,7 +48,7 @@
         [value]="value"
         [outlet]="filteringService.grid.outlet"
         [locale]="filteringService.grid.locale"
-        (onSelection)="onDateSelected($event)"
+        (valueChange)="onDateSelected($event)"
         (onClosed)="datePickerClose()">
         <ng-template igxDatePickerTemplate let-openDialog="openDialog">
```
damyanpetev, line 53: Oh boy, just noticed this isn't the simplest of uses and it's the deprecated template...

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
jackofdiamond5: I see what you mean and I think this could be a little bug in the input group itself. Because right now if you focus the input group and then blur it and make it invalid, the proper style will be applied, however, if you have not focused it then, yes, the model will become invalid but no styles will be applied. I'll test it to be sure.

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
```diff
@@ -450,6 +456,11 @@ export class IgxDateTimeEditorDirective extends IgxMaskDirective implements OnCh
       this.updateMask();
     }
 
+    // TODO: remove when formatter func has been deleted
+    if (this.displayValuePipe) {
```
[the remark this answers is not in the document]
jackofdiamond5: We need it because we don't want to call the mask's blur handler if a `displayValuePipe` is set. Because in the case with the date picker, the pipe will call the `formatter` function which is supposed to manipulate dates, while the argument for the mask's blur handler is a `string`. This will be removed as soon as the `formatter` is no longer present in the date picker.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
damyanpetev: In the end, there's some value to this method so we're keeping it after all

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
damyanpetev: Yeah, our editors are configured to show validation errors (through color) only after the model has been touched/is dirty and that's by design. If you modify the conditions with a value that's not entered by the user (not dirty) I think it's fairly reasonable to count this as app scenario that might as well update the value to a valid one or mark the model dirty if that;s the intention

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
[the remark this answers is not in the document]
damyanpetev: Still `displayValuePipe` functionality is inherited, apparently not very well though. Should be improved in the future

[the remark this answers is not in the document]
damyanpetev: Still displayValuePipe functionality is inherited, apparently not very well though. Should be improved in the future

projects/igniteui-angular/src/lib/date-common/picker-base.directive.ts
```diff
@@ -138,6 +138,17 @@ export abstract class PickersBaseDirective extends DisplayDensityBase implements
...
+    /**
+     * Gets/Sets the default template editor's tabindex.
+     *
+     * @example
+     * ```html
+     * <igx-date-picker editorTabIndex="1"></igx-date-picker>
+     * ```
+     */
+    @Input()
+    public tabindex: number | string;
```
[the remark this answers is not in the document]
jackofdiamond5: We can include the functionality for `tabIndex` in a separate PR.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -862,7 +865,10 @@ export class IgxDatePickerComponent extends PickerBaseDirective implements Contr
 
         this._overlayService.onOpened.pipe(...this._overlaySubFilter).subscribe((eventArgs) => {
             this.opened.emit(eventArgs as IBaseEventArgs);
-            this._calendar?.daysView?.focusActiveDate();
+            const errors = DateTimeUtil.validateMinMax(this.value, this.minValue, this.maxValue);
+            if (Object.keys(errors).length === 0) {
+                this._calendar?.daysView?.focusActiveDate();
```
damyanpetev, line 870: Except, if you don't have an active date and you don't focus... the document active element goes to.. body? And then keyboard nav won't work?
Might make a lot of sense to check when setting the `viewDate` if the current value is out of the valid range, to perhaps shift the calendar to a valid date and show that instead

jackofdiamond5: That's the better option and the behavior will be consistent with the time picker - if you are greater than the max value, then after opening the max value in the calendar will be focused. Same thing for min value.

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.html
Lipata: Will be fixed in a separate PR.

Issue #8881, opened 2021-02-02
Themes: Reduce compiled CSS size


simeonoff:
## Suggestion
The size of a single theme is ~370kB. We can reduce the size substantially by revising the produced CSS and pinpointing potential areas of improvement.

At a first glance we can:
- [x]  Reduce the nesting of placeholder selectors
- [x]  Stop extending base placeholder selectors in modifier classes
- [x]  Extract common style patterns into their own classes
- [ ] Printing Styles

simeonoff:
@desig9stein Printing styles need updating too.

PR #8951, opened 2021-02-10
feat(button): icon + label button styling

### LINKED ISSUE #8904: Icon + label Button within the Button Group
[The current implementation](https://www.infragistics.com/products/ignite-ui-angular/angular/components/button-group) of the button group (IgxButtonGroupModule) does not allow correct styling of the Icon+label button type when used within a button-group.
https://stackblitz.com/edit/github-c2skkg?file=src/app/dialog-sample-3/dialog-sample-3.component.html 

The Icon+Label/Label+Icon buttons when used in the button-group should look like the Icon+Label button component in terms of padding. 
https://material.io/components/buttons#anatomy

SisIvanova:
Closes #8904 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


desig9stein:
@SisIvanova  I am ok with your changes, however, due to the [task](https://github.com/IgniteUI/igniteui-angular/issues/8881) to find a way to reduce the CSS output I want to point something out.

I would rather use the igx-icon selector rather than %igx-icon-display like it's done for the Fab button. This will have a lot better CSS output. Take a look at the screenshots.
output of: [igx-icon](https://imgur.com/IV8cH3m)
output of: [%igx-icon-disply](https://imgur.com/WtmXjZA)

Using a placeholder selector `%igx-icon-display` that comes from another file, in this case, the icon-theme can cause problems. As you can see from the screenshots, this will add all icon modifiers that are extending the base `%igx-icon-disply` to the button selector. Also If someone changes the placeholder in the icon-theme it has to somehow know that this will break the icon override in the button theme and not only :). I noticed that we use the same approach in the list, grid, and input-group. and this is only for the icon base.

Summary: In my opinion, we should stop borrowing placeholder selectors in order to override styles from one component to another,  this will reduce selectors length in the outputted CSS.

 @simeonoff what do you think?

simeonoff:
> @SisIvanova I am ok with your changes, however, due to the [task](https://github.com/IgniteUI/igniteui-angular/issues/8881) to find a way to reduce the CSS output I want to point out something.
> 
> I would rather use the igx-icon selector rather than %igx-icon-display like it's done for the Fab button. This will have a lot better CSS output. Take a look at the screenshots.
> output of: [igx-icon](https://imgur.com/IV8cH3m)
> output of: [%igx-icon-disply](https://imgur.com/WtmXjZA)
> 
> Using a placeholder selector `%igx-icon-display` that comes from another file, in this case, the icon-theme can cause problems. As you can see from the screenshots, this will add all icon modifiers that are extending the base `%igx-icon-disply` to the button selector. Also If someone changes the placeholder in the icon-theme it has to somehow know that this will break the icon override in the button theme and not only :). I noticed that we use the same approach in the list, grid, and input-group. and this is only for the icon base.
> 
> Summary: In my opinion, we should stop borrowing placeholder selectors in order to override styles from one component to another, this will reduce selectors length in the outputted CSS.
> 
> @simeonoff what do you think?

I absolutely agree with you. We should be mindful of the implications of using placeholders selectors and use them sparsely both when extending and using them with other selectors.

simeonoff:
Perhaps the requirements were not clear.

Here's a rundown:

1. We should be able to have icons on both the left and the right of the text:

```html
<button igxButton>
  <igx-icon>smile</igx-icon>
  Text
  <igx-icon>smile</igx-icon>
</button>
```

2. We should be able to have more than one icon at a time on either side:
```html
<button igxButton>
  <igx-icon>smile</igx-icon>
  <span>Text</span>
  <igx-icon>smile</igx-icon>
  <igx-icon>smile</igx-icon>
</button>
```
3. The content next to the icon doesn't matter, it could be any element or text.

projects/igniteui-angular/src/lib/core/styles/components/button/_button-theme.scss
```diff
@@ -604,28 +604,44 @@
...
             font-size: $icon-in-button-size;
-            display: inline-flex;
-            order: -1;
+        }
+
+        igx-icon {
             margin-#{$right}: rem(12px);
         }
+
+        * ~ igx-icon {
```
simeonoff, line 613:
```suggestion
        * + igx-icon {
```
I'd suggest you use the `+` (next sibling) operator as we'd want to match only siblings that follow the icon.

simeonoff (changes requested):
I see this snippet repeated many times throughout the document:
```scss
igx-icon {
    margin-#{$right}: rem(8px);
}

* ~ igx-icon {
    margin-#{$right}: 0;
    margin-#{$left}: rem(8px);
}
```

You can extract it into a private (`_` prefixed) mixin with a single param for margin and then include it in all the places you need it. You can also create a map with the margins for the different display densities (similar to how it's done in the FAB).

simeonoff (approved):

desig9stein (approved):

PR #9030, opened 2021-02-22
fix(Grid Navigation): Adding a null check in focusTBody() after filtering out all grid records.

### LINKED ISSUE #8930: Getting  TypeError: Cannot convert undefined or null to object when clicking on empty row after filtering grid
## Description  
The browser console is logging error : "ERROR TypeError: Cannot convert undefined or null to object". This may not cause any problems for others. Usually the users would not notice this, but the project I am working on with my team has an error handler that displays a dialog with error message when any error occurs, including this one. 

 * igniteui-angular version:  11.0.x, 10.2.x
 * browser:  Google Chrome

## Steps to reproduce  
1. You can find this behavior even on the documentation page - https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/filtering , Take a look at the first example 
1.1 Or open this stackblitz example https://stackblitz.com/github/IgniteUI/igniteui-live-editing-samples/tree/master/angular-demos/grid/grid-filtering-sample?file=src%2Fapp%2Fapp.component.html
2.  Enter for example 'cha' as a filter for first column (Product Name) and hit enter.
3.  For the same column enter another filter that you know that would not match anything and give you an empty grid. For example enter 'xyz'.
4.  The grid should be empty now. Click somewhere in the empty grid. For now no error is logged.
5.  Remove the last filter that you applied ('xyz').
6.  You should now have two rows - Chai and Chang. Click on empty white space bellow the rows.

## Result  
ERROR TypeError: Cannot convert undefined or null to object is logged to the console. After you select any cell with data, this error doesn't appear any more after clicking the same whitespace area below the filtered rows.
In our project a dialog with error message is displayed.

## Attachments  
ignite

gedinakova:
Closes #8930   

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
@@ -35,6 +35,7 @@ export class IgxGridNavigationService {
     }
 
     public set activeNode(value: IActiveNode) {
+        // if (!value) debugger;
         this._activeNode = value;
     }
 
@@ -221,7 +222,8 @@ export class IgxGridNavigationService {
     focusTbody(event) {
         const gridRows = this.grid.verticalScrollContainer.totalItemCount ?? this.grid.dataView.length;
         if (gridRows < 1) { this.activeNode = null; return; }
-        if (!Object.keys(this.activeNode).length || this.activeNode.row < 0 || this.activeNode.row > gridRows - 1) {
+        if (!this.activeNode || !Object.keys(this.activeNode).length || this.activeNode.row < 0 || this.activeNode.row > gridRows - 1) {
+            debugger;
             const hasLastActiveNode = Object.keys(this.lastActiveNode).length;
             const shouldClearSelection = hasLastActiveNode && (this.lastActiveNode.row < 0 || this.lastActiveNode.row > gridRows - 1);
             this.setActiveNode(this.lastActiveNode.row >= 0 && this.lastActiveNode.row < gridRows ?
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -221,7 +221,7 @@ export class IgxGridNavigationService {
     focusTbody(event) {
         const gridRows = this.grid.verticalScrollContainer.totalItemCount ?? this.grid.dataView.length;
         if (gridRows < 1) { this.activeNode = null; return; }
```
gedinakova, line 223: @ddincheva The other option here is to change: this.activeNode = null to this.activeNode = {} as IActiveNode so we won't need the additional check in the next if. Both produce the same result in the scenario under test, so Desi, please, share your thoughts on whether one is a better option than the other.

ddincheva: hi @gedinakova, according to me is better to set the active node to an object rather than null, because if there are some places where is missed the check is active node is defined this will most probably cause runtime errors. I have made a fast check and in the hierarchical grid there are such scenarios and the following errors appear 
hGrid
Also as far as I know in the next major version of Angular strictNullChecks will be enabled by default so I believe is better to avoid explicitly set an object to null.

gedinakova: Yes, this is just another issue I am currently addressing. Will go for the second option.

ddincheva (approved):

ddincheva (approved):

PR #9069, opened 2021-02-25
fix(igx-dialog): max width

### LINKED ISSUE #9038: Dialog does not expands with content when templated
## Description  
Dialog does not expands with content  when templated 

 * igniteui-angular version: 11.1.0
 * browser: 88.0.4324.182

## Steps to reproduce  

1. Open example
- [example](https://stackblitz.com/edit/github-3zfykb?file=src%2Fapp%2Fdialog-sample-3%2Fdialog-sample-3.component.html)
- [example-dialog-11.1.0.zip](https://github.com/IgniteUI/igniteui-angular/files/6024597/sample-dialog-11.1.0.zip)
2. Click on Show Custom Dialog button.

## Result  
The dialog has a maximum width and the content is partially visible.

## Expected result  
The dialog expands with the content.

In version 11.0.7 the dialog was adjusted to the width of the content .

-  [example ](https://stackblitz.com/edit/github-3zfykb-mhjkkr?file=src%2Fapp%2Fdialog-sample-3%2Fdialog-sample-3.component.html)
- [example-dialog-11.0.7.zip](https://github.com/IgniteUI/igniteui-angular/files/6024595/sample-dialog-11.0.7.zip)

An alternative could be adding a field to allow the dialog to expand with the content.

## Attachments  

desig9stein:
fix: (igx-dialog) make sure that if the dialog content is templated we don't set max-width on the window content.

Closes #9038

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/dialog/dialog-content.component.html
+++ b/projects/igniteui-angular/src/lib/dialog/dialog-content.component.html
@@ -7,7 +7,7 @@
         <ng-content *ngIf="!title" select="igx-dialog-title,[igxDialogTitle]"></ng-content>
 
         <div class="igx-dialog__window-content">
-            {{ message }}
+            <span *ngIf="message" class="igx-dialog__window-message">{{ message }}</span>
             <ng-content *ngIf="!message"></ng-content>
         </div>
 
--- a/projects/igniteui-angular/src/lib/core/styles/components/dialog/_dialog-component.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/dialog/_dialog-component.scss
@@ -23,6 +23,10 @@
         @extend %igx-dialog-content !optional;
     }
 
+    @include e(window-message) {
+        @extend %igx-dialog-message !optional;
+    }
+
     @include e(window-actions) {
         @extend %igx-dialog-actions !optional;
     }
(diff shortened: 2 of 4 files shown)


kdinev (approved):

projects/igniteui-angular/src/lib/dialog/dialog-content.component.html
```diff
@@ -7,7 +7,7 @@
         <ng-content *ngIf="!title" select="igx-dialog-title,[igxDialogTitle]"></ng-content>
 
         <div class="igx-dialog__window-content">
-            {{ message }}
+            <span *ngIf="message" class="igx-dialog__window-message">{{ message }}</span>
             <ng-content *ngIf="!message"></ng-content>
```
kdinev, line 11: I think you can still leave the `<ng-content></ng-content>` be rendered even if there's a provided message. There could be a use case with elements going below a message which is programmatically populated!

simeonoff (approved):

projects/igniteui-angular/src/lib/dialog/dialog-content.component.html
desig9stein: I was thinking the same thing, but decide not to do it, because the message will have that max-width of 40ch which was the problem for the templated area, and if we have both, they will look misaligned, broken. The whole idea is to keep the current behavior of the dialog if it's not templated.

kdinev: @desig9stein Alright! I'm perfectly fine with this decision. I know that we're keeping the `[message]` input just for backwards compatibility anyways, so the second example that I gave would be the way to achieve the scenario that I've described. Thanks!

PR #9143, opened 2021-03-15
fix(radio): can't select radio on click

### LINKED ISSUE #9136: igx-radio emits multiple change event with ngModel
## Description  
Using igx-radio and ngModel does not work as expected. Inside the method triggered by event "change" we set up the value and trigger again the event. In version 10.1.4 wasn't doing that.

 * igniteui-angular version: 10.2.15
 * browser: Microsoft Edge Version 89.0.774.50

## Steps to reproduce  

1. Navigate to stackblitz example: https://stackblitz.com/edit/angular-igniteui-radio?file=src%2Fapp%2Fapp.module.ts
2. Select Option 1
3. check logs, select again option 2

## Result  
Changing the ngModel does not trigger the change event

## Expected result  
Changing the ngModel does not trigger the change event

## Attachments  
1. Navigate to stackblitz example:
https://stackblitz.com/edit/angular-igniteui-radio?file=src%2Fapp%2Fapp.module.ts

simeonoff:
Closes #9136  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/radio/radio.component.html
+++ b/projects/igniteui-angular/src/lib/radio/radio.component.html
@@ -9,10 +9,10 @@
     [attr.aria-checked]="checked"
     [attr.aria-labelledby]="ariaLabelledBy"
     [attr.aria-label]="ariaLabel"
-    (click)="_clicked($event)"
     (blur)="onBlur()" />
 
 <label #nativeLabel class="igx-radio__composite" igxRipple
+    (click)="_clicked($event)"
     igxRippleTarget=".igx-radio__ripple"
     [igxRippleDisabled]="disableRipple"
     [igxRippleCentered]="true"
--- a/projects/igniteui-angular/src/lib/radio/radio.component.ts
+++ b/projects/igniteui-angular/src/lib/radio/radio.component.ts
@@ -142,7 +142,7 @@ export class IgxRadioComponent implements ControlValueAccessor, EditorProvider {
      * @memberof IgxRadioComponent
      */
     @Input()
-    public value: any;
+    public value: string;
 
     /**
      * Sets/gets the `name` attribute of the radio component.
@@ -323,6 +323,7 @@ export class IgxRadioComponent implements ControlValueAccessor, EditorProvider {
     public onKeyUp(event: KeyboardEvent) {
         event.stopPropagation();
         this.focused = true;
+        this.select();
     }
 
     /**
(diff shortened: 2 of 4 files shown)


projects/igniteui-angular/src/lib/radio/radio.component.html
```diff
@@ -9,10 +9,10 @@
     [attr.aria-checked]="checked"
     [attr.aria-labelledby]="ariaLabelledBy"
     [attr.aria-label]="ariaLabel"
-    (click)="_clicked($event)"
```
kdinev, line 12: Shouldn't the input itself still be handling click together with the label?

simeonoff: The native input is hidden, hence why it doesn't make sense to listen for the click even on it.

kdinev (dismissed):

kdinev (dismissed):

wnvko (changes requested):
`igxRadio` now calls `select` method on each click. This forces `ngModelChange` to fire even when value was not changed which is not correct. Also changing the value of the bound property does not selects the radio.  Having this radios:
```
        <igx-radio [(ngModel)]="selected" value="option1">Option 1</igx-radio>
        <igx-radio [(ngModel)]="selected" value="option2">Option 2</igx-radio>
        <button (click)="selectSecond()">Select second</button>
```
and this handler
```
    public selectSecond() {
        this.selected = 'Option 2';
    }
```
does not selects second option.

simeonoff:
> `igxRadio` now calls `select` method on each click. This forces `ngModelChange` to fire even when value was not changed which is not correct.

The radio has always called `select` on each click. This is nothing new - no behavior change with this PR from that perspective. What's being changed now is that `select` is also called on `keyUp` to account for changing the radios using keyboard as if they were in a group.

>Also changing the value of the bound property does not selects the radio. Having this radios:
> 
> ```
>         <igx-radio [(ngModel)]="selected" value="option1">Option 1</igx-radio>
>         <igx-radio [(ngModel)]="selected" value="option2">Option 2</igx-radio>
>         <button (click)="selectSecond()">Select second</button>
> ```
> 
> and this handler
> 
> ```
>     public selectSecond() {
>         this.selected = 'Option 2';
>     }
> ```
> 
> does not selects second option.

It should be `this.selected = 'option2'` instead. The value is `option2` not `Option 2`.

wnvko (dismissed):

wnvko (approved):

PR #9285, opened 2021-04-07
remove content padding overlap on other elements  #9272

### LINKED ISSUE #9272: Tab items' content's top padding overlaps with the element above.
## Description  
When padding is applied to the igx-bottom-nav-content, the top padding overlaps with the element above.

 * igniteui-angular version: 12.0
 * browser: N/A

## Steps to reproduce  

1. Run [this sample](https://stackblitz.com/edit/github-zog9mm?file=src/app/tabbar-sample-1/tabbar-sample-1.component.html)
2. Inspect the tab's content "This is Item 1 content".
3. Check the padding rectangle

## Result  
The top padding overlaps with the div above the bottom nav component.
image

## Expected result  
Padding should not overlap with other page elements.

zhosti:
Closes #9272

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/bottom-nav/_bottom-nav-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/bottom-nav/_bottom-nav-theme.scss
@@ -86,8 +86,12 @@
     $item-max-width: rem(168px);
     $item-padding: 0 rem(12px, 16px);
 
-    %igx-bottom-nav-panel:focus {
-        outline-style: none;
+    %igx-bottom-nav-panel {
+        display: block;
+
+        &:focus {
+            outline-style: none;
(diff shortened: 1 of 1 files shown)


simeonoff (approved):

simeonoff:
This fixes the problem, but if you apply padding directly to the `igx-bottom-nav-content` element selector as shown in the example from the issue, it will cause the padding to always be there when another bottom nav item is selected. This is because we stack the items. It's another bug that needs fixing. Instead of just manipulating the z-index, we should ensure all hidden items also have `display: none` applied to them on TransitionEnd.

simeonoff:
Closing in favor of https://github.com/IgniteUI/igniteui-angular/pull/9288 as it also fixes the issue in my comment.

simeonoff (approved):

PR #9610, opened 2021-05-28
refactor(migration): ensure absolute paths ServerHost, ngLS fixes

### LINKED ISSUE #9572: Some changes are not applied properly in template files with Angular LS 12.0
## Description  
Angular have two different language services, a standard language service and one dedicated for ivy. After Angular 12, our migrations fail to locate the changes that need to be made (in template files) and so some changes are not applied at all. Changes are still properly applied to TypeScript files.

 * igniteui-angular version: 12.0.x
 * browser: n/a

## Steps to reproduce  

1. Run the migrations in the samples with the latest version of Angular's LS (ivy enabled)

## Result  
Everything should be migrated properly.

## Expected result  
Some changes are not applied in template files.

ViktorSlavov:
Closes #9572 

Files were not properly resolved when checking w/ relative routes (`/src/...`).
Patch all paths to be resolved as absolute (w/ `process.cwd()` as root), patch them back when checking ng schematics tree.
Resolve minor issue w/ replace position in template files.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -34,6 +34,17 @@ export class UpdateChanges {
     public get projectService(): tss.server.ProjectService {
         if (!this._projectService) {
             this._projectService = createProjectService(this.serverHost);
+            // Force Angular service to compile project on initial load w/ configure project
+            // otherwise if the first compilation occurs on an HTML file the project won't have proper refs
+            // and no actual angular metadata will be resolved for the rest of the migration
+            const mainPath = path.resolve(this._projectService.currentDirectory, 'src/app/app.component.ts');
```
jackofdiamond5, line 40: `'src/app/app.component.ts'` this path doesn't seem right to me as the file may not be present. Why don't we force the LS to resolve to `main.ts` instead? We could also get its relative location from the `angular.json` file since I don't think hard coding the whole path is a good idea.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -34,6 +34,17 @@ export class UpdateChanges {
...
+            // const wsProject = this.workspace.projects[this.workspace.defaultProject] || this.workspace.projects[0];
+            // path.posix.join('/', wsProject.root, wsProject.architect?.build?.options['main']);
```
jackofdiamond5, line 42: This seems like the better alternative to me.

projects/igniteui-angular/migrations/common/UpdateChanges.spec.ts
```diff
@@ -321,7 +321,7 @@ describe('UpdateChanges', () => {
             ]
         };
         const jsonPath = path.join(__dirname, 'changes', 'classes.json');
-        spyOn(fs, 'existsSync').and.callFake((filePath: string) => {
+        spyOn(fs, 'existsSync').and.callFake<any>((filePath: string) => {
```
jackofdiamond5, line 324: Why is this `any`?

projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -415,19 +426,28 @@ export class UpdateChanges {
     }
 
     protected updateClassMembers(entryPath: string, memberChanges: MemberChanges) {
-        const langServ = this.getDefaultLanguageService(entryPath);
         let content = this.host.read(entryPath).toString();
         const changes = new Set<{ change; position }>();
         for (const change of memberChanges.changes) {
```
jackofdiamond5, line 432:
```suggestion
```

projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -185,7 +185,7 @@ const patchHostOverwrite = (host: Tree, fileVersions: Map<string, number>) => {
  */
 export const createProjectService = (serverHost: tss.server.ServerHost): tss.server.ProjectService => {
     // set traceToConsole to true to enable logging
-    const logger = new Logger(false, tss.server.LogLevel.verbose);
+    const logger = new Logger(true, tss.server.LogLevel.verbose);
```
jackofdiamond5, line 188: Might want to set that back to `false` so we don't flood the console with too many messages.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
ViktorSlavov: I assume this was about calculating `absPath` multiple times (while it would never change) and assigning the `langServ` multiple times per file?

jackofdiamond5: Actually it was just about the empty row😄

jackofdiamond5:
@ViktorSlavov, as we discussed, since we're using the `ivy` language service, we will have to explicitly set `strictTemplates` to `true` in the `angularCompilerOptions` in the `tsconfig.json` file of the project that is executing our migrations. This is a requirement by the `ivy` LS and if it's not set then changes which target members in the template files will not be applied. 

We may also want to include a message which notifies the user that `strictTemplates` will be temporarily set to `true` and will be switched back/removed when the update finishes.

jackofdiamond5 (dismissed):

jackofdiamond5 (approved):

PR #9650, opened 2021-06-03
igxAccordion

### LINKED ISSUE #9559: Include accordion component to Ignite UI for Angular
Include accordion components as a part of Ignite UI for Angular toolset

Ability to add IgxAccordion Module similar to IgxGridModule

Related #307

mmart1n:
Closes #9559   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/accordion/accordion.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="igx-accordion__root" role="accordion" (keydown)="handleKeydown($event)">
```
ViktorSlavov, line 1: Not sure if this wrapper is needed. It could be
```typescript
/**
 * The role attribute of the accordion.
**/
@HostBinding('attr.role')
public role="accordion";
...
/** @hidden @internal **/
@HostBinding('class.igx-accordion__root')
public cssClass = "igx-accordion__root"
...
/** @hidden @internal **/
@HostListener('keydown', [$event])
public handleKeydown(event: KeyboardEvent) {
   ...
}
```

ViktorSlavov: This can be discussed w/ the design team, once (if) styles for the accordion are being implemented

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,276 @@
...
+    selector: 'igx-accordion',
+    templateUrl: 'accordion.component.html',
+    providers: [IgxAccordionService]
+})
+export class IgxAccordionComponent implements AfterContentInit, AfterViewInit, OnDestroy {
+    /**
+     * Sets/gets the `id` of the accordion component.
+     * If not set, `id` will have value `"igx-accordion-0"`;
+     * ```html
+     * <igx-accordion id = "my-first-accordion"></igx-expansion-panel>
```
ViktorSlavov, line 33: Please remove the extra space, when you get to it

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+     * Gets/Sets accordion expansion mode
+     *
+     * @remarks
+     * By default the accordion expansion mode is 'Single'
+     * @param expansionMode: IgxAccordionExpansionMode
+     */
+    @Input()
+    public get expansionMode() {
```
ViktorSlavov, line 77: Should be strictly typed, as opposed to inferred
```suggestion
    public get expansionMode(): IgxAccordionExpansionMode {
```

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,276 @@
...
+            this.accordionService.expandedPanels.forEach(panel => {
+                if (panel !== lastExpanded) {
+                    panel.collapse();
+                }
+            });
+        }
+        this._expansionMode = expansionMode;
+    }
+
+    /** Emitted when a panel is expanding, before it finishes
```
ViktorSlavov, line 92: Move this to the next line

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,276 @@
...
+     *
+     * ```html
+     * <igx-accordion (panelExpanding)="handlePanelExpanding($event)">
+     * </igx-accordion>
+     * ```
+     *
+     *```typescript
+     * public handlePanelExpanding(event) {
```
ViktorSlavov, line 100: Type the event in the example handler
```suggestion
     * public handlePanelExpanding(event: IExpansionPanelCancelableEventArgs) {
```

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    /** Emitted when a panel is expanded, after it finishes
+     *
+     * ```html
+     * <igx-accordion (panelExpanded)="handlePanelExpanded($event)">
+     * </igx-accordion>
+     * ```
+     *
+     *```typescript
+     * public handlePanelExpanded(event) {
```
ViktorSlavov, line 119: Type the event args of the example handler
```suggestion
     * public handlePanelExpanded(event: IExpansionPanelCancelableEventArgs) {
```

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -212,7 +215,7 @@ export class IgxAccordionComponent implements AfterContentInit, AfterViewInit, O
         }
 
         this.panels.forEach(panel => {
-            if (panel.collapsed) {
+            if (panel.collapsed && !panel.header.disabled) {
```
ViktorSlavov, line 218: The check for `collapsed` is a bit redundant, as the panels' expand/collapse methods also check that

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -231,14 +234,68 @@ export class IgxAccordionComponent implements AfterContentInit, AfterViewInit, O
...
         });
     }
 
-    public handleKeydown(ev) {
+    public handleKeydown(event: KeyboardEvent, panel: IgxExpansionPanelComponent) {
+        const key = event.key.toLowerCase();
+        if (!(ACCORDION_NAVIGATION_KEYS.has(key))) {
+            return;
+        }
+        this.enabledPanels = this._panels.filter(p => !p.header.disabled);
```
ViktorSlavov, line 248: You could leave a comment here - if we ever want to improve the performance of the accordion, `enabledPanels` could also be cached (by making a `disabledChange` emitter on the panel header).
However, filtering the panels is trivial in most cases - leave it as is for the initial implementation.

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -231,14 +234,68 @@ export class IgxAccordionComponent implements AfterContentInit, AfterViewInit, O
...
+            setTimeout(() => this.handleNavigation(event, panel), 1);
+        } else {
+            this.handleNavigation(event, panel);
+        }
+    }
+
+    private handleNavigation(event: KeyboardEvent, panel: IgxExpansionPanelComponent) {
+        switch (event.key.toLowerCase()) {
+            case 'home':
+                this.enabledPanels[0].header.elementRef.nativeElement.children[0].focus();
```
ViktorSlavov, line 260: Expose `panelheader.elementRef.nativeElement.children[0]` in a hidden internal getter on IgxExpansionPanelHeaderComponent

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -265,6 +322,11 @@ export class IgxAccordionComponent implements AfterContentInit, AfterViewInit, O
             panel.contentCollapsing.pipe(takeUntil(this.unsubChildren$)).subscribe((args: IExpansionPanelCancelableEventArgs) => {
                 this.panelCollapsing.emit(args);
             });
+            fromEvent(panel.header.elementRef.nativeElement, 'keydown')
```
ViktorSlavov, line 325: If you expose a getter for the focused element in the panel header (see previous comment), I think it could be used here as well (would look a bit better)

ViktorSlavov (changes requested):

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,330 @@
...
+     */
+    @HostBinding('attr.id')
+    @Input()
+    public id = `igx-accordion-${NEXT_ID++}`;
+
+    /**
+     * The role attribute of the accordion.
+     **/
+    @HostBinding('attr.role')
+    public role = 'accordion';
```
damyanpetev, line 47: Just noticed this in a previous comment - fairly certain there's no such role defined in the WAI-ARIA spec, so this should be removed. Looking at the [practices](https://www.w3.org/TR/wai-aria-practices/#accordion) info the requirements are for headers to be buttons linked to the content and as far as I can tell the expansion panel already covers that? The only specific thing I saw was a suggestion to wrap the panels in a list (outside of suggested practices) to aid in navigation/orientation, though I think using headings on our end should serve a similar purpose.

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,324 @@
...
+        const panelIndex = this.enabledPanels.indexOf(panel);
+        return this.enabledPanels[panelIndex + dir] || panel;
+    }
+
+    private subToChanges() {
+        this.unsubChildren$.next();
+        this._panels.forEach(panel => {
+            panel.contentExpanded.pipe(takeUntil(this.unsubChildren$)).subscribe((args: IExpansionPanelEventArgs) => {
+                this.expandedPanels.add(args.owner);
+                this.panelExpanded.emit(args);
```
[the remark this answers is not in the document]
MonikaKirkova: image
We have implemented the event this way because of the specification
https://github.com/IgniteUI/igniteui-angular/wiki/Accordion-Specification#events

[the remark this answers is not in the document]
jackofdiamond5: Yes, I've seen the spec but I still think this should be changed. Just look at any of our controls which uses the overlay service and you'll see what I mean. Take the `igxToggle` for example - it essentially reemits the same events from the overlay service but it changes the `owner`. That is because the developer who will handle the event, will attach their handler to the toggle directive and from their point of view the toggle directive was the one who emitted the event, not the overlay service. 

And it makes sense to expect that the `owner` of the event is the control that you've attached your handler to since it was the one who actually emitted it. So in this case I think that we may want to revise this section of the specification, considering it's been signed off. 

@kdinev, can you share your thoughts on this?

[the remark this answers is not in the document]
kdinev: I'll let @damyanpetev share his thoughts and if you both agree the spec needs to change, then go ahead!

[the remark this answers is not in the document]
damyanpetev: True the toggle/drop down re-emit, however they do that _internally_ and and out users are not even necessarily aware of that, thus the owner swap :) There's no precedence I can think of for such re-emit.

I too assume `owner` is the reference to the component that emits the event and that's true for other components in the product. I understand there's very little API to interact with accordion itself and it's easier to re-emit as is, but that `owner` may be confusing for those already used to it being one way.

The use of those events is kind of limited, considering they'll barely (if any) benefit if the panels are repeated; as in `<igx-accordion (panelExpanding)="handler()>` is identical to `<igx-expansion-panel *ngFor=".." (expanding)="handler()">`, right? So this is only a convenience for the cases where each panel is manually declared, yet some common handling is needed, which should likely be reflected in the samples. Just my 2c

PS: I was almost right about the no precedence... however I went looking and figured the Hierarchical Grid re-emits a lot of events, and sure enough - inherited grid events do emit with the grid as owner intentionally, so we've already strayed from the rule. I'm still going to count that as an exception for the HGrid, and still lean towards being consistent elsewhere :)

projects/igniteui-angular/src/lib/core/styles/components/expansion-panel/_expansion-panel-theme.scss
```diff
@@ -114,6 +114,10 @@
         overflow: hidden;
     }
 
+    %igx-expansion-panel--expanded {
```
damyanpetev, line 117: Does this apply to.. every expansion panel even outside of an accordion? Ignore if this is just temporary

mmart1n: Тhe igx-accordion-component set this class only to its expanded children(igxExpansionPanels). However if our customers decide to set manually this class to an igxExpansionPanel outside of an igxAccordion - it will be applied as well.

projects/igniteui-angular/src/lib/core/styles/components/expansion-panel/_expansion-panel-theme.scss
```diff
@@ -114,10 +114,6 @@
         overflow: hidden;
     }
 
-    %igx-expansion-panel--expanded {
```
ViktorSlavov: @simeonoff 
I believe that we discussed this on one of the initial meetings - we will **not** be providing the margin / class out of the box, as it would be difficult for users to remove it, if they so wished.
@StefanIvanov 
There's a line in the [spec](https://github.com/IgniteUI/igniteui-angular/wiki/Accordion-Specification#3-functionality) that specifically mentions the margin of `8px` for expanded panels - can we remove it? It's easy enough to achieve in a sample by adding an `<igx-expansion-panel #panel [ngClass]="{'margin-class':  isPanelExpanded(panel) }">` and we'll make sure to showcase it in the docs.
We could add an `exportAs` to the `igx-expansion-panel`, so it's easier to get the panel state, but that's a separate enhancement.

simeonoff: I remember we decided that the users will be setting this margin themselves. So, yes, removing this is the right thing to do.

StefanIvanov: We want the default expanded state to be 8px apart from nearby panels, so I was thinking we would have an input for that that can be set to 0px if customers wanted to have no gaps like that.

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,339 @@
...
+import { takeUntil } from 'rxjs/operators';
+import { ACCORDION_NAVIGATION_KEYS } from '../core/utils';
+import { IExpansionPanelCancelableEventArgs,
+    IExpansionPanelEventArgs, IgxExpansionPanelBase } from '../expansion-panel/expansion-panel.common';
+import { IgxExpansionPanelComponent } from '../expansion-panel/expansion-panel.component';
+import { ToggleAnimationSettings } from '../expansion-panel/toggle-animation-component';
+
+interface IAccordionEventArgs extends IExpansionPanelEventArgs {
+    owner: IgxAccordionComponent;
+    panel: IgxExpansionPanelBase;
```
jackofdiamond5, line 13: For both of the interfaces we may want to add a comment which informs the user that the `panel` is actually the expansion panel which was expanded/collapsed.  Something like this maybe:
```suggestion
    /** Provides a reference to the `IgxExpansionPanelComponent` which was expanded/collapsed. */
    panel: IgxExpansionPanelBase;
```

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -0,0 +1,339 @@
...
+}
+
+interface IAccordionCancelableEventArgs extends IExpansionPanelCancelableEventArgs {
+    owner: IgxAccordionComponent;
+    panel: IgxExpansionPanelBase;
```
jackofdiamond5, line 18: Same as the above but the comment here should say that it's the panel which is currently expanding or collapsing.

CHANGELOG.md
```diff
@@ -5,6 +5,42 @@ All notable changes for each version of this project will be documented in this
 ## 12.1.0
 
 ### New Features
+- Added `IgxAccordion` component
+    - A collection of vertically collapsible igx-expansion panels that provide users with data and the ability to drill into it in a compact manner. The control is **not** data bound and takes a declarative approach, giving users more control over what is being rendered.
```
jackofdiamond5, line 9: I'm not sure what **drill into it in a compact manner** has to mean. Can we think of another way to say it?

CHANGELOG.md
```diff
@@ -5,6 +5,42 @@ All notable changes for each version of this project will be documented in this
...
+    - Exposed two expansion modes, templating, easy-to-use keyboard navigation, option for nested accordions and more.
```
jackofdiamond5, line 10: Even though I quite like the `and more` ending, this document is not related to marketing. So why don't we include everything else that a potential user will want to know about the accordion? One thing that comes to mind is the fact that the user can say whether they want single branch expanding or not. So instead of saying **expansion modes** we can simply state them.

Also, I don't think we need to say **templating** because the accordion is a wrapper around the expansion panels.

projects/igniteui-angular/src/lib/accordion/README.md
```diff
@@ -0,0 +1,88 @@
+
+# IgxAccordion
+
+
+**IgxAccordion** is a container-based component that contains a collection of collapsible **IgxExpansionPanels**.
+
+A walkthrough of how to get started can be found [here](https://www.infragistics.com/products/ignite-ui-angular/angular/components/accordion)
+
```
jackofdiamond5, line 8: Might be a good idea to include a `Dependencies` section somewhere in the beginning since one of the first things a user will want to see is how to actually set up the accordion.

CHANGELOG.md
```diff
@@ -5,6 +5,42 @@ All notable changes for each version of this project will be documented in this
...
+    - Code example below:
+
+    ```html
+    <igx-accordion>
```
jackofdiamond5, line 14: This sample right here is way too big for the changelog. Let's make it as simple as possible. We can always add a little note stating that multiple expansion panels are supported.

projects/igniteui-angular/src/lib/accordion/README.md
```diff
@@ -0,0 +1,88 @@
...
+# Usage
+
+```html
+<igx-accordion>
```
jackofdiamond5, line 12: Same as with the sample in the changelog.

projects/igniteui-angular/src/lib/accordion/README.md
```diff
@@ -0,0 +1,88 @@
...
+The following tables summarize the **igx-accordion** inputs, outputs and methods.
+
+### Accessors
+The following accessors are available in the **igx-accordion** component:
+| Name | Type | Description |
+| :--- | :--- | :--- |
+| `panels` | `IgxExpansionPanelComponent[]` | All IgxExpansionPanel children of the accordion |
+### Inputs
+The following inputs are available in the **igx-accordion** component:
+
```
jackofdiamond5, line 50:
```suggestion
```
Remove that line to visually distinguish that the table below is related to the `Inputs` title.

projects/igniteui-angular/src/lib/accordion/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+# API Summary
+The following tables summarize the **igx-accordion** inputs, outputs and methods.
+
+### Accessors
+The following accessors are available in the **igx-accordion** component:
+| Name | Type | Description |
+| :--- | :--- | :--- |
+| `panels` | `IgxExpansionPanelComponent[]` | All IgxExpansionPanel children of the accordion |
```
jackofdiamond5, line 47:
```suggestion
| `panels` | `IgxExpansionPanelComponent[]` | All IgxExpansionPanel children of the accordion |

```

Add a new line after this so that the `Inputs` title is not close to the table above it as it's not related to it.

projects/igniteui-angular/src/lib/accordion/README.md
```diff
@@ -0,0 +1,88 @@
...
+            <div>
+                Cascading Style Sheets (CSS) is a style sheet language used for describing the presentation of a document written in a markup language like HTML      
+            </div>
+        </igx-expansion-panel-body>
+    </igx-expansion-panel>
+</igx-accordion>
+```
+
+# API Summary
+The following tables summarize the **igx-accordion** inputs, outputs and methods.
```
jackofdiamond5, line 41: We may want to arrange the tables below the way that they are described here. For example - this line states that we have tables for `inputs`, `outputs` and `methods` but the first table after this line is about `Accessors`.

projects/igniteui-angular/src/lib/accordion/README.md
ViktorSlavov: I feel a `Dependencies` section would be redundant, as it would either need to state that you have to import the appropriate module (`IgxAccordionModule`) in your project (which is pretty self-explanatory, is featured in the docs and is fundamental for any component) or that you have to import `BrowserAnimationsModule` in your app root (which is stated in the `getting started` section and is already part of our `ng add` migration).
I'll agree that there are some inconsistency in these sections and a `Dependency` section is not features in most of the other components' READMEs, but I think it's not necessary to add one.

README.md
```diff
@@ -21,6 +21,7 @@ You can find source files under the [`src`](https://github.com/IgniteUI/igniteui
 
 |Components|Status|||Directives|Status|||
 |:--|:--:|:--|:--|:--|:--:|:--|:--|
+|accordion|:white_check_mark:|[Readme](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/accordion/README.md)|[Docs](https://www.infragistics.com/products/ignite-ui-angular/angular/components/accordion)|||||
```
ViktorSlavov, line 24:
```suggestion
|accordion|:white_check_mark:|[Readme](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/accordion/README.md)|[Docs](https://www.infragistics.com/products/ignite-ui-angular/angular/components/accordion)|
```
Remove the extra columns at the end, as it [adds an empty line](https://github.com/IgniteUI/igniteui-angular/tree/mevtimov/feat-9559-master) under `Directives` section

src/app/accordion/accordion.sample.ts
```diff
@@ -0,0 +1,30 @@
...
+    public singleBranchExpand = false;
+
+    public panelExpanding(event) {
+        console.log(event);
+    }
+    public panelExpanded(event) {
+        console.log(event);
+    }
+
+    public contentExpanding(event) {
```
ViktorSlavov, line 26: Bit of a nit-pick, but this method doesn't seem to be used, you can remove it.

src/app/accordion/accordion.sample.ts
```diff
@@ -0,0 +1,30 @@
+import { Component, ViewChild } from '@angular/core';
+import { slideInLeft, slideOutRight } from 'igniteui-angular';
+import { IgxAccordionComponent } from 'projects/igniteui-angular/src/lib/accordion/accordion.component';
```
ViktorSlavov, line 3: This should be changed as the accordion component is the public API - the import for `IgxAccordionComponent` should be from `'igniteui-angular'`

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.component.ts
```diff
@@ -178,6 +178,13 @@ export class IgxExpansionPanelComponent extends ToggleAnimationPlayer implements
         return this.header ? `${this.id}-header` : '';
     }
 
+    /**
+     * @hidden
```
ViktorSlavov, line 182:
```suggestion
     * @hidden @internal
```

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.component.ts
```diff
@@ -192,7 +199,7 @@ export class IgxExpansionPanelComponent extends ToggleAnimationPlayer implements
 
     private _collapsed = true;
 
-    constructor(private cdr: ChangeDetectorRef, protected builder: AnimationBuilder) {
+    constructor(private cdr: ChangeDetectorRef, private elementRef: ElementRef, protected builder: AnimationBuilder) {
```
ViktorSlavov, line 202: Adding a new parameters to a public class's constructor is a breaking change.
Though I highly doubt that would affect anyone, `elementRef` could still be moved as the last parameter and be made options (`?`) so as to avoid it.

ViktorSlavov (changes requested):

ViktorSlavov (approved):

jackofdiamond5 (approved):

PR #9690, opened 2021-06-09
Update IgxMask's placeholder when mask property is updated

### LINKED ISSUE #7490: igxMask cannot be changed dynamically
## Description  
igxMask cannot be changed dynamically

 * igniteui-angular version: 9.0.1
 * browser: 

## Steps to reproduce  

1. Create input field with [igxMask]="mask"
2. Initialize mask variable on ngOnInit() with value (for example '000-000') 
3. Assign a new value (for example '(000)0000') to the mask variable for any event (for example, click on a button)

## Result  
When you press the button the mask on input field does not change

## Expected result  
When you press the button the mask on input field should change

jackofdiamond5:
Closes #7490

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -273,9 +273,24 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
...
+    public ngOnChanges(changes: SimpleChanges) {
+        // B.P. 9th June 2021 #7490
+        if (changes['mask'] && !changes['mask'].firstChange) {
+            this.setPlaceholder();
+        }
+    }
+
     /** @hidden */
     public ngOnInit(): void {
         if (!this.nativeElement.placeholder) {
```
wnvko, line 286: you do not need this check here anymore

wnvko (approved):

projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -274,12 +277,19 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
     }
 
     /** @hidden */
-    public ngOnInit(): void {
-        if (!this.nativeElement.placeholder) {
-            this.renderer.setAttribute(this.nativeElement, 'placeholder', this.maskOptions.format);
+    public ngOnChanges(changes: SimpleChanges): void {
```
wnvko, line 280: looking further at this we should not use `ngOnChanges`. IMO you should change `mask` to property with getter and setter. Then you can move all the logic from `ngOnChanges` to the mask's setter.

_Note: keeping this in `ngOnChanges` will not work if one calls `inputWithMask.mask = 'some_new_mask'` as this will not fire `ngOnChanges`._

wnvko (changes requested):

projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -135,13 +146,17 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
     private _end = 0;
     private _start = 0;
     private _key: string;
+    private _mask: string;
```
wnvko, line 149: having `_mask` private field we can remove both `_oldMask` and `defaultMask` now. First initialize `_mask` to be equal to `CCCCCCCCCC` and default may gone. Then in `setPlaceholder` check against `_mask`. To make this work you should also call in `mask` setter first `setPlaceholder` and then set `_mask` to provided value.

projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -23,7 +23,18 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
...
      */
     @Input('igxMask')
-    public mask: string;
+    public get mask(): string {
+        return this._mask;
+    }
+
+    public set mask(val: string) {
+        // B.P. 9th June 2021 #7490
+        this._oldMask = this._mask || this.defaultMask;
```
wnvko, line 32: having `_mask` we may remove `_oldMask` and `defaultValue` so this line becomes needless.

projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -23,7 +23,18 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
...
+        if (val) {
```
wnvko, line 33: check if provided value is different from `_mask` and set the mask and placeholder only in this case.

wnvko (changes requested):

wnvko (dismissed):

damyanpetev:
This only updates the placeholder though, it should update the displayed masked value as well.

projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -23,7 +23,20 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
...
+    }
+
+    public set mask(val: string) {
+        // B.P. 9th June 2021 #7490
+        const oldOptions = this.maskOptions;
+        if (val !== this._mask) {
+            this.setPlaceholder(val);
+            this._mask = val;
+        }
+        const cleanInputValue = this.maskParser.parseValueFromMask(this.inputValue, oldOptions);
```
wnvko, line 37: if you call this before change `_mask` you will not need to cache `this.maskOptions`. Also, why do not call this inside the if statement. Should we update the input value if mask did not change at all?

projects/igniteui-angular/src/lib/directives/mask/mask.directive.ts
```diff
@@ -23,7 +23,20 @@ export class IgxMaskDirective implements OnInit, AfterViewChecked, ControlValueA
      * ```
      */
     @Input('igxMask')
-    public mask: string;
+    public get mask(): string {
+        return this._mask || this.defaultMask;
```
wnvko, line 27: do not like that getter is actually sets the mask if there is no initial value 😢

wnvko (approved):

PR #9877, opened 2021-07-19
Add row refactor and integration with transaction service.

### LINKED ISSUE #9792: Implement row adding in/with the edit flow
## Description  
There seems to be multiple issues stemming from the fact that row adding doesn't implement the edit flow, especially around the pending transactions / transactions state handling. When it calls functions such as to create event args, the common function for that expects that it is called with the state from row editing and produces incorrect results.

Also there is a lot of code that can be removed if the two flows are integrated into one.

 * igniteui-angular version: 12.1
 * browser: n/a

## Steps to reproduce  

1. Open the attached sample
2. Add a row and click done - the event is cancelled but the args are logged
3. Observe the logged (in console) args

## Result  
Neither of the three main args - newValue, oldValue and rowData are correct . newValue in particular is not populated at all because it should come from the pendingTransactions open/close process.

## Expected result  
Args should be correct

## Attachments  
[igx-grid-cancel-rowedit-on-new-row-issue.zip](https://github.com/IgniteUI/igniteui-angular/files/6709810/igx-grid-cancel-rowedit-on-new-row-issue.zip)

skrustev:
Closes #9792 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


skrustev:
@damyanpetev @wnvko Could you look into the changes since in order to preserve the exact same current user experience I had to make some small additions to the transactions and if you are ok with them or if we need to consider other approach.

wnvko:
Looking at this what you need is `insert` method in transactions service, or `add(... atIndex..)`. Even if we implement this in transactions service I do not see a way to save this in the back end, e.g. how you will tell SQL to insert a row in a given table at given index?
Added row will always be the last row in the table. If you need to show added row, or any other row, at some specific position you can sort the data by some visible or not visible column.

skrustev:
@wnvko Yes in general there is no 'insert at index' in SQL, but one can always manage their database manually and have column that saves the positions of a row and sort by it when making a request.

Anyways this is not the reason why I've added the 'pendingIndex', but because in the way we show pending transactions in the grid. This is main reason for it since we want to show the newly added row initially at an index and once the user ends editing it, the transaction is added at the end anyway. Maybe it can be renamed to `showIndex` or something more related.

skrustev:
We've discussed and decided that it is maybe better to get the dummy for from the `dataWithAddedInTransactionRows` method, until finished adding it instead of using the pending transactions, since they aren't meant to render anything. The latest changes should reflect that.

hanastasov:
Currently the way to add row through grid.neginAddRowByIndex does not work for `IgxTreeGrid`, so let's just verufy that explicitly when this PR is done.

wnvko (approved):
I am ok with the changes in `transactions`.
Looking at the other changes it seams like adding a row does not use existing API, but adds  new methods and classes. Can we skip this?

skrustev:
@wnvko Can you be more specific on what do you mean existing API and which methods and classes? The add row requires additional things from editing only and is basically an extension of it. Is is intentionally as much separated as possible to not pollute the edit row logic too much.

wnvko:
IMO adding a row should go through these steps:

1. `AddRowViaUI` method should create an empty row, generate its PK, and save it in some grid's field. If grid's data source is empty add this row to the data source. If grid's data source is not empty AddRowPipe should insert the empty row in the data source at correct position. Update grid and add row should be there.
2. Enter edit mode for the added row. This should start adding each edition in pending transaction. Also should fire all the events: `rowEditEnter`, `cellEditEnter`, `cellEdit`, `cellEditDone` and `cellEditExit`. The last three `rowEdit`, `rowEditDone` and `rowEditExit` should be handled separately.
3. On click of `done` button call `addRowFromUIHandler`. This handler should end pending transaction and fire `rowEdit`, `rowEditDone` and `rowEditExit`. Then it should add the row to data source, or generate ADD transaction. On click of `cancel` button remove the added row. and fire `rowEditExit`. Note: to do this we should use custom row edit template with custom row edit actions.

About the cancelable events:
- `rowEditEnter` - if this is canceled add row should be removed.
- `cellEditEnter` - if this is canceled nothing should be done. Still row is in edit mode.
- `cellEdit` - if this is canceled nothing should be done. Still row is in edit mode.
- `rowEdit` - if this is canceled add row should be removed. Note, having custom row edit template this should be easily handled there.

`rowEditEnter` is the only place where some custom logic, related to add row should be added.

hanastasov:
Is it necessary to add/insert empty row in the grid, and then start edit mode for this added row, so that the user can enter values? It results in flickering.

Cant we just wait for some action, like `rowEditDone`, `rowEditExit`, before adding the row at once?

skrustev:
@wnvko This is almost exactly what it is doing currently. 
The `rowEdit`, `rowEditDone` and `rowEditExit` events don't need to handled separately because they are already part of the edit process. What you are describing actually further separates the Add Row logic from the Edit Row one, opposite on what you said earlier and would lead to the old way of duplicating logic.
We also already have a custom row edit template which omits the edit message, but the actions since looks the same don't need to changed in my opinion.

@hanastasov We cannot insert the row after the events you mentioned, because the row needs to exist somewhere so editing knows about it and perform all edit actions per cell and track any changes.

hanastasov:
> @wnvko This is almost exactly what it is doing currently.
> The `rowEdit`, `rowEditDone` and `rowEditExit` events don't need to handled separately because they are already part of the edit process. What you are describing actually further separates the Add Row logic from the Edit Row one, opposite on what you said earlier and would lead to the old way of duplicating logic.
> We also already have a custom row edit template which omits the edit message, but the actions since look the same, don't need to changed in my opinion.
> 
> @hanastasov We cannot insert the row after the events you mentioned, because the row needs to exist somewhere so editing knows about it and perform all edit actions per cell and track any changes. The inserted row isn't added to the data or as add transaction until after the user click the `Done` button, so the row exists only in the grid.
> Not sure where you are seeing that flicker you mention, but if you post a gif will be better to understand it. Are you sure you are not just mistaking the row appearing animation we have for the Add Row UI?

Sorry for late reply. Yes, indeed I mean the row appearing. It looks like flickering, because first we add row, and the very next moment this row enters edit mode. Your answer addresses just that, so I understand.

ChronosSF:
For igx-treegrid, the rowEditDone/Exit event arguments' newValue is undefined when both adding a child and adding a top-level row.

hanastasov:
Last thing I noted - it seems that `rowAdding` event is missing, thus we are not able to cancell adding the row based on a business rule. Could we add that too?

projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.ts
```diff
@@ -24,7 +25,7 @@ export const RadioGroupAlignment = mkenum({
     horizontal: 'horizontal',
     vertical: 'vertical'
 });
-export type RadioGroupAlignment = (typeof RadioGroupAlignment)[keyof typeof RadioGroupAlignment];
+export type RadioGroupAlignment = typeof RadioGroupAlignment[keyof typeof RadioGroupAlignment];
```
ChronosSF, line 28: @skrustev , are you sure these radio group changes aren't coming from some issue with the target branch change?

ChronosSF (changes requested):

ChronosSF:
@damyanpetev , could you please check this PR?

ChronosSF (dismissed):

ChronosSF (approved):

kdinev (approved):

wnvko (approved):

PR #9995, opened 2021-08-12
Enable row editing overlay customization for child-level records

### LINKED ISSUE #9925: IgxHierarchicalGrid: Customization of the row-editing overlay works only for the root level records
## Description  
If you try to customize the row editing overlay of a row island in IgxHierarchicalGrid you will find out that it will be customized only for the root records.

 * igniteui-angular version: 12.0.7
 * browser: all

## Steps to reproduce  

1. Open the demo in StackBlitz - [here](https://stackblitz.com/edit/github-csqf59?file=src%2Fapp%2Fhierarchical-grid%2Fhierarchical-grid-add-row%2Fhierarchical-grid-add-row.component.html)
2. Expand the first row of the grid.
3. Start editing a row on the second level of the grid.

## Result
The row editing overlay is not customized.

## Expected result
The row editing overlay should be customized.

3phase:
Closes #9925  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
```diff
@@ -328,6 +328,27 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
...
+     */
+    public get resolveRowEditActions() {
+        return this.rootGrid ? this.rootGrid.rowEditActions : this.rowEditActions;
+    }
+
+    /**
+     * @hidden
+     */
+    public get resolveRowEditText() {
+        return this.rootGrid ? this.rootGrid.rowEditText : this.rowEditText;
```
MayaKirova, line 349: Shouldn't we allow setting different templates per rowIsland level? With the current implementation child grids will always use the template from the root grid so they will not take into account any template set on the row island level.

MayaKirova (changes requested):

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
3phase: I'm not sure since I'm not able to find that part in the spec. 
If that's the case however, it has to be decided whether a custom template in the parent grid and a lack of one in the child grids should mean that the child outlets should resolve to the parent's outlet (or to the default one). 
In case templating of the child grids' overlay should not be possible then this approach is cool, unless we decide to disable templates in the child grids at all, which makes no sense to me.

MayaKirova: IMO it should follow the same logic as the other templates:
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts#L397

It should use the template defined in its own row island definition, otherwise use the default one.
Using the one from the parent doesn't make sense to me.

3phase: The initial problem was that a row editing overlay was declared in a parent grid and was not shown in their successors. In my opinion, this is a legit use case scenario. If you don't think so, we can discuss it in the issue.
If you think that it's suitable, I can edit my fix to make a check if a template has been declared in the row island too, in order to allow customization of the overlay per children.

ChronosSF: I don't think that was the original issue. AFAIU the original issue states that all child layouts show the parent action strip instead of their own customized ones.

3phase: Now I see, I'm sorry, I got the original issue wrong.

MayaKirova (approved):

ChronosSF (approved):

ChronosSF (approved):

Issue #9429, opened 2021-04-26
How to start add row?


Eralmidia:
## Question  
I'm trying to use the new add row functionality of the igx grid, but I can't the method to simply start the adding.

I see that the igx row component has a begin add row, but that would require the grid to actually have a row? Then there is the CRUD service, which also has a begin add row method, but this, for some reason seems to require a cell. How do I simply start the process of adding a new row? Note: we want to trigger this programtically, not using built in buttons in the grid, like the igx-grid-editing-actions.

 * igniteui-angular version: 10.2
 * browser:

ChronosSF:
@Eralmidia , the add row UI is always dependent on another row so that we don't have a second implementation for TreeGrid where it allows adding children for the said row as well. What the empty template button does and what will work for your case is an undocumented method inherited by all grids that should be called like this:

```typescript
this.grid.beginAddRowByIndex(null, -1, false);
```

What it actually does internally is to create a default record with a generated PK and start edit mode for it using:

```typescript
this.crudService.enterEditMode(cell, event);
```

Where cell is just the first editable cell.

You should either use the undocumented method or create a wrapper around `enterEditMode` that does something similar. However, please, don't close the issue as we should make this easier. I'll transform it to a feature request tomorrow so that we can create better APU for this.

Eralmidia:
@ChronosSF Yes, before the add row functionality came along, we have been doing the same. Adding a blank row and setting that in edit mode. It's just that that brings along a bit of cleanup requirements. We use a state management library for the application state, which streams the records to the row through an observable. So typically that means adding a new entity to the state, which then adds a new row in the grid. And if the user cancels the adding, we need to remove it in the store etc. Add row is nice, since we can simply let the grid handle the input data of the new row until its ready to be committed to the store.

Having this reviewed, and hopefully a section about it the the docs, would be very good. For now, I'll check out the undocumented beginAddRowByIndex method. Thanks.

PR #10070, opened 2021-08-27
feat(add-row): implementing context-free beginAddRow method #9675 - master

### LINKED ISSUE #9675: Add row at top of grid
## Question  
I'm replacing an old jquery ignite grid with a new igx one. In the old one, we added a new row at the top of the grid like this: 
image

In the new grid, I'm starting the adding of a new row with the following code:
```
var row = this.grid.getRowByIndex(0);
row.beginAddRow();
```
This however seems to add a new row below the first row, which look odd:
image

How do I add a new row at the top of the grid?

 * igniteui-angular version: 10.2

ChronosSF:
Closes #9675 #9429  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


skrustev:
Things I've found so far:

1. You can specify row index even if there is primary key set for the grid, which results in wrong row [here](https://github.com/IgniteUI/igniteui-angular/pull/10070/files#diff-440a9d274756495c7bb4e7acf64e5ff3757d060999c43e932fe7654b565f793eR5985) and throws warning. I think having a separate methods for specifying `index` or `rowID` will resolve this like we talked earlier about it.

2. You cannot use the API if there are no rows added for the similar case as 1, but here you have no rowID to base it off of and I think default behavior should be triggered in this case.

3. If the default behavior is changed like it is at the moment in the PR, [here](https://github.com/IgniteUI/igniteui-angular/pull/10070/files#diff-546e39078a4b678b1598e0e135010a8081b02f4f63bc9997650a08633caa5879L440) the second part of the shorthand if is obsolete and can be just `null`.

4. The API doesn't seem to work for IgxTreeGrid, since its calling `get_rec_index_by_id` for the base gridAPI service because it is in the base class and not the IgxTreeGrid one and also the dataView in the IgxTreeGrid is different and would need separate implementation. 

5.  A small note currently for the IgxTreeGrid is that you cannot show the Add Row at the top of an inner level inside the tree. I think its far from a limitation, but maybe in the context of the Tree Grid it might make sense.

ChronosSF:
@skrustev , think I addressed your comments. Please, check the PR again.

MayaKirova (approved):
Looks good but decide what to do with the xit-ted test. Either remove it or maybe log a separate task to fix it.

skrustev:
Other things from further testing:

1. Add row does not start when you pin a row and try to add a row at index 2 for example

ChronosSF:
These are low impact as they involve row pinning. I'll log them separately.

skrustev (approved):

PR #10092, opened 2021-09-08
fix(IgxLinearProgressBar): fix max value of progress bar

### LINKED ISSUE #10067: IgxLinearProgressBar: progress doesn't change when max property changes dynamically
## Description  
If you change `IgxLinearProgressBar.value` property then progress changes.
If you change `IgxLinearProgressBar.max` property then progress doesn't change.

 * igniteui-angular version: 12.1.0
 * browser: Chrome

## Steps to reproduce  
Change `IgxLinearProgressBar.max` property dynamically.

## Result  
Progress doesn't update.

## Expected result  
Progress should update.

## Attachments  
https://stackblitz.com/edit/github-gvbevv

katherinedragieva:
Closes #10067

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
```diff
@@ -667,7 +668,7 @@ export class IgxCircularProgressBarComponent extends BaseProgressDirective imple
 
 export const valueInRange = (value: number, max: number, min = 0): number => Math.max(Math.min(value, max), min);
 
-export const toPercent = (value: number, max: number) => Math.floor(100 * value / max);
+export const toPercent = (value: number, max: number) => Math.floor(100 * (value / max));
```
kdinev, line 671:
```suggestion
export const toPercent = (value: number, max: number) => Math.floor(100 * value / max);
```
This is not needed, as it doesn't matter what precedence multiplication and division are executed with.

kdinev (changes requested):

kdinev (dismissed):

projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
```diff
@@ -173,7 +172,7 @@ export abstract class BaseProgressDirective {
      * ```
      */
     public set valueInPercent(value: number) {
-        this._valueInPercent = value;
+        value = toValue(value, this._max);
```
hanastasov, line 175: This setter does nothing. See if possible to remove it, or move it to the base class.

projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
```diff
@@ -402,6 +401,18 @@ export class IgxLinearProgressBarComponent extends BaseProgressDirective impleme
     constructor() {
         super();
     }
+    public set valueInPercent(value: number) {
+        const val = toValue(value, this._max);
+        if (value < 100) {
```
hanastasov, line 406: This check value < 100 can be safely removed. If `value` exceeds 100, then `val` will be such a value, that will not pass the check inside the `set value`.

projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
```diff
@@ -402,6 +401,18 @@ export class IgxLinearProgressBarComponent extends BaseProgressDirective impleme
...
+            this.value = val;
+        } else {
+            this.value = 100;
+        }
+    }
+
+    public get valueInPercent(): number {
```
hanastasov, line 413: `get valueInPercent` seems the same for both extended classes. Maybe just move it under the base class.

projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
```diff
@@ -431,13 +442,12 @@ export class IgxLinearProgressBarComponent extends BaseProgressDirective impleme
         if (isNaN(valInRange) || this._value === val || this.indeterminate) {
```
hanastasov, line 442: should checj this._value against valInRange

hanastasov (changes requested):
`set value` also seems the same for both extended classes - just move it under the base class.

projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
katherinedragieva: One option is to delete it from BaseProgressDirective, since in the extending class (IgxLinearProgressBarComponent) we are calling the setter of the value.

hanastasov: Oh, what I Show is actually in the base class. So, just move the `set valueInPercent` from extended to base class, using the implementattion from the base class.

projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
```diff
@@ -475,6 +475,23 @@ describe('IgLinearBar', () => {
...
+            const fix = TestBed.createComponent(InitLinearProgressBarComponent);
+            fix.detectChanges();
+
+            const bar = fix.componentInstance.linearBar;
+            const maxVal = 3.25;
+            const val =  3.55;
+
+            bar.max = maxVal;
+            bar.value = val;
+            const valueInPercent = Common.calcPercentage(bar.value, bar.max);
```
Aleksandyr, line 489: You can simulate max value increase and double-check the valueInPercent prop to make the test more valid.
### Example:
**first scenario:**
1. set max to 100
2. set value to 50
3. check valueInPercent to equal 50%

**second scenario:**
1. set max to 200
2. check valueInPercent to equal 25%

Aleksandyr (changes requested):

hanastasov (changes requested):
1) `set value` seems the same for both extended classes. Just move it to the base class.
2)  in the `max` setter, don't assign the value to the `valueInPercent`. First convert the passed value to a percentage value, and then assign it:
```typescript
    public set max(maxNum: number) {
        if (maxNum < this._value) {
            this.value = maxNum;
        }
        this._max = maxNum;
        this.valueInPercent = toPercent(this.value, this._max);
    }
```

3) IMO, the `_valueInPercent` var is excessive, no need to keep it. Try removing it. Then, the `get valueInPercent` will look like:
```typescript
    public get valueInPercent(): number {
        return toPercent(this.valueInPercent, this.max);
    }
```

projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
```diff
@@ -475,6 +474,30 @@ describe('IgLinearBar', () => {
...
+            let maxVal = 100;
+            const val =  50;
+
+            bar.max = maxVal;
+            bar.value = val;
+            let valueInPercent = Common.calcPercentage(bar.value, bar.max);
+
+            tick(tickTime);
+            fix.detectChanges();
+            expect(valueInPercent).toBe(50);
```
hanastasov, line 492: Here, we calculated the expected valueInPercent on line 488. Now, we should compare this to the `valueInPercent` holded by the LinearBar itself:

typescript```
expect(valueInPercent).toBe(bar.valueInPercent);
```

hanastasov (approved):

Issue #9975, opened 2021-08-06
Icon is not shown when addSvgIcon is used without being passed a fontFamily 


Aleksandyr:
## Description  
Icon is not shown when addSvgIcon is used without being passed a fontFamily 

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open this [sample](https://stackblitz.com/edit/github-veijhk-pitaeu?file=src%2Fapp%2Fsvg-icon-sample%2Fsvg-icon-sample.component.ts)
2. Try removing fontFamily prop definition from either the template or from addSvgIcon method

## Result  
Icon is not shown

## Expected result  
Icon should be shown as it is described in the [documentation](https://www.infragistics.com/products/ignite-ui-angular/angular/components/icon#svg-icons)
`The icon name and file URL path are the method's mandatory parameters; family can be specified as well.`

PR #10202, opened 2021-09-27
refactor(icon-service): cache svg icons as text

simeonoff:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/icon/icon.component.html
+++ b/projects/igniteui-angular/src/lib/icon/icon.component.html
@@ -5,9 +5,7 @@
 </ng-template>
 
 <ng-template #svgImage>
-    <svg>
-        <use [attr.href]="getSvgKey"></use>
-    </svg>
+    <svg [innerHTML]="getSvg"></svg>
 </ng-template>
 
 <ng-container *ngTemplateOutlet="template"></ng-container>
--- a/projects/igniteui-angular/src/lib/icon/icon.component.ts
+++ b/projects/igniteui-angular/src/lib/icon/icon.component.ts
@@ -238,9 +238,9 @@ export class IgxIconComponent implements OnInit, OnDestroy {
      * }
      * ```
      */
-    public get getSvgKey(): string {
+    public get getSvg(): string {
         if (this.iconService.isSvgIconCached(this.name, this.family)) {
-            return '#' + this.iconService.getSvgIconKey(this.name, this.family);
+            return this.iconService.getSvgIcon(this.name, this.family);
         }
 
         return null;
--- a/projects/igniteui-angular/src/lib/icon/icon.service.ts
+++ b/projects/igniteui-angular/src/lib/icon/icon.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, SecurityContext, Inject, OnDestroy, Optional } from '@angular/core';
+import { Injectable, SecurityContext, Inject, Optional } from '@angular/core';
 import { DomSanitizer } from '@angular/platform-browser';
 import { DOCUMENT } from '@angular/common';
 import { HttpClient } from '@angular/common/http';
@@ -32,7 +32,7 @@ export interface IgxIconLoadedEvent {
 @Injectable({
     providedIn: 'root'
 })
-export class IgxIconService implements OnDestroy {
+export class IgxIconService {
     /**
      * Observable that emits when an icon is successfully loaded
      * through a HTTP request.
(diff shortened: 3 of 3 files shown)


projects/igniteui-angular/src/lib/icon/icon.component.html
```diff
@@ -5,9 +5,13 @@
...
 <ng-template #svgImage>
-    <svg>
-        <use [attr.href]="getSvgKey"></use>
-    </svg>
+    <svg
+        viewBox="0 0 24 24"
+        preserveAspectRatio="xMidYMid meet"
+        xmlns="http://www.w3.org/2000/svg"
+        xmlns:xlink="http://www.w3.org/1999/xlink"
+        [innerHTML]="getSvg"
```
pmoleri, line 13: I don't think that it's safe to assume that any SVG will have a 24x24 viewbox.
Can you use `[outerHTML]` instead?

simeonoff: We can't bind to outerHTML in Angular. It throws an error.

projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -162,18 +153,25 @@ export class IgxIconService implements OnDestroy {
...
+     *   const svgIcon = this.iconService.getSvgIcon('aruba', 'svg-flags');
      * ```
      */
-    public getSvgIconKey(name: string, family: string = '') {
-        return family + '_' + name;
+    public getSvgIcon(name: string, family: string = '') {
+        if(this._cachedSvgIcons.has(family)) {
+            const familyRegistry = this._cachedSvgIcons.get(family) as Map<string, string>;
+            return familyRegistry.get(name);
+        }
```
pmoleri, line 174: You can rewrite the whole thing as:
```
return this._cachedSvgIcons.get(family)?.get(name);
```

projects/igniteui-angular/src/lib/icon/icon.component.html
pmoleri: It does work for me: https://stackblitz.com/edit/angular-1kvoe2?file=src/app/app.component.html

In any case, I'm not able to understand the code fully.
As I see in the cache you are saving: `this._cachedSvgIcons.get(family).set(name, svg.outerHTML);`
And then here you create your own SVG and put the string from the Cache inside.
So you would get a duplicate `<svg>` tag? Or I'm missing something?

simeonoff: Technically you can use outerHTML and it will work, however you will get the following error in the browser console:

```
ERROR DOMException: Failed to set the 'outerHTML' property on 'Element': This element has no parent node.
```

You can refer to [this issue](https://github.com/angular/angular/issues/41576).

No duplicate svg tags are being created. The icon service creates a div and places the content of the fetched string (svg) inside of it (`div.innerHTML = value`). Then we query the div for the svg element we just added as innerHTML to modify some of its attributes before caching it. Finally we cache the icon for the specified family.

kdinev (approved):

projects/igniteui-angular/src/lib/icon/icon.component.ts
```diff
@@ -226,21 +231,21 @@ export class IgxIconComponent implements OnInit, OnDestroy {
...
+     *    let svg: SafeHtml = this.icon.getSvg;
      * }
      * ```
      */
-    public get getSvgKey(): string {
+    public get getSvg(): SafeHtml {
         if (this.iconService.isSvgIconCached(this.name, this.family)) {
-            return '#' + this.iconService.getSvgIconKey(this.name, this.family);
+            const svgText = this.iconService.getSvgIcon(this.name, this.family);
+            return this.sanitizer.bypassSecurityTrustHtml(svgText);
```
pmoleri, line 248: I'm not completely sure about this. But I've experienced excesive DOM refresh issues with getter + sanitizer.
The issue is that the sanitizer returns a brand new object each time which is not strictly equal to the previous one and Angular keeps refreshing it.
What I've done is cache the sanitized result. In this case I would keep: `lastIconKey: this.name + "/" + this.family;` and `lastIconContent = sanitizedString;`
Then compare if the key has changed and if not return the lastIconContent.

zdrawku:
Issue https://github.com/IgniteUI/igniteui-angular/issues/9975 might be related

PR #10224, opened 2021-10-01
Add handling for multiple column dimensions.

MayaKirova:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
@@ -21,6 +21,7 @@ import { IPivotConfiguration, IPivotDimension } from './pivot-grid.interface';
 import { IgxPivotHeaderRowComponent } from './pivot-header-row.component';
 import { IgxColumnGroupComponent } from '../columns/column-group.component';
 import { IgxColumnComponent } from '../columns/column.component';
+import { PivotUtil } from './pivot-util';
 
 let NEXT_ID = 0;
 const MINIMUM_COLUMN_WIDTH = 200;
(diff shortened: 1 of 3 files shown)


dkamburov:
Expanding and collapsing in certain cases breaks the state
multi-cols

dkamburov:
Do we need to support child levels for `columns`? It seems if I define two of them:
```
columns: [{
member: 'Country',
enabled: true,
childLevels:[{
    member: 'Date',
    enabled: true,
    childLevels: []
    },
    {
    member: 'SellerName',
    enabled: true,
    childLevels:[]
    }]
}
```
I get the following output:
image

dkamburov:
Got a bit confused, seems like the following configurations generate the same output:
```
{
    member: 'Country',
    enabled: true,
    childLevels: [{
        member: 'SellerName',
        enabled: true,
        childLevels: []
    }]
}
```
```
{
    member: 'Country',
    enabled: true,
    childLevels: []
},
{
    member: 'SellerName',
    enabled: true,
    childLevels: []
}
```
Can't figure out how they should be different and if we need the `childLevels` for column dimensions

MayaKirova:
In ignite ui it seems that if column dimension is sibling it renders again as a sub group but it does not render an expander. It renders expander only if it has a child dimension:
http://jsfiddle.net/arzhncj4/2/

So I'll apply the same logic here.

MayaKirova:
Regarding having sibling children on the same level:

```
columns: [{
member: 'Country',
enabled: true,
childLevels:[{
    member: 'Date',
    enabled: true,
    childLevels: []
    },
    {
    member: 'SellerName',
    enabled: true,
    childLevels:[]
    }]
}
```

I don't think this is supported in the ignite grid or the wpf one.
In both cases when you define levels they are always nested.
http://jsfiddle.net/gajpL235/1/
https://www.infragistics.com/help/wpf/xampivotgrid-us-defining-hierarchies-and-providing-metadata-with-flatdata

Hence you cannot have 2 columns on level 1. They always form a hierarchy ( so Date is level 1 and Seller will be level 2).

Same goes for the rows as far as I can tell.
So I'm not sure if we should support it here. 

If we do I guess the expected result would be something like this:

|Bulgaria----------------------------------|
|01/01/2021| 02/19/2021|Stenly| Walter |

Since they would be siblings - each unique value would form its own column.

Let me know what you think.

dkamburov:
@MayaKirova I agree on not supporting it and having them always nested.

PR #10427, opened 2021-11-04
Remove deprecateDecorators

### LINKED ISSUE #10319: Use @deprecated JSDoc properly.
You're not using the @deprecated JSDoc comment properly. Looks like you're just specifying the version there that it was deprecated, and then you're using a custom directive to contain the real deprecation message.

The @deprecated JSDoc comment is what all IDE know to look at, and thus *that* is where the deprecation messages should be. Your directive is a run-time thing, which isn't really helpful unless I even know I should be looking in the console to see a warning. Instead, when I'm in the IDE, and I see that the method is deprecated from there, I want to see the "use this instead" message right in the IDE, not have to go run the app to get a runtime warning.

### LINKED ISSUE #10365: Stop showing invalid deprecations
You're spitting out deprecation warnings to the console for code that I am not using. If I haven't utilized a deprecated method or property, there shouldn't be any output. It's fine to deprecate something and leave it in the public API for a few releases, but it should *not* be marked deprecated until your code has itself stopped using it.

It's not just annoyance that it fills up the console with warnings, it makes unit testing harder as well as every test prints out all those invalid warnings.  I end up having to put something like this in every test file:

```ts
beforeAll(() => jest.spyOn(console, 'warn').mockImplementation(jest.fn()))
```

Which then means *my* warnings won't show up.

For example, I get warnings about color being deprecated on IgxIconComponent, and pageChange being deprecated on IgxGridBaseDirective, along with all the other paging related items.  My code uses the `igx-paginator` already.

mmart1n:
Closes #10365
Closes #10319

 - [ ] Bug fix
 - [ ] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/core/deprecateDecorators.ts
```diff
@@ -1,108 +1,5 @@
 import { isDevMode } from '@angular/core';
```
ChronosSF, line 2: This file becomes pointless with just `showMessage`  in it. The function itself is also only used in only 2 places, one being a deprecation that's 2 years old (you can just remove the warning I guess) . And another which just shows a warning that can be implemented in another way.  Even if you choose to keep `showMessage` move it to some util file and remove this file entirely.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -719,6 +705,20 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy {
         });
     }
 
+    /**
+     * @deprecated in version 8.2.0. Use 'selected' property instead
```
ChronosSF, line 709: Was it really deprecated in 8.2.0? @hanastasov , this should've been long removed.

ChronosSF (changes requested):

projects/igniteui-angular/src/lib/grids/cell.component.ts
hanastasov: True, should have been long removed, but is getting removed now. The PR with updates to angular 13 removes this prop.

ChronosSF (approved):

kdinev (approved):

PR #10444, opened 2021-11-09
feat(exporters): added migration comments and test

### LINKED ISSUE #10247: Add migrations for the new way of providing exporter services

igdmdimitrov:
Closes #10247   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


Lipata (dismissed):

gedinakova:
@igdmdimitrov It seems we need to add the trailing comma (if it exists) to the comment, because it causes an error:
image

gedinakova:
@igdmdimitrov @jackofdiamond5 @damyanpetev I'm not sure If we should comment the services out in the imports as well? Maybe only in the providers array?

damyanpetev:
> @igdmdimitrov @jackofdiamond5 @damyanpetev I'm not sure If we should comment the services out in the imports as well? Maybe only in the providers array?

May be missing something, but is the comment message correct - `IgxExcelExporterService has been removed.`?
I thought it was available publicly as stand-alone service and if so that message is a bit misleading, no? And more to the point the use of it in provides can still have valid uses like overriding the service like `{ provide: IgxExcelExporterService , useClass: CustomService }`, right?
In that case, especially if we can't detect with absolute certainty what the code is doing, adding just an explainer comment should be enough, just to indicate to updated project they can drop the provider most likely. IIRC that was kind of the general idea behind this. And even if we could, commenting out the service in code is kind of odd - like why leave it there? Either add the comment or remove it, doing both with the remove done with comment looks kinda messy in the end.

PS: The linked issue is not a _bug_ btw, more like a nice-to-have enhancement for migrations
PPS: Not a huge fan of editing TypeScript code using Regex string manipulation - as you've seen it can get complicated handling code structure in just string without context (commas, formatting, etc). Ideally those changes should really target something very specific and be sure about it - there are some methods to find references to identifiers in the migration's tsUtils and perhaps some of that can be re-used.

igdmdimitrov:
> > @igdmdimitrov @jackofdiamond5 @damyanpetev I'm not sure If we should comment the services out in the imports as well? Maybe only in the providers array?
> 
> May be missing something, but is the comment message correct - `IgxExcelExporterService has been removed.`? I thought it was available publicly as stand-alone service and if so that message is a bit misleading, no? And more to the point the use of it in provides can still have valid uses like overriding the service like `{ provide: IgxExcelExporterService , useClass: CustomService }`, right? In that case, especially if we can't detect with absolute certainty what the code is doing, adding just an explainer comment should be enough, just to indicate to updated project they can drop the provider most likely. IIRC that was kind of the general idea behind this. And even if we could, commenting out the service in code is kind of odd - like why leave it there? Either add the comment or remove it, doing both with the remove done with comment looks kinda messy in the end.
> 
> PS: The linked issue is not a _bug_ btw, more like a nice-to-have enhancement for migrations PPS: Not a huge fan of editing TypeScript code using Regex string manipulation - as you've seen it can get complicated handling code structure in just string without context (commas, formatting, etc). Ideally those changes should really target something very specific and be sure about it - there are some methods to find references to identifiers in the migration's tsUtils and perhaps some of that can be re-used.

Addressed - now commenting only at the start of the file and mentioning the services used in the file.

projects/igniteui-angular/migrations/update-13_0_0/index.ts
```diff
@@ -16,6 +16,8 @@ export default (): Rule => (host: Tree, context: SchematicContext) => {
     const GRIDS = ['IgxGridComponent', 'IgxTreeGridComponent', 'IgxHierarchicalGridComponent'];
     const TAGS = ['igx-grid', 'igx-tree-grid', 'igx-hierarchical-grid'];
     const tsFiles = update.tsFiles;
+    const SERVICES = ['IgxCsvExporterService', 'IgxExcelExporterService'];
+    const moduleTsFiles = update.moduleTsFiles;
```
damyanpetev, line 20: this could easily be `update.tsFiles.filter(x => x.endsWith('.module.ts'))`, no? Kind of no need to go through a separate getter that loops over the directories again IMO - doubt it's meaningfully slower, but it's still duplicate work :)

projects/igniteui-angular/migrations/update-13_0_0/index.ts
```diff
@@ -61,5 +63,22 @@ export default (): Rule => (host: Tree, context: SchematicContext) => {
...
+        const servicesInFile = [];
+        SERVICES.forEach(service => {
+            if (content.indexOf(service) > -1) {
+                servicesInFile.push(service);
+            }
+        });
+
+        if (servicesInFile.length > 0) {
+            const comment = '// ' + servicesInFile.join(' and ') + ' no longer need to be manually provided and can be safely removed.';
+            content = comment + content;
```
damyanpetev, line 78: Might be missing something here, but if you directly prepend the comment to the content - shouldn't there be a new line somewhere?

damyanpetev (approved):

PR #10503, opened 2021-11-15
Add indentation based on level and set correct collapsed state icon as per the design

MayaKirova:
Minor UI changes.

Note: Indentation size depends on display density, which will probably be smaller in the pivot grid.

Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


skrustev:
I am not sure but for a configuration like the one bellow, it appears that the first row dimension receives and extra indentation for the first rows of each `All Products`,  while I think there should be no indentation at all for the first dimension:

UpperLevelIndentation

Configuration:


```typescript
        rows: [
            {
                memberName: 'City',
                enabled: true
            },
            {
                memberFunction: () => 'All',
                memberName: 'AllProducts',
                enabled: true,
                childLevel:
                {
                    memberFunction: (data) => data.ProductCategory,
                    memberName: 'ProductCategory',
                    enabled: true,
                }
            }

        ]
```

skrustev:
1) For column dimensions it seems to apply the width specified to each value column. 
For example if there is column dimension with width 200px specified and there are two values in the config, the column dimension actually has 400px width.
It can be logged separately though

2) When moving column dimension chip to row dimension area the grid does not reflow. This is not reproduced in the pivot master.
PivotGrid

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row.component.html
```diff
@@ -98,15 +98,25 @@
...
+            <igx-icon [attr.draggable]="false"
+                      (click)="grid.toggleRow(getRowDimensionKey(column))">
+                      {{ getExpandState(column) ? 'expand_more' : 'chevron_right'}}</igx-icon>
             {{column.header}}
         </div>
     </ng-template>
 
     <ng-template #headerTemplateGray let-column>
-        <div class="igx-grid__tr--header">
+        <div class='igx-grid__tr--header igx-grid__row-indentation--level-{{getLevel(column)}}'>
```
mddragnev, line 110: Do we still need both default template and gray template ?

mddragnev (changes requested):

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row.component.html
dkamburov: I guess this comment is for @desig9stein and @StefanIvanov. Should we drop completely gray styles before having the cell merging feature?

StefanIvanov: We can drop it for now, this version is just a preview, so let's not add things that we wouldn't like to be there later on if this is the case here :)

MayaKirova: Ok. In that case I'll remove the whole template.

mddragnev (approved):

PR #10558, opened 2021-11-18
Use typechecker to determine property types

jackofdiamond5:
Closes https://github.com/IgniteUI/igniteui-angular/issues/10572

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -264,6 +264,30 @@ export const getTypeDefinitionAtPosition =
...
         }
+        if (entryPath.endsWith('.ts')) {
+            // for ts files we can use the type checker to look up a specific node
+            // and attempt to resolve its actual type
+            const sourceFile = langServ.getProgram().getSourceFile(entryPath);
+            const typeChecker = langServ.getProgram().getTypeChecker();
+            // const node = (tss as any).getTouchingPropertyName(sourceFile, position); -> tss internal that looks up a node
+            const node = findNodeAtPosition(sourceFile, position);
+            const symbolFromNode = typeChecker.getSymbolAtLocation(node)
+            const nodeType = typeChecker.getTypeOfSymbolAtLocation(symbolFromNode, node)
```
damyanpetev, line 275: PS: There's also `getTypeAtLocation` that only uses the node, w/o a symbol btw. Not sure why this API needs both, but it hints to me maybe it can do something we don't necessarily expect

projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -264,6 +264,30 @@ export const getTypeDefinitionAtPosition =
...
+            if (nodeType) {
+                const typeArguments = typeChecker.getTypeArguments(nodeType as tss.TypeReference);
+                if (typeArguments && typeArguments.length < 1) {
+                    // it's not a generic type so try to look up its name and fileName
+                    // atm we do not support migrating union/intersection generic types
+                    // a type symbol (type) should have only one declaration
+                    // if the type is 'any' or 'some', there will be no type symbol
+                    const name = nodeType.getSymbol()?.getName();
+                    const fileName = nodeType.getSymbol()?.getDeclarations()[0].getSourceFile().fileName;
```
damyanpetev, line 284: `getDeclarations()` can return undefined btw, not sure if it can return empty array

projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -339,6 +362,24 @@ export const isMemberIgniteUI =
...
+const findNodeAtPosition = (sourceFile: tss.SourceFile, position: number): tss.Node | null => {
+    if (!sourceFile) {
+        return null;
+    }
+
+    return findInnerNode(sourceFile, position);
+}
+const findInnerNode = (node: tss.Node, position: number): tss.Node | null => {
+    if (position <= node.getEnd()) {
+        return node.forEachChild(cn => findInnerNode(cn, position)) || node;
```
damyanpetev, line 377: This seems like it can be improved, since the way it's set up right now, there's no way to short-circuit the recursion (might not be ideal for this case) from processing larger part of the tree than necessary.

damyanpetev (changes requested):

projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -45,7 +50,9 @@ export const getIdentifierPositions = (source: string | ts.SourceFile, name: str
                 return false;
             }
         }
-        return node.text === name;
+        // for methods the node.text will not contain characters like ()
+        const cleanName = name.match(/\w+/g)[0] || name;
```
damyanpetev, line 54: @jackofdiamond5 for future improvement: this can probably be resolved with some `kind` check and likely the method nodes will have `name` or something

projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -12,12 +12,17 @@ export const NG_CORE_PACKAGE_NAME = '@angular/core';
...
 export const CUSTOM_TS_PLUGIN_NAME = 'igx-ts-plugin';
 
-enum SynaxTokens {
+enum SyntaxTokens {
     ClosingParenthesis = ')',
     MemberAccess = '.',
     Question = '?'
 }
 
+export class MemberInfo implements Pick<tss.DefinitionInfo, 'name' | 'fileName'> {
```
damyanpetev, line 21: while this is related to members, this isn't the member info - it's still the definition info for whatever we looked up at a position and it's usually the container type of the member instead :) Don't much care for the name ATM, just saying it's not quite right.

projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -339,6 +358,48 @@ export const isMemberIgniteUI =
...
+    const nodeType = typeChecker.getTypeAtLocation(node);
+    const typeArguments = typeChecker.getTypeArguments(nodeType as tss.TypeReference);
+    if (typeArguments && typeArguments.length < 1) {
+        // it's not a generic type so try to look up its name and fileName
+        // atm we do not support migrating union/intersection generic types
+        // a type symbol (type) should have only one declaration
+        // if the type is 'any' or 'some', there will be no type symbol
+        const name = nodeType.getSymbol()?.getName();
+        const declarations = nodeType.getSymbol()?.getDeclarations();
+        if (declarations) {
```
damyanpetev, line 372: Found one instance of this in ts server that was also checked for length, so it seems there might be an obscure case where it's not undefined but also empty?

projects/igniteui-angular/migrations/common/tsUtils.ts
```diff
@@ -339,6 +358,48 @@ export const isMemberIgniteUI =
             && change.definedIn.indexOf(typeDef.name) !== -1;
     };
 
+const resolveMemberInfo = (langServ: tss.LanguageService, node: tss.Node): MemberInfo | null => {
+    const typeChecker = langServ.getProgram().getTypeChecker();
+    const nodeType = typeChecker.getTypeAtLocation(node);
```
damyanpetev, line 363: Any chance this will return something `getTypeArguments` won't like? The errorType?

projects/igniteui-angular/migrations/common/tsUtils.ts
damyanpetev: Alright, the docs can clear it up, it's just the `forEachChild` method is not that intuitively name, but it does exactly what we need and internally terminated the loop

projects/igniteui-angular/migrations/common/tsUtils.ts
jackofdiamond5: If you try to call it on a token, for example, It will return an intrinsic type (`error`) and then `getTypeArguments` will return an empty array. So it shouldn't throw.

PR #10611, opened 2021-11-23
fix(icon-service): don't initialize DOMParser in SSR mode

### LINKED ISSUE #10602: igx-icon SSR ReferenceError: DOMParser is not defined
## Description  
Using the `igx-icon` component in an Angular application with SSR causes the following error:

 > ReferenceError: DOMParser is not defined

 * igniteui-angular version: 12.3.0
 * browser: N/A

## Steps to reproduce  

1. Create an Angular app with SSR and `igx-icon` component in the app.component

## Result  
The view is not rendered because of the error.

## Expected result  
There should be no error thrown.

## Attachments  
image

simeonoff:
Closes #10602  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/icon/icon.service.ts
+++ b/projects/igniteui-angular/src/lib/icon/icon.service.ts
@@ -59,7 +59,7 @@ export class IgxIconService {
     ) {
         this.iconLoaded = this._iconLoaded.asObservable();
 
-        if(this._platformUtil.isBrowser) {
+        if(this._platformUtil?.isBrowser) {
             this._domParser = new DOMParser();
         }
     }
@@ -194,7 +194,7 @@ export class IgxIconService {
     private cacheSvgIcon(name: string, value: string, family = this._family, stripMeta: boolean) {
         family = !!family ? family : this._family;
 
-        if (this._platformUtil.isBrowser && name && value) {
+        if (this._platformUtil?.isBrowser && name && value) {
             const doc = this._domParser.parseFromString(value, 'image/svg+xml');
             const svg = doc.querySelector('svg') as SVGElement;
 
--- a/projects/igniteui-angular/src/lib/icon/icon.service.ts
+++ b/projects/igniteui-angular/src/lib/icon/icon.service.ts
@@ -3,6 +3,7 @@ import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
 import { DOCUMENT } from '@angular/common';
 import { HttpClient } from '@angular/common/http';
 import { Observable, Subject } from 'rxjs';
+import { PlatformUtil } from '../core/utils';
 
 /**
  * Event emitted when a SVG icon is loaded through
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -48,14 +48,19 @@ export class IgxIconService {
...
     private _iconLoaded = new Subject<IgxIconLoadedEvent>();
-    private _domParser = new DOMParser();
+    private _domParser: DOMParser;
 
     constructor(
         @Optional() private _sanitizer: DomSanitizer,
         @Optional() private _httpClient: HttpClient,
-        @Optional() @Inject(DOCUMENT) private _document: any
+        @Optional() @Inject(DOCUMENT) private _document: any,
+        @Inject(PLATFORM_ID) private platformId: any
```
kdinev, line 57: Directly inject the `PlatformUtil` we have. It has an `isBrowser` field.

projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -48,14 +48,19 @@ export class IgxIconService {
...
     ) {
         this.iconLoaded = this._iconLoaded.asObservable();
+
+        if(isPlatformBrowser(this.platformId)) {
```
kdinev, line 61: Don't make the checks where the `cacheSvgIcon` is called. Make the check inside the first `if` in the `cacheSvgIcon` method instead. This way it would be more future proof if the method is called from somewhere else in the future.

kdinev (changes requested):

projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -164,7 +161,7 @@ export class IgxIconService {
      */
     public isSvgIconCached(name: string, family: string = ''): boolean {
         const familyClassName = this.familyClassName(family);
-        if (this._cachedSvgIcons.has(familyClassName)) {
+        if (this._cachedSvgIcons.has(familyClassName) && this._platformUtil.isBrowser) {
```
kdinev, line 164: Remove this check. In platform server the check would just return `false`, so nothing will throw. If we implement server handling in the future, then the method will return the icon properly on the server also.

projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -180,7 +177,7 @@ export class IgxIconService {
      */
     public getSvgIcon(name: string, family: string = '') {
         const familyClassName = this.familyClassName(family);
-        return this._cachedSvgIcons.get(familyClassName)?.get(name);
+        return this._platformUtil.isBrowser && this._cachedSvgIcons.get(familyClassName)?.get(name);
```
kdinev, line 180: Remove this check. In platform server the method would just return `undefined`, so nothing will throw. If in the future we implement server-handling for this, then this method will return the icon properly.

kdinev (changes requested):

kdinev (approved):

kdinev (approved):

PR #10653, opened 2021-11-29
Emit panelCollapsing event once when a panel starts collapsing

### LINKED ISSUE #10520: IgxAccordion panelCollapsing event is emitted multiple times
## Description  
When a panel is collapsing and panel.collapse() is being called, panelCollapsing event is emitted multiple times.

 * igniteui-angular version: 13.0.x, 12.3.x

## Steps to reproduce  

1. Open the following [sample](https://stackblitz.com/edit/github-ajopib?file=src%2Fapp%2Faccordion-sample-1%2Faccordion-sample-1.component.ts)
2. Expand one of the panels
3. Click on the expanded panels multiple times

## Result  
panelCollapsing is emitted multiple times.

## Expected result  
panelCollapsing should be emitted only once

## Attachments  
accordion2

accordion1

MonikaKirkova:
Closes #10520 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/accordion/accordion.component.ts
```diff
@@ -387,7 +387,9 @@ export class IgxAccordionComponent implements AfterContentInit, AfterViewInit, O
             });
             panel.contentCollapsing.pipe(takeUntil(this._unsubChildren$)).subscribe((args: IExpansionPanelCancelableEventArgs) => {
                 const evArgs: IAccordionCancelableEventArgs = { ...args, owner: this, panel: args.owner };
-                this.panelCollapsing.emit(evArgs);
+                if (!args.owner.closeAnimationPlayer?._started) {
```
jackofdiamond5, line 390: Do we want to emit `panelCollapsing` if `closeAnimationPlayer` is falsy? I'm asking because right now if the `closeAnimationPlayer` is falsy, this may not behave as expected as the entire thing will resolve to `undefined` and then the negation will make it `true`, so the event will emit. If we want to emit the event only if the closing animation hasn't started then it might be a good idea to separate the check to be something like `args.owner.closeAnimationPlayer && !args.owner.closeAnimationPlayer._started`.

projects/igniteui-angular/src/lib/accordion/accordion.component.spec.ts
```diff
@@ -232,6 +232,7 @@ describe('Rendering Tests', () => {
                 argsIng = { event: expArgs.event, cancel: expArgs.cancel, owner: accordion, panel: expArgs.owner };
             });
             accordion.panels[0].collapse();
+            accordion.panels[0].collapse();
```
jackofdiamond5, line 235: Why are we calling `collapse` on the first panel twice?

MonikaKirkova: I am calling collapse method twice, in order to check if panelCollapsing is emitted again after the panel already started collapsing.

projects/igniteui-angular/src/lib/accordion/accordion.component.ts
MonikaKirkova: When a panel starts collapsing the closeAnimationPlayer is still null and the event should be emitted, however, if the check is `args.owner.closeAnimationPlayer && !args.owner.closeAnimationPlayer._started` the event would not be emitted.

wnvko (changes requested):
the issue here is not in the accordion. The issue is coming from the panel. Panel should not emit `contentCollapsing` if its close animation has already started. You need to extend the logic in panel's `collapse` method and try to collapse the panel only if close animation has not started. Doing this panel will fire `contentCollapsing` once, and finally accordion will reemit it once.

wnvko (approved):

PR #10739, opened 2021-12-15
fix(pivot): Ensure that when columns are created for row dimensions w…

### LINKED ISSUE #10498: Pivot grid vertical scrolling performance issue.
## Description  
Currently when there are more dimensions that produce a vertical scrollbar for the Pivot Grid, the scrolling performance is slower than expected

 * browser: any

## Steps to reproduce  

1. Open the pivot sample
2. Setup more row dimensions so vertical scroll is rendered
3. Scroll vertically

## Result  
Scrolling performance is slower than expected

## Expected result  
Scrolling performance should be faster and not too different than the regular igxGrid

MayaKirova:
…hile scrolling the grid does not repaint.

Closes #10498   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


skrustev:
Completely disabling the repainting for the Pivot Grid currently results in some cases ending up with the grid needing to repaint, but not doing so.

For example:
1. Set the following column dimensions to the Pivot sample: 
 ```
[
           {
                memberName: 'AllProduct',
                memberFunction: () => 'All Products',
                enabled: true,
                childLevel: {
                    memberName: 'ProductCategory',
                    enabled: true
                }
            },
            new IgxPivotDateDimension(
                {
                    memberName: 'Date',
                    enabled: true
                },
                {
                    months: false,
                    total: true
                }
            ),
            {
                memberName: 'SellerName',
                enabled: true
            },
]
```

2. Move 'All products' dimension from row to column area
3. Remove the 'All Products' dimension

The bottom grid area has a black area that results from not repainting the grid after removing of the dimension
Maybe this needs to false always when creating the `IgxColumnComponent` in the rows only?

MayaKirova:
@skrustev The problem is that this is triggered while the column is still in the process of being created (during createComponent execution) so the column is not yet added anywhere (its not in any ContentChildren collection nor in the viewRef children yet, nor in the DOM) so it will be difficult at this point to find where it will be added (in the rows or in the root grid). 

In general it seems that notifyChanges decorator is meant to trigger reflow on change for particular column properties, but I cannot think of scenarios where you would want it to trigger during initialization with the default prop values. I'm not sure that is a desired effect. And since all columns in the pivot are auto-generated (user cannot directly change prop values on them) I think it should be safe to disable this for the pivot and just call reflow when needed (for example when columns are re-generated as the result of adding/removing column dimensions). 

Let me know if that sounds ok or if you have any other ideas.

dkamburov (approved):

PR #10753, opened 2021-12-20
fix(tooltip): remove unnecessary div creation

### LINKED ISSUE #10723: Tooltip is moving when it is used with flex and gap
## Description  
When tooltip is used with flex with gap, it is initially hidden, but once it's triggering is leaving a lingering non-hidden div behind.

## Result  
When the component appear it is moving slightly to the left side.

## Expected result  
The tooltip component should stay on the same position as when it is not hovered.

## Attachments  
Here is an example of the bug: [StackBlitz](https://stackblitz.com/edit/github-h8flju?file=src/app/app.component.html)

Volen99:
Closes #10723

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


Svetloslav15 (approved):

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -577,16 +576,6 @@ export class IgxOverlayService implements OnDestroy {
         return info;
     }
 
-    private placeElementHook(element: HTMLElement): HTMLElement {
-        if (!element.parentElement) {
-            return null;
-        }
-
-        const hook = this._document.createElement('div');
```
wnvko, line 585: we are using element hook to return the overlay element at its original position. If we do not do this overlay will change the original markup which is not desired.
To fix the issue you can set the `display` of the hook element to `none` like this:
```
hook.style.display = 'none';
```

wnvko (changes requested):

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
Volen99: Oh, I see. That was very helpful, I was really wondering what was the reason behind this hook. Will do so ✔

wnvko (approved):

Svetloslav15 (approved):

hanastasov (approved):

PR #10787, opened 2022-01-06
feat(igxAutocomplete): expose empty template

### LINKED ISSUE #10681: Autocomplete does not expose emptyTemplate
Autocomplete does not support emptyTemplate

Expose emptyTemplate

ddaribo:
Closes #10681 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


Aleksandyr:
12 test are failing:
image

Aleksandyr:
Please update the CHANGELOG and the WIKI with the new empty template.

Aleksandyr:
Seems LGTM to me. @wnvko can we have a review here?

wnvko (changes requested):
Do we really need empty template in the drop down?
You can add one by simply put in the drop down something like this:
```
<igx-drop-down>
    <igx-drop-down-item *ngFor="let item of items>{{item}}<igx-drop-down-item>
    <span *ngIf="items.length === 0">There are no items to show</span>
</igx-drop-down>
```
As the issue was logged for autocomplete with custom filtering pipe you can do the same like this:
```
    <igx-input-group>
        <input #input igxInput [igxAutocomplete]='items'/>
    </igx-input-group>
    <igx-drop-down #attractionsPanel>
        <igx-drop-down-item *ngFor="let item of items | contains: input.value" [value]="item.value">
            <span>{{item.name}}</span>
        </igx-drop-down-item>
        <igx-drop-down-item *ngIf="hasItems(input.value)">There is no item to show here!</igx-drop-down-item>
    </igx-drop-down>
```
and in the code behind:
```
    public hasItems(item: string): boolean {
        return filterContains(this.attractions, item).length === 0;
    }
```

ddaribo:
> 

True, although this solution shows the empty drop down item if there are valid suggestions, because the filterContains checks against exact matches. This could probably be handled by altering the contains logic ofc:
image


The other thing is the selection behavior on Enter, which with the above solution would be:
 
close-on-enter

Which attempts to select the highlighted “empty item”, however it is not in the suggestions list and results in the value being discarded. Probably additional logic needs to be introduced to handle this and closing/selection on app level and depending on the use case - whether a non-mathcing value should be kept, etc.

@Aleksandyr what do you think?

wnvko:
@ddaribo that is right.
However, dropdown is just a container. If you provide it drop down items you have navigation between items. However, you can put inside whatever you want. So you can put the empty message in a div, and will not be able to navigate to it.
The main idea here is - you have an empty container and you can put inside whatever you want. No need to implement specific templates for container where I can put anything inside it 😄

Aleksandyr:
@wnvko I remember how we were discussing this functionality as we were trying to integrate the dropdown into the AB workspace dialog where we needed an empty template when no users are available for an invite.

I see that just passing an element in the container of either autocomplete or a dropdown is not working as expeted e.g combo. This is not the acceptance criteria for this issue and this is not the desired behavior.

ddaribo:
As per @wnvko’s suggestion, a “no matches” indicator can be provided as a drop down item on application level with a few additions, concerning the selection and closing behavior, for example:
  ```
<igx-input-group>
       <input #input igxInput [igxAutocomplete]='items' (selectionChanging)='autocompleteSelectionChanging($event)'/>
</igx-input-group>
<igx-drop-down #attractionsPanel>
    <igx-drop-down-item *ngFor="let item of items | contains: input.value" [value]="item.value">
        <span>{{item.name}}</span>
    </igx-drop-down-item>
    <igx-drop-down-item *ngIf="hasItems(input.value)" [value]="'noMatches'">There is no item to show here!</igx-drop-down-item>
</igx-drop-down>
```


```
@ViewChild('attractionsPanel', { read: IgxDropDownComponent, static: true })
private dropdown: IgxDropDownComponent;
```
```
public autocompleteSelectionChanging(event){
   if (this.dropdown.children.length === 1 && this.dropdown.children.first.value === 'noMatches') {
        event.cancel = true;
   }
}
```
```
this.dropdown.closing.subscribe((args: IBaseCancelableBrowserEventArgs) => {
  if (this.dropdown.children.length === 1
      && this.dropdown.children.first.value.toString() === 'noMatches'
      && args.event?.type === "keydown" && (args.event as KeyboardEvent).key.toLowerCase() === 'enter') {
             args.cancel = true;
  }
});
```

Canceling the autocomplete’s selectionChanging event prevents setting the input’s value to the “empty” dropdown item’s value. By subscribing to the dropdown’s closing event, the developer could cancel closing if required – for example, on pressing “Enter”/”Escape” when the “empty” item is shown. Such behavior would be similar to the [Simple ComboBox](https://www.infragistics.com/products/ignite-ui-angular/angular/components/simple-combo))

However, the IBaseCancelableBrowserEventArgs' event property is not passed down to the overlay service’s hide method when closing the dropdown. As discussed, this PR will address this instead and an emtpy template for the dropdown will not be exposed.

wnvko (approved):

PR #10874, opened 2022-01-21
Render values even if no dimensions specified

### LINKED ISSUE #10810: [Pivot] If there are no columns or row dimensions declared no values are rendered
## Description  
If there are no columns or row dimensions declared no values are rendered

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open pivot sample from dev samples
2. Remove all row dimensions

Same applies for column dimensions

## Result  
No values are rendered

## Expected result  
There should be rendered values

## Attachments  
image

there should be aggregations even if the row dimensions are not specified(as in igPivotGrid)
image

3phase:
Closes #10810   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.html
```diff
@@ -27,10 +27,24 @@
...
-        <igx-pivot-row-dimension-content  role='row' class="igx-grid-thead" [grid]="this"
-            [dimension]='dim'
+            <igx-pivot-row-dimension-content role='row' class="igx-grid-thead" [grid]="this"
+                [dimension]='dim'
+                [rowIndex]='rowIndex' [rowData]='rowData'
+                [density]="displayDensity" [width]="resolveRowDimensionWidth(dim)">
+            </igx-pivot-row-dimension-content>
+        </ng-template>
+    </div>
+    <div tabindex="0" *ngIf='rowDimensions.length === 0' #rowDimensionContainer role="rowgroup" class='igx-grid__tbody-pivot-dimension' (focus)="navigation.focusTbody($event)" (keydown)="navigation.handleNavigation($event)">
```
MayaKirova, line 37: I think this template can be simplified a bit. For example both can be extracted in their own templates so that it is more clear which is used when and they can be switched via ng-content, for example:

```
    <ng-content *ngTemplateOutlet="rowDimensions.length === 0 ? emptyRowDimensionsTemplate : defaultRowDimensionsTemplate"></ng-content>
```

Also the empty dimension does not need an igxForOf, it always renders just 1 record. It can be simplified to just something like this:

```
<ng-template #emptyRowDimensionsTemplate>
    <div tabindex="0" class='igx-grid__tbody-pivot-dimension' (focus)="navigation.focusTbody($event)">
        <igx-pivot-row-dimension-content role='row' class="igx-grid-thead" [grid]="this"
        [dimension]='emptyRowDimension'
        [rowIndex]='0' [rowData]='{}'
        [density]="displayDensity" [width]="resolveRowDimensionWidth(emptyRowDimension)">
    </igx-pivot-row-dimension-content>
    </div>
</ng-template>
```

Also passing undefined as dimension is not ideal. I would recommend implementing a default empty row dimension, which can store information like widths, since some features like resizing would not work with undefined.

projects/igniteui-angular/src/lib/data-operations/pivot-strategy.ts
```diff
@@ -40,6 +40,16 @@ export class PivotRowDimensionsStrategy implements IPivotDimensionStrategy {
         let prevDimTopRecords = [];
         const currRows = cloneArray(rows, true);
         PivotUtil.assignLevels(currRows);
+
+        if (currRows.length === 0) {
+            for (const row of collection) {
```
MayaKirova, line 45: No need to loop through each record.  Also `row` here is not a dimension so there's no point in passing it in `getFieldsHierarchy` or pushing in prevRowDims.

This could be simplified to something like:

```
        if (currRows.length === 0) {
            hierarchies = PivotUtil.getFieldsHierarchy(collection, [{memberName: '', enabled: true}], PivotDimensionType.Row, pivotKeys);
            // generate flat data from the hierarchies
            data = PivotUtil.processHierarchy(hierarchies, collection ?? [], pivotKeys, 0, true);
            return data;
        }
```

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.pipes.ts
```diff
@@ -74,7 +74,10 @@ export class IgxPivotRowExpansionPipe implements PipeTransform {
             PivotUtil.flattenHierarchy(data, config, row, expansionStates, defaultExpand, pivotKeys, totalLlv, prevDims, 0, lvl);
             prevDims.push(row);
         }
-        const finalData = config.columnStrategy ? data : data.filter(x => x[pivotKeys.records]);
+        const finalData = config.columnStrategy ? data :
+            enabledRows.length ? data.filter(x => x[pivotKeys.records]) :
+                PivotUtil.flattenHierarchy(data, config, {} as IPivotDimension, expansionStates, defaultExpand, pivotKeys, totalLlv, prevDims, 0, 0);
```
MayaKirova, line 79: Not sure what's the point of flattening the data here.  If there are no row dimensions, there shouldn't be any hierarchy to flatten. There should be just 1 record and it should not have a hierarchy.

MayaKirova (changes requested):

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
```diff
@@ -345,6 +345,23 @@ export class IgxPivotGridComponent extends IgxGridBaseDirective implements OnIni
      */
     public rowDimensionResizing = true;
 
+    /**
+     * @hidden @internal
+     */
+    private _emptyRowDimension: IPivotDimension;
```
MayaKirova, line 351: Could just be a getter, I don't think we need to set it. Something like:

```
    /**
     * @hidden @internal
     */
    private _emptyRowDimension: IPivotDimension = {memberName: '', enabled: true};
    public get emptyRowDimension(): IPivotDimension {
        return this._emptyRowDimension;
    }
```

Re-using the same object instance will allow resizing to set width to it.
You might also need to update some of the methods that get the total widths, like `pivotRowWidths` to use this instance to reflect resizing via the dimension cell correctly.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row-dimension-content.component.ts
```diff
@@ -175,18 +180,26 @@ export class IgxPivotRowDimensionContentComponent extends IgxGridHeaderRowCompon
...
     }
 
     protected _createColComponent(field: string, header: string, dim: IPivotDimension, lvl) {
         const ref = this.viewRef.createComponent(IgxColumnComponent);
         ref.instance.field = field;
         ref.instance.header = header;
-        ref.instance.width = this.grid.resolveRowDimensionWidth(this.dimension) + 'px';
         ref.instance.resizable = this.grid.rowDimensionResizing;
+
+        if (!dim) {
```
MayaKirova, line 195: With the latest changes, there should always be a dimension, so maybe the all changes in this file that check for undefined dimension are no longer necessary.

MayaKirova (changes requested):

MayaKirova:
Just one more thing, when resizing the empty row dimension the headers and the column don't change size accordingly:

empty-dimension-resize


Probably the related column object and the `pivotRowWidths` should take into consideration the width that is stored in the empty row dimension object.

3phase:
> Just one more thing, when resizing the empty row dimension the headers and the column don't change size accordingly:
> 
> empty-dimension-resize
> 
> Probably the related column object and the `pivotRowWidths` should take into consideration the width that is stored in the empty row dimension object.

It's resolved now, thanks.

MayaKirova (dismissed):

mddragnev:
@3phase
Remove all row and filters dimensions and leave only column dimension. Move column dimension to filters. Observe that there is no data in the grid and a `All periods` column exists.
image

Almost the same happens if you remove all rows and move the column dimension to the rows or if you move the column to the rows without removing them.
image
or 
image

3phase:
@mddragnev I'm not able to reproduce any of the issues you're mentioning.
Issue 1:
pivot-issue-1

Issue 2:
pivot-issue-2

Issue 3:
pivot-issue-3

Can you give it another shot?

MayaKirova (approved):

PR #10892, opened 2022-01-24
test(progress): refactoring unit tests

kdinev:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [x] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/progressbar/circularbar.component.spec.ts
+++ b/projects/igniteui-angular/src/lib/progressbar/circularbar.component.spec.ts
@@ -1,8 +1,7 @@
 import { Component, ViewChild } from '@angular/core';
 import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
-import { IgxCircularProgressBarComponent, IgxProgressBarModule } from './progressbar.component';
-import { Common } from './common.spec';
+import { IgxCircularProgressBarComponent, IgxProgressBarModule, toPercent } from './progressbar.component';
 
 import { configureTestSuite } from '../test-utils/configure-suite';
 
@@ -242,11 +241,16 @@ describe('IgCircularBar', () => {
 
         fix.detectChanges();
 
-        const percentValue = Common.calcPercentage(value, max);
+        const percentValue = toPercent(value, max);
         expect(bar.value).toBe(value);
         expect(bar.step).toBe(step);
         expect(bar.max).toBe(max);
         expect(bar.valueInPercent).toBe(percentValue);
+
+        // Should not set a step larger than max value
+        bar.step = 20;
+        expect(bar.step).toBe(step);
+        expect(bar.max).toBe(max);
     });
 
     it(`when step value is not divisble to passed value the result returned from the
@@ -257,7 +261,7 @@ describe('IgCircularBar', () => {
         const bar = fix.componentInstance.circularBar;
         const step = 3.734;
         let value = 30;
-        let valueInPercent = Common.calcPercentage(value, bar.max);
+        let valueInPercent = toPercent(value, bar.max);
         bar.step = step;
         bar.value = value;
 
@@ -267,7 +271,7 @@ describe('IgCircularBar', () => {
         expect(bar.valueInPercent).toBe(valueInPercent);
 
         value = 10;
-        valueInPercent = Common.calcPercentage(value, bar.max);
+        valueInPercent = toPercent(value, bar.max);
         bar.value = value;
         fix.detectChanges();
 
@@ -381,7 +385,7 @@ describe('IgCircularBar', () => {
             tick(tickTime);
             fix.detectChanges();
 
-            const progressRepresentation = Common.calcPercentage(val, maxVal);
+            const progressRepresentation = toPercent(val, maxVal);
             const progressBarElem = fix.debugElement.query(By.css('svg'));
             const valueInPercent = progressBarElem.query(By.css(`.${CIRCULAR_TEXT_CLASS}`)).nativeElement;
             expect(valueInPercent.textContent.trim()).toBe(`${progressRepresentation}%`);
--- a/projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
+++ b/projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
@@ -1,8 +1,5 @@
-import { Component, ViewChild } from '@angular/core';
-import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
-import { By } from '@angular/platform-browser';
-import { IgxLinearProgressBarComponent } from './progressbar.component';
-import { Common } from './common.spec';
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { IgxLinearProgressBarComponent, IgxProgressBarModule, toPercent } from './progressbar.component';
 
 import { configureTestSuite } from '../test-utils/configure-suite';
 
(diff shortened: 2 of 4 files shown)


projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
```diff
@@ -184,10 +184,6 @@ export abstract class BaseProgressDirective {
 
         this._internalState.newVal = Math.round(toValue(toPercent(this.value, maxNum), maxNum));
         this._value = this._internalState.oldVal = Math.round(toValue(this.valueInPercent, maxNum));
-        if (maxNum < this._value) {
```
hanastasov, line 187: Setting the `max` runtime with a value lesser than current value would potentially enter this if. However, in such case, `this._value` will be recalculated wrongly to 0 on line 186. IMO we should place the `maxNum < this._value` check in the  the first if statement in the max setter.

hanastasov (changes requested):

projects/igniteui-angular/src/lib/progressbar/progressbar.component.ts
kdinev: @hanastasov I have added some tests and they pass without this snippet. Have you tested that the logic doesn't work without it, or is it an assumption. Either something is wrong with the tests, or there's some other logic also normalizing the `_value` depending on the new `max`.

hanastasov: I tested it and it seemed to lead the progressbar in wrong state. Looking at `Value should not exceed the max limit when operating with floating numbers` test now.

hanastasov: OK, turns out what I mentioned is an edge case when max is set to 0. I have added a test that fails now and will add a fix in next commit.

hanastasov: I've added a fix for the edge case I explained. When outside the edge case, the logic in `progressValue` normalizes the `this._value`, as you assumed.

hanastasov (dismissed):

projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
```diff
@@ -45,405 +44,259 @@ describe('IgxLinearBar', () => {
...
+        fixture.detectChanges();
 
-            fix.detectChanges();
-            expect(bar.classList.contains(INDETERMINATE_CLASS)).toEqual(true);
-        });
+        expect(bar.classList.contains(INDETERMINATE_CLASS)).toEqual(true);
     });
 });
-
-@Component({ template: `<igx-linear-bar [animate]="false"></igx-linear-bar>` })
```
kdinev, line 431: @Aleksandyr Take a look at the component before - animations were still off, so nothing has changed here.

Aleksandyr: You are right.
I saw that 2 months ago I have disabled all animations when they were improved and that has been merged without noting that.
image
We need to change that because we miss several important cases when progress is updated with animation.
Especially when setting `max` and `value` runtime.

kdinev: @Aleksandyr I actually tried to add a few tests with animations on, and since we use the native browser animations, it made the tests really unstable. I think this can be left for a more "UI test" scenario, while keeping the current test suite as unit tests only.

Aleksandyr: I tried as well and it faced the same result as you. I am going to add an issue for UI animation tests

projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
```diff
@@ -45,405 +44,259 @@ describe('IgxLinearBar', () => {
...
-
         expect(progress.value).toBe(progressBarValue);
     });
 
-    it('Should update value when we try to decrese it', () => {
-        const fixture = TestBed.createComponent(LinearBarComponent);
-        fixture.detectChanges();
-        const progressBar = fixture.componentInstance.progressbar;
-        let expectedValue = 50;
+    it('when update step is bigger than passed value the progress indicator should follow the value itself', () => {
```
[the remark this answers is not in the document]
kdinev: I looked through those tests and neither tests the `increment` functionality. All of them are testing logic which is applicable even without animations like setting a larger step than max value, which should revert to the previous step value, etc. I think these tests should stay.

projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
```diff
@@ -45,405 +44,259 @@ describe('IgxLinearBar', () => {
...
-        const fix = TestBed.createComponent(InitLinearProgressBarComponent);
-        fix.detectChanges();
-
-        const progressbar = fix.componentInstance.linearBar;
-        expect(progressbar.value).toEqual(0);
-
-        progressbar.indeterminate = true;
-        progressbar.value = 30;
-        fix.detectChanges();
+    it('Should change class suffix which would be relevant to the type that has been passed', () => {
```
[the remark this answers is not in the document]
kdinev: There are getters in the code, which are host bound to the template. I can break each of them by changing the enumeration value the getters are comparing the `type` against. Seems like a unit test to me.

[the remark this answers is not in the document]
Aleksandyr: Agree

projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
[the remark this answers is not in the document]
Aleksandyr: This test should stay as you suggested.

projects/igniteui-angular/src/lib/progressbar/circularbar.component.spec.ts
```diff
@@ -11,449 +10,360 @@ const CIRCULAR_OUTER_CLASS = 'igx-circular-bar__outer';
...
+            // Should not set a step larger than max value
+            progress.step = 20;
+            expect(progress.step).toBe(step);
+            expect(progress.max).toBe(max);
+        });
 
-        expect(bar.value).toBe(expectedRes);
-        expect(bar.valueInPercent).toBe(expectedRes);
-    });
+        it(`when step value is not divisble to passed value the result returned from the value getter should be the same as the passed one`, () => {
```
Aleksandyr, line 175: I see that we test increment/decrement functionality which does not include the step property.

projects/igniteui-angular/src/lib/progressbar/linearbar.component.spec.ts
```diff
@@ -45,405 +44,259 @@ describe('IgxLinearBar', () => {
...
 
-        expectedValue = 20;
-        fixture.componentInstance.value = expectedValue;
+    it('Prevent constant update of progress value when value and max value differ', () => {
+        const maxVal = 3.25;
+        const value = 2.55;
+        const progressBar = fixture.debugElement.nativeElement;
+
+        expect(progressBar.attributes['aria-valuenow'].textContent).toBe(progress.value.toString());
+        progress.step = 0.634;
```
Aleksandyr, line 175: Passing this value doesn't affect the test. It will always be valid no matter the step.

kdinev: I have not changed this test from the previous batch.

Aleksandyr:
I thought that we might apply some changes that might improve the test suite. But as we logged another issue for [that](https://github.com/IgniteUI/igniteui-angular/issues/10965) we might apply them there.

Aleksandyr (approved):

Aleksandyr (approved):

PR #10972, opened 2022-02-08
feat(hgrid): allow to prevent exiting of edit mode when row is toggled through the UI - 13.0.x

### LINKED ISSUE #10634: Cancelable toggle event that emits before expand state changes
I am trying to disable row toggle events when I am editing content in a hierarchical grid. This causes the row to commit changes, which I need to prevent.

Any way to disable expansion state changes in the grid programaticly. 

Is it possible to implement something like this?

viktorkombov:
Closes #10634 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/api.service.ts
@@ -397,6 +397,13 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
         if (args.cancel) {
             return;
         }
+
+        const isHierarchicalGrid = grid.nativeElement.tagName.toLowerCase() === 'igx-hierarchical-grid';
+
+        if (isHierarchicalGrid) {
+            grid.hgridAPI.endEditAll();
+        }
+        
         expandedStates.set(rowID, expanded);
         grid.expansionStates = expandedStates;
         this.crudService.endEdit(false);
--- a/projects/igniteui-angular/src/lib/grids/api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/api.service.ts
@@ -398,7 +398,7 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


dkamburov (approved):

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -397,6 +397,13 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
         if (args.cancel) {
             return;
         }
+
+        const isHierarchicalGrid = grid.nativeElement.tagName.toLowerCase() === 'igx-hierarchical-grid';
```
kdinev, line 401: If we ever turn this into a web component and use a different tag name, this won't function. I suppose a different way should be figured to recognize a type of grid inside the api service from a tag name check.

kdinev (changes requested):

projects/igniteui-angular/src/lib/grids/api.service.ts
kdinev: @skrustev indeed. If the service needs to know what type of grid it is instantiated for, this can be injected as a token or input/property.

skrustev: For now one way this could be resolved even without properties is to introduce `endEdit` method call in `set_row_expansion_state` in the `gridAPI` and override it in the `hgridAPI to call `endEditAll` in it as well. 
For the other similar cases of `'igx-hierarchical-grid' can be logged separately.

kdinev: @skrustev That's what I was also thinking. It should have an override. I was looking at the code today to see the more general picture and there's a bunch of things that need to be done - `hgridAPI` and `gridAPI` being two references to one and the same thing being one of them...

PR #10998, opened 2022-02-11
Fixed cascading ESF

### LINKED ISSUE #10904: Excel Style Filter showing incorrect options
## Description  
I was working on my grid and applying some filters.
Suddenly a set of filters rendered my view blank. I had to reset my filters to load the grid data.
I investigated the excel style filter and some issues.

 * igniteui-angular version: 12.3.0
 * browser: Chrome

## Steps to reproduce  

1. Apply some filters on a column. For eg. Column A
2. Apply some filters on another column. For eg. Column B
3. Reopen the filter for Column A. It still has some options which should not be there. Applying those options renders the grid to show no data.

## Result  

https://user-images.githubusercontent.com/79090369/150990904-1b8b5a6b-8796-4807-abc3-6c50ac6c297d.mp4

## Expected result  
The filters should automatically remove invalid values.

## Attachments  
After applying the invalid filter the filter menu refreshes.
As shown in the image below the option for Blanks is not available now, but as shown in the video it was still there.
image

gedinakova:
Closes #10904  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts
```diff
@@ -550,20 +546,18 @@ export class IgxFilteringService implements OnDestroy {
...
-                }
-            } else if (condition) {
-                // create expressions tree for this field and add the new expression to it
-                const newExprTree: FilteringExpressionsTree = new FilteringExpressionsTree(filteringState.operator, fieldName);
-                newExprTree.filteringOperands.push(newExpression);
-                newExpressionsTree.filteringOperands.push(newExprTree);
+        // no expressions tree found for this field
+        if (expressionsTree) {
+            if (insertAtIndex > -1) {
+                newExpressionsTree.filteringOperands.splice(insertAtIndex, 1, expressionsTree);
```
DiyanDimitrov, line 552: Could we simplify it like this?
```suggestion
                newExpressionsTree.filteringOperands[insertAtIndex] = expressionsTree;
```

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
```diff
@@ -303,12 +303,10 @@ export class IgxExcelStyleSearchComponent implements AfterViewInit, OnDestroy {
         const searchAllBtn = this.esf.listData[0];
 
         if (!this.searchValue) {
-            const anyFiltered = this.esf.listData.some(i => i.isFiltered);
-            const anyUnfiltered = this.esf.listData.some(i => !i.isFiltered);
+            const listData = this.esf.listData.slice(1);
```
DiyanDimitrov, line 306: Is the actual fix here removing the select all item? If that is the case I would still use the `anyFiltered`/`anyUnfiltered` approach on the sliced array since it is faster than filtering the whole collection.

DiyanDimitrov (changes requested):

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-search.component.ts
gedinakova: The fix actually is to always update the indeterminate state of the `Select All` checkbox, not setting it to true only when both anyFiltered and anyUnfiltered are true. Otherwise, the state that used to be indeterminate, stays so even when on subsequent opening all items are actually selected.

DiyanDimitrov (approved):

PR #11162, opened 2022-03-04
fix(typedoc): adding to excluded paths

kdinev:
Closes #  

 - [x] Bug fix
 - [ ] New functionality
 - [x] Documentation
 - [ ] Demos
 - [x] CI/CD

### Checklist:

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -53,7 +53,7 @@ jobs:
         uses: coverallsapp/github-action@v1.1.2
         with:
             github-token: ${{ github.token }}
-      - name:
+      - name: Build TypeDoc & SassDoc
         run: |
             npm run build:typedoc:en:production
             npm run build:sassdoc:en:production
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -55,5 +55,5 @@ jobs:
             github-token: ${{ github.token }}
       - name:
         run: |
-            build:typedoc:en:production
-            build:sassdoc:en:production
+            npm run build:typedoc:en:production
+            npm run build:sassdoc:en:production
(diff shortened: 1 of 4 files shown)


.github/workflows/nodejs.yml
```diff
@@ -53,3 +53,7 @@ jobs:
         uses: coverallsapp/github-action@v1.1.2
         with:
             github-token: ${{ github.token }}
+      - name:
```
damyanpetev, line 56:
```suggestion
      - name: Build TypeDoc & SassDoc
```
Wondering if we should run this step for all node versions or just one (like coveralls above)

tsconfig.json
```diff
@@ -31,7 +31,10 @@
     "exclude": [
         "dist",
         "projects/igniteui-angular/migrations/",
-        "projects/igniteui-angular/schematics/"
+        "projects/igniteui-angular/schematics/",
+        "projects/igniteui-angular/cypress/",
+        "cypress/",
```
damyanpetev, line 36: Oh so this is the main tsconfig? Assuming cypress run would have their own tsconfig this is fine, but also kind of weird. As far as I can see typedoc has options below and those are based on this exclude list, yet there are entry point options.

So the config can be improved in the future. We're interested in API for the project lib only

damyanpetev: Actually, there's a [`tsconfig.typedoc.json`](https://github.com/IgniteUI/igniteui-angular/blob/f9db7af5e3b6b7504ecbc6ca3cb0bfee0f94fd22/tsconfig.typedoc.json) already present in the repo. What's lacking is it doesn't seem to be in use for the actual command:
https://github.com/IgniteUI/igniteui-angular/blob/f9db7af5e3b6b7504ecbc6ca3cb0bfee0f94fd22/gulpfile.js#L234-L240
Replace the `tsconfig.json` on L240 w/ `tsconfig.typedoc.json` and it should work fine

ChronosSF (approved):

damyanpetev (approved):

Issue #10205, opened 2021-09-28
IGX Datagrid autosize feature with performance improvements


BotanicalGarden:
## Is your feature request related to a problem? Please describe.
We are using IGX Grid in Angular with numerous rows and columns.  Our grids are required to have all columns autosized once data is loaded.  We do this by manually calling `column.autosize` and notice that is a performance bottleneck, sometimes taking up to 2.5 seconds to complete for a particularly large datagrid.  This frustrates the user, especially when there are numerous datagrids being displayed and autosized.

## Describe the solution you'd like
We'd like:
1. A feature that automatically has the datagrid columns autosized when data is loaded.  E.g. `@Input autosize: boolean` on the igx-grid component.
2. We require the autosize to be fairly quick and unnoticeable to the user.

## Describe alternatives you've considered
We have made our own performance improvements on the IGX datagrid by avoiding `datagrid.reflow()` from being called until all columns have been resized.  Please see our stackblitz example to compare the old way to our new way:
https://stackblitz.com/edit/github-jxvrgd-9sbr3f?file=src/app/grid/grid-resizing-sample/grid-resizing-sample.component.ts
We'd like this performance to be further improved.  We've found that the main bottleneck is happening in the `column.getAutoSize()` method.

## Additional context
Product Details
COMPONENT / igx-grid
PLATFORM / Angular
BUILD NUMBER / 10.2.24

PR #11267, opened 2022-03-22
Аuto-size of all columns on init.

MayaKirova:
Closes https://github.com/IgniteUI/igniteui-angular/issues/10205
This is a POC.

Idea for initial auto-sizing is simple:
1) Render all cells from a column with width:fit-content so that they render with the exact size needed to fit.
2) Measure the width and apply the max width as the new auto-size of the column.

Same logic is also applied to columns that are not in the initial virtualization frame (they are out of view, hence not rendered) when they get scrolled in view - they render to fit the content, get measured and then get assigned the max width of their related cells and header.

Reason to consider this:

The current auto-size method in the grid requires a full grid reflow, which is a slow operation. Also the actual measuring currently used via a Range, triggers a style recalculation for the each cell that is measured. This is marked as a potential performance bottleneck: 
image
And for a large number of cells can become slow.


rkaraivanov (approved):
LGTM

Do you have any benchmarks to share? I'm interested in seeing if this
skips the initial batch of reflows triggered by the old behavior.

damyanpetev:
This is pretty neat, certainly the native `fit-content` is way more optimized, mind you measuring the width still costs a small layout, but as far as I can tell it's pretty quick still.

@rkaraivanov  The initial load on the perf sample is barely affected with this approach - I see 4 calls to `autoSizeColumnsInView` on load (initial one + additional adjust scroll) and they're all in the range of 18-50ms.
image
Looks like a decent improvement from the old approach.

@MayaKirova Are there any concerns with the approach? Particularly the effect it can have on horizontal scroll. I also seem to see some odd bunching of columns before they shift in their place when scrolling, not sure if just me.

MayaKirova:
@damyanpetev Yes, indeed. We need to first render the content before measuring so first time you scroll a new chunk in view it will have width:fit-content hence the bunching of the cells. After measuring is done, the cells will take their max width. 
It's not ideal but can maybe be hidden with some smoke and mirrors ( maybe with some loading animation on top or something).

damyanpetev:
> [...] Which is not ideal but can maybe be hidden with some smoke and mirrors ( maybe with some loading animation on top or something). But in any case there'll always be a moment where the rendered content is not yet sized and will cause a flicker when it takes its final size. Not sure if that is something we can live with or not, but I though we might explore it as an idea regardless.

Yup, I think it's worth to check out if there's something to be done. As @rkaraivanov mentioned perhaps rAF might help as well

MayaKirova:
@kdinev @dkamburov  After a short discussion with @damyanpetev and @rkaraivanov the suggested API for enabling auto-sizing is as follows:

Allow setting a special `width` as "*" for columns that should auto-size according to content, for example:

```
        <igx-column ... width='*'>
        </igx-column>
```
Also in case there is any `minWidth` or `maxWidth` set for the column, it should adhere to those bounds (auto-size should not go below the minWidth or above the maxWidth).

Let me know if that sounds ok and I'll update the spec accordingly and apply the change.

kdinev:
@MayaKirova For `auto`sizing, I would prefer `width="auto"` over `width="*"`.

MayaKirova:
@kdinev We discussed 'auto' as an option too but @damyanpetev pointed out that across our other products we've used "*" for auto-sizing already so it might  be better to keep it consistent.

damyanpetev:
The `width="*"` seems intuitive enough (especially with prev experience with grids that use it), not bad for consistency and it's also shorter :)
I initially said it's equivalent with `auto` for me, however, `auto` is an actual valid CSS keyword (the default value for width IIRC). I don't know if that can even do much in our virtualized layout, but still - seems a bit odd to re-use it

skrustev:
There are two things I found:
1) When the column width hasn't been calculated it cannot be pinned:
    * Open the Grid Performance sample and try to pin a column 100 or something like that.
2) When there is deeper column group hierarchy something seems to break and the cells are not rendering. Not sure if is supposed to work
   * In the Column Groups sample I used the following grid config:
  
  ```
      <igx-grid [allowFiltering]="true" [moving]="true" [pinning]="pinningConfig" [rowSelection]="selectionMode" #grid [data]="data" displayDensity="compact">
        <igx-column-group header="Address Information">
            <igx-column-group header="Location">
                <igx-column [width]="'auto'" [filterable]="true" [sortable]="true" [resizable]="true" field="Country"></igx-column>
                <igx-column [width]="'auto'" [filterable]="true" [sortable]="true" [resizable]="true" field="Region"></igx-column>
                <igx-column [width]="'auto'" [filterable]="true" [sortable]="true" [resizable]="true" field="City"></igx-column>
                <igx-column [width]="'auto'" [filterable]="true" [sortable]="true" [resizable]="true" field="Address"></igx-column>
            </igx-column-group>
            <igx-column-group header="Contact Information">
                <igx-column [width]="'auto'" [filterable]="true" [sortable]="true" [resizable]="true" field="Phone"></igx-column>
                <igx-column [width]="'auto'" [filterable]="true" [sortable]="true" [resizable]="true" field="Fax"></igx-column>
                <igx-column [width]="'auto'" [filterable]="true" [sortable]="true" [resizable]="true" field="PostalCode"></igx-column>
            </igx-column-group>
        </igx-column-group>
    </igx-grid>
  ```
  
  
image

skrustev:
1) The pinning now works but there is some weird scrollbar calculation happening and the pinned area scroll thumb is bigger than expected. When the grid reflows its recalculated correctly though:

scrollbar

2) More related to the column group sample itself I think its a leftover change. We should keep the old sample in place imo.

mddragnev:
So far I tested what svetoslav did and those things seem good to me. I have managed to find one more thing:
If you open the column group sample and change the template to this
```
<igx-grid igxPreventDocumentScroll #grid
        [width]="'100%'"
        height="720px"
        [data]="data"
      >

        <igx-column-layout [pinned]="true" [header]="'ID'">
            <igx-column [rowStart]="1" [colStart]="1" width='auto' [rowEnd]="3" field="ID" [filterable]="false"></igx-column>
        </igx-column-layout>
        <igx-column-layout [pinned]="true" [header]="'Contact Details'">
            <igx-column [rowStart]="1" [colStart]="1" [colEnd]="3" field="CompanyName" [header]="'Company Name'" [sortable]="true" [width]="'350px'"></igx-column>
            <igx-column [rowStart]="2" [colStart]="1" [colEnd]="2" field="ContactName" [header]="'Contact Name'" [groupable]="true"></igx-column>
            <igx-column [rowEnd]="3" [rowStart]="2" [colStart]="2" [colEnd]="3" field="ContactTitle" [header]="'Contact Title'" [groupable]="true"></igx-column>
        </igx-column-layout>
        <igx-column-layout [header]="'Address Details'">
            <igx-column [rowStart]="1" [colStart]="1" [colEnd]="3" field="Country" [groupable]="true" [filterable]="false" [width]="'220px'"></igx-column>
            <igx-column [rowStart]="1" [colStart]="3" [colEnd]="5" field="Region" [groupable]="true" [filterable]="false" [width]="'220px'"></igx-column>
            <igx-column [rowStart]="1" [colStart]="5" [colEnd]="7" field="PostalCode" [header]="'Postal Code'" [groupable]="true" [filterable]="false" [width]="'220px'"></igx-column>
            <igx-column [rowStart]="2" [colStart]="1" [colEnd]="4" field="City" [groupable]="true" [filterable]="false"></igx-column>
            <igx-column [rowStart]="2" [colStart]="4" [colEnd]="7" field="Address" [filterable]="false"></igx-column>
        </igx-column-layout>
        <igx-column-layout [header]="'Phone Details'">
                <igx-column [rowStart]="1" [colStart]="1" [colEnd]="2" field="Phone" [filterable]="false" [width]="'220px'"></igx-column>
                <igx-column [rowStart]="2" [colStart]="1" [colEnd]="2" field="Fax" [filterable]="false"></igx-column>
            </igx-column-layout>
    </igx-grid>
```
You can observe that there is an error in the console - stackoverflow.

MayaKirova:
@mddragnev Should be fixed now.

mddragnev:
Another thing I noticed in hierarchical grid. If you go to the dev sample of hgrid and add `[width]="'auto'"` to the first column of the first rowisland you can observe that the cells are missaligned
Screenshot 2022-07-20 at 16 11 40
And so are each and every child grids if any of their columns is set to`[width]="'auto'"`

dkamburov:
@MayaKirova Bigger cell view outside of the current grid view:
big-content-outside-view
Doesn't seem to reapply the autosizing for these values. (All columns are with `width: 'auto'`)

dkamburov:
What about exposing `autoSizeColumnsInView`? There might be scenarios where data or values change and the users would to trigger autosizing of the columns without recreating the whole grid.
content-change

dkamburov (changes requested):
Comments in the conversation

MayaKirova:
@dkamburov Currently the column size is calculated only once on initialization so any changes that happen post initialization like editing a cell value or scrolling down to another chunk of data will not recalculate the sizes. We could also cover such scenarios but it would require additional handling, such as measuring and caching all cell sizes per column. Not sure if this is within the scope of this task, as the previous column auto-size was also limited to the currently visible cells as per the [docs](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-resizing#auto-size-columns-on-double-click). 

Regarding exposing the `autoSizeColumnsInView` method.

In the scenario where the user edits a single cell, it might be better to autosize just the related column, not all columns. So the existing autoSize method on the column itself can be used on some event after edit cell is done. That should not have a large performance hit since just 1 column will be auto-sized.

We could expose it for the scrolling down scenario, to use on an event like the `gridScroll` but since currently we have no caching, on every scroll all cells will have to be rendered once with 'fit-content' width and measured, which will impact performance. So not sure if it is worth it without caching. Also not sure if changing the column widths while scrolling down the rows will be good from an UX perspective, it might look strange for the columns to change sizes while you scroll.

dkamburov (approved):

rkaraivanov (approved):

PR #11319, opened 2022-03-31
themes(fluent): add new fluent theme

### LINKED ISSUE #11302: Make themes consistent
I noticed that the fluent word palette is not using the correct color #2b579a but uses #0078d4. We decided to introduce a fluent default palette with the #0078d4 and use it for both light and dark schemas, preserving the excel palette and preceding a correct  #2b579a, but having the new one as a default for both light and dark schemas, you can also find all color definitions in [this table](https://infragistics.visualstudio.com/Design-System-Model/_wiki/wikis/Design-System-Model.wiki/562/Palettes)

Please check all other palettes against the table in case I have missed anything when reviewing them.

didimmova:
add a new fluent theme, update material, and fluent-word themes

Closes #11302 

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@ All notable changes for each version of this project will be documented in this 
 - `IgxToast`
     - **Breaking Changes** - The toast deprecated `position` property have been removed. We suggest using `positionSettings` property instead.
 
+### New palette
+
+A new fluent light and dark palettes that use the default fluent colors - `$light-fluent-palette` and `$dark-fluent-palette`.
+
 ## 13.0.5
 
 ### New Features
--- a/projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
@@ -97,19 +97,11 @@ $dark-palette: $default-dark-palette;
 /// @see $light-palette
 $light-material-palette: $light-palette;
 
-/// Dark material palette
-/// @type Map
-/// @prop {Color} primary [#09f] - The 500 variant of the `primary` color (default).
-/// @prop {Color} secondary [#72da67] - The 500 variant of the `secondary` color (default).
-/// @prop {Color} grays [#fff] - The base color for  the `grays` shades.
-/// @prop {Color} surface [#222] - The color used as a background in components, such as cards, sheets, and menus.
-/// @access private
-$dark-material-palette: palette(
-    $primary: #09f,
-    $secondary: #72da67,
-    $grays: #fff,
-    $surface: #222,
-);
+/// An alias of $dark-palette
+/// @group palettes
+/// @alias dark-palette
+/// @see $dark-palette
+$dark-material-palette: $default-dark-palette;
 
 /// @access private
 $green-palette: palette(
@@ -200,7 +192,7 @@ $fluent-excel-palette: palette(
     $primary: #217346,
     $secondary: #217346,
     $success: #107c10,
-    $warn: #797673, 
+    $warn: #797673,
     $error: #a80000,
     $variant: 'fluent'
 ) !default;
--- a/projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
@@ -317,7 +317,7 @@ $fluent-dark-palette: palette(
     $variant: 'fluent'
 ) !default;
 
-/// Light Fluent Word palette
+/// Light Fluent palette
 /// @type Map
 /// @prop {Color} primary [#0078d4] - The 500 variant of the `primary` color (default).
 /// @prop {Color} secondary [#0078d4] - The 500 variant of the `secondary` color (default).
@@ -331,7 +331,7 @@ $fluent-dark-palette: palette(
 /// @group palettes
 $light-fluent-palette: $fluent-palette;
 
-/// Dark Fluent Word palette
+/// Dark Fluent palette
 /// @type Map
 /// @prop {Color} primary [#0078d4] - The 500 variant of the `primary` color (default).
 /// @prop {Color} secondary [#0078d4] - The 500 variant of the `secondary` color (default).
--- a/projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
@@ -56,7 +56,7 @@ $default-palette: palette($primary: #09f, $secondary: #e41c77) !default;
 /// @access private
 $default-dark-palette: palette(
     $primary: color($default-palette, 'primary'),
-    $secondary: #72da67,
+    $secondary: color($default-palette, 'secondary'),
     $grays: #fff,
     $surface: #222,
 ) !default;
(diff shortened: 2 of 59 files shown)


StefanIvanov:
@didimmova can you also update the info color for all three fluent theme presets?

projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
```diff
@@ -287,6 +287,56 @@ $light-fluent-word-palette: $fluent-word-palette;
...
+    $secondary: color($fluent-palette, 'secondary'),
+    $success: color($fluent-palette, 'success'),
+    $warn: color($fluent-palette, 'warn'),
+    $error: color($fluent-palette, 'error'),
+    $surface: #222,
+    $grays: #fff,
+    $variant: 'fluent'
+) !default;
+
+/// Light Fluent Word palette
```
StefanIvanov, line 312:
```suggestion
/// Light Fluent palette
```

projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
```diff
@@ -287,6 +287,56 @@ $light-fluent-word-palette: $fluent-word-palette;
...
+/// @prop {Color} surface [#fff] - The color used as a background in components, such as cards, sheets, and menus.
+/// @prop {Color} info [#1377d5] - The `info` color. Default for every palette if not specified.
+/// @prop {Color} success [#107c10] - The `success` color. Default for every palette if not specified.
+/// @prop {Color} warn [#797673] - The `warn` color. Default for every palette if not specified.
+/// @prop {Color} error [#a80000] - The `error`. Default for every palette if not specified.
+/// @access public
+/// @group palettes
+$light-fluent-palette: $fluent-palette;
+
+/// Dark Fluent Word palette
```
StefanIvanov, line 326:
```suggestion
/// Dark Fluent palette
```

didimmova:
> @didimmova can you also update the info color for all three fluent theme presets?

It's not explicitly set because it uses the same color as the default palette and it's inherited.

StefanIvanov:
> > @didimmova can you also update the info color for all three fluent theme presets?
> 
> It's not explicitly set because it uses the same color as the default palette and it's inherited.

OK for now let's leave it as is

StefanIvanov (dismissed):
This solves the problems I reported, thanks

projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
```diff
@@ -56,7 +56,7 @@ $default-palette: palette($primary: #09f, $secondary: #e41c77) !default;
 /// @access private
 $default-dark-palette: palette(
     $primary: color($default-palette, 'primary'),
-    $secondary: color($default-palette, 'secondary'),
+    $secondary: #72da67,
```
simeonoff, line 59: Why did we change this color in the default palette?

simeonoff (changes requested):

projects/igniteui-angular/src/lib/core/styles/themes/_palettes.scss
didimmova: Because material uses the default palette and the color in the table is different from this one. All the other themes override it. If you want I can change it just for material?

StefanIvanov (approved):

CHANGELOG.md
```diff
@@ -46,6 +46,10 @@ All notable changes for each version of this project will be documented in this
 - `IgxToast`
     - **Breaking Changes** - The toast deprecated `position` property have been removed. We suggest using `positionSettings` property instead.
 
+### New palette
+
+A new fluent light and dark palettes that use the default fluent colors - `$light-fluent-palette` and `$dark-fluent-palette`.
+
```
simeonoff, line 52: I think we should target this to be a part of the 13.2.x release, i.e. change the base to master.

simeonoff (commented):
Change target to master(13.2.x).

StefanIvanov:
> @StefanIvanov this PR changes the dark material palette to look identical to the dark green palette, why? Doesn't make much sense to me.

@simeonoff by definition the material dark theme was using green and blue, so I am not sure why we had that approach initially, in any case I think that the correct approach would be to merge these into one i.e. mat dark that comes with the green color and maybe remove the dark green palette as such

simeonoff:
> > @StefanIvanov this PR changes the dark material palette to look identical to the dark green palette, why? Doesn't make much sense to me.
> 
> @simeonoff by definition the material dark theme was using green and blue, so I am not sure why we had that approach initially, in any case I think that the correct approach would be to merge these into one i.e. mat dark that comes with the green color and maybe remove the dark green palette as such

I must disagree. Using the dark green palette as the default in the dark material version of the theming widget was the mistake, not that we have two different palettes. I don't think we should be changing anything in the dark material palette, nor that we should be deleting presets or palettes.

StefanIvanov:
> > > @StefanIvanov this PR changes the dark material palette to look identical to the dark green palette, why? Doesn't make much sense to me.
> > 
> > 
> > @simeonoff by definition the material dark theme was using green and blue, so I am not sure why we had that approach initially, in any case I think that the correct approach would be to merge these into one i.e. mat dark that comes with the green color and maybe remove the dark green palette as such
> 
> I must disagree. Using the dark green palette as the default in the dark material version of the theming widget was the mistake, not that we have two different palettes. I don't think we should be changing anything in the dark material palette, nor that we should be deleting presets or palettes.

I see your point and had our approach been otherwise as we laid the foundation of the dark material theme, I would have agreed. When the task to design a dark material palette a few years ago was active, its outcome was a dark palette with secondary color that was green unlike the secondary color used in the light theme. Therefore as we take that design and output document as the definition for what the dark material theme was defined to be it seems to me that implementation did not follow definition and I don't see how what we did in the widget and plugin was a mistake.

simeonoff:
> When the task to design a dark material palette a few years ago was active, its outcome was a dark palette with secondary color that was green unlike the secondary color used in the light theme.

This is absolutely not true. You can go back to the inception of the dark green theme preset (6.2.1) and you will see in the code  that it has always been distributed as a separate preset and palette and definitely not as 'material dark'. The `igx-dark-theme` mixin never assumed any predefined colors or palettes. We have released the official `$material-light-palette` and `$material-dark-palette` in 11.0.0 (almost 2 years ago). In this version the material dark palette was identical to the light one with the exception of the gray and surface colors.

> Therefore as we take that design and output document as the definition for what the dark material theme was defined to be it seems to me that implementation did not follow definition and I don't see how what we did in the widget and plugin was a mistake.

If you are talking about the color definitions as part of the DSM repo, those are being defined as we speak, while the all dark palettes have been around for almost 2 years. Actually, the theming widget showing the dark green palette as the dark material theme was admittedly a mistake by the grinders team who decided to name the dark green preset "dark material" for some reason.

simeonoff:
> > > When the task to design a dark material palette a few years ago was active, its outcome was a dark palette with secondary color that was green unlike the secondary color used in the light theme.
> > 
> > 
> > This is absolutely not true. You can go back to the inception of the dark green theme preset (6.2.1) and you will see in the code that it has always been distributed as a separate preset and palette and definitely not as 'material dark'. The `igx-dark-theme` mixin never assumed any predefined colors or palettes. We have released the official `$material-light-palette` and `$material-dark-palette` in 11.0.0 (almost 2 years ago). In this version the material dark palette was identical to the light one with the exception of the gray and surface colors.
> > > Therefore as we take that design and output document as the definition for what the dark material theme was defined to be it seems to me that implementation did not follow definition and I don't see how what we did in the widget and plugin was a mistake.
> > 
> > 
> > If you are talking about the color definitions as part of the DSM repo, those are being defined as we speak, while all dark palettes have been around for almost 2 years. Actually, the theming widget showing the dark green palette as the dark material theme was admittedly a mistake by the grinders team who decided to name the dark green preset "dark material" for some reason.
> > I hope you can see why I can't just merge this PR with the change it introduces to the `$dark-material-palette`. We should not be doing anything to it. We already expose this palette, it's called `$dark-green-palette` and it's a modification of the `$green-palette` introduced in 6.2.1.
> 
> I was talking about the design of the dark theme done in Sketch 3-4 years ago by @sdimchevski that defined a different secondary color so my point is that we made the mistake thereafter by not respecting that and introducing not just a material dark theme with green but a dark theme with no change and another theme with green. Now if both of you can find an agreement about what is right, we have to make sure that this truth is applied across the board where there are inconsistencies.

I think I already expressed my opinion - we shouldn't be introducing such dramatic breaking changes to the included palettes, nor should we be making two palettes that have different names the same in terms of colors. It's confusing for the end users of the theming system and it also makes things inconsistent compared to the rest of the included palettes and presets where all primary and secondary colors are identical in both the dark and light variants. There was never a material dark theme 3-4 years ago in any sketch file that I know of, designed by @sdimchevski, myself or Spas back then. We didn't really have an official UI library besides the one Spas and I created and it did not contain anything related to a dark palette, only the default light palette was covered in that sketch file.

PR #11388, opened 2022-04-13
refactor(themes): use logical properties

didimmova:
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/list/_list-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/list/_list-theme.scss
@@ -419,6 +419,12 @@
         igx-icon {
             color: var-get($theme, 'item-action-color')
         }
+
+        [dir='rtl'] & {
+            igx-icon {
+                transform: scaleX(-1);
+            }
+        }
     }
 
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/core/styles/components/list/_list-theme.scss
```diff
@@ -457,11 +454,11 @@
         color: var-get($theme, 'item-thumbnail-color');
 
         > igx-icon {
-            @include if-ltr() {
+            [dir='ltr'] & {
```
simeonoff, line 457: This will and doesn't work if the `dir` attribute is not explicitly set, which it is not by default and requiring the users to explicitly be setting direction is a breaking change in terms of behavior, plus we shouldn't make them explicitly set the direction when LTR is used. Additionally, I would prefer to switch to logical properties for padding, instead of using the `padding` shorthand

simeonoff (changes requested):
List theme needs more attention.

simeonoff (approved):

PR #11441, opened 2022-04-19
fix(igxGrid): Add handling for remote virtualization navigation.

### LINKED ISSUE #11416: igx-grid virtualization - Tabbing not working in cell edit mode when past rowView
## Description  
I have a problem where if I'm editing a cell in a grid with virtualization, if I press the tab key when I'm at row 21 or greater, the next cell doesn't start editing and it exits edit mode.

 * igniteui-angular version: 13.1.9
 * browser: Chrome

## Steps to reproduce  

1. [Stackblitz](https://stackblitz.com/edit/angular-euxhf7-shluaz?file=src/app/grid/grid-sample-4/grid-sample-4.component.html). This stackblitz was taken from the [remote virtualization demo](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/remote-data-operations#remote-virtualization-demo), and modified to set `batchEditing` to true, and a couple of the columns to `editable`
2. Start editing any cell in row 21 or greater (you don't need to make any modifications, just enter edit mode)
3. Press the tab key

## Result  
The exits edit mode and doesn't move to the next cell.

## Expected result  
The grid should not exit edit mode, and the next editable cell should become active

## Extra info
I dug into the code and I see where the problem happens. I think the grid's `getNextCell` function is the one that's having the problem

MayaKirova:
Closes #11416 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
@@ -7207,7 +7207,7 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
         // find next/prev record that is editable.
         const nextRowIndex = previous ? this.findPrevEditableDataRowIndex(currentRowIndex) :
             this.dataView.findIndex((rec, index) =>
-                index > resolvedIndex && this.isEditableDataRecordAtIndex(resolvedIndex));
+                index > resolvedIndex && this.isEditableDataRecordAtIndex(index));
         const nextDataIndex = this.getDataIndex(nextRowIndex);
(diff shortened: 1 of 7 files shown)


mddragnev:
@MayaKirova,
Having the same configuration within hierarchicalGridRemoteVirtualization sample, it throws an error which is not really related to the fix you are applying but it does not work because of that. I am attaching a screenshot of the method and the fix I suggest. It is located in hierarchical-grid.compononet.ts. IMO, we should just check whether the record exists.
Screenshot 2022-04-20 at 13 17 00

skrustev (approved):

PR #11686, opened 2022-06-07
IgxTreeGrid - emit rowDelete and rowDeleted events

### LINKED ISSUE #10940: igx-tree-grid rowDelete and rowDeleted events are not emitted
## Description  
When deleting a row through the API method or through the action-strip delete button neither **rowDelete**, nor **rowDeleted** events are emitted. However, when using **igx-grid** or **igx-hierarchical-grid** the events are emitted and this leads to inconsistency.

 * igniteui-angular version: 13.0.x, 12.3.x
 * browser: all

## Steps to reproduce  

1. Define a tree grid with action-strip and rowDelete and rowDeleted events set. 
2. Log 'delete' on the console when the events are emitted.
3. Delete a row through the action-strip delete button or through the API.

## Result  
Event is not emitted and the console does not log anything.

## Expected result  
The event should be emitted and the console should log 'delete'.

MonikaKirkova:
Closes #10940 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
```diff
@@ -720,8 +720,23 @@ export class IgxTreeGridComponent extends IgxGridBaseDirective implements GridTy
...
+            oldValue: null
+        };
+        this.rowDelete.emit(args);
+        if (args.cancel) {
+            return;
+        }
 
+        const record = this.gridAPI.deleteRowById(rowId);
+        if (record !== null && record !== undefined) {
+            //  TODO: should we emit this when cascadeOnDelete is true for each row?!?!
```
ChronosSF, line 736: We shouldn't merge changes with TODOs

I'd say we should probably fire the event for all rows when cascading deletes them. @damyanpetev , what do you think?

ChronosSF (changes requested):

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid.component.ts
damyanpetev: I think it'll be a bit excessive firing a potentially pretty large amount of events for a single user interaction. For the most part I'm assuming the cascade delete reflects the default behavior of the data source (backend/DB) and the root record deletion is enough information considering the user is the one that configured the cascade so it shouldn't be a surprise.
What could be useful actually, since `rowData` may not be hierarchical is to also provide the public row interface in the arguments - in this case the `IgxTreeGridRow` which will also give access to the `children` should the data not contain those. So even if the cascade is only in the client grid, it shouldn't be much of a hassle to grab all the children

damyanpetev: So after a quick discussion, we've settled on this: The delete event is fired once delete on flat data using primary and foreign keys - that keeps it consistent with the equivalent interaction on actual hierarchical data and also with cascade on (default) the presumption is the client grid is mimicking the behavior of the backend/DB. Should there be need to access the child structure under a deleted row - my initial suggestion to add the row interface to all events was a bit off, as after a delete the grid APIs no longer make sense and the event interfaces won't accommodate that easily ATM which is something we might tackle on down the road.
The children should still be available in the row data (if hierarchical) or should be accessible based on the row key from the grid API (flat data). The latter applies similarly to access an added row and its parent structure.

damyanpetev (commented):
Some minor comments, otherwise LGTM

projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -477,8 +477,8 @@ export class IgxRowAddCrudState extends IgxRowCrudState {
         if (isAddRow) {
             this.endAddRow();
             if (commit) {
-                this.grid.rowAddedNotifier.next({ data: args.newValue });
-                this.grid.rowAdded.emit({ data: args.newValue });
+                this.grid.rowAddedNotifier.next({ data: args.newValue, owner: this.grid });
```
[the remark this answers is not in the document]
teodosiah: Both rowAddedNotifier and rowDeletedNotifier use IRowDataEventArgs, so the owner is required:
image

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.spec.ts
```diff
@@ -291,12 +291,17 @@ describe('igxGridEditingActions #grid ', () => {
...
+            };
+
             // select delete
-            addChildBtn.actionClick.emit();
+            deleteChildBtn.actionClick.emit();
             fixture.detectChanges();
 
-            expect(treeGrid.rowDelete.emit).toHaveBeenCalledWith(rowDeleteArgs);
-            expect(treeGrid.rowDeleted.emit).toHaveBeenCalled();
+            expect(treeGrid.rowDelete.emit).toHaveBeenCalledOnceWith(rowDeleteArgs);
```
damyanpetev, line 303: nice :)

damyanpetev (approved):

ChronosSF (approved):

Issue #11298, opened 2022-03-28
Using *ngIf in Master-Detail resets grid's state


RivaIvanova:
## Description  
When using ***ngIf** directive in the Master-Detail template, the grid's state gets reset when expanding the rows.

 * igniteui-angular version: 13.1.x, 12.3.x
 * browser: all

## Steps to reproduce  

1. Open this [sample](https://stackblitz.com/edit/angular-em3xzv-kmc1rq?file=src%2Fapp%2Fgrid%2Figx-sample%2Figx-sample.component.html).
2. Apply filtering/sorting/moving/etc. on any column
3. Save the grid's state and restore it.
4. Expand the first row (state gets reset).
5. Restore state again.
6. Expand the first row again (state remains).
7. Expand any of the other rows (state gets reset again).

## Result  
The grid's state gets reset when expanding the rows.

## Expected result  
The state should be preserved.

## Attachments  
This behavior could be observed in the attachment below:
![state-issue-demo](https://user-images.githubusercontent.com/49126110/160392101-d33c3aed-be1f-43a7-a0cd-607d40169bab.gif)

MayaKirova:
@dkamburov 
In general the problem is that the *ngIf triggers all ContentChildren in the grid to be re-calculated at the time that the ngIf template is resolved (so in this case when the details template is expanded).

Since the columnList collection is ContentChildren it gets re-calculated and changed to the original columns as defined in the template, which are different object instances than the ones that the state directive applies. Hence all generated columns from the state directive are considered removed and their feature's state is cleared - https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts#L6370

This can be worked around but in general it would be best to not modify and rely on a ContentChildren collection to persist anything related to the grid state, since that collection changes in many instances out of our control (based on angular's internal logic) and any previous manual changes may be lost.

Issue #11116, opened 2022-02-28
When moving a templated column in grid it does not stay in its new position.


MayaKirova:
## Description  
When a column that has a cell template is moved, on mouse move it goes back to its original position.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open the following sample: https://www.infragistics.com/angular-demos/grid/grid-moving-sample
2. Try moving the "Change" column.

![moving](https://user-images.githubusercontent.com/10397980/156003477-a179c85c-b6a0-4ff1-a59b-a9bfebc42fda.gif)

## Result  
It seems to move, but then it goes back to its old position.

## Expected result  
It should be moved as expected.

## Attachments  
Attach a sample if available, and screenshots, if applicable.

Issue #7681, opened 2020-06-25
When use grid.columns getter the error is thrown in the console


ddincheva:
## Description  

 * browser: all

## Steps to reproduce  

1. Open the following [sample](https://stackblitz.com/edit/angular-pskbsu)
2. Open the browser console

## Result  

![image](https://user-images.githubusercontent.com/34240583/85696990-163a3900-b6e2-11ea-9b75-26ebcffaae32.png)

## Expected result  

No errors should be thrown in the console

hanastasov:
Let's try to fix it in following way:
- keep `columns` prop, drop new `columnsCollection`, drop migrations
- use `columnList` internally

hanastasov:
Resolved in version 13. Will not fix for 12.3.x as it turns out to be requiring a lot of changes in code and tests.

PR #11701, opened 2022-06-09
fix(igxGrid): Re-work fetaures that mutate or use the columnsList Con…

### LINKED ISSUE #11493: Ordering grid columns based on the order of ngFor array
## Description  
When changing the order of columns in the view, using an *ngFor directive, the columns in the grid itself aren't reordered.

 * igniteui-angular version: 12.3.6
 * browser: Chrome

## Steps to reproduce  
See this stackblitz:
https://stackblitz.com/edit/angular-bejvgc-pxqlom

Clicking "reverse" will reverse the column order, and you can see the changes in the field display, but they aren't seen in the grid.

## Result  
Columns don't move.

## Expected result  
Columns move based on the updated array order.

MayaKirova:
…tentChildren collection. Use internal state collections instead.

Closes #11493  

**Note**: When testing re-test all recent columnList related issues:
https://github.com/IgniteUI/igniteui-angular/issues/11116
https://github.com/IgniteUI/igniteui-angular/issues/11298 

Also re-test the original issue for which the collumnList collection was used:
https://github.com/IgniteUI/igniteui-angular/issues/7681

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


rkaraivanov (approved):

MayaKirova:
@dkamburov `columnsList` still exists and will be populated as before in cases where the columns are defined in the grid's template. So not sure if we would need a migration for the samples, they mostly define their columns in the template so nothing should change for them.

In general I'm not sure if we need to migrate something that is not a public API. `columns` is public and should be the one customers use anyway.

If you feel like we should still write migrations for this, I'm ok with that, but since it will need to be a custom migration it will probably take a few more day to write and test for all versions.

dkamburov (approved):

mddragnev:
@MayaKirova This issue https://github.com/IgniteUI/igniteui-angular/issues/7681 is not reproducible in master but is still reproducible in your branch. Could you take a look at it, please?

mddragnev:
@MayaKirova Sounds fair enough to me

PR #11799, opened 2022-06-30
fix(pivot): Set correct data type when creating custom IgxPivotDateDi…

### LINKED ISSUE #11738: [PivotGrid] Date Sorting
## Question  

For date type dimensions, we are using IgxPivotDateDimension with everything on false in terms of options apart from "fullDate", which is on true. We've noticed that if we use "memberFunction" and try to sort, it seems it will do a string sorting rather than a date one. Without "memberFunction", the sorting is done okay. Should the method return a different type than string?

 * igniteui-angular version: 13.2.1
 * browser: Chrome

date-sorting

mddragnev:
…mension

Closes #11738   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-dimensions.ts
```diff
@@ -46,6 +47,11 @@ export interface IPivotDateDimensionOptions {
 export class IgxPivotDateDimension implements IPivotDimension {
     /** Enables/Disables a particular dimension from pivot structure. */
     public enabled = true;
+    
+    /**
+     * Gets/Sets data type
+     */
+    public dataType?: GridColumnDataType;
```
MayaKirova, line 54: I think for the date dimensions we should set the column type to be Date by default, so that the user does not have to explicitly set it every time. In most cases you would always use this for date values anyway.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-dimensions.ts
```diff
@@ -94,6 +100,8 @@ export class IgxPivotDateDimension implements IPivotDimension {
             return;
         }
 
+        this.dataType = inBaseDimension.dataType;
```
MayaKirova, line 103: Shouldn't the child dimensions below also apply a data type? Otherwise I assume they won't be sorted as dates but as strings as well.

MayaKirova (changes requested):

MayaKirova (dismissed):

mddragnev:
@dkamburov please note that this pr should be tested after https://github.com/IgniteUI/igniteui-angular/pull/11820 this is merged.

skrustev (approved):

MayaKirova (approved):

PR #11841, opened 2022-07-08
Expose combo filtering strategy

### LINKED ISSUE #9884: IgxCombo: ability to filter multiple fields
If you have a IgxCombo with a complex object you can only search one property.

see https://www.infragistics.com/products/ignite-ui-angular/angular/components/combo-templates#angular-combo-templates-example

In the example Combo the fields 'state' and 'region' are displayed but you can only search for 'state'.

How about adding a property `searchFields` where you can specify all fields to search like `['state', 'region']`.

An alternatve solution would be to let the programmer customize the search with `[customFilter]="true"` and a `(filter)` event.

### LINKED ISSUE #11532: Add 'searchKey' to Combo
Please add a "searchKey" input to the IGX Combo component which tells the combo which data key to use to search/filter by. Right now the search automatically uses the "displayKey" to search and filter the drop down options. However, if you are using custom templates for displaying drop down items in the combo, you might have additional info you'd like to include in the search. By providing a "searchKey" (which defaults to the displayKey value) you can have the search filter by a different data key value which might have more information.

## Additional context

Example: I have combo using an array of employee data. The employee data array consists of objects with id, name, division, name_and_division. The "displayKey" is set to "name". I have a custom template to display which prefixes the employee's division to their name (i.e. "Corporate: John Doe"). I'd like to be able to type "corporate" in the search field and have it also match employees at corporate by searching the "name_and_division" property values instead of the just the "name" property. Then when I select an employee, just their "name" appears in the input field box and not their "name and division".

### LINKED ISSUE #11810: IgxCombo's filtering strategy is not exposed limiting extensibility
Filtering for combo is entirely internal and implemented in a pipe. Instead, to allow extension points opening a slew of additional use-cases, filtering should be exposed as a strategy to users similar to igxGrid.

n/a

n/a

## Additional context
n/a

PlamenaMiteva:
Closes #9884
Closes #11532
Closes #11810  

 - [ ] Bug fix
 - [X] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -42,6 +42,38 @@ export interface IComboFilteringOptions {
...
+                e[displayKey]?.toString().toLowerCase().includes(searchTerm));
+        } else {
+            return data.filter(e => filteringOptions.caseSensitive ? e.includes(searchTerm) :
+                e.toString().toLowerCase().includes(searchTerm));
+        }
+    }
+}
+
+/** Filters a collection by both display and value key */
+export class AdvancedFilteringStrategy implements IComboFilteringStrategy {
```
kdinev, line 64: This could be used for unit tests but should not be exported by the library.

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -42,6 +42,38 @@ export interface IComboFilteringOptions {
     caseSensitive: boolean;
 }
 
+export interface IComboFilteringStrategy {
```
kdinev, line 45: Do we need to export an interface and expose the strategy through a wrapping interface?

projects/igniteui-angular/src/lib/combo/combo.common.ts
```diff
@@ -386,6 +389,24 @@ export abstract class IgxComboBaseDirective extends DisplayDensityBase implement
...
+    /**
+     * Gets/Sets the filtering strategy of the combo.
+     *
+     * @example
+     * ```html
+     *  <igx-comb #combo [data]="localData" [filterStrategy]="filterStrategy"></igx-combo>
+     * ```
+     */
+     @Input()
+     public get filterStrategy(): IComboFilteringStrategy {
```
kdinev, line 402: The strategy should be of type `(args...) => returnType`

projects/igniteui-angular/src/lib/combo/combo.component.ts
Lipata: It seems like an unnecessary wrapper, knowing the fact that it contains only one method. Two reasons for having it in that way were to be consistent with the grid, and the other is to be extensible for the future, without introducing a breaking change. But a method is also extensible, knowing the fact that you can add an additional parameter, and we can choose simplicity over consistency with the grid. @wnvko will proceed with exposing a function, @kdinev we just need your approval for its name.

Lipata (approved):

Lipata (approved):

PR #11853, opened 2022-07-14
[igxPivot] Integration with igxGridState directive to store config.

### LINKED ISSUE #11741: [PivotGrid] State Persistence
Similar to Data Grid, would be possible for Pivot Grid to also have the option for setting and retrieving its state object?

MayaKirova:
Closes #11741

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@hanastasov @ChronosSF @dkamburov  I've updated the specification of the [Grid State Persistence](https://github.com/IgniteUI/igniteui-angular/wiki/Grid-State-Persistence) with information on persisting a PivotGrid state. I've also updated the known issues. Please review it when you have the chance.

hanastasov:
> @hanastasov @ChronosSF @dkamburov I've updated the specification of the [Grid State Persistence](https://github.com/IgniteUI/igniteui-angular/wiki/Grid-State-Persistence) with information on persisting a PivotGrid state. I've also updated the known issues. Please review it when you have the chance.

I did, IMO is written well and clear - it made sense to me even though I have not worked with the pivot yet.

hanastasov (dismissed):

mddragnev:
@MayaKirova Everything looks perfect to me except that  I thought of noop pivot grid. What are we doing with this. Probably most of the customers would use some strategy that extends `NoopPivotDimensionsStrategy`. Should we expose one more event where we can set both strategies because they are not persisted in the current configuration state? Have you thought anything about that or we just missed it while talking about pivot state feature?

MayaKirova:
@mddragnev @dkamburov I don't think we have discussed persisting/re-setting custom strategies. Not sure if there's such a mechanism in the base grid for custom sort/filter strategies either. @hanastasov is there any way in the base grid to set back a custom `filterStrategy`/`sortStrategy` after restoring a previous state from a string? If not then maybe we should add some event or events for strategies since for fully remote scenarios all strategies will need to be set again.

dkamburov:
@MayaKirova @mddragnev @hanastasov Let's do a meeting about it and discuss restoring the strategies next sprint.

hanastasov:
> @mddragnev @dkamburov I don't think we have discussed persisting/re-setting custom strategies. Not sure if there's such a mechanism in the base grid for custom sort/filter strategies either. @hanastasov is there any way in the base grid to set back a custom `filterStrategy`/`sortStrategy` after restoring a previous state from a string? If not then maybe we should add some event or events for strategies since for fully remote scenarios all strategies will need to be set again.

Strategies are class references so fall under the limitations I believe and need to be set with application code. Will try to prepare an example for the meeting.

MayaKirova:
@mddragnev @dkamburov @hanastasov I have added a new event for the igxGridState directive that can be used to modify the state before applying it when it is passed as a string. Changelog and tests are also updated.

mddragnev:
@MayaKirova 
The new approach with the strategies looks very good to me. One thing I noticed in the noop pivot sample is that if we run the exact sample you have pushed (with stringified state) - saving the initial state of the grid, removing all row dimensions and then restoring the state works fine. Once I decide to use `IGridState` on application level - create a public state variable and set it like this in the `saveState` method - `this.state1 = this.state.getState(false);` and then when restoring having this logic to restore the state
```
this.state1.sorting.forEach(x => x.strategy = NoopSortingStrategy.instance());
  this.state1.pivotConfiguration.rowStrategy = NoopPivotDimensionsStrategy.instance();
  this.state1.pivotConfiguration.columnStrategy = NoopPivotDimensionsStrategy.instance();
  this.state.setState(this.state1);
```

The same steps from above does not lead to restoring the rows.

MayaKirova:
@mddragnev Issue should be fixed now.

hanastasov (approved):

dkamburov (approved):

PR #11912, opened 2022-07-28
fix(tree): Make the tree scroll automatically when it reaches its parent height.

### LINKED ISSUE #11848: igxTree overflow auto
## Description  
Can we consider adding overflow-y and overflow-x to the Tree component? This decision can be made based on the question - is it already present in the WC Tree.
 - If it is, we should add it for the Angular tree as well. This will bring a better consistency as platforms like the AB are relying on this to produce quality code without additional CSS changes depending on the picked framework for code generation. Currently, If overflow is present in WC and missing from Angular, upon code generation this CSS Property should be manually added.
 - If it is not present in both, should we add it or we can "workaround" this on App builder level and export the CSS props based on design surface configuration?

**Example:**
image

desig9stein:
Closes #11848

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/tree/_tree-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/tree/_tree-theme.scss
@@ -127,6 +127,8 @@
     %tree-display {
         display: block;
         z-index: 0;
+        height: 100%;
+        overflow-y: auto;
     }
 
     %tree-node,
--- a/src/app/tree/tree.sample.scss
+++ b/src/app/tree/tree.sample.scss
@@ -14,7 +14,6 @@
 
 .sample-scroll {
     max-height: 300px;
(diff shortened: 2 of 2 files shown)


zdrawku (approved):

projects/igniteui-angular/src/lib/core/styles/components/tree/_tree-theme.scss
```diff
@@ -127,6 +127,8 @@
     %tree-display {
         display: block;
         z-index: 0;
+        height: 100%;
```
zdrawku, line 130: @desig9stein we've reconsidered the whole approach. Could you please make an additional change and remove the whole height definition. Let the change be only adding `overflow-y: auto`

Issue #6501, opened 2020-01-21
Validator service for igxGrid editing


kdinev:
## Is your feature request related to a problem? Please describe.
As a developer, I would like to be able to reuse some or all of my forms validation logic with the `igx-grid` editing.

## Describe the solution you'd like
Currently, if I want to validate the igx-grid row or cell editing and display different validation messages, stop user navigation, etc. I have to write an entirely custom validation logic. There's also no unified way to plug validation into the grid editing. I would like some mechanism, like the grid transaction service, which allows me to define validation logic and events it triggers on, similar to forms validation.

## Describe alternatives you've considered
Write an entirely custom validation logic.

## Additional context
This affects maintainability of applications using igniteui-angular.
Consider #4135

Eralmidia:
@kdinev Any idea approximately when this feature will be prioritized? We're planning on converting quite a few grids from the jquery version to igx this fall, and I need to decide if we should wait for builtin support, or we need to start planning for a custom solution for validation.

kdinev:
@Eralmidia I'll have @radomirchev @rkaraivanov and @damyanpetev look at the item in order to consider it for prioritization for the next milestone. We have a few larger items that we're already looking at for it, like a pivot data selector and a pdf exporter for the grid, and after the evaluation we will update the issues/roadmap to reflect which items we will be working on and when!

mikerentmeister:
Any idea when this tentatively might be released?

kdinev:
@mikerentmeister This item is in our backlog plans for the current quarter. We will update the item with an estimated release date soon! @radomirchev

mikerentmeister:
Any updates on this in terms of estimated delivery?

kdinev:
@mikerentmeister We kicked this item off this week. What we're planning is to expose `FormControl` instances for the grid row editors, which will allow the same validation logic, which is passed into other forms in an application, to also be accepted by the grid columns. We also discussed updating the read-only rendering to reflect validation state (invalid column values) after cell editing has been performed on them and the editor is no longer visible. We're currently in the state of updating the editing specification to reflect these. @MayaKirova and @damyanpetev will update the issue further after the specification update is in place, so there can be a discussion on it.

MayaKirova:
[...]

PR #11923, opened 2022-08-02
Grid Validation POC

MayaKirova:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -90,6 +92,19 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
...
+    /**
+     * Sets/gets the validators used when editing cell from this column.
+     * ```typescript
+     * let validators = this.column.validators;
+     * ```
+     * ```html
+     * <igx-column [validators] = "validatorsArray"></igx-column>
+     * ```
+     */
+    public validators: Validator[] = [];
```
mddragnev, line 106: Is this supposed to be an input? If not we should remove the `html` doc

mddragnev:
Should we expose the default error template ? In case we have something like this:
```
<igx-column appForbiddenName='bob' minlength="3" maxlength='8' email required *ngFor="let c of columns" [editable]='true' [sortable]="true" [filterable]="true" [field]="c.field"
            [header]="c.field" [width]="c.width" [resizable]='true'>
            <ng-template igxCellValidationError let-cell='cell'>
                <div *ngIf="cell.formGroup?.get(cell.column?.field).errors?.['required']">
                    This name is required.
                </div>
                <div *ngIf="cell.formGroup?.get(cell.column?.field).errors?.['forbiddenName'] else test">
                    This name is forbidden.
                </div>
                <ng-template #test>
                    <div>Default</div>
                </ng-template>
            </ng-template>
        </igx-column>
```
There is a case where we want to show a specific message only if a specific error occurs in any other case we may want to display the default erorr message. So IMO something like this should be possible
```
<ng-template igxCellValidationError let-cell='cell'>
                <div *ngIf="cell.formGroup?.get(cell.column?.field).errors?.['forbiddenName'] else cell.defaultErrorTemplate">
                    This name is forbidden.
                </div>
            </ng-template>
```

mddragnev:
In the grid validation dev sample if you pin the first row and make for example product name of the pinned row invalid (set a string with length > 8). Then click done. It shows that exactly the edited cell is invalid. Now if we try to delete that row from the action strip it just visually makes the cell valid again and does not delete it.

mddragnev:
Did you noticed the warning in the console when trying to enter a edit mode in the dev sample
Screenshot 2022-08-12 at 16 21 16
This warning is thorwn only when there are validators on the column.

mddragnev:
One other thing, If we go on the dev sample with transactions enabled and cell editing:
If you enter edit mode on some cell and make it invalid, then press enter and enter edit mode on any other cell of that same row. An error is thrown:
Screenshot 2022-08-12 at 19 17 28

This might be something related to the sample because i havent investigated it much

MayaKirova:
@mddragnev 

1) Regarding the default error template. 

I don't mind exposing it as a public property. @dkamburov Unless there are any objections we can add it to the spec.

2) About the row delete issue.

Data changes should directly reflect in the data of the non-transaction grid IMO. I've added an `autoCommit` property for this in the transaction service, that is true for the base transaction service. In that case it will track transactions and their validity but will reflect the changes in the data directly.

3) About the warning.

I'm not sure whether changing the ngModel binding with what they've recommended in the docs:
https://angular.io/guide/reactive-forms#updating-parts-of-the-data-model
Will will work in all scenarios since I think we've removed the [(ngModel)] binding before but then had to re-add it back due to some issues. 
I'll look into it a bit more.

damyanpetev (commented):
The positioning of the error message is odd - when you enter a cell with active message that is longer and change the value to trigger a shorter message the tooltip is not repositioned to be center aligned under the icon and it looks odd:
image
Same happens really when you go from a shorter message to a longer one and it just shifts to the right instead. A reposition on message change (error changes) will do away with both.

Also not sure if sample issue (on the first sample w/o transactions), but when you enter a value like this the message goes empty:
image

MayaKirova:
@damyanpetev Regarding the empty error message.
It's because that's a custom validator (appForbiddenName='bob') so we have no default error message for it.
We have only for the default validator directives we've extended for the columns. Hence in the sample a custom error template can be added:
<ng-template igxCellValidationError let-cell='cell'>
                <div *ngIf="cell.formGroup?.get(cell.column?.field).errors?.['forbiddenName'] else cell.defaultErrorTemplate">
                    This name is forbidden.
                </div>
  </ng-template>

projects/igniteui-angular/src/lib/services/transaction/igx-transaction.ts
```diff
@@ -166,12 +172,16 @@ export class IgxTransactionService<T extends Transaction, S extends State> exten
...
         this._transactions.splice(this._transactions.length - lastActions.length);
+        this._validationTransactions.splice(this._validationTransactions.length - lastActions.length);
         this._redoStack.push(lastActions);
 
         this._states.clear();
+        this._validationStates.clear();
         for (const currentActions of this._undoStack) {
             for (const transaction of currentActions) {
                 this.updateState(this._states, transaction.transaction, transaction.recordRef);
+                const validationTransaction = this._validationTransactions.find(x => x.id === transaction.transaction.id);
```
damyanpetev, line 183: Right, so I'm of the opinion that this form of state tracking for validity will likely slip at one point or another - unlike the value states being tracked that should aggregate neatly, validity doesn't. With more involved custom validation, where for example you have checks that depend on multiple cell values and the ability to commit transactions out of order per record id, grabbing the previous validity on undo in not guaranteed to be correct. That of course, is true for cell edits, but the same case can be made for cross-row checks even if less likely. If the validation depends on some external factor - then we have no idea if the old state is still valid regardless. 

I think the most prudent way to avoid having to deal with this is simply to keep a singular current state of validity and simply re-trigger validation on such state changes to update it - there aren't that many and it seems about in line with how most validation I've seen handles changes anyway.
Given that, you'll no longer need to store separate validation transactions, and with them gone and the state stored already being pretty foreign to the transaction service I don't see a reason to keep it there. Transactions have no control over the validation process and them being the owner of a state they makes even less sense to me this way, so we might as well move it out.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -722,6 +823,48 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
...
+    }
+
+    public ngAfterViewInit() {
+        this.errorTooltip.changes.subscribe(() => {
+            if (this.errorTooltip.length > 0 && this.active) {
+                // error ocurred
+                this.cdr.detectChanges();
+                this.openErrorTooltip();
+            }
+            this.grid.validationStatusChange.emit(
```
damyanpetev, line 836: @MayaKirova I'm a bit confused - this is a grid level event, right? Yet it emits for every cell edit change?

MayaKirova: It's because we agreed that it should trigger the moment an invalid value is entered in an editor. Here:
https://github.com/IgniteUI/igniteui-angular/issues/6501#issuecomment-1177878614

dkamburov:
Clicking multiple times on the icon of the validator is throwing an error:
click-icon

MayaKirova:
@dkamburov Issue should be fixed now.

projects/igniteui-angular/src/lib/grids/grid-public-row.ts
```diff
@@ -51,6 +52,16 @@ abstract class BaseRow implements RowType {
             this.grid.crudService.row.id === this.key;
     }
 
+    /** Gets the validation errors if any.
+    * ```typescript
+    * let errors = row.errors;
+    * ```
+    */
+    public errors(): ValidationErrors {
```
dkamburov, line 60: This should be a getter

projects/igniteui-angular/src/lib/grids/grid-public-row.ts
```diff
@@ -51,6 +52,16 @@ abstract class BaseRow implements RowType {
...
+        const formGroup = this.grid.validation.getFormGroup(this.key);
```
dkamburov, line 61: This requires additional check. `FormGroup` might not be created yet.

dkamburov (changes requested):

projects/igniteui-angular/src/lib/grids/grid-public-row.ts
MayaKirova: @dkamburov There is a check on the line below:
```
return formGroup?.errors;
```
If there's no formGroup it will return undefined.

dkamburov (dismissed):

projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -200,6 +205,29 @@ export interface ColumnType {
     populateVisibleIndexes?(): void;
 }
 
+export interface IFieldValid {
+    field: string;
+    valid: boolean;
+    formGroup: FormGroup;
+}
+
+export enum Validity {
```
damyanpetev, line 214: Wondering if should we make this compatible with Angular's `FormControlStatus`?

projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -63,6 +65,7 @@ export interface CellType {
     grid: GridType;
     id?: { rowID: any; columnID: number; rowIndex: number };
     cellID?: any;
+    errors?: ValidationErrors;
```
damyanpetev, line 68: Oh almost forgot - because the errors are getters only, do mark them as readonly in the interface

projects/igniteui-angular/src/lib/services/transaction/base-transaction.ts
```diff
@@ -111,12 +112,14 @@ export class IgxBaseTransactionService<T extends Transaction, S extends State> i
     /**
      * @inheritdoc
      */
-    public commit(_data: any[], _id?: any): void { }
+    public commit(_data: any[], _id?: any): void {
+        this.clear(_id);
+    }
```
damyanpetev, line 117: Mind reverting the transaction changes in general?

projects/igniteui-angular/src/lib/grids/public_api.ts
```diff
@@ -13,7 +13,7 @@ export * from './grid-base.directive';
 export * from './grid.common';
 export * from './grid-public-row';
 export * from './grid-public-cell';
-export { CellType, RowType, IGX_GRID_BASE } from './common/grid.interface';
+export { CellType, RowType, IGX_GRID_BASE, Validity, IRecordValidationState, IFieldValidationState } from './common/grid.interface';
 export * from './summaries/grid-summary';
 export * from './grid-common.module';
 export * from './grid.rowEdit.directive';
```
damyanpetev, line 19: Is the Validation service exposed somewhere in the public API? Because I don't see it and it is part of the Grid's public API

projects/igniteui-angular/src/lib/grids/grid/grid-validation.service.ts
```diff
@@ -0,0 +1,164 @@
...
+    private _valid = true;
+
+
+    /** Gets whether state is valid.
+    */
+    public get valid() : boolean {
+        return this._valid;
+    }
+
+    public set valid(value: boolean) {
```
damyanpetev, line 22: Not entirely sure we need this to be settable. Any use for it?

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-api.service.ts
```diff
@@ -287,14 +288,14 @@ export class IgxTreeGridAPIService extends GridBaseAPIService<GridType> {
...
         rowNewValue: { [x: string]: any }) {
-        if (grid.transactions.enabled) {
             const path = grid.generateRowPath(rowID);
             const transaction: HierarchicalTransaction = {
                 id: rowID,
                 type: TransactionType.UPDATE,
                 newValue: rowNewValue,
                 path
             };
+        if (grid.transactions.enabled) {
```
damyanpetev, line 298: This seems like a bit of a leftover change

projects/igniteui-angular/src/lib/grids/cell.component.html
```diff
@@ -109,52 +78,72 @@
...
+    <div *ngIf="formGroup?.get(column?.field).errors?.['max']">
+        {{grid.resourceStrings.igx_grid_max_validation_error | igxStringReplace:'{0}':formGroup.get(column.field).errors.max.max }}
+    </div>
+    <div *ngIf="formGroup?.get(column?.field).errors?.['email']">
+        {{grid.resourceStrings.igx_grid_email_validation_error }}
+    </div>
+    <div *ngIf="formGroup?.get(column?.field).errors?.['pattern']">
+        {{grid.resourceStrings.igx_grid_pattern_validation_error}}
+    </div>
+</ng-template>
```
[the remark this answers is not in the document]
damyanpetev: Actually a few files are like that now, so valid throughout

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -421,6 +472,45 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
...
+    /** @hidden @internal */
+    @HostBinding('attr.aria-describedby')
+    public get describeBy() {
+        return this.grid.id + '_' + this.column.field;
+    }
+
+    /** @hidden @internal */
+    @HostBinding('attr.aria-errormessage')
+    public get ariaErrorMessage() {
+        return this.grid.id + '_' + this.column.field + '_' + this.intRow.index + '_error';
```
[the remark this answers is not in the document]
damyanpetev: Further details, looking at this compat table: https://a11ysupport.io/tech/aria/aria-errormessage_attribute#expectations
We should definitely forgo using error message for now.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -956,6 +1118,9 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
 
         if (this.editable && editMode && !this.intRow.deleted) {
             if (editableCell) {
+                if (this.grid.validationTrigger === 'blur') {
+                    this.grid.tbody.nativeElement.focus({ preventScroll: true });
+                }
```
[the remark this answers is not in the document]
MayaKirova: It does become a timing issue, since validationTrigger tells the formcontrol when it can actually update the value. So until a blur happens the old value will remain in the form control (no changes will be pushed)  and in the editValue of the cell as well. Hence the updateCell api will not have the up-to-date value if a blur has not happened yet.

 I'll check if it's possible to force the form control to update it's value even if the trigger has not been called.

projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -584,6 +616,7 @@ export interface GridType extends IGridDataBindable {
     resetColumnCollections(): void;
     triggerPipes(): void;
     repositionRowEditingOverlay(row: RowType): void;
+    resizeAndRepositionOverlayById(overlayId: string, newSize: number): void;
```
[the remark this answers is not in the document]
MayaKirova: I needed access to the grid's overlayService, which is protected in the base grid. That's why I moved it there. 
Not sure if there's any other way to get the overlay service...

projects/igniteui-angular/src/lib/grids/public_api.ts
MayaKirova: It's exposed now,

projects/igniteui-angular/src/lib/grids/common/events.ts
```diff
@@ -31,6 +31,7 @@ export interface IGridEditDoneEventArgs extends IBaseEventArgs {
     column?: ColumnType;
     owner?: GridType;
     isAddRow?: boolean;
+    valid?: boolean;
```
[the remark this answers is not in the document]
MayaKirova: Same type seems to be used in some row events like rowDelete where valid does not really make sense. I'll leave it as optional for now.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1264,6 +1281,32 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
...
+     * Sets the validation error template.
+     * ```html
+     * <ng-template igxCellValidationError let-cell="cell">
+     *     <div *ngIf="cell.errors?.['forbiddenName']">
+     *      This name is forbidden.
+     *     </div>
+     * </ng-template>
+     */
+    public set errorTemplate(template: TemplateRef<any>) {
+        this._errorTemplate = template;
```
mddragnev, line 1307: @MayaKirova Isn`t this an input setter ? Do we need to expose such thing? I mean this gives the possibility the error template to be defined outside the igx-column template. Is this why this is exposed? Also, if what I think is true maybe we should change the `html` doc

MayaKirova: All other template are exposed so we might as well expose this one as well. I'll update the html doc.

mddragnev:
On the dev grid validation sample, if you make the first row invalid by one of the cells and then use the public row api to get the errors it returns undefined.
`grid.getRowByIndex(0).errors` is undefined, if you use the public cell api, the error is there. So the formGroup seems to not persist the errors that any of the form controls have.

MayaKirova:
> On the dev grid validation sample, if you make the first row invalid by one of the cells and then use the public row api to get the errors it returns undefined. `grid.getRowByIndex(0).errors` is undefined, if you use the public cell api, the error is there. So the formGroup seems to not persist the errors that any of the form controls have.

@damyanpetev should we collect both the errors from the `formGroup` and the child `formControl`'s for the public row API, or should the row be associated only with the  `formGroup`?

mddragnev:
Open the hierarchical grid dev sample, on the first grid template add [primaryKey]="'ID'" and [rowEditable]='true' , then on the first row island add [primaryKey]="'ID'", in the rowIsland see the column group with header Information and set maxlength='5' validator on the ProductName column. Now in the sample, Expand the first row, double click on the first row of the child grid on the productname column and enter some string longer than 5 chars. Press enter and observe the console for error.

mddragnev:
On the tree grid add row sample, if you make the first two rows (with rowId 0 and 4) invalid. Afterwards if you call grid.validation.getInvalid() you will get an array with 2 invalid records. Then use the grid validation service clear method to clear only the validation errors of the first (grid.validation.clear(0)). Call again grid.validation.getInvalid() and observe that it returns empty array. This is because 
```
 public clear(rowId?: any) {
        if (rowId) {
            this._validityStates.delete(rowId);
        } else {
            this._validityStates.clear();
        }
        this.updateStatus();
    }
```
the rowId is 0 and it is fale. And the if block hits the else statement

damyanpetev:
> > On the dev grid validation sample, if you make the first row invalid by one of the cells and then use the public row api to get the errors it returns undefined. `grid.getRowByIndex(0).errors` is undefined, if you use the public cell api, the error is there. So the formGroup seems to not persist the errors that any of the form controls have.
> 
> @damyanpetev should we collect both the errors from the `formGroup` and the child `formControl`'s for the public row API, or should the row be associated only with the `formGroup`?

I think we should stick to what normal form groups do: the errors are specific to where the validation is - if there was a validator on the form group itself (cross-field), then I'd expect the row to have errors. Otherwise, the errors are under the specific cell and you can't really expose them on the row without changing the error type (since there would need to be indication for which field the error object is, which is a map of maps, essentially what row.cell.errors is already :) )

MayaKirova:
@mddragnev Issues should be fixed now.

mddragnev:
On the grid validation dev sample, when trying to add row make any of the cells invalid, the observe that `grid.validation.getInvalid()` does contain that row which is not added yet. If you cancel the adding, the validation service should not contain the invalid record for thar particular row, but it does.

mddragnev (approved):

projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -205,21 +205,23 @@ export interface ColumnType {
...
 
-export enum Validity {
-Valid,
-Invalid
+export interface IGridValidityStatusEventArgs {
+    status: ValidityStatus,
+    owner: GridType
 }
 
+export declare type ValidityStatus = 'VALID' | 'INVALID';
```
damyanpetev, line 218: Since this is not a definition and I don't thing we need it to be ambient, just the export will suffice:
```suggestion
export type ValidityStatus = 'VALID' | 'INVALID';
```

src/app/grid-cellEditing/grid-cellEditing.component.html
```diff
@@ -25,15 +25,14 @@ <h3>Grid with primary key ProductID</h3>
...
     <igx-column field="ProductName" width="200px" [groupable]="kk" [header]="pname" [sortable]="true" [dataType]="'string'" [editable]="true" [resizable]="true">
-        <ng-template igxCell let-cell="cell">
-            <button> {{cell.value}}</button>
-        </ng-template>
-    </igx-column>
-    <igx-column field="UnitsInStock" header="UnitsInStock" width="200px" [dataType]="'number'" [editable]="true" [sortable]="true" [hasSummary]="false">
-        <ng-template igxCellEditor let-cell="cell">
-            <input name="units" [(ngModel)]="cell.value" style="color: black" />
+        <ng-template igxCellValidationError let-cell='cell'>
+            <div *ngIf="cell.formGroup?.get(cell.column?.field).errors?.['forbiddenName']">
```
damyanpetev, line 29: That's a bit of a left over, can be fixed later though

damyanpetev (approved):
Still a few nit to clean up around ARIA and dev demos, but overall LGTM

skrustev (dismissed):

skrustev (approved):

dkamburov (approved):

damyanpetev (approved):

Issue #11849, opened 2022-07-13
Bug: radio-group breaks when changing options after initial render of directive


ntziolis:
Use case:
- We have an observable that returns a list of options
  - The list of options changes based what the user has selected in previous form inputs
- The first time the options are rendered everything works as expected 
  - Options are rendered
  - And only single option can be selected)
- The second time the options are rendered the directive breaks
  - Options are properly rendered
  - But now all options can be selected at the same time 

Looking at the source of the directive this is likely due to only iterating through the options after initial render of the directive

Likely solution:
The `QueryList` has a property `changes` which is an observable that captures any subsequent changes to the list of `radioButtons`

https://github.com/IgniteUI/igniteui-angular/blob/fa8c887b72fde85bb9ccb4a5e8b309b0083151ef/projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.ts#L377

Please note:
- We have also tried if removing `async` form the html
- Instead we saved the new options array into a normal non obs variable
- But the directive still breaks
- So this is not an observable based issue it will happen with declarative programming as well

ntziolis:
We have found a (very ugly) workaround to be able to keep using the radio button group, by  re-rending the radio group fresh each time.

How:
- We created a structural directive based on `ngIf`
- The main difference to `ngIf` is: 
  - every time there is new data 
  - it clears the view container 
  - afterwards it creates an embedded view with the `templateRef`

This directive `renderIf` can be used like `ngIf` (without the else part).

```ts
import {
  Directive,
  Input,
  OnInit,
  TemplateRef,
  ViewContainerRef,
} from '@angular/core';
import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
import { ReplaySubject, tap } from 'rxjs';

@UntilDestroy()
@Directive({
  selector: '[renderIf]',
})
export class RenderIfDirective<T = unknown> implements OnInit {
  private _context: RenderIfContext<T> = new RenderIfContext<T>();
  private _data$ = new ReplaySubject<T | null>();

  constructor(
    private _viewContainer: ViewContainerRef,
    private templateRef: TemplateRef<RenderIfContext<T>>
  ) {}

  @Input()
  set renderIf(data: T | null) {
    this._context.$implicit = this._context.ncsRenderIf = data as T;
    this._data$.next(data);
  }

  ngOnInit(): void {
    this._data$
      .asObservable()
      .pipe(
        untilDestroyed(this),
        tap((data) => {
[...]

Issue #11942, opened 2022-08-04
igx-radio and igx-radio-group don't work with dynamic content


yl-endress:
Dear igx-team,

## Description  
There is some weird behavior with the igx-radio and igx-radio-group with formControls and dynamic content

 * igniteui-angular version: "12.3.9", "12.3.33" and also "14.0.11"
 * browser: Chrome Version 103.0.5060.134 x64

## Steps to reproduce  

Simple have a look here https://stackblitz.com/edit/angular-mgp2nr-9ka7qq

a) the ideal solution is not working at all
1. press the button below
2. click on all radios
3. all options are selected -> scope is lost

b) workaround is having issues with the reset
1. click Banana
2. click on reset
3. mango is selected, but the form-control is still dirty
4. click reset again: now it works as expected

## Result  
a) the scope of the radios get lost -> multi selection
b) formControl+formGroup stays dirty

## Expected result  
a) no multi select, single select as radios normally do
b)  formControl+formGroup are reset properly (pristine and not dirty) -> reset works with the first click 

## Attachments  
Here a screenshot containing both errors
![image](https://user-images.githubusercontent.com/44203645/182863114-3adea85d-0eba-4cde-9848-76ad8995e7c0.png)

ChronosSF:
Likely related to #11849

PR #11959, opened 2022-08-09
fix(radio): allowing radio group to pick dynamically added radios - master

ChronosSF:
Closes https://github.com/IgniteUI/igniteui-angular/issues/11849, https://github.com/IgniteUI/igniteui-angular/issues/11942

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/radio/radio-group.directive.ts
```diff
@@ -301,8 +306,10 @@ export class IgxRadioGroupDirective implements AfterContentInit, ControlValueAcc
         // the OnInit of the NgModel occurs after the OnInit of this class.
         this._isInitialized = true;
 
-        setTimeout(() => {
-            this._initRadioButtons();
+        this.radioButtons.changes.pipe(startWith(0), takeUntil(this.destroy$)).subscribe(() => {
+            this.queryChange$.next();
+            this.queryChange$.complete();
```
skrustev, line 311: I am not sure what the purpose of the `queryChange$` is in here. When the radio button change it gets completed immediately and not used again.

ChronosSF: We discussed this in a meeting that it might not be needed to have this Subject at all. But the idea here is that if the same button instance is kept after an ngFor change, it will already have a subscription from a previous `changes` handler and that needs to be removed before subscribing it again.

You can try testing this for memory leaks without the queryChange handler too and if it's good, I'll remove the whole thing.

skrustev: I see, still though the `queryChange$` actually doesn't have any effect in the `button.change.pipe` from what I can see. In the following scenario `_selectedRadioButtonChanged` get triggered multiple times, as many as the added new radios to the `ngFor`. How to reproduce it:

1. Push one by one values to a list used in `ngFor` inside the igxRadioGroup.
2. Click on one of the old radios to select it
3. Observe how many times the `_selectedRadioButtonChanged` triggers.

I think the `merge` combined with `takeUntil` in this scenario acts as `takeUntil` all the Observables in the merge have completed.
I would suggest another approach like using a `IterableDiffer ` to determine the added radios from the `this.radioButtons.changes` by comparing `ids` and the calling `_initRadioButtons` for a specified list of radios or something like that.

simeonoff (approved):

PR #12001, opened 2022-08-22
Refactor validation in separate service.

MayaKirova:
Related to: https://github.com/IgniteUI/igniteui-angular/pull/11923#discussion_r950276246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,10 +77,10 @@ All notable changes for each version of this project will be documented in this 
         ```html
           <igx-column required minlength="4" ...>
           ```
-        2. Via reactive forms using the FormGroup exposed via the `onFormGroupCreate` event of the grid:
+        2. Via reactive forms using the FormGroup exposed via the `formGroupCreated` event of the grid:
 
         ```html
-        <igx-grid (onFormGroupCreate)='formCreateHandler($event)' ...>
+        <igx-grid (formGroupCreated)='formCreateHandler($event)' ...>
         ```
 
         ```ts
--- a/projects/igniteui-angular/src/lib/grids/api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/api.service.ts
@@ -558,10 +558,12 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
         } else {
             mergeObjects(rowValueInDataSource, rowNewValue);
         }
-        const isAddRow = this.crudService.row instanceof IgxAddRow;
-        if (isAddRow) { return; }
-        const validation = grid.validation as IgxGridValidationService;
-        validation.update(rowID, rowNewValue);
+
+        if (!this.grid.transactions.isPending) {
+            const validation = grid.validation as IgxGridValidationService;
+            validation.update(rowID, rowNewValue);
+        }
+
     }
 
 
--- a/projects/igniteui-angular/src/lib/grids/api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/api.service.ts
@@ -301,6 +301,7 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
             grid.data.push(rowData);
         }
         const validation = grid.validation as IgxGridValidationService;
+        validation.markAsTouched(transactionId);
         validation.update(transactionId, rowData);
     }
 
--- a/projects/igniteui-angular/src/lib/grids/api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/api.service.ts
@@ -4,7 +4,7 @@ import { cloneArray, reverseMapper, mergeObjects } from '../core/utils';
 import { DataUtil, GridColumnDataType } from '../data-operations/data-util';
 import { IFilteringExpressionsTree } from '../data-operations/filtering-expressions-tree';
 import { Transaction, TransactionType, State } from '../services/transaction/transaction';
-import { IgxCell, IgxGridCRUDService, IgxEditRow, IgxAddRow } from './common/crud.service';
+import { IgxCell, IgxGridCRUDService, IgxEditRow } from './common/crud.service';
 import { CellType, ColumnType, GridServiceType, GridType, IFieldValid, RowType } from './common/grid.interface';
 import { IGridEditEventArgs, IRowToggleEventArgs } from './common/events';
 import { IgxColumnMovingService } from './moving/moving.service';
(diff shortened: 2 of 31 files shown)


projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -568,7 +559,6 @@ export class IgxGridCRUDService extends IgxRowAddCrudState {
 
             } else {
                 this.createCell(cell);
-                this.grid.onFormGroupCreate.emit(this.cell.row.rowFormGroup);
```
damyanpetev, line 571: We might actually need to keep this place in the process rather than general init, just to make the validity state adequate, but will come back to you on this.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6200,6 +6203,12 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
         this.summaryService.clearSummaryCache();
         this.pipeTrigger++;
         this.notifyChanges();
+        if (event.origin === TransactionEventOrigin.REDO || event.origin === TransactionEventOrigin.UNDO) {
+            event.actions.forEach(x => {
+                const value = this.transactions.getAggregatedValue(x.transaction.id, true);
+                this.validation.update(x.transaction.id, value ?? x.recordRef);
+            });
+        }
```
damyanpetev, line 6211: Would move this above the pipe trigger and notify just for sake of consistency with the rest of the handling, though both are async but still.

Also interested if this can be combined with the selection handling that sort of checks for similar state change

projects/igniteui-angular/src/lib/grids/common/crud.service.ts
damyanpetev: Actually you might need to hook into the end edits just to know when an edit passes to keep the validity for it

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -557,20 +544,20 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
...
         };
         if (grid.transactions.enabled) {
             grid.transactions.add(transaction, rowCurrentValue);
         } else {
             mergeObjects(rowValueInDataSource, rowNewValue);
         }
-        grid.transactions.addValidation(transaction, rowCurrentValue);
+        const formGroup = this.grid.validation.getFormGroup(rowID);
+        const validation = grid.validation as IgxGridValidationService;
+        validation.add(rowID, formGroup)
```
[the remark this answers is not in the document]
damyanpetev: What we might want to do though (after checking with the CRUD service to avoid in-edit double dips, since UI edits end up here too), is to check if value is updated for a validated record/cell and update it's status accordingly.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
MayaKirova: The existing selection handling seems to care only about ADD undo or DELETE redo actions, while we need to update validity on any kind of undos/redos. So I'd rather not combine them.

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -557,20 +548,23 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
...
+            newValue: rowNewValue
         };
         if (grid.transactions.enabled) {
             grid.transactions.add(transaction, rowCurrentValue);
         } else {
             mergeObjects(rowValueInDataSource, rowNewValue);
         }
-        grid.transactions.addValidation(transaction, rowCurrentValue);
+
+        if (!this.grid.transactions.isPending) {
```
damyanpetev, line 563: Really not a fan of adding this as hidden implementation to the service (considering it's both public API and replaceable). There should be a good way to check if there's an active row edit (presuming that's what you're after) from the CRUD service, right?

CHANGELOG.md
```diff
@@ -77,10 +77,10 @@ All notable changes for each version of this project will be documented in this
         ```html
           <igx-column required minlength="4" ...>
           ```
-        2. Via reactive forms using the FormGroup exposed via the `onFormGroupCreate` event of the grid:
+        2. Via reactive forms using the FormGroup exposed via the `formGroupCreated` event of the grid:
 
         ```html
-        <igx-grid (onFormGroupCreate)='formCreateHandler($event)' ...>
+        <igx-grid (formGroupCreated)='formCreateHandler($event)' ...>
```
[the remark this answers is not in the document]
damyanpetev: After discussion, most are of the opinion this is fairly obvious in purpose, so closing

projects/igniteui-angular/src/lib/grids/api.service.ts
MayaKirova: @damyanpetev 
I cannot think of a better place since we want to have this work with any api calls and all calls (api and ui) go through here. The only time we don't want to trigger an update is when this is a pending transaction that may be discarded in the future. 

The pending state change does not seem to be the same as the actual exiting of the edit mode, since exitCellEdit/exitRowEdit is called after the data is updated in the data/transactions. And we don't really track whether the current state is pending in the crud service (at least as far as I can tell).

Also, not sure if this is a viable scenario, but `startPending`/`endPending` are public apis, so in theory the user can use them to build some pending transactions himself. In that case any state we store in the crud service would not apply to those so we may trigger validation for something that is still pending, I guess. Not sure if viable, but still might be worth considering.

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -244,7 +239,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
      * @memberof IgxGridCellComponent
      */
     public get template(): TemplateRef<any> {
-        if (this.editMode) {
+        if (this.editMode && this.formGroup) {
```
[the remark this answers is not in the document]
MayaKirova: Yes. It was a strange timing/change detection mismatch in one of the test scenario that I could not reproduce outside of the tests:
https://dev.azure.com/IgniteUI/igniteui-angular/_build/results?buildId=42173&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=59a85588-b0ba-5043-24c4-d9e29d89c6f6&l=7086

For some reason when forcing editing with:
cell.editMode = true;
And then doing changing the data view so that this cell now has a new data context and should not be in editMode, in the test it still tried to render it as if it is in edit mode, hence the error.

dkamburov (approved):

PR #12011, opened 2022-08-23
fix(combos): Add predefined selection to the input when the data is bound after the initialization

### LINKED ISSUE #11811: Combo as cell editor value disappears
## Description  
I have a grid with `IgxCombo` as cell editor. The data of the combo is loaded from the backend and is async. When the cell containing the combo enter edit mode the value of the combo input disappears.

 * igniteui-angular version: 14.0.0
 * browser: all

## Steps to reproduce  

1. Run the attached sample.
2. Put the first row in edit mode.
3. Click on friends column.

## Result  
The value of the cell disappears.

## Expected result  
Value of the cell should not disappear.

## Attachments  
[grid_combo_issue.zip](https://github.com/IgniteUI/igniteui-angular/files/9033044/grid_combo_issue.zip)

viktorkombov:
Closes #11811 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/combo/combo.common.ts
```diff
@@ -306,6 +306,10 @@ export abstract class IgxComboBaseDirective extends DisplayDensityBase implement
     }
     public set data(val: any[] | null) {
         this._data = (val) ? val : [];
+
+        if (this._data?.length && this.selectionService.get(this.id) && this.virtDir) {
+            this._value = this.createDisplayText(this.selection, []);
```
jackofdiamond5, line 311: We could move this in a `ngDoCheck` call for example which will be invoked after the `data` changes and is populated with some values, in which case we shouldn't need the check above.

Additionally, we may want to note somewhere in the docs that when a combo is bound to remote data and put in the grid, then depending on the delay for the requested data one can expect a few moments when the combo will remain blank without any items and without anything in its field. I think this is somewhat of a limitation since when enter edit mode we essentially recreate the cell component and respectively the combo, so it has to load everything from scratch.

projects/igniteui-angular/src/lib/combo/combo.common.ts
```diff
@@ -980,6 +986,18 @@ export abstract class IgxComboBaseDirective extends DisplayDensityBase implement
         });
     }
 
+        /** @hidden @internal */
+        public ngDoCheck() {
+            if (this._init) {
+                return;
+            }
+    
+            let changes = this._iterableDiffer.diff(this.data);
```
jackofdiamond5, line 995: I don't think we actually need to be invoking the IterableDiffer API as the only thing it's doing right now is it is checking if we have populated data. Which you can easily do by just checking `this.data?.length`, which you had already done in the previous change. We can also check for `this.selection` just to be sure that we actually have some selection before calling `createDisplayText`. In any case all of this is going to prevent `createDisplayText` from being called a few times before the data has loaded, so it's more of an optimization check.

jackofdiamond5 (approved):

kdinev (approved):

PR #12023, opened 2022-08-29
chore(*): Move validation updates on update_cell/update_row if the ed…

MayaKirova:
Related to https://github.com/IgniteUI/igniteui-angular/pull/12001#discussion_r955881647

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/api.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/api.service.ts
@@ -146,7 +146,11 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
 
         this.grid.summaryService.clearSummaryCache(args);
         const data = this.getRowData(cell.id.rowID);
-        this.updateData(this.grid, cell.id.rowID, data, cell.rowData, reverseMapper(cell.column.field, args.newValue));
+        const newRowData = reverseMapper(cell.column.field, args.newValue);
+        this.updateData(this.grid, cell.id.rowID, data, cell.rowData, newRowData);
+        if (!this.grid.rowEditable) {
+            this.grid.validation.update(cell.id.rowID, newRowData);
+        }
         if (this.grid.primaryKey === cell.column.field) {
             if (this.grid.selectionService.isRowSelected(cell.id.rowID)) {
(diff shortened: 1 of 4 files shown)


projects/igniteui-angular/src/lib/grids/grid/grid-validation.service.ts
```diff
@@ -126,14 +126,14 @@ export class IgxGridValidationService {
...
         const rowGroup = this.getFormGroup(rowId);
         if (!rowGroup) return;
         rowGroup.markAsTouched();
         for (const col of this.grid.columns) {
-            const field = col.field;
-            const control = rowGroup?.get(field);
-            control.markAsTouched();
+            if(!field || (field && col.field === field)) {
+                rowGroup?.get(col.field)?.markAsTouched();
+            }
```
damyanpetev, line 136: Mnor: You could easily short-circuit the loop if a field is passed to avoid looping all columns. I highly doubt it'll make much difference, but still. If you want to skip extending the code into if-else clause, you could loop over fields assigned in advance with a check though, like:
```ts
const fields = field ? [field] : this.grid.columns.map(x => x.field);
for (... fields)
```

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -4528,6 +4528,7 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
 
                 const cell = new IgxCell(id, index, col, rowData[col.field], value, rowData, this as any);
                 this.gridAPI.update_cell(cell);
+                this.validation.markAsTouched(rowSelector, column);
```
damyanpetev, line 4531: Not entirely sure on this one, I'm kind of okay with both versions, though the existing behavior seems quite reasonable to me. Might need to check with UX still

MayaKirova: After discussing it, we decided to not mark them as touched automatically, but expose the API as public so that the user could mark them if he so chooses.

damyanpetev (approved):

dkamburov (approved):

PR #12051, opened 2022-09-08
chore(*): Add aria-describedby reference to error message element as …

### LINKED ISSUE #12043: Validation aria attributes are not read in screen readers.
Related to: https://github.com/IgniteUI/igniteui-angular/pull/11923#discussion_r960622227

Most readers don't seem to follow the spec and don't read the `aria-errrormessage` attribute.
Check if it can be worked around to read the validation message in some other way (with aria-describeby or some other attribute).

MayaKirova:
…well since readers are more likely to read it than aria-errormessage.

Note: Works with more advanced readers like NVDA. More basic ones like the built-in win Narrator don't seem to read it though. 

Closes #12043  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -472,6 +472,9 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
     /** @hidden @internal */
     @HostBinding('attr.aria-describedby')
     public get describeBy() {
```
damyanpetev, line 474: PS: There's also a describedby in the row templates

mddragnev:
This does work with Macos VoiceOver and NVDA.

MayaKirova:
@damyanpetev I think it reads better when there's no aria-describedby on the cell that point to the header (does not repeat the header twice in NVDA) but it might be a breaking change since there are some tests that explicitly expect the old describedby attribute. Should we remove it or just set both the error and the header ref for describedby?

MayaKirova:
> @damyanpetev I think it reads better when there's no aria-describedby on the cell that point to the header (does not repeat the header twice in NVDA) but it might be a breaking change since there are some tests that explicitly expect the old describedby attribute. Should we remove it or just set both the error and the header ref for describedby?

Actually it also seem to depend on the browser and reader.

- In NVDA (win)
  Chrome and Edge now read the header text 1 time, but Firefox does not read the header.

-  In VoiceOver (macOs)
 Chrome does not to read the header.


So I guess it's better to go with both so that there's no breaking change and the header is read at least once on every browser/reader.

damyanpetev (approved):
LGTM

PR #12076, opened 2022-09-13
chore(*): Sync editValue changes with the formControl value and mark …

### LINKED ISSUE #12066: Allow for cell edit templates binding `editValue` to function fully with validation
Currently, if the cell edit template is provided, we guide users towards the newly exposed formControl that can be directly attached to save on extra handling which is all nice and good. However, if the editor is bound to the `cell.editValue` - the validation does work in the sense that tha grid becomes invalid and the validation service correctly returns the errors, except the error UI tooltips are not displayed. Likely because we don't manipulate the touched status of the internal form control, yet that seems like a fairly low handing fruit to me, as we could handle the `editValue` as a proxy for the form control and mark as touched when it changes (based on update mode) and unconditionally when exiting edit mode.

That'll reduce the friction of adding validation to existing grids with custom templates at least, but also make it somewhat consistent across the board too, with the direct formControl still being offered to achieve optimal handling.

MayaKirova:
…as touched for better custom template handling.

Closes #12066  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -870,7 +870,6 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
             // while in edit mode subscribe to value changes on the current form control and set to editValue
             this.formControl.valueChanges.pipe(takeWhile(x => this.editMode)).subscribe(value => {
                 this.editValue = value;
```
damyanpetev, line 872: If this being a setter also goes and sets the value right back to the form control, wouldn't it sort of double dip with the validation logic? Wondering if we could use the form control as the backing for `editValue` entirely (assuming getting it from the service is quick), which might allow us to even skip this sync sub?

projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -97,10 +97,30 @@ export class IgxCell {
         public rowIndex: number,
         public column,
         public value: any,
-        public editValue: any,
+        public _editValue: any,
```
damyanpetev, line 100: Assuming `IgxCell` is internal (fairly certain it is), I guess this can remain public too. Could've been done with an extra method to assign the value too.

projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -97,10 +97,30 @@ export class IgxCell {
...
         public rowData: any,
         public grid: GridType) {
         this.grid.validation.create(id.rowID, rowData);
+        this.editValue = this._editValue;
```
damyanpetev, line 104: This will likely trigger an extra validation, though I think I saw the public API creating this with different value/editValue initially, so we might be stuck with this.

projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -97,10 +97,30 @@ export class IgxCell {
...
+    }
+
+    public get editValue() {
+        const formControl = this.grid.validation.getFormControl(this.id.rowID, this.column.field);
+        if (formControl) {
+            return formControl.value;
+        }
```
damyanpetev, line 111: Kind of surprised, this doesn't break the model bind with blur if the value changes, yet the getter still returns the old one. It works, but still perhaps it'd be better if we return the pending value for blur for consistency?

damyanpetev (commented):
Minor comments, of which only the getter is of some importance.

We should still leave an explanation in the docs that using the `[formControl]` directive to bind edit templates in combination with blur trigger will leave the cell `editValue` unsynced until blur/edit exit due to the way the form control handles its value.

skrustev:
Thee only thing I found s that an error is throw when `validationTrigger` is set to `change` only and adding row
 Steps:
 1) Click on add row
 2) Focus a cell with templated editor
 3) Observe the console log

projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -91,16 +91,39 @@ export interface IgxAddRowParent {
...
     constructor(
         public id,
         public rowIndex: number,
         public column,
         public value: any,
-        public editValue: any,
+        public _editValue: any,
         public rowData: any,
         public grid: GridType) {
         this.grid.validation.create(id.rowID, rowData);
```
damyanpetev, line 104: @MayaKirova There's one odd case where the grid public `updateCell` API creates a new `IgxCell` with a different edit value and the form control will be out of sync here. That's actually not new it seems, however, with these changes the so will be the new value used by the API call (it just uses the args) to perform the update.
That's fortunately for blur only and could be logged separately, if we consider it still a minor use, though it's API that should work regardless.
Ideally, we'd initialize the form control with the edit value regardless instead of setting it additionally.

projects/igniteui-angular/src/lib/grids/grid/grid-validation.service.ts
```diff
@@ -32,6 +32,7 @@ export class IgxGridValidationService {
                 const value = resolveNestedPath(data || {}, col.field);
                 const field = this.getFieldKey(col.field);
                 const control = new FormControl(value, { updateOn: this.grid.validationTrigger });
+                control.setValue(value);
```
damyanpetev, line 35: Um.. if the control was just initialized with the same value, I'm kind of missing the point of assigning it again. Any specific reason?

MayaKirova: Yes. It's due to a rather weird behavior when the initial value is undefined (which we use when creating an adding row for example, all fields have undefined as value).  When the form control is created for undefined it actually sets it to null instead:

So this:
```
new FormControl(undefined, { updateOn: this.grid.validationTrigger });
```

Then returns null for control.value.

Which breaks some tests.

dkamburov (approved):

PR #12084, opened 2022-09-16
Add some more template guard types.

MayaKirova:
Added template guards for:
- IgxRowEditTemplateDirective
- IgxRowEditTextDirective
- IgxRowAddTextDirective
- IgxRowEditActionsDirective
- IgxDragIndicatorIconDirective
- IgxRowDragGhostDirective
- IgxPivotValueChipTemplateDirective

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid.rowEdit.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid.rowEdit.directive.ts
@@ -1,5 +1,5 @@
 import { Directive, ElementRef, HostListener, Inject } from '@angular/core';
-import { GridType, IgxGridRowEditTemplateContext, IGX_GRID_BASE } from './common/grid.interface';
+import { GridType, IgxGridRowEditTemplateContext, IgxGridRowEditTextTemplateContext, IGX_GRID_BASE } from './common/grid.interface';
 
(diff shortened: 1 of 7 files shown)


projects/igniteui-angular/src/lib/grids/grid.rowEdit.directive.ts
```diff
@@ -1,29 +1,49 @@
...
+ }
 
 /** @hidden @internal */
 @Directive({
     selector: '[igxRowAddText]'
 })
-export class IgxRowAddTextDirective { }
+export class IgxRowAddTextDirective {
+    public static ngTemplateContextGuard(_directive: IgxRowAddTextDirective,
+        context: unknown): context is null {
```
mddragnev, line 32: I think that this should not be null but for example - 
Screenshot 2022-09-19 at 16 16 49
Because:
1. We do set `$implicit = null` if we are adding row
2. We should be consistence for example with `IgxDragIndicatorIconDirective` where we do not set `implicit` and we do not set context at all but we have set that the type of the context is `IgxGridEmptyTemplateContext`

mddragnev (changes requested):

mddragnev (approved):

PR #12094, opened 2022-09-16
feat(igx-grid): add isSelectedAllClicked to rowSelectionChanging event

### LINKED ISSUE #10961: igx-grid rowSelectionChanging isSelectAll flag
This is probably both a bug and feature request, because the igx-grid "select all" checkbox doesn't play nice with remote data sources

I am working on building an igx-grid with a remote data source (pretty much exactly like the examples, no paginator, but virtually scrolled), but I'm also adding row selections to the grid. One thing that I've noticed is that when you click the "select all" checkbox, it will only select the rows that the grid knows about at the time (whatever is loaded for the "page"). 

The ability to know if the "select all" checkbox was clicked in the `rowSelectionChanging` event, and add the ability to handle the event _asynchronously_ (promise/observable) to get the row keys of all items.

I have a work around by looking at the event's `currentTarget`, and seeing if it's the select all checkbox, but it would be nice to not have to do that just in case it changes in the future.

## Additional context
My page size is 50, as you can see the grid thinks that all items are selected, but when I scroll down to load the other items, it's obviously not selected, because the grid knew nothing about that record other than it existing based on the `totalItemCount`

image

Svetloslav15:
Closes #10961 

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


hanastasov (changes requested):
Let's write unit test for the new stuff:
1. Verify isSelectAllClicked contians the expected value
2. Test for remote data + paging/virtualization scenario - Verify that if the Select all checkbox is selected, newly fetched rows (fetched from scrolling or paging) will be automatically selected .

projects/igniteui-angular/src/lib/grids/headers/grid-header-row.component.ts
```diff
@@ -211,9 +211,9 @@ export class IgxGridHeaderRowComponent implements DoCheck {
         }
 
         if (this.grid.selectionService.areAllRowSelected()) {
-            this.grid.selectionService.clearRowSelection(event);
+            this.grid.selectionService.clearRowSelection(event, false);
```
hanastasov, line 214: It seems obvious that on clearRowSelection, isSelectAll would be false. Maybe this can be omitted.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -1175,6 +1175,16 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
 
+    /**
+     * A template reference for the template if all rows are selected.
+     *
+     * @example
+     * ```
+     * const areAllRowsSelected = this.grid.isSelectAllClicked;
+     * ```
+     */
+    public isSelectAllClicked = false;
```
kdinev, line 1186: There shouldn't be a property inside the grid for this. This is an event driven flag and it should exist only in the lifecycle of the interaction raising the event.

Also, `isSelectAllClicked` is not an appropriate name for the flag. There could be other interactions, not only `click` that cause all rows to be selected.

kdinev (changes requested):

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
hanastasov: I didn't quite understand what you mean. What defines " the lifecycle of the interaction raising the event."? 

The issue to solve -  subscribe on the `dataPreload` event - we need to know if the Select all checkbox is/was selected. Where would this flag live, if not on the grid?

kdinev: The issue described doesn't subscribe to the `dataPreload`, instead it subscribes to the `rowSelectionChanging` and tries to determine whether any data loaded in the future should also be selected, or not. The application scenario can use the flag from the `rowSelectionChanging` event to populate a component-level property with a flag of whether newly loaded data is to be selected, or not, but that shouldn't be done on a library level by populating a flag belonging to the grid itself.

hanastasov: We went ahead to see if implementing this scenario on library level would be ok. In the end, it seemed possible even without populating a flag in the grid itself.

Can evaluate the resolution after @Svetloslav15  make the final commits/

Svetloslav15: I've removed the flag from the component and now if the user wants to automatically select remote data when scrolling, they should subscribe to rowSelectionChanging and dataChanged event, and handle selection on application level.

hanastasov (changes requested):
image
Just clicked the header checkbox - the flag reports false, which is wrong.

hanastasov:
Along with the code, please prepare a sample (can use some of the existing samples) where to demonstrate the application level logic to achieve the following:

1) remote data with scroll > select all rows > scroll to new data > select the new data if all rows should be selected.
2) remote data with paging -> select all rows > go to next page > have all rows selected

kdinev:
I'm thinking about a scenario where the flag in the event wouldn't function properly. If there's an already loaded chunk `x` of data of size `y`, where `x < y`, and I, as a user, check the checkboxes for all `x` items individually, before scrolling to load more items, then on the last item, I will receive `allRowsSelected = true` in the selection event. In fact, I would have selected all currently rendered rows, but I still have more that should come from the server. 

I'm starting to think that we should create a new event for the select all checkbox itself, and fire it separate from the selection event, which was one of the suggestions in the issue. Have you considered that?

projects/igniteui-angular/src/lib/grids/selection/selection.service.ts
```diff
@@ -528,17 +529,11 @@ export class IgxGridSelectionService {
...
-        }
-
-        const dataItemsID = this.getRowIDs(this.allData);
-        return this.allRowsSelected = Math.min(this.rowSelection.size, dataItemsID.length) > 0 &&
-            new Set(Array.from(this.rowSelection.values()).concat(dataItemsID)).size === this.rowSelection.size;
+    public areAllRowSelected(newSelection?): boolean {
+        const selectedData = newSelection ? newSelection : [...this.rowSelection]
+        const allData = this.getRowIDs(this.allData);
+        const unSelectedRows = allData.filter(row => !selectedData.includes(row));
+        return this.allRowsSelected = this.allData.length > 0 && unSelectedRows.length === 0;
```
hanastasov, line 536: I don't see no place where `this.allRowsSelected` is read. It is only being set couple of times. Seems we should go without it at all.

hanastasov (changes requested):
Bug:
https://www.facebook.com/watch/?v=834989784341211

ddincheva:
The implemented solution seems to fix the issue but I have noticed a significant performance decrease so this definitely needs to be checked in detail:
Performance on master without these changes:
ScrollOnMaster

Scroll with these changes:
ScrolllAfterTheChanges

PR #12125, opened 2022-09-21
Blazor angular elements

gmurray81:
This is for tracking the changes required to emit the blazor/react grid api based on the elements API


projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -632,8 +671,11 @@ export interface GridType extends IGridDataBindable {
     isHierarchicalRecord?(record: any): boolean;
     columnToVisibleIndex(key: string | number): number;
     moveColumn(column: ColumnType, target: ColumnType, pos: DropPosition): void;
+    /* blazorSuppress */
```
mddragnev, line 674: Can we CS suppress those methods instaed of suppressing them for all platforms because they are not included in the type definitions for web components too but are 100% viable and easy to use. There are many more like that.

mddragnev: @gmurray81 Could you please take a look into this?

PR #12134, opened 2022-09-27
Add handling for column groups in elements grids.

MayaKirova:
Add handling for column groups in elements grids.


projects/igniteui-angular-elements/src/analyzer/elements.config.ts
```diff
@@ -16,11 +17,12 @@ export const registerComponents = [
...
...umnVisibleIndex", "groupBy", "clearGrouping", "isExpandedGroup", "toggleGroup", "selectRowsInGroup", "deselectRowsInGroup", "fullyExpandGroup", "toggleAllGroupRows", "getSelectedData", "getRowByIndex", "getRowByKey", "getCellByColumn", "getCellByKey", "pinRow", "unpinRow", "isRecordPinnedByIndex", "toggleColumnVisibility", "expandAll", "collapseAll", "expandRow", "collapseRow", "toggleRow", "getHeaderGroupWidth", "getColumnByName", "getColumnByVisibleIndex", "nextPage", "previousPage", "moveColumn", "paginate", "markForCheck", "addRow", "deleteRow", "updateCell", "updateRow", "getRowData", "sort", "filter", "filterGlobal", "enableSummaries", "disableSummaries", "clearFilter", "clearSort", "pinColumn", "unpinColumn", "reflow", "findNext", "findPrev", "refreshSearch", "clearSearch", "getPinnedWidth", "selectRows", "deselectRows", "selectAllRows", "deselectAllRows", "selectedColumns", "selectColumns", "deselectColumns", "deselectAllColumns", "selectAllColumns", "getSelectedColumnsData", "combineSelectedCellAndColumnData", "navigateTo", "getNextCell", "getPreviousCell", "openAdvancedFilteringDialog", "closeAdvancedFilteringDialog", "endEdit", "beginAddRowById", "beginAddRowByIndex"] },
```
damyanpetev, line 22: Hm, fairly certain this isn't picking up the content queries for the parents (since the query is for the column base class and the group overrides it). Does it actually work if the grid isn't notified of the groups? I'd assume there would be a case where something would break due to that

damyanpetev: Oh just saw the change in the strategy. Okay, I guess that'll do, though strictly speaking the query won't load subclass components without them explicitly providing themselves as the concrete query type. Wondering if there are cases where components look for such components separately, though for now we could just leave a comment to look into this down the line

damyanpetev (approved):

PR #12204, opened 2022-10-18
fix(action-strip): Hide action strip on mouse leave event of row

### LINKED ISSUE #12077: When leaving row hover igx-action-strip stays visible until cursor leaves the grid component.
## Description  
When using an `igx-action-strip` component inside of an `igx-grid` the action strip will stay visible after the cursor is no longer over a row. This is best seen when the grid is partially empty and has white space at the bottom. The expected behavior is that when the cursor is no longer over a row that row should not display the `igx-action-strip` component anymore.

Please find a [stackblitz](https://stackblitz.com/edit/angular-nlf1jc) of the issue and the attached screenshots.

 * igniteui-angular version: "^13.2.8"
 * browser: Google Chrome 105.0.5195.52 (Official Build) (64-bit) (cohort: Stable)

## Steps to reproduce  
(There is a working example in the stackblitz above)
1. Create a partially empty grid with an `igx-action-strip`
2. hover cursor over last row in grid and observe the presense the `igx-action-strip` content.
3. Move the cursor down until it is no longer over the last row but still hovering over the grid.
4. Observe that the `igx-action-strip` content is still visible on the last row.
5. Move cursor out of the grid and observe that the `igx-action-strip` content is no longer visible.

* It also occurs when hovering over the first row and moving the cursor up over the column headers.
## Result  

You should see the `igx-action-strip` stay visible even if the cursor is not over that row but still hovering over the grid. (see first screen  shot below.

## Expected result  
 The expected behavior is the `igx-action-strip` content should be hidden once the cursor leaves the row.

## Attachments  
2022-09-13 16_23_20-Infragistics Angular Components - StackBlitz

[StackBlitz Example](https://stackblitz.com/edit/angular-nlf1jc)

georgianastasov:
Closes #12077 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


dkamburov:
> The customer reported the issue for v 13.2 . Shouldn't this have PRs to 13.2.x and 14.2.x as well?

Sure, @georgianastasov can you apply the fix for the mentioned versions as well?

dkamburov (dismissed):

mddragnev:
Couple of things I noticed:
1. `hideOnRowLeave` getter does not work correct if we have action strip like this:
```
<igx-action-strip #actionstrip>
            <igx-grid-pinning-actions [asMenuItems]="true"></igx-grid-pinning-actions>
            <igx-grid-editing-actions [addRow]="true"></igx-grid-editing-actions>
</igx-action-strip>
```

When we click on the dropdown icon and hover over the pinning icon the action strip hides.
Note that this does not happen if we add menu items like this:
```
<igx-action-strip #actionstrip>
            <igx-grid-pinning-actions></igx-grid-pinning-actions>
            <igx-grid-editing-actions [addRow]="true"></igx-grid-editing-actions>
            <ng-template igxActionStripMenuItem>
                <button>test</button>
                <button>testsss</button>
            </ng-template>
</igx-action-strip>
```
There the check in the getter works correctly because we have that contentchildren collection populated.

2.  Also, when we have that directive (from the second example I mentioned) the getter returns false each and every time which results in actionstrip that does not hide at all.

3. The second thing is that if we want to hide action strip on row leave we should remove the grid handler - the `HostListener` on `mouseleave` in grid-base

dkamburov (approved):

ChronosSF (approved):

PR #12400, opened 2022-11-29
fix(grid): update response from emitRowSelectionEvent to have whole record

### LINKED ISSUE #12371: Grid row selection is inconsistent with and without primaryKey
## Description  
Selection API returns different types of data depending on whether `primaryKey` is set to `igx-grid`, `igx-tree-grid`, `igx-hierarchical-grid`, etc. In the case of `primaryKey` set, I get an array of primitive values. In the case of `primaryKey` not being set, I get an array of data objects behind the grid.

 * igniteui-angular version: 15.0.0
 * browser: all

## Steps to reproduce  

1. Create an app with any grid
2. Turn row selection on
3. Log `args.newSelection` on selection changing
4. Set a `primaryKey`
5. Log `args.newSelection` on selection changing

## Result  
Inconsistent results between `primaryKey` and no `primaryKey`.

## Expected result  
Consistent results. Data members should be provided in both cases.

### LINKED ISSUE #12502: owner of rowSelectionChanging is always undefined
## Description  
When handling grid's `rowSelectionChanging` event the owner of the event arguments is always undefined.

 * igniteui-angular version: 15.0.0
 * browser: all

## Steps to reproduce  
In [this](https://stackblitz.com/edit/angular-5qggu6?file=package.json) demo I am logging the owner in `rowSelectionChanging` like this:
``` typescript
  public handleRowSelection(event: IRowSelectionEventArgs) {
    console.log(event.owner);
    this.selectedRowsCount = event.newSelection.length;
    this.selectedRowIndex = event.newSelection[0];
    this.snackbarRowCount.open();
    this.snackbar.close();
  }
```
Open the demo, change the selection of any row, and observe the console. It is always logs `undefinde`.

## Result  
`owner` is `undefined`.

## Expected result  
`owner` should point to the grid component.

Svetloslav15:
This PR is fixing the issue, but it is only changing the response format of the event, to always return the whole rows, not only ids. 
We should consider if it is fine, or we should refactor the whole logic around grids selection and make it consistent
regardless of whether there is a primaryKey or not.

Closes #12371
Closes #12502

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


Svetloslav15 (commented):
grid-virtual
When we use grid remote virtualization with scroll sample, select a row then when we scroll down and the new chunk of data is loaded, select row from it and we can see that the newSelection array is storing only the selected rows from the current chunk and currently selected rows are missing.

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-selection.service.ts
```diff
@@ -183,8 +183,8 @@ export class IgxTreeGridSelectionService extends IgxGridSelectionService {
      * populates the rowsToBeSelected and rowsToBeIndeterminate sets
      * with the rows which will be eventually in selected/indeterminate state
      */
-    private calculateRowsNewSelectionState(args: any) {
-        this.rowsToBeSelected = new Set<any>(args.oldSelection ? args.oldSelection : this.getSelectedRows());
+    private calculateRowsNewSelectionState(args: any, shouldConvert = false) {
+        this.rowsToBeSelected = new Set<any>(args.oldSelection ? shouldConvert ? args.oldSelection.map(r => r[this.grid.primaryKey]) : args.oldSelection : this.getSelectedRows());
```
hanastasov, line 187: this.rowsToBeSelected will contain one undefined item when PK is not set on the grid

projects/igniteui-angular/src/lib/grids/tree-grid/tree-grid-selection.service.ts
```diff
@@ -160,15 +159,16 @@ export class IgxTreeGridSelectionService extends IgxGridSelectionService {
      * @param visibleRowIDs list of all visible rowIds
      * @returns a new set with all direct parents of the rows within rowsToBeProcessed set
      */
-    private collectRowsChildrenAndDirectParents(rowsToBeProcessed: Set<any>, visibleRowIDs: Set<any>, adding: boolean): Set<any> {
+    private collectRowsChildrenAndDirectParents(rowsToBeProcessed: Set<any>, visibleRowIDs: Set<any>, adding: boolean, shouldConvert: boolean): Set<any> {
         const processedRowsParents = new Set<any>();
-        Array.from(rowsToBeProcessed).forEach((rowID) => {
+        Array.from(rowsToBeProcessed).forEach((row) => {
+            const rowID = shouldConvert ? row[this.grid.primaryKey] : row;
```
hanastasov, line 165: if PK is not set this result in rowID === undefined, which is then added to selection Set and possibly causing bugs. SHould be fixed.

hanastasov (changes requested):

projects/igniteui-angular/src/lib/grids/selection/selection.service.ts
```diff
@@ -420,57 +432,100 @@ export class IgxGridSelectionService {
...
+        this.emitRowSelectionEvent(newSelection, [this.getRowDataById(rowID)], removed, event, selectedRows);
+    }
 
+    public selectPivotRowById(rowID, clearPrevSelection: boolean, event?): void {
         const selectedRows = this.getSelectedRows();
-        const newSelection = clearPrevSelection ? [rowID] : this.rowSelection.has(rowID) ?
-            selectedRows : [...selectedRows, rowID];
-        const removed = clearPrevSelection ? selectedRows : [];
-        this.emitRowSelectionEvent(newSelection, [rowID], removed, event);
+        const newSelection = clearPrevSelection ? [rowID] : this.rowSelection.has(rowID) ? selectedRows : [...selectedRows, rowID];
```
hanastasov, line 448: PivotGrid supports only single selection, so maybe this line can be simpler. I could not think of a scenario where newSelection would contain more than one row.

Anyway, for Pivot I was not able to see any selection in the event arguments:

image


I used the Pivot Grid sample.

hanastasov (changes requested):

projects/igniteui-angular/src/lib/grids/selection/selection.service.ts
ddincheva: PivotGridSelection

hanastasov (changes requested):
1) Add this sample https://stackblitz.com/edit/angular-z9ynz5?file=src/app/grid/grid-remote-paging-defaultTemplate-sample/remote-paging-default-template.component.html with `primaryKey="ID"`

1) Select rows 2,5,6 on first page
2) Put a breakpoint in selection.service.ts, line 453
3) Go to next page and add row to selection:
image

When you go back to previous page, only row 6 is kept in selection

hanastasov (dismissed):

MayaKirova (approved):

hanastasov (approved):

PR #12423, opened 2022-12-06
Don't allow entering edit mode if PrimaryKey is not set - 14.2

### LINKED ISSUE #12393: Grid throws error when trying to edit
1. Enable rowEditing, or rowActions for grid without PrimaryKEy set
2. Enter edit mode for a row

## Result  
image

## Expected result  
Not to throw error and:
1) either display the editing overlay

OR

2) dont allow edntering edit mode if PrimaryKey is not set

georgianastasov:
Closes #12393   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
+++ b/projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
@@ -117,7 +117,9 @@ export class IgxGridEditingActionsComponent extends IgxGridActionsBaseDirective 
         // be sure row is in view
         if (grid.rowList.filter(r => r === row).length !== 0) {
             grid.gridAPI.crudService.enterEditMode(firstEditable, event);
-            firstEditable.activate(event);
+            if(!grid.gridAPI.crudService.nonEditable){
+                firstEditable.activate(event);
+            }
         }
         this.strip.hide();
     }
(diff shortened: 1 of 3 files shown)


hanastasov (approved):

hanastasov:
This PR partially changes current grid behavior in the following way:

**Current behavior for grid with enabled editing, no `primaryKey` set**
If a cell or row is editable, user is able to enter edit mode. This edit mode is buggy (for example, for row edit UI will lack the Done, Cancel buttons due to bug), and in the end no succesful transaciton is done). In the console, a warning is displayed that `primaryKey` needs to be set to the grid.

**New behavior**
If `primaryKey` is not set, same warning is displayed in the console, but grid will not even try to display the editing UI.

dkamburov:
@hanastasov @georgianastasov @sdimchevski 
I'm not sure about it. Here my thoughts(please correct me if I'm wrong). We currently do throw such warning, but only when the row editing mode is enabled. Now this will apply to cell editing as well, right?

Also do we really need this for both editing modes? Isn't cell editing working without primary key?

And the bigger concern If we go with this fix. We previously didn't stop users from entering edit mode and now we do, sounds like a **breaking change**. Should be included in the changelog and drop fixes for 14.2. And maybe there were some scenarios were this isn't required, but now it will be.

georgianastasov:
@hanastasov @dkamburov @sdimchevski 
The fix will only apply when `rowEditing `is set to true, but without `primaryKey`.

On the other hand, when the `rowEditing `property is not set, and editing is performed as cell editing again without `primaryKey`, then the fix is not applied and the given cell enters edit mode.

The given changes are related to this issue #12393 in order not to enter row edit mode without _editing overlay_ and without displaying the following error _Cannot read properties of undefined (reading 'getBoundingClientRect')_, but only to display a warning.

This also relates perhaps to **breaking chang**e as @dkamburov mentioned.

PR #12517, opened 2023-01-12
fix(chip): icons alignment in prefix/suffix containers

### LINKED ISSUE #12516: IgxChip icons are misaligned when added in igx-prefix and igx-suffix
## Description  
IgxChip icons are misaligned when added in 'igx-prefix' and 'igx-suffix' trough the content projection instead of using 'igxPrefix' and 'igxSuffix' directives directly on the chip icons.

## Steps to reproduce  

1. Add chip with 'igx-prefix' and 'igx-suffix' components instead of using 'igxPrefix' and 'igxSuffix' directives.
` <igx-chip class="chip">
     <igx-prefix>
         <igx-icon class="icon">
                    place
          </igx-icon>
       </igx-prefix>
      <igx-suffix>
           <igx-icon class="icon">
                clear
            </igx-icon>
        </igx-suffix>
                Country
</igx-chip>`

## Result  
Icons are misaligned in the chip.
chip-icons

## Expected result  
Icons to be aligned inside of 'igx-prefix' and 'igx-suffix'

## Attachments  

dobromirts:
Closes #12516 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
@@ -401,7 +401,7 @@
     %igx-chip__prefix,
     %igx-chip__suffix {
         @include ellipsis();
-        display: inline-block;
+        display: flex;
         vertical-align: middle;
         max-width: $chip-max-width;
     }
--- a/projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
@@ -401,9 +401,13 @@
     %igx-chip__prefix,
     %igx-chip__suffix {
         @include ellipsis();
-        display: flex;
+        display: inline-block;
         vertical-align: middle;
         max-width: $chip-max-width;
+
+        > igx-icon { 
+            display: block;
+        }
     }
 
     %igx-chip__suffix {


simeonoff:
@SisIvanova Please verify all scenarios in which the chip can be used, why we've set the display type to inline-block in the first place.

projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
```diff
@@ -401,7 +401,7 @@
     %igx-chip__prefix,
     %igx-chip__suffix {
         @include ellipsis();
-        display: inline-block;
+        display: flex;
```
simeonoff, line 404: I don't think this is a good solution as it will force sibling chip elements to appear one below the other. The `inline-flex` type would be better if we want to make it a flex element, but I believe there is a very specific reason why we went with `inline-block` in the first place. It could have been IE11, but I can't recall anymore.

SisIvanova: It was `inline-block` so that the `ellipsis` mixin can work. If we still want to have `ellipsis` both in the chip content and in the prefix/suffix, the `inline-flex` won't work either.
https://github.com/IgniteUI/igniteui-angular/pull/5783

simeonoff: @SisIvanova can we fix the issue then in another way without changing the display type?

SisIvanova: I believe this should work:

%igx-chip__prefix,
%igx-chip__suffix {
        @include ellipsis();
        display: inline-block;
        vertical-align: middle;
        max-width: $chip-max-width;

        > igx-icon { 
            display: block;
        }
}

SisIvanova (approved):

PR #12523, opened 2023-01-13
fix(paginator): Display '1 of 1' when there are no records to show

### LINKED ISSUE #12335: Grid Paginator state when there are no records to show
Related to https://github.com/IgniteUI/igniteui-angular/pull/12253 that removes the condition to hide the paginator where there's no data or visible columns. Although as it turned out that didn't apply to to some grids:
image
While hiding in the grids certainly works around the "non-valid" state of the paginator, it's a standalone component and we might want to have it handle such scenarios and thus improve the component itself and its use in the grids.
For example:
```html
<igx-paginator></igx-paginator>
<igx-paginator [totalRecords]="0"></igx-paginator>
```
currently produce:
image
Without setting total records is probably a bit harsh but should still be handled, and as for `0` records could easily be achieved when bound to a collection `length` and that collection is empty or not yet populated.

I was going to suggest to at least match the zero pages with `0 of 0`, but I figured it'd probably be more user-friendly to go with `1 of 1`. After all, when you're doing pages, there's theoretically at least one page - that one you're currently on, even if there's nothing on it.
I'll take some form of invalid state indication like `1 of -` or similar, though I feel the previous suggestion would be much easier to implement.

viktorkombov:
Closes #12335 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/paginator/pager.component.html
+++ b/projects/igniteui-angular/src/lib/paginator/pager.component.html
@@ -29,7 +29,7 @@
             paginator.resourceStrings.igx_paginator_pager_text
         }}&nbsp;</span
     >
-    <span>{{ paginator.totalPages }}</span>
+    <span>{{ paginator.totalPagesIndicator }}</span>
 </div>
 <button
     [title]="paginator.resourceStrings.igx_paginator_next_page_button_text"
--- a/projects/igniteui-angular/src/lib/paginator/pager.component.html
+++ b/projects/igniteui-angular/src/lib/paginator/pager.component.html
@@ -29,7 +29,7 @@
             paginator.resourceStrings.igx_paginator_pager_text
         }}&nbsp;</span
     >
-    <span>{{ paginator.totalPagesIndicator }}</span>
+    <span>{{ paginator.totalPages || 1 }}</span>
 </div>
 <button
     [title]="paginator.resourceStrings.igx_paginator_next_page_button_text"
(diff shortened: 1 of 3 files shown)


katherinedragieva:
The solution works as expected.

In my opinion I think we should leave it `"x of x"`, because currently this `-` that we added looks a bit strange to me. What do you think @damyanpetev?

image

katherinedragieva:
@viktorkombov let's display it  `1 of 1`, instead of `1 of -`

damyanpetev:
@viktorkombov @katherinedragieva The preferred option I put in the issue description was `1 of 1`. Technically an UX decision, but I still think it makes the most sense.

projects/igniteui-angular/src/lib/paginator/pager.component.html
```diff
@@ -29,7 +29,7 @@
             paginator.resourceStrings.igx_paginator_pager_text
         }}&nbsp;</span
     >
-    <span>{{ paginator.totalPages }}</span>
+    <span>{{ paginator.totalPagesIndicator }}</span>
```
damyanpetev, line 32: PS: Since this is simple enough, it can easily go without one more getter like `paginator.totalPages || 1` :)

hanastasov:
@viktorkombov , could you please make a new PR withiut this commits history hoise?

PR #12587, opened 2023-01-26
Date-picker: fix selection on focus

### LINKED ISSUE #12566: IgxDatePicker is not selected on focus
## Description  
When the IgxDatePicker is clicked or tab over, that is, when the focus event fires, the IgxDatePicker is not selected. For example the
IgxTimePicker is selected as expected.

 * igniteui-angular version: 14.2.x 15.0.x

## Steps to reproduce  

Open [this sample](https://stackblitz.com/edit/angular-agmzk2?file=src/app/datepicker-sample/datepicker-sample.component.html).

1. Click or tab navigate to the igx-date-picker.

## Result  
IgxDatePicker is not selected on focus.

## Expected result  
IgxDatePicker should be selected on focus.

## Attachments  
issue

PlamenaMiteva:
Closes #12566  

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/date-picker/date-picker.module.ts
+++ b/projects/igniteui-angular/src/lib/date-picker/date-picker.module.ts
@@ -24,12 +24,12 @@ import { IgxDatePickerComponent } from './date-picker.component';
         FormsModule,
         CommonModule,
         IgxIconModule,
+        IgxDateTimeEditorModule,
         IgxMaskModule,
         IgxCalendarModule,
         IgxInputGroupModule,
         IgxPickersCommonModule,
-        IgxTextSelectionModule,
-        IgxDateTimeEditorModule,
+        IgxTextSelectionModule,        
         IgxCalendarContainerModule,
     ]
 })


ChronosSF:
This cannot be the fix. Please, discuss alternatives in how the mask behaves with @damyanpetev .

Also I am pretty certain that if the bundle order is what ultimately decides which functionality takes precedence, then that bundle order is not guaranteed because of tree shaking in an user app that can specify these modules in a different order in the app's own dependencies.

valadzhov (approved):

damyanpetev:
@ChronosSF @PlamenaMiteva @Lipata 
I agree actually, it's very doubtful this would be a solution we can rely upon.
Keep in mind, this is a regression - i.e. the selection directive was added, verified and working for a while at least. So this behavior/handling could've changed recently, though can't be sure. The actual runtime handling depending on module registration is quite likely tied to the actual runtime order - i.e. the compiled js code, not our module definitions. Which makes this look kinda shaky in the sense that those can and likely will be rearranged during compilation and tree-shaking and there's a good chance this will still break in certain app uses. In a dependency system it doesn't make sense to have expectations of order, probably a reason there's no guidance to do this.

In any case, while this could help some cases, I wouldn't really consider the issue fixed.
If this is a locked-in focus behavior, we could simply handle the selection in the components - calling `select()` at the right time instead of relying on the directive will hardly weight them down. Could also consider the mask behavior implemented in the `igc-mask` if we feel it's pretty much used always used with the select directive anyway.

ChronosSF:
Closing these PRs as both me and Damyan agree this should be fixed in a different way.

PR #12593, opened 2023-01-27
fix(IgxDialog, IgxOveralyService): change closeOnEscape/closeOnOutsideSelect while the dialog is open - master

### LINKED ISSUE #12537: Dialog: Change close configuration after opening is not working
## Description  
Whenever a dialog is already opened, the close behavior cannot be changed anymore.
"closeOnEscape" and "closeOnOutsideSelect" are ignoring new incoming values.

 * igniteui-angular version: "@infragistics/igniteui-angular": "14.1.1",
 * browser: Version 109.0.5414.74 (64-Bit)

## Steps to reproduce  

1. Click the button to open the dialog
2. In the dialog, click the button to disable the close behavior
3. Press ESC-Key or click besides the dialog

## Result  
The dialog closes

## Expected result  
The dialog shouldn't close

## Attachments  
Code sample of the template part (html)
```
<button (click)="myDialog.open()">open dialog</button>

<igx-dialog #myDialog [closeOnEscape]="!disableActions"
            [closeOnOutsideSelect]="!disableActions" message="test message">
   <div class="dialog-footer" igxDialogActions>
       <button (click)="disableActions = !disableActions">Escape key and outside shouldn't work anymore</button>
  </div>
</igx-dialog>
```

in the .ts part you just need to declare the var
```
...
export class MyComponent {
   disableActions: boolean = false;
   ...
}
```

georgianastasov:
Closes #12537   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -938,4 +946,44 @@ export class IgxOverlayService implements OnDestroy {
             info.closeAnimationPlayer.finish();
         }
     }
+
+    /**
+     * @hidden
+     */
+    public proxy = new Proxy(
```
wnvko, line 953: the proxy should not be exposed to the outside world. Only the overlay should know of the existence of the Proxy.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -807,21 +812,24 @@ export class IgxOverlayService implements OnDestroy {
         }
     }
 
-    private removeOutsideClickListener(info: OverlayInfo) {
-        if (info.settings.modal === false) {
```
wnvko, line 811: when the overlay is `modal` we do not add event to the `document` but to the modal overlay element. Do not remove this check!
Instead, we should remove the handler from the overlay element. Look at the `addOutsideClickListener` to see where the event is added when the overlay is `modal`

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -328,10 +329,40 @@ export class IgxOverlayService implements OnDestroy {
             return null;
         }
 
+        let proxy = {
```
wnvko, line 332: add the `info` object here as well as all the methods you need like this:
```ts
let proxy = {
   ...your code...
   info: info,
   addOutsideClickListener: this.addOutsideClickListener,
   ... more functions here ...
}

This will allow you then to call methods like this:
```ts
this.addOutsideClickListener(this.info);
```

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -327,11 +328,44 @@ export class IgxOverlayService implements OnDestroy {
             console.warn('Overlay was not able to attach provided component!');
             return null;
         }
+        
+        settings = Object.assign({}, this._defaultSettings, settings);
+
+        let target = {
```
wnvko, line 334: No need of the `target`. All the properties and methods added in the `target` should be added in the handler.

Actually the `settings` is the target. If we set `info.settings = new Proxy(target, handler);` anyone having access to the `info.settings` will not receive `OverlaySettings` interface but some strange object, which is not ok.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -327,11 +328,44 @@ export class IgxOverlayService implements OnDestroy {
...
+            settings: settings,
+            addOutsideClickListener: this.addOutsideClickListener,
+            addCloseOnEscapeListener: this.addCloseOnEscapeListener,
+            removeOutsideClickListener: this.removeOutsideClickListener,
+            removeCloseOnEscapeListener: this.removeCloseOnEscapeListener
+        };
+
+        let handler = {
+            get(obj, prop) {
```
wnvko, line 343: no need to override `get` as you return exactly what base getter will return.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -684,6 +714,8 @@ export class IgxOverlayService implements OnDestroy {
                 this._overlayElement = null;
             }
             this.removeCloseOnEscapeListener();
+        } else if (!this._overlayInfos[this._overlayInfos.length - 1].settings.closeOnEscape) {
```
wnvko, line 717: if there are two overlays open, e.g. dialog in another dialog, and if the first one has `closeOnEscape` set to `true`, when the second one is closed this will remove the handler of the first one, right?
This does not look ok.

projects/igniteui-angular/src/lib/services/overlay/utilities.ts
```diff
@@ -130,6 +130,14 @@ export interface OverlayAnimationEventArgs extends IBaseEventArgs {
     animationType: 'open' | 'close';
 }
 
+export interface Handler extends ProxyHandler<OverlaySettings> {
```
wnvko, line 133: can we go with a better name here?
`OverlaySettingsProxyHandler` or something similar but shorter - just to know what this is when someone see it.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
georgianastasov: When the overlay, e.g. the dialog is closed in the `detach()` method, only the click listener is removed `this.removeOutsideClickListener(info);`.

Then the `cleanUp()` method is called to clear everything else. In this method, when checking `if (this._overlayInfos.length === 0) {}` or that is, only when there is an overlay, e.g. a dialog, then the escape listener is also removed `this.removeCloseOnEscapeListener();`. Removing the escape listener only happens here.

When another dialog or nested dialog is open, if the first one has `closeOnEscape` set to `false` and the last open dialog is closed, then the escape listener will never be removed - we will not enter the check because we have another open dialog, and this dialog has set `closeOnEscape` to `false` and there will still be an escape listener.

This check checks, after closing and detaching the last open overlay, e.g. dialog, whether the one before it has the `closeOnEscape ` property set to `false`, and if so, the escape listener is removed.

Regarding the question, if there are two overlays open, e.g. dialogs, and if the first one has `closeOnEscape ` set to `true`, when the second one is closed this will not remove the handler of the first one.

projects/igniteui-angular/src/lib/services/overlay/utilities.ts
georgianastasov: Yes of course. Тhe name will be changed.

ChronosSF:
As discussed with @damyanpetev , this user's request has potential downsides. He'll communicate his concerns in the issue and see if there aren't better solutions to the customer's case that don't involve such changes.

PR #12626, opened 2023-02-08
feat(transactions): clean pending state changes resulting in initial

### LINKED ISSUE #12270: Change detection cell value in cellEditor
Hi,
If change the value of a cell in edit mode, the dropdown menu tells me i have 1 change like expected. If i then change the value back to the original value before i entered edit mode(still in the same edit mode, same row), it still tells me i have 1 change. 

If possible it would be nice to have the cell remember the value in the cell from before entering edit mode so that it shows the changes from how it used to be before editing. 

Regards

RivaIvanova:
Closes #12270   

**NOTE**: Test when `batchEditing` is enabled as well.

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/services/transaction/base-transaction.ts
```diff
@@ -157,6 +157,8 @@ export class IgxBaseTransactionService<T extends Transaction, S extends State> i
             state = { value: this.cloneStrategy.clone(transaction.newValue), recordRef, type: transaction.type } as S;
             states.set(transaction.id, state);
         }
+
+        this.cleanState(transaction.id, states);
```
damyanpetev, line 161: This seams reasonable. Since the base only supports the pending stack (for row edit) it also didn't compare against the original. Guessing that's the note about testing with `batchEdit` on and off.

I'd actually expand on that - there's also Row Adding that relies on this that should be verified.
There's also the Hierarchical variant of the service that's specialized for the Tree Grid - so that should also be tested. As far as I can see the HierarchicalTransactionService also overrides this `updateState` method, so it _could_ be affected, though it seems like it's handling a very specific additional case.

projects/igniteui-angular/src/lib/services/transaction/igx-transaction.ts
```diff
@@ -283,46 +283,7 @@ export class IgxTransactionService<T extends Transaction, S extends State> exten
...
-                if (state.type !== TransactionType.DELETE && Object.keys(state.value).length === 0) {
-                    states.delete(id);
-                }
-            } else {
-                if (state.recordRef === state.value) {
-                    states.delete(id);
-                }
-            }
-        }
+        this.cleanState(transaction.id, states);
```
damyanpetev, line 286: Didn't find any references to the old behavior in the specs and nothing was explicitly testing or relying on it looking at the passing tests.
Unless @wnvko can recall what the comment was about, I think cleaning up the state makes sense for consistency between reported pending and applied state and we should update the spec to actually describe the state handling in such scenarios.

wnvko (approved):

damyanpetev (approved):

PR #12229, opened 2022-10-25
Angular elements project

damyanpetev:
Closes #  

### Additional information (check all that apply):
 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [ ] All relevant tags have been applied to this PR
 - [ ] This PR includes unit tests covering all the new code ([test guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Test-implementation-guidelines-for-Ignite-UI-for-Angular))
 - [ ] This PR includes API docs for newly added methods/properties ([api docs guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines))
 - [ ] This PR includes `feature/README.MD` updates for the feature docs
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [ ] This PR includes `CHANGELOG.MD` updates for newly added functionality
 - [ ] This PR contains breaking changes
 - [ ] This PR includes `ng update` migrations for the breaking changes ([migrations guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Update-Migrations))
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

PR #12641, opened 2023-02-14
Fixing angular elements tests

skrustev:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
+++ b/projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
@@ -181,7 +181,6 @@ export class IgxActionStripComponent extends DisplayDensityBase implements After
         @Optional() @Inject(DisplayDensityToken) protected _displayDensityOptions: IDisplayDensityOptions,
         public cdr: ChangeDetectorRef) {
         super(_displayDensityOptions);
-        this._originalParent = this._viewContainer.element.nativeElement?.parentElement;
     }
 
     /**
@@ -267,6 +266,7 @@ export class IgxActionStripComponent extends DisplayDensityBase implements After
                 button.actionClick.emit();
             }
         });
+        this._originalParent = this._viewContainer.element.nativeElement?.parentElement;
     }
 
     /**
@@ -305,9 +305,11 @@ export class IgxActionStripComponent extends DisplayDensityBase implements After
     public hide(): void {
         this.hidden = true;
         this.closeMenu();
-        if (this.context && this.context.element && !!this._originalParent) {
+        if (this._originalParent) {
             // D.P. fix(elements) don't detach native DOM, instead move back. Might not matter for Angular, but Elements will destroy
             this.renderer.appendChild(this._originalParent, this._viewContainer.element.nativeElement);
+        } else if (this.context && this.context.element) {
+            this.renderer.removeChild(this.context.element.nativeElement, this._viewContainer.element.nativeElement);
         }
     }
 
--- a/projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
+++ b/projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
@@ -138,7 +138,7 @@ export class IgxActionStripComponent extends DisplayDensityBase implements After
 
     /**
      * Hide or not the Action Strip based on if it is a menu.
-     * 
+     *
      * @hidden
      * @internal
      */
@@ -305,7 +305,7 @@ export class IgxActionStripComponent extends DisplayDensityBase implements After
     public hide(): void {
         this.hidden = true;
         this.closeMenu();
-        if (this.context && this.context.element) {
+        if (this.context && this.context.element && !!this._originalParent) {
             // D.P. fix(elements) don't detach native DOM, instead move back. Might not matter for Angular, but Elements will destroy
             this.renderer.appendChild(this._originalParent, this._viewContainer.element.nativeElement);
         }
--- a/projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.spec.ts
+++ b/projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.spec.ts
@@ -1202,7 +1202,7 @@ describe('igxExpansionPanel', () => {
             expect(grid.attributes.getNamedItem('ng-reflect-auto-generate').nodeValue).toEqual('true');
             expect(grid.attributes.getNamedItem('ng-reflect-width').nodeValue).toEqual(fixture.componentInstance.width);
             expect(grid.attributes.getNamedItem('ng-reflect-height').nodeValue).toEqual(fixture.componentInstance.height);
-            expect(grid.childElementCount).toEqual(6);
+            expect(grid.childElementCount).toEqual(7);
         }));
         it('Should apply all appropriate classes on combo initialization_image + text content', fakeAsync(() => {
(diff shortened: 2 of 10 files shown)


dkamburov (approved):

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
```diff
@@ -305,7 +305,7 @@ export class IgxActionStripComponent extends DisplayDensityBase implements After
     public hide(): void {
         this.hidden = true;
         this.closeMenu();
-        if (this.context && this.context.element) {
+        if (this.context && this.context.element && !!this._originalParent) {
```
damyanpetev, line 308: As discussed, as Angular can initialize the component before it's attached to the DOM - the ctor is a bad place to gather the original parent. To mirror this, it should be available to save during `show` actually.
This also got me thinking that it's probably not a bad idea to keep the old detach behavior as fallback if there's no parent - I think that might be technically possible if you dynamically create the actions and don't attach them to a view, right?

projects/igniteui-angular/src/lib/expansion-panel/expansion-panel.spec.ts
```diff
@@ -1202,7 +1202,7 @@ describe('igxExpansionPanel', () => {
             expect(grid.attributes.getNamedItem('ng-reflect-auto-generate').nodeValue).toEqual('true');
             expect(grid.attributes.getNamedItem('ng-reflect-width').nodeValue).toEqual(fixture.componentInstance.width);
             expect(grid.attributes.getNamedItem('ng-reflect-height').nodeValue).toEqual(fixture.componentInstance.height);
-            expect(grid.childElementCount).toEqual(6);
+            expect(grid.childElementCount).toEqual(7);
```
damyanpetev, line 1205: lol, won't ask why is the expansion panel checking the grid's children... but erm, okay

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1402,7 +1402,7 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
             col = this.allChildren.filter(c => !c.columnGroup && !c.hidden)[0] as any;
         }
         if (this.columnLayoutChild) {
-            return this.parent.childrenVisibleIndexes.find(x => x.column === this).index;
+            return this.parent.childrenVisibleIndexes.find(x => x.column === this)?.index;
```
damyanpetev, line 1405: Would it not be better if this returned `|| -1` if the find fails? Otherwise not really doing what the getter signature implies.
And the fail I assume is to do with the async population of children? Just trying to figure out if there's a bigger problem this would mask.

damyanpetev: Actually, strike the last, the actual Angular tests should have no issue with async children

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3772,8 +3768,8 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
         }
         this.resetColumnsCaches();
         this.resetColumnCollections();
-        this.resetCachedWidths();
         this.resetForOfCache();
+        this.resetCachedWidths();
```
damyanpetev, line 3772: Didn't see a test that specifically relates to this, wondering if it's relevant

projects/igniteui-angular/src/lib/grids/grid/grid.master-detail.spec.ts
```diff
@@ -704,10 +704,12 @@ describe('IgxGrid Master Detail #grid', () => {
 
     describe('Integration', () => {
         describe('Paging', () => {
-            it('Should not take into account expanded detail views as additional records.', fakeAsync(() => {
+            it('Should not take into account expanded detail views as additional records.', async () => {
                 fix = TestBed.createComponent(DefaultGridMasterDetailComponent);
                 grid = fix.componentInstance.grid;
+
                 fix.detectChanges();
+                await wait(); // trigger full ngAfterViewInit cycle to be sure rendered = true
```
damyanpetev, line 712: Wouldn't a `tick()` do the trick as well?
I mean `ngAfterViewInit` is run with the first `fix.detectChanges()` before this, so we're not really waiting on it. Is this about the rendered flag? That's a promise so wait/tick would do indeed, but it's somewhat obscure. If you want the event, just wait for it?
```ts
await grid.rendered.toPromise();
```
Then again if that's to get the pipes to populate the `pagingState`, again I'm kind of confused what necessitated this

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
skrustev: It's related to tests:
 * `should correctly initialize when there are initially pinned columns` 
 * `should correctly initialize when there are initially pinned columns` 

It is due to the `resetCachedWidths()` using the grid API in order to calculate the grid container size and when the test scenario calls `grid.unpinnedWidth` the last width calculated doesn't include if scrollbar is rendered, since the widths are not reset after `resetCachedWidths()`.

The initial issue was that the grid throws error when columns don't have width set and was addressed with 0c09e9f commit.

projects/igniteui-angular/src/lib/grids/grid/grid.master-detail.spec.ts
skrustev: As discussed this is related to enabling paging after the grid is rendered and is related to [this](https://github.com/IgniteUI/igniteui-angular/pull/12229/files#diff-9511c230760b50c458554bf036523d4de868b74f0f9fbfcb80d5428068b938caL75) change, because it initially has null values for `page` and `perPage` and when paginator is added this triggers the paging pipe. For now the pipe is triggered by [this pipeTrigger](https://github.com/IgniteUI/igniteui-angular/pull/12229/files#diff-440a9d274756495c7bb4e7acf64e5ff3757d060999c43e932fe7654b565f793eR3867), but since this is a change that should affect mainly angular elements.

We discussed that maybe best would be to introduce a separate paging pipe trigger when the paginator is added like `pipeTrigger`, to keep the old behavior of the grid.

damyanpetev: Yes, at this point the best thing to do is revert this change in the template (and related tests) for this PR and leave any changes to https://github.com/IgniteUI/igniteui-angular/pull/12647 since I moved the paging changes there anyway

projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
skrustev: Yes when created dynamically the `originalParent` would be null on show, but when it is rendered at least in Angular it would go through the AfterViewInit hook and it will set the `originalParent `. I've added the old behavior just in case though.

projects/igniteui-angular/src/lib/grids/grid/grid.master-detail.spec.ts
damyanpetev: reverted

damyanpetev (approved):

PR #12656, opened 2023-02-16
fix(row-pinning): preserve pinned state when editing the primary key

### LINKED ISSUE #12569: Grid - Editing the primary key in a pinned row unpins the row
## Description  
Editing the primary key in a pinned row unpins the row

## Steps to reproduce  
in the following sample edit the ID column in one of the pinned rows
https://stackblitz.com/edit/angular-ujej8f?file=src/app/grid/grid-row-pinning/grid-row-pinning.component.html

## Result  
After the edit is complete the row is no longer pinned

## Expected result  
The row remains pinned after editing the primary key

dobromirts:
Closes #12569 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


ddincheva:
When you try to edit a column which is set as primary key from a pinned row an error is thrown:
Pinned row editing error

ddincheva:
With the latest fix, there is little inconsistent behavior in some cases when we edit the cell which is the Primary key the pinned rows are reordered:
PinningRowReorder

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -151,6 +151,15 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
             this.grid.validation.update(cell.id.rowID, newRowData);
         }
         if (this.grid.primaryKey === cell.column.field) {
+            if (this.grid.pinnedRecords.length > 0) {
+                const cellRow = this.grid.getRowByIndex(cell.rowIndex);
```
ddincheva, line 155: why do you need to get the row component .. I believe that all the necessary info used below can be directly accessed from the cell and the edit event args like in the selection below

ddincheva (changes requested):

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -543,6 +551,49 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
...
+        grid._pinnedRecordIDs.splice(insertIndex, 0, rowID);
+    }
+
+    public unpin_row(rowID: any): void {
+        const grid = (this.grid as any);
+        const index = grid._pinnedRecordIDs.indexOf(rowID);
+        if (index === -1) {
+            return;
+        }
+        const eventArgs = this.get_pin_row_event_args(rowID, null , null, false);
```
hanastasov, line 575: Why two null values are passed for the `index` and `row` param? Both are available.

projects/igniteui-angular/src/lib/grids/api.service.ts
```diff
@@ -543,6 +551,49 @@ export class GridBaseAPIService<T extends GridType> implements GridServiceType {
...
+    }
+
+    public get_pin_row_event_args(rowID: any, index?: number, row?: RowType, pinned?: boolean) {
+        const eventArgs: IPinRowEventArgs = {
+            isPinned: pinned ? true : false,
+            rowID,
+            row,
+            cancel: false
+        }
+        if (typeof index === 'number') {
```
hanastasov, line 591: It is strange to keep a check checking if index would not be a number. Is this needed if no null is passed as index from the unpin_row method?

hanastasov (commented):
1) I used the row pinning dev demo
2) Disable rowEditing, set ID column (the PK) to be visible and editable
3) Pin the first row
4) Edit the value of ID column in the pinned row

Result:
`rowPinning` event is emitted twice. In both emissions, the value of `row` argument is missing - first time it is null, second time it is undefined,

PR #12708, opened 2023-02-28
Pivot bidi sync

gmurray81:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -133,16 +124,33 @@ export const cloneValue = (value: any): any => {
     }
 
     if (isObject(value)) {
+        if (seen.has(value)) {
+            return seen.get(value);
```
damyanpetev, line 128: _Right, so you're essentially adding what [`structuredClone()`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) does - i.e. deep clone that keeps map of refs and reuses them. Bit newer API, but since IE is out and at least for now we don't need this in a worked we might consider it for deep _object_ clones? Seems to do exactly the same thing https://stackblitz.com/edit/js-ys6dff?file=index.js_

Anyway, wondering why did maintaining of same refs become important? Cuz this clone is pretty central I believe, possibly used to process grid's data. Point being, with large enough data this could add noticeable enough overhead to unrelated operations without any benefit. _Then again it's quite likely the native clone will be optimized and could be faster than the old code even with ref maps._

Since this is related to the Pivot, I only see one use in the util and with my limited pivot exp and little to no typing, kind of hard to figure out what that's about. In all honesty, if we're deep cloning something it should be to mess with it internally, i.e. it's not coming back to the public API without a good reason, right?

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
```diff
@@ -151,6 +151,21 @@ export class IgxPivotGridComponent extends IgxGridBaseDirective implements OnIni
...
+     * @remarks
+     * Returns the new dimension collection and its type:
+     * @example
+     * ```html
+     * <igx-pivot-grid #grid [data]="localData" [height]="'305px'"
+     *              (dimensionsChange)="dimensionsChange($event)"></igx-grid>
+     * ```
+     */
+    @Output()
+    public configurationChange = new EventEmitter<IPivotConfigurationChange>();
```
damyanpetev, line 168: @gmurray81 would this also work as `<prop>Change`-type event that directly emits the config itself so it could also work with two way binding? Though it being an object makes it a bit iffy with immutability and change detection.

damyanpetev: Actually, there are already events for dimension and value changes, right? Assuming the other bit that can change through the UI would be rows and values? If we've started down that path, would it actually be useful to expose those too instead?

projects/igniteui-angular/src/lib/core/utils.ts
gmurray81: The issue was in the pivot pipeline where it was cloning the pivot dimensions before transforming the data. It was calling this method. But because we have a add a back reference to the pivot dimension in the blazor logic we were getting a stack overflow here, because the ref cycle wound up causing this clone logic to enter a visitation loop. It may be that that one place in the pipeline could just call structured clone instead possibly, rather than altering the implementation of this clone method. However, it seemed dangerous that this clone method doesn't like reference cycles if it is used elsewhere where one might be introduced.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
gmurray81: yeah, if you wanted it to be public, you could make it a bananabox style event and that would work just fine for our purposes. We'd just need to change the attribute path that describes where to find the updated value.

yeah, I saw the other events, but for our purposes it would be useful to have an event that fires whenever anything changes in the pivot config, and that the args is the entire pivot config payload for us to marshal accross (unless we start trying to play games with synchronizing only portions of the tree when the grid updates them). I wanted to start with something reasonably brute force though, since I want to get to the point where this behaves acceptably for people that want to make updates both programatically and via the data selector. 

There's some remaining work to be done still possibly since we need to make sure that if we replicate new items coming from the data selector or the grid outward, that these can roundtrip back in appropriately.

This isn't necessarily straightforward though since the grid or selector may create new dimensions, I'm guessing, that have callback methods on them, without necessarily also setting an identifier that would allow for the callback to be fetched again if the dimension were round-tripped through blazor. The best option would be if we could make it always set enough information on the dimension that it can definitely round trip through blazor, failing that we could possibly consider something sneaky, whereby when a dimension is round tripping outward, and it has callback functions we need to ensure round trip, we could technically cache them somewhere and put an identifier in something like aggregatorScript, so we can fetch the callback function later when it round trips back.

I can't tell if I'm making sense or not :)

gmurray81:
I adjusted to not impact the perf of the clone method. There could be some subtle impact since it needs to check if it has an omit set within the loop. This could be avoided by bifurcating the loop into 2 versions, or duplicating the clone method to use a different one if the omit set it needed. 

Again, making externalObject non-enumerable might possibly be another option too, but I think we should avoid that for now since there's some risk associated with it since me _might_ enumerate that in various places. Not to mention there are multiple places where we create those back-refs, and for different purposes. So I'd propose the current state of this PR is the safer route, and we can consider other options at a future date.

skrustev (approved):

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.pipes.ts
```diff
@@ -82,7 +82,9 @@ export class IgxPivotAutoTransform implements PipeTransform {
     protected processCollectionToPivotRecord(config: IPivotConfiguration, collection: any[]): IPivotGridRecord[] {
         const pivotKeys: IPivotKeys = config.pivotKeys || DEFAULT_PIVOT_KEYS;
         const enabledRows = config.rows.filter(x => x.enabled);
-        const allFlat: IPivotDimension[] = PivotUtil.flatten(enabledRows);
+        const allFlat: IPivotDimension[] = cloneArray(
+            PivotUtil.flatten(enabledRows), true, new Set<string>(["externalObject"]));
```
damyanpetev, line 86: Oh, um, yeah... can't say I'm a huge fan of baking in an external 'magic' property name in the source.

damyanpetev: The more I look at this, the more I'm thinking we should consider a pivot (tehe) to some sort of clone strategy that can be overridden to serve per-case needs. I seem to recall there already is something like that added, though obviously not used for this clone.

damyanpetev: In any case, the above suggestion might end up taking more time and API adjustments, so hey - might have to be okay with this for now and just plan on replacing it with a strategy down the line

gmurray81:
Sorry, had meant to comment before. I think that should be fine.

damyanpetev:
> I guess this one can be closed as the other got merged #12859

Actually we still need the meta flags I believe and I've cleaned the changes to just those I believe and will still merge those :)

PR #12711, opened 2023-03-01
fix(grid cell edit exit): add revert for new value on esc

### LINKED ISSUE #12568: Grid should not report "new value" in cell edit exit event when edit is canceled
## Description  
If a cell edit is canceled (for example by pressing the escape key) when the cell edit exit event is called the event args contain the edited value in the "newValue" property even though this value is not the true value. 
 

## Steps to reproduce  
create a grid and handle the cellEditExit event then type a value into one of the columns and press escape. For example in the following sample. double click in the price column and change 18 to 23 then press escape

https://stackblitz.com/edit/angular-jf8qui?file=src/app/grid/grid-editing-events/grid-editing-events.component.ts

## Result  
The cell edit exit event reports the old value as 18 and the new value as 23

## Expected
the cell edit exit event does not report the new value as 23. the new value should either not be reported or it should be the current value of 18.

monacodelisa:
Closes #12568 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -411,6 +411,9 @@ export class IgxGridNavigationService {
                 }
 
                 if (this.grid.crudService.cellInEditMode || this.grid.crudService.rowInEditMode) {
+                    // when esc is pressed and editing is canceled, new value is reverted to the old value
+                    const editCellEsc = this.grid.crudService.cell;
+                    editCellEsc.editValue = editCellEsc.value;
                     this.grid.crudService.endEdit(false, event);
```
ddincheva, line 417: This fix needs to be addressed in the crudService .. there are a lot of operations in the grid that triggered the escape of edit mode for the cell .. /so the cellEditExit event is emitted/ .. you can see the table in the [wiki](https://github.com/IgniteUI/igniteui-angular/wiki/Grid-Editing#common-behaviors)

monacodelisa: Ok thank you will look into implementing that change

ddincheva:
Also, I have some concerns regarding this issue, because if this fix is introduced we are making impossible the following scenario for example: when we type in some cell and for some reason the user decade to perform soring/ filtering or whatever other action in the grid which discard the value .. we can not be able to cancel the exit event based on the fact if the current value in the cell is changed or not... so maybe we need to think further and discuss the logged issue.

PR #12748, opened 2023-03-09
feat(grid): add a property to filter out columns from autogen

### LINKED ISSUE #12577: Way to filter auto-generated columns
It would be helpful if there's a way to select columns to be auto-generated or filter out unwanted columns. Sometimes the data can carry meta props (e.g. `__tag`) or perhaps it could be useful to just drop the primary key column so it can't be visible in the UI.

Initially was thinking of simply opting to ignore private/meta patterns like `__`-prefixed fields, but perhaps a hook/event into the column key gathering process can be more flexible and avoid any behavior changes in the process.
Currently, `Object.keys` are gathered from the first item in data for auto-generate - that is simple enough to be wrapped in a default callback that can be overridden (doesn't need to be a full-blown strategy I think). Being able customize this might be useful in other situations too where perhaps the first item is perhaps not ideal to generate from - I've seen such inconsistency in the wild with only the first item with extra meta fields.

Another suggestion was to make `columnInit` cancellable, though that'd be breaking with the current emitter arg. 

Looked at both `columnInit` and `rendered` events as possible hooks - the init event fires per-column so it can't be used to reject a column. The rendered event fires a bit late and the full set of columns is generated and you get a flash of the old, before the filtered columns are re-rendered.

## Additional context
While this is more of an inconvenience for other platforms atm (Blazor), in general outside declarative columns, actual grid API to manage them is a bit lacking.

ChronosSF:
Closes #12577 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


kdinev:
@ChronosSF `igx-tree-grid` and `igx-hierarchical-grid` have logic already to exclude certain columns from generation. WIth the newly added input, the excluded values would be ignored by the tree and hierarchical grids.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -212,6 +212,17 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
     @Input()
     public autoGenerate = false;
 
+    /**
+     * Gets/Sets a list of property keys to be excluded from the generated column collection
+     * 
+     * @example
+     * ```html
+     * <igx-grid [autoGenerate]="true" [autoGenerateExclude]="['prop1', 'prop2']"></igx-grid>
```
kdinev, line 220: Add an example of a record in the data source.

kdinev (changes requested):

ChronosSF:
> @ChronosSF `igx-tree-grid` and `igx-hierarchical-grid` have logic already to exclude certain columns from generation. WIth the newly added input, the excluded values would be ignored by the tree and hierarchical grids.

I don't think that's correct. I autogenerated both and both generated their "special" props as columns. There is also no custom implementation of the `autogenerateColumns` function to be found in hierarchical/tree grids. There is one for Pivot but it is very custom and additionally autogeneration has a different purpose entirely there so after discussing it with Maya I decided to simply hide the property for pivot.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -7142,7 +7159,7 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
         const fields = this.generateDataFields(data);
         const columns = [];
 
-        fields.forEach((field) => {
+        fields.filter((field) => !this.autoGenerateExclude.includes(field)).forEach((field) => {
```
damyanpetev, line 7162: @ChronosSF Any reason this is not baked into `generateDataFields()` called above? Since that's where all the field filtering happens in overrides for Tree/HGrid too, sort of feels like that's where it belongs

ChronosSF: Because I am blind, though it would work the same either way.

damyanpetev (dismissed):

Issue #6931, opened 2020-03-19
igx-grid overrides any applied css classes


kdinev:
## Description  
The class attribute value applied to the `igx-grid` is overridden by the component.

 * igniteui-angular version: 9.0.3
 * browser: N/A

## Steps to reproduce  

1. Define an igx-grid and apply a class
2. Inspect the grid element

## Result  
Class is not applied.

## Expected result  
Class should be applied.

PR #12749, opened 2023-03-09
Host class binding refactor

damyanpetev:
This re-does the fix for #6931 and covers it properly. The `class` input with `attr.class` host binding approach works with Angular's virtual tree, but it's not compatible with Elements where the attribute becomes the input as well and causes a circular feedback loop.

The pattern was trying to work around an issue that has long been resolved with Ivy https://github.com/angular/angular/issues/7289, so the built-in class binding seems to do just fine tracking the classes it applied and removing them without messing with customer-applied ones through various means. Also, the "custom classes should be at the end" looks like a misconception to me, order of the classes applied doesn't matter AFAIK. Quick demo: https://stackblitz.com/edit/angular-8c54ab?file=src%2Fmain.ts

Refactored other components using this pattern and added coverage where needed.

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/.eslintrc.json
```diff
@@ -41,7 +41,14 @@
...
         "id-match": "off",
-        "no-underscore-dangle": "off"
+        "no-underscore-dangle": "off",
+        "no-restricted-syntax": [
+          "error",
+          {
+            "selector": "Decorator[expression.callee.name=HostBinding] Literal[value='attr.class']",
+            "message": "Do not use `attr.class` HostBinding in public-facing components"
+          }
+        ]
```
damyanpetev, line 51: Threw in a lint rule for good measure. Might be a bit overkill, but really don't want to leave it to chance this is never used since Angular's test won't catch it. Adapted 
Also it was quite fun, [`no-restricted-syntax`](https://eslint.org/docs/latest/rules/no-restricted-syntax) with AST selectors seems to be pretty much all-powerful 😁
Playground link for the curious: [https://typescript-eslint.io/play/#...](https://typescript-eslint.io/play/#ts=4.5.2&sourceType=module&showAST=es&code=PQKgsAUABFCmDOAbAlgOwC5VQewLQCcF19kBjdWAE13gE8MBDADwC4oBtSGbqAclnz5s+XgBouPKAG8JkqACJ4sRLHLD5beQBFVwhumHtYTAA6F48ZNlQA6Ug0QrYN1AwC2sALwAJbPHQAQmiUaADmALpQADLIFPgO7ABuDgCuXrz6xHaIDBa84fLi0HIKHhYMobAaCgCMUFrYWNiYKUpQAAaZ+Nm58O1Qvv5BqCGooVBoUCYpAEYopLgAZgykYVCk2G4m1rAY8PKy3AC+RXIyxXKKyqoG+NXauvG3RqbmltZ2Dk4u7l6DgcEwuEXmYEO9bAx8KEUh49jYAAw2ZKINKeDLoLKkHJ5AqnEryMrwCpVTQAJnqjRwLTanQx3SxvX6-2Go3Gk2mczISxWaw2Wx2ewOF2OEnCkBAwEgkEoqhyhHW1n8Az8AJGYQA3FKIKBwNAQFAAAIAC2QlBlqC4EsgDIsUAoSvOMANzMBYwAFOjMdj4LwAJQSDnzKCVTBGlUAYW9bt9bH8JDG0kOhHQKXwqD4vE1xSOkCOQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGYBfEWg&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3QAacDUjoAHoizJ46IngCGqACL9YmNXl3Cwo9FIC+IM0A)

projects/igniteui-angular/src/lib/grids/column-actions/column-actions.component.ts
```diff
@@ -62,7 +62,7 @@ export class IgxColumnActionsComponent implements DoCheck {
      * this.columnHidingUI.cssClass = 'column-chooser';
      * ```
      */
-    @HostBinding('attr.class')
+    @HostBinding('class')
```
damyanpetev, line 65: Would also be nice if someone can tell me if/why this component is public API

dkamburov: `cssClass` property is public in most of our components(i.e., badge, avatar, banner). Looks to me like a forgotten `@hidden @internal` tags and documented by mistake.

dkamburov (approved):

projects/igniteui-angular/src/lib/grids/column-actions/column-actions.component.ts
damyanpetev: Oh just noticed this is actually not hidden/internal. I only replaced the binding, because I can't figure out the deal with the component itself - it's in the public API, but no mention/usage in the docs/samples I can see. Just two grid dev demos.

Anyways, I could just mark this particular binding as private like the rest if you'd like. The other components and refactoring their host bindings to be private - that goes bit out of scope for this PR I think, but could be done separately  for sure.

PR #12766, opened 2023-03-14
Elements analyzer refactor

rkaraivanov:


projects/igniteui-angular-elements/src/analyzer/printer.ts
```diff
@@ -0,0 +1,132 @@
...
+    #printer: ts.Printer;
+    #configPath: string;
+    #configFile: ts.SourceFile;
+
+    // TODO: Change to the default output path
+    public out = 'elements.config.ts';
+
+    constructor(public config = 'projects/igniteui-angular-elements/src/analyzer/config.template.ts') {
+        this.#configPath = path.join(process.cwd(), config);
+        this.#program = ts.createProgram([this.#configPath], {});
```
damyanpetev, line 21: Unless TypeScript optimizes this behind the scenes (almost certain it doesn't cuz it crating this takes a bit and afterwards again), this could also be a point of optimization - I think it should be possible to use the config file + main entry for the lib to create a single program that has the it and can resolve the entire lib source.

projects/igniteui-angular-elements/src/analyzer/printer.ts
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        return ts.factory.createObjectLiteralExpression(properties);
+    }
+
+    run(config: Map<ts.InterfaceType, ComponentMetadata>) {
+        const outFile = path.join(path.dirname(this.#configPath), this.out);
+        const placeholder = ts.createSourceFile(outFile, '', ts.ScriptTarget.Latest);
+
+        // REVIEW: We the current approach this seems unnecessary
```
damyanpetev, line 83: Not sure I get it, the config can add files that are not part of the original, thus not imported. Those are now hardcode in the template, but it seems cumbersome to add those instead of letting the analyzer discover them. So this could very well still be needed I think

damyanpetev (approved):
Some suggestions, but I'll still approve it as is regardless as it's excellent progress none the less :)

PR #12770, opened 2023-03-15
Chip - fix discrepancy between WebC and Angular

### LINKED ISSUE #12695: Chip Component does not have a disabled state
## Description  
When using the chip component in an App Builder app, i enabled its disabled state, but it's not working, not in App Builder or in the generated apps.

image

So i tested in an angular sample in Stackblitz, and added the disabled property, and i couldnt get it to work either:
image

### LINKED ISSUE #12771: Chip - styles discrepancy between Ignite UI for Angular and WC in all dark themes
## Description  
When an app is generated with App Builder there are styles discrepancy between Ignite UI for Angular and WC in all dark themes. The chip in WC looks different compared to the chip in the Angular product:

WC

image

Angular

image

## Steps to reproduce  

1. Run the following samples - [WC](https://filetransfer.infragistics.com/public.php?service=files&t=0575de27770f5b598b352fef815689ed), [Angular](https://filetransfer.infragistics.com/public.php?service=files&t=467528bc2550be95a16e438579c14de7)
2. See the differences between them.

## Expected result  
There should no be differences between Ignite UI for Angular and WC.

desig9stein:
Closes #12695
Closes #12771

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/chips/chip.component.html
```diff
@@ -14,27 +14,31 @@
     (over)="onChipOverHandler($event)"
     (dropped)="onChipDrop($event)">
 
-    <div #selectContainer [ngClass]="selectClass(selected)">
-        <ng-container *ngTemplateOutlet="selectIconTemplate"></ng-container>
-    </div>
+    <div class="igx-chip__start" #selectContainer>
+        <div *ngIf="selected" [ngClass]="selectClass(selected)">
```
skrustev, line 18: So with this change you are pretty much eliminating the select icon animation. Is this the desired effect?

desig9stein: Yes, the animation is eliminated

projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
```diff
@@ -319,32 +307,102 @@
...
+                    color: var-get($theme, 'hover-selected-text-color');
+                    background: var-get($theme, 'hover-selected-background');
+                    border-color: var-get($theme, 'hover-selected-border-color');
+                }
+            }
+        }
+    }
+
+    %igx-chip__item {
+        display: grid;
```
skrustev, line 364: I think because of the use of display: grid, currently the grid resizing does not autosize the filter header cells with the default filter chips. 
The main issue is that when the grid autosizes a column(either through api or double click), it resets the header group cell styles like `width` and `minWidth` temporarily and then calculates the size however the browser renders the headers. Here's exact [link](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/core/utils.ts#L287)

This can be reproduced on the Grid Filtering demo, if you switch to filter row and double click the firs column.
If you know a way this can be dealt with through styles only?

desig9stein: Test with the last commit please, the problem should be solved.

skrustev: Thanks, that seem to did it. I've updated the tests and all should be fine now.

simeonoff (dismissed):

ChronosSF:
Discussing this with @kdinev and @damyanpetev , we'd like to see some "before & after" screenshots for example in grid groupby, filter row and other places chips are used as these visual changes always come with potential issues.

simeonoff (approved):

simeonoff:
@ChronosSF @kdinev @damyanpetev 

The requested screenshots:
chips-comparison

kdinev:
> @ChronosSF @kdinev @damyanpetev
> 
> The requested screenshots: chips-comparison

The change looks good to me. I notice that the new font is darker, which makes the chips more readable and they stand out better in the grid context, however I'm not sure if this is the desired effect. We seem to have a lot of different font colors in the grid at the moment. Should this be so?

simeonoff:
> > @ChronosSF @kdinev @damyanpetev
> > The requested screenshots: chips-comparison
> 
> @simeonoff The change looks good to me. I notice that the new font is darker, which makes the chips more readable and they stand out better in the grid context, however I'm not sure if this is the desired effect. We seem to have a lot of different font colors in the grid at the moment. Should this be so?

We merely implemented the chip as it is in the design system. Perhaps this question should be addressed to @sdimchevski.

kdinev:
@sdimchevski Could you take a look at the result and tell us whether there's any discrepancies with the current design system definitions?

simeonoff:
@viktorkombov why did you remove the verified label?

sdimchevski:
@ChronosSF @kdinev @damyanpetev @simeonoff 
Looks good to me too, because of the font color change it's more readable now.

Also it's normal for the disabled chip to not have optimal contrast, if the contrast is great then it will be enabled. Disabled chip does not pass contrast requirements in Material v2 too. 

Font has always been 14px Titillium Web Regular 400 in our material design system. I'm not sure why it was implemented in Angular as SemiBold 600 at the first place.

I think we need to update only the disabled chip's background in the context of the grid as it blends with the grid's background color. So in this case the background color of the disabled chip should be adjusted to grays 300 instead of 200.

simeonoff:
> @ChronosSF @kdinev @damyanpetev @simeonoff Looks good to me too, because of the font color change it's more readable now.
> 
> Also it's normal for the disabled chip to not have optimal contrast, if the contrast is great then it will be enabled. Disabled chip does not pass contrast requirements in Material v2 too.
> 
> Font has always been 14px Titillium Web Regular 400 in our material design system. I'm not sure why it was implemented in Angular as SemiBold 600 at the first place.
> 
> I think we need to update only the disabled chip's background in the context of the grid as it blends with the grid's background color. So in this case the background color of the disabled chip should be adjusted to grays 300 instead of 200.

@desig9stein Could you contextually update the background of the chip. I know this is not how it's supposed to be, but this is on Svilen's request.

desig9stein:
@sdimchevski, What about material dark and other themes, what should be the colors there?

sdimchevski:
> @sdimchevski, What about material dark and other themes, what should be the colors there?

In the context of the grid the background color of the disabled chip in material should be adjusted to grays 300 in both light and dark. For the other themes I have to check

sdimchevski:
@desig9stein 
**Material**:
dark - let's keep the background to grays 200 and update the content to become grays 300
**Fluent**:
dark - content should become grays 400
**Bootstrap**:
dark - content should become grays 400
**Indigo**: 
light - border should become grays 300
dark - content should become grays 300

sdimchevski (approved):

PR #12774, opened 2023-03-15
fix(hgrid): Remove extra code setting columns to incorrect column list if there are no layouts.

### LINKED ISSUE #12252: Hierarchical Grid dynamic row islands can remove auto-generated columns
## Description  
If the top level H Grid is loaded with auto-generated columns and row layouts are added with delay, their possible columns trigger the columns list update handling, and since the root list is empty - the empty list gets assigned to the internal grid collection, essentially deleting the generated columns.

This is related to #12251 and to do with the same handling block. Still, quite a lot of stars to align for this one, thus the low prio.

 * igniteui-angular version: 15.0.x
 * browser: N/A

## Steps to reproduce  

1. Use the sample attached to #12251
3. Set `autoGenerate` to true on the HGrid so the grid renders all columns initially
4. Open the sample
5. Hit the Toggle child layouts button

## Result  
After child layouts are enabled the grid ends up empty

## Expected result  
The grid columns and layouts should render correctly

skrustev:
Fix not detecting columns change for elements grid.

Closes #12252

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
@@ -1116,14 +1116,9 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
         const colLength = this.columnList.length;
         const topCols = this.columnList.filter((item) => colsArray.indexOf(item) === -1);
         if (topCols.length > 0) {
+            const columnsChanged = this.columns.length !== colLength;
             this.updateColumns(topCols);
-            if (recalcColSizes && this.columns.length !== colLength) {
-                this.calculateGridSizes(false);
-            }
-        } else if (colLength) {
-            // temp(D.P.): Without layouts (or not loaded), colsArray is empty, so init w/ gathered columns
-            this.updateColumns(this.columnList.toArray());
-            if (recalcColSizes && this.columns.length !== colLength) {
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
```diff
@@ -1116,14 +1116,9 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
         const colLength = this.columnList.length;
         const topCols = this.columnList.filter((item) => colsArray.indexOf(item) === -1);
         if (topCols.length > 0) {
+            const columnsChanged = this.columns.length !== colLength;
```
damyanpetev, line 1119: As far as I can see this is pretty much cleaning up the changes in this branch as they should no longer be needed, in which case I'd actually appreciate if there are no changes left here - as in just delete the else case and drop the extra `columnsChanged` variable so it matches the master source. Thanks :)

skrustev: The `columnsChanged` is regarding the issues described [https://github.com/IgniteUI/igniteui-angular/issues/12252#issuecomment-1470388757](here).
This is because once the `updateColumns` finished, the columns will be the same and `this.columns.length !== colLength` will be false.

PR #12781, opened 2023-03-16
fix(grid): invoking pipe recalc on column change

ChronosSF:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grouping/group-by-area.component.html
```diff
@@ -2,16 +2,16 @@
     <ng-container *ngFor="let expression of chipExpressions; let last = last;">
         <igx-chip
             [id]="expression.fieldName"
-            [title]="(expression.fieldName | igxGroupByMeta:grid).title"
+            [title]="(expression.fieldName | igxGroupByMeta:grid:grid.pipeTrigger).title"
             [displayDensity]="grid.displayDensity"
-            [removable]="(expression.fieldName | igxGroupByMeta:grid).groupable"
-            [draggable]="(expression.fieldName | igxGroupByMeta:grid).groupable"
-            [disabled]="!(expression.fieldName | igxGroupByMeta:grid).groupable"
+            [removable]="(expression.fieldName | igxGroupByMeta:grid:grid.pipeTrigger).groupable"
```
kdinev, line 7: Instead of propagating the whole grid through the pipe, can't just the column `groupable`, `header` and `key` properties be passed in as arguments?

kdinev (changes requested):

projects/igniteui-angular/src/lib/grids/grouping/group-by-area.component.html
damyanpetev: That's the entirety of the pipe though - it's only there to cache the `grid.getColumnByName` and returns a stripped-down version of it with just `groupable` and field, which is why the UI doesn't update if the columns or their props change.
Which is why I'm sort of interested in the need of the pipe in the first place - it's caching a call to a `find` basically and sort of wondering how much perf it saves or perhaps it was done for convenience. If the latter - the solution could easily be to drop it

PR #12783, opened 2023-03-17
refactor(pivot-data-selector): allow init without initially set grid

damyanpetev:
Changes for Elements extracted from #12229
The Pivot Data Selector is quite tied to the grid by design, however if the grid is not assigned initially (as it happens in Elements) there were no checks to allow the component to safely initialize so it can accept the grid with a short delay.

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


damyanpetev:
@dkamburov Yeeeah, not entirely sure that'll help all that much. The second one is _attached to a grid_ with the empty config :)
Also a bit odd being a full-on object, having default like that - if those were separate array props perhaps, but the object feels off to have a value when a user hasn't provided one.

Assuming the issue with the first one is not showing the panel headers text and empty text - that'd be because the Data Selector is using the same resource strings as the grid I believe? What's more, it's trying to load those from the assigned grid itself:
https://github.com/IgniteUI/igniteui-angular/blob/b2a3d1ca3053e6a82e0abd6208e375782f399931/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-data-selector.component.html#L62
Whether those belong to the grid aside, I believe the Data selector can still access the global resource strings as fallbacks if that's required.
And the badge `0` could be a better fallback here I'm assuming?
https://github.com/IgniteUI/igniteui-angular/blob/b2a3d1ca3053e6a82e0abd6208e375782f399931/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-data-selector.component.html#L66
I added this a bit hastily to make it not throw without the grid, didn't really pay attention that was length instead of text.

Again, my main concern was to allow the Data Selector to merely initialize without the grid being immediately assigned. I'm under no impression the component makes sense on its own without being attached to a grid and expect one to be assigned still.
It's just that happens with a slight delay in Elements and other frameworks that can't gather ref as they render.

skrustev (approved):

PR #12812, opened 2023-03-23
fix(select) Update the select size inside the context of the paginator.

### LINKED ISSUE #12805: The default paging page-size dropdown is too big
## Description  
The paging dropdown seems to be too big compared to other elements of the Pager in 15.1.0 . It's default width also makes the editing area too small  now so not all options are fully visible when chosen.

 * igniteui-angular version: 15.1.0-rc.0
 * browser: n/a

## Steps to reproduce  

1. You can observe the behavior in any sample with paging in the staging samples browser but the easiest to investigate is probably http://localhost:4200/pagination from the dev demos 
2. It looks especially broken when you choose 100 and 500 because the edit area can't fit the numbers and displays an ellipsis

## Result  
Screenshot 2023-03-22 172552

## Expected result  

I guess it doesn't have to look exactly like before but at least somewhat more reasonable. For reference this is 15.0

Screenshot 2023-03-22 172634

desig9stein:
- make the select inside the paginator to always be in compact density.
- change the width of the select to be a bit wider in order to show at least four digits before showing an ellipsis.

Closes #12805 

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/paginator/_paginator-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/paginator/_paginator-theme.scss
@@ -117,8 +117,8 @@
 
     %igx-page-size__select {
         display: flex;
-        max-width: rem(90px);
-        min-width: rem(80px);
+        max-width: rem(114px);
+        min-width: rem(100px);
     }
 
     %igx-page-nav {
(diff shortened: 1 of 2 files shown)


ChronosSF:
@desig9stein , wouldn't it be better if the select receives the paginator's display density?

I tried the paginator with all 3 and with compact the select takes the whole height and looks a bit out place

ChronosSF:
> @desig9stein , wouldn't it be better if the select receives the paginator's display density?
> 
> I tried the paginator with all 3 and with compact the select takes the whole height and looks a bit out of place

Nvm, I checked it and it then covers the whole space in all densities.

PR #12856, opened 2023-04-04
refactor(*): enable noImplicitOverride and clean up

damyanpetev:
Enabled `noImplicitOverride` in tsconfig.
Tried to remove all unintentional overrides (constructor parameter auto-properties from copy-pastes, identical property definitions). Also caught a few components overriding Angular hooks like `ngOnInit` without calling super, thus missing on logic or cleanup, so added those where it looked like it made sense.

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/moving/moving.drag.directive.ts
```diff
@@ -23,35 +23,35 @@ export class IgxColumnMovingDragDirective extends IgxDragDirective implements On
...
         private cms: IgxColumnMovingService,
         _platformUtil: PlatformUtil,
     ) {
         super(cdr, element, viewContainer, zone, renderer, _platformUtil);
     }
 
-    public ngOnDestroy() {
+    public override ngOnDestroy() {
         this._unsubscribe();
+        super.ngOnDestroy();
```
damyanpetev, line 46: @hanastasov There seems to be quite a lot of cleanup in the base `ngOnDestroy` so I added a call. Doesn't seem like those are overridden explicitly, but also behind some checks so not sure if they are in use in this directive and if the base call is really needed. Won't mind if someone takes a look,

damyanpetev: IIRC, same for the moving drop directive

hanastasov: What's behind a check does not seem to be called even for the base directive. bcause the ghostElement is already cleared in `onTransitionEnd`. I believe the call is excessive, but does not do any harm

Lipata (dismissed):

Lipata (approved):

PR #12859, opened 2023-04-05
enhancement(pivotGrid): Add pivotConfiguration change event and clone Value strategy.

skrustev:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/data-operations/data-clone-strategy.ts
+++ b/projects/igniteui-angular/src/lib/data-operations/data-clone-strategy.ts
@@ -1,4 +1,4 @@
-import { cloneValue } from "../core/utils";
+import { cloneValue, cloneValueCached } from "../core/utils";
 
 export interface IDataCloneStrategy {
     clone(data: any): any;
@@ -7,5 +7,11 @@ export interface IDataCloneStrategy {
 export class DefaultDataCloneStrategy implements IDataCloneStrategy {
     public clone(data: any): any {
         return cloneValue(data);
+    }
+}
+
+export class CachedDataCloneStrategy implements IDataCloneStrategy {
+    public clone(data: any): any {
+        return cloneValueCached(data, new Map<any, any>);
     }
 }
--- a/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
@@ -33,7 +33,7 @@ import { IgxForOfSyncService, IgxForOfScrollSyncService } from '../../directives
 import { ColumnType, GridType, IGX_GRID_BASE, RowType } from '../common/grid.interface';
 import { IgxGridCRUDService } from '../common/crud.service';
 import { IgxGridSummaryService } from '../summaries/grid-summary.service';
-import { DEFAULT_PIVOT_KEYS, IDimensionsChange, IgxPivotGridValueTemplateContext, IPivotConfiguration, IPivotDimension, IPivotValue, IValuesChange, PivotDimensionType } from './pivot-grid.interface';
+import { DEFAULT_PIVOT_KEYS, IDimensionsChange, IgxPivotGridValueTemplateContext, IPivotConfiguration, IPivotConfigurationChange, IPivotDimension, IPivotValue, IValuesChange, PivotDimensionType } from './pivot-grid.interface';
 import { IgxPivotHeaderRowComponent } from './pivot-header-row.component';
 import { IgxColumnGroupComponent } from '../columns/column-group.component';
 import { IgxColumnComponent } from '../columns/column.component';
@@ -74,6 +74,7 @@ import { IFilteringStrategy } from '../../data-operations/filtering-strategy';
 import { IgxPivotValueChipTemplateDirective } from './pivot-grid.directives';
 import { IFilteringOperation } from '../../data-operations/filtering-condition';
 import { IgxGridValidationService } from '../grid/grid-validation.service';
+import { DefaultDataCloneStrategy, IDataCloneStrategy } from '../../data-operations/data-clone-strategy';
 
 let NEXT_ID = 0;
 const MINIMUM_COLUMN_WIDTH = 200;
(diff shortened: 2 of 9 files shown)


MayaKirova (dismissed):

gmurray81 (dismissed):

projects/igniteui-angular/src/lib/data-operations/data-clone-strategy.ts
```diff
@@ -9,3 +9,9 @@ export class DefaultDataCloneStrategy implements IDataCloneStrategy {
         return cloneValue(data);
     }
 }
+
+export class CachedDataCloneStrategy implements IDataCloneStrategy {
```
damyanpetev, line 13: This file is exported to the public API, so unless we want to make this public you might want to set hidden/internal on it and modify the API exports to exclude it. Perhaps it can be useful as public, though the issue it solves is a particularly narrow and not the goal of this PR anyway, so I'd go with hiding it.
I'd also throw a hidden on the `cloneValueCached` for good measure

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
```diff
@@ -184,6 +185,20 @@ export class IgxPivotGridComponent extends IgxGridBaseDirective implements OnIni
...
+     * @remarks
+     * Returns the new dimension collection and its type:
+     * @example
+     * ```html
+     * <igx-pivot-grid #grid [data]="localData" [height]="'305px'"
+     *              (dimensionsChange)="dimensionsChange($event)"></igx-grid>
+     * ```
+     */
+    @Output()
+    public pivotConfigurationChange = new EventEmitter<IPivotConfigurationChange>();
```
damyanpetev, line 200: Okay, so we're doing a `<prop>Change` pattern thing but not fully following it? If it was, this would have to emit the same type as `pivotConfiguration`. Was this intentional? Something messing with the two-way binding because of object refs?

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
```diff
@@ -374,6 +389,25 @@ export class IgxPivotGridComponent extends IgxGridBaseDirective implements OnIni
...
+    /**
+     * Gets/Sets the values clone strategy of the pivot grid when assigning them to different dimensions.
+     *
+     * @example
+     * ```html
+     *  <igx-pivot-grid #grid [data]="localData" [pivotValueCloneStrategy]="customCloneStrategy"></igx-pivot-grid>
+     * ```
+     */
+    @Input()
+    public get pivotValueCloneStrategy(): IDataCloneStrategy {
```
damyanpetev, line 401: Oh, okay. 
So this becomes an actual input. In that case I suppose disregard part of my previous comment since I assume the idea _is_ to make the strategy public. In which case.. do document it a bit at least to explain what the use of it is.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
```diff
@@ -374,6 +389,25 @@ export class IgxPivotGridComponent extends IgxGridBaseDirective implements OnIni
...
+        return this._pivotValueCloneStrategy;
+    }
+
+    public set pivotValueCloneStrategy(strategy: IDataCloneStrategy) {
+        if (strategy) {
```
damyanpetev, line 406: Minor nit: Bit odd to need this, but okay. Ideally, the logic using this should be able to handle the prop being null/unset and sue a fallback

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
```diff
@@ -597,8 +631,9 @@ export class IgxPivotGridComponent extends IgxGridBaseDirective implements OnIni
         return this._emptyRowDimension;
     }
 
+    protected _pivotValueCloneStrategy: IDataCloneStrategy = new DefaultDataCloneStrategy();
     protected override _defaultExpandState = false;
-    protected override _filterStrategy: IFilteringStrategy = new DimensionValuesFilteringStrategy();
+    protected override _filterStrategy: IFilteringStrategy = new DimensionValuesFilteringStrategy(this.pivotValueCloneStrategy);
```
damyanpetev, line 636: Fairly certain this won't refresh properly if the clone strategy is reset, no? Since you have a setter anyway, you might want to refresh this one too or use an alternative approach

damyanpetev (changes requested):
Do describe stuff for the changelog - apparently we have a new pivot event (that may or may not support two way binding), new property and a new strategy available to all. And a new prop/param for the pivot dimension, right?

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.ts
skrustev: Yeah and since its a complex object two way data binding would be kinda weird and the users could get confused and need to set the whole object to a new instance while they will receive the same one. That's why I intentionally didn't follow the pattern so it is one way only but could easily allow for two way binding.

damyanpetev: In that case we might want to avoid using the patterns name if we're not following it, like `pivotConfigurationChanged` (with a `-d`). Still a bit unfortunate. 
Also if we're keeping the event args do name them as such -> IPivotConfigurationChangeEventArgs

dkamburov:
@skrustev there is a conflict in the changelog, can you take a look?

gmurray81 (dismissed):

damyanpetev (approved):
We discussed we might still want to keep the event args for future expansion, so I've restored that. And also since we're not entirely sure we like the clone strategy input as it and might use the cached one by default if possible - I've marked it as hidden/internal for now. That can be addressed later on

Issue #12786, opened 2023-03-17
Migrations for 15.1.0 hang


ChronosSF:
## Description  
Executing migrations for 15.1.0 seems to hang on 

the "Migration completed" message and don't the control to the user in the console.

 * igniteui-angular version: 15.1.0-alpha.0
 * browser: n/a

## Steps to reproduce  

1. You can use this pr that is already loaded to use latest angular alpha - https://github.com/IgniteUI/igniteui-angular-samples/pull/3266/files
2. Run `ng update igniteui-angular --from 15.0 --to 15.1 --migrate-only`

## Result  
Three files do migrate which is great but the command needs to be stopped with Ctrl+C as it doesn't end

## Expected result  
The command to both migrate and return control to the user

Lipata:
Check the following migrations - if they run properly: https://github.com/IgniteUI/igniteui-angular/pull/12651/files#diff-95c4a23fc45b5581a0dfee1d0787585f02b4eaea52a397273bcdb8ced1d75447

and revert back https://github.com/IgniteUI/igniteui-angular/pull/12798/files

PR #12925, opened 2023-04-24
Do not add file & directory watchers in ServerHost

jackofdiamond5:
Closes https://github.com/IgniteUI/igniteui-angular/issues/12786

The member migrations hang because the `tsserver` adds file watchers to all files that are to be migrated and these watchers are later never disposed of. Atm, it seems to me like the intended behavior is for whoever adds the watcher(s), to get rid of them themselves after they are not longer needed. And since we are implementing the `ServerHost` which is what's communicating with the file system, we are the ones that have to dispose of the file & dir watchers when we're done modifying files.

Though, as our implementation revolves mainly around working with Angular's virtual file system, we do not need to be adding file/directory watchers at all.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/migrations/common/ServerHost.ts
+++ b/projects/igniteui-angular/migrations/common/ServerHost.ts
@@ -38,15 +38,24 @@ export class ServerHost implements ts.server.ServerHost {
         return ts.sys.getFileSize(path);
     }
 
-    public watchFile(path: string, callback: ts.FileWatcherCallback, pollingInterval?: number):
+    //#region Watchers
+    // Atm we do not need to have file or dir watchers that access the actual FS
+    // as we are only working with NG's virtual FS
+    // Prior to https://github.com/microsoft/TypeScript/pull/49990 we were more or less required
+    // to add a watcher since it threw an error otherwise
+
+    public watchFile(_path: string, _callback: ts.FileWatcherCallback, _pollingInterval?: number):
         ts.FileWatcher {
-        return ts.sys.watchFile(path, callback, pollingInterval);
+        // return ts.sys.watchFile(path, callback, pollingInterval);
+        return undefined;
     }
 
-    public watchDirectory(path: string, callback: ts.DirectoryWatcherCallback, recursive?: boolean):
+    public watchDirectory(_path: string, _callback: ts.DirectoryWatcherCallback, _recursive?: boolean):
         ts.FileWatcher {
-        return ts.sys.watchDirectory(path, callback, recursive);
+        // return ts.sys.watchDirectory(path, callback, recursive);
+        return undefined;
     }
+    //#endregion
 
     public resolvePath(path: string): string {
         return ts.sys.resolvePath(path);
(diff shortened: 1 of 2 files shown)


jackofdiamond5:
For future ref - an easy way that I found to get rid of the file watchers is to create an [async hook](https://nodejs.org/api/async_hooks.html#class-asynchook), something like this:

```typescript
import * as asyncHooks from 'node:async_hooks';

/* ... */

const activeResources = new Map<number, any>();
const hook = asyncHooks.createHook({
   init(asyncId, type, triggerAsyncId, resource) {
      activeResources.set(asyncId, resource);
   },
   destroy(asyncId) {
      activeResources.delete(asyncId);
   }
}).enable();
```

Then after the migrations have completed, we can just look at the hooks that are related to FS changes (those will be the watchers) and close them, something like this:
```typescript
[...activeResources.values()]
   .filter((resource): resource is fs.FSWatcher => typeof resource.hasRef === 'function' && resource.hasRef())
   .forEach(watcher => watcher.close());

// dispose of the hook after it's no longer needed
hook.disable();
```
It's worth mentioning that this should not affect any other watchers that are running concurrently as they should be on a separate thread and Angular has no business poking around in the process that's been spawned for us, for the sake of our migrations. Additionally, calling consecutive updates should spawn different processes which will have different watchers, etc. so closing all of the previous' migration file watchers should conflict with the currently running migration.

Lipata (approved):

projects/igniteui-angular/migrations/common/ServerHost.ts
```diff
@@ -38,15 +38,24 @@ export class ServerHost implements ts.server.ServerHost {
...
+    // Atm we do not need to have file or dir watchers that access the actual FS
+    // as we are only working with NG's virtual FS
+    // Prior to https://github.com/microsoft/TypeScript/pull/49990 we were more or less required
+    // to add a watcher since it threw an error otherwise
+
+    public watchFile(_path: string, _callback: ts.FileWatcherCallback, _pollingInterval?: number):
         ts.FileWatcher {
-        return ts.sys.watchFile(path, callback, pollingInterval);
+        // return ts.sys.watchFile(path, callback, pollingInterval);
+        return undefined;
```
damyanpetev, line 50: PS: This relies a little bit on the internal TypeScript handling, which is not guaranteed in the future. Would've preferred if we still return something that still satisfies the public interface (`watchFile`'s signature doesn't expect undefined result), e.g. a simple mock with noop close method - just to be safe. Regardless, if we monitor this regularly (or if automation covers it) this should be okay for now.

damyanpetev: And just for ref, happened as expected and the undefined watcher is now checked and throws so this resolved it https://github.com/IgniteUI/igniteui-angular/pull/13478/commits/f6fd00e4c3ca84157ae54026ebeb71fe8ef7281a

PR #12949, opened 2023-05-02
fix(grid): In case of external data change sync cell's formControl wi…

### LINKED ISSUE #12724: Wrong edit value after refreshing the grid data
## Description  
Edit a cell and then update `igx-grid` data property to refresh the whole grid, the previous edit value remains.
If you open up the previously edited cell, instead of the current value, it shows the last edit value.

 * igniteui-angular version: 15.0.0+
 * browser: Chrome 110.0.5481.178

## Steps to reproduce  

1. Go to https://stackblitz.com/edit/angular-sw373h
2. The "Units in stock" value of the product "Chai" is 39. Edit the cell and update it to 50.
3. Click the "Refresh" button at the top
4. The grid is refreshed. Click the "Units in stock" cell of the product "Chai" again.

## Actual Result  
The default edit value of the cell is 50.

## Expected result  
The default edit value of the cell should be 39 because the cell value before editing was 39.

MayaKirova:
…th new value.

Closes #12724   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -889,6 +889,14 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
                 this.highlight.lastSearchInfo.caseSensitive = this.grid.lastSearchInfo.caseSensitive;
                 this.highlight.lastSearchInfo.exactMatch = this.grid.lastSearchInfo.exactMatch;
             }
+
+            // add check in case value changes externally and form control needs to update.
+            if (!this.editMode && this.formControl && this.formControl.value != changes.value.currentValue) {
```
damyanpetev, line 894: While this will certainly fix the specific scenario described in the issue, it relies on the cell being in view to fix the state, which might not always be the case. If the cell is out of view when the data changes (out of scroll or another page, etc) this won't happen and leave the validation state of the grid out of sync until the cell comes in view.

So it stands to reason the cell component shouldn't be handling the state update, just displaying it. The state update should be driven by the data, so when that changes any existing row states affected should be updated with the new data w/ any possible transactions active on top

damyanpetev (changes requested):

projects/igniteui-angular/src/lib/grids/cell.component.ts
MayaKirova: @damyanpetev  That will be rather hard to do without re-running validation on all formControls that have matching record id in the new data since we don’t really know if the data change was due to you wanting to set a fully new data to the grid, do some remote operation (like remote paging, filtering etc.)  or maybe do some live data update in existing records.

Would we be ok with that?

damyanpetev: Yeah I don't think we distinguish between those uses. Setting data generally not something that happens in a rapid succession so the hit shouldn't be that significant. I see no other option but to check the new data for cached validation state keys and update the state if needed. And there is a soft check to avoid running validation if the value is the same.. or at least used to be one, cuz I think it just got removed for some reason.

BTW, for the issue's use case if the goal is to 'refresh' the grid to initial state, a call to the validation service to clear existing state could probably be part of that. Perhaps something we should offer to be used.

projects/igniteui-angular/src/lib/grids/grid/grid-validation.service.ts
```diff
@@ -137,6 +137,22 @@ export class IgxGridValidationService {
         this.updateStatus();
     }
 
+    /**
+     * @hidden
+     * @internal
+     * Update validity based on new data.
+     */
+    public updateAll(newData: any) {
+        if (!newData) return;
```
damyanpetev, line 146: You can also add another short-circuit here to avoid looping the data -  if the `_validityStates` map is empty

projects/igniteui-angular/src/lib/grids/grid/grid.component.ts
```diff
@@ -321,6 +321,10 @@ export class IgxGridComponent extends IgxGridBaseDirective implements GridType,
         const dataLoaded = (!this._data || this._data.length === 0) && value && value.length > 0;
         this._data = value || [];
         this.summaryService.clearSummaryCache();
+        if (!this._init) {
+            this.validation.updateAll(this._data);
```
damyanpetev, line 325: One slight issue with this - it's missing the "w/ any possible transactions active on top" from my original comment. As in, if there's an active edit on a row in transactions and those are still active, they won't be reflected in this data.
Ideally, I was hoping we can hook into some place along the pipeline, but I don't see a cached result from the transactions pipe like some others do and no easy way to grab the merged data result :/
This will certainly complicate the fix a fair amount, will let you know if I see something useful

MayaKirova: The updateAll method does get the aggregated values from the transaction service, which should include all transactions (pending and active) :
const recAggregatedData = this.grid.transactions.getAggregatedValue(rowId, true) || rec;
Shouldn't that be enough?

damyanpetev (approved):

dkamburov:
@MayaKirova can you please add this behavior detail in the specification of the validator and into the topic for validation.
We need to have it written that changing the data source will revalidate, previously updated cells/rows.

damyanpetev:
> @MayaKirova can you please add this behavior detail in the specification of the validator and into the topic for validation. We need to have it written that changing the data source will revalidate, previously updated cells/rows.

@dkamburov Spec update - sure, as something we didn't note initially. Don't think it's really needed as explicit explanation in docs though, it's still accepted as a bug fix, right? 'Resetting' the data in this manner could be considered a shortcut for going through a bunch of row updates in a single go, which will revalidate otherwise we could end up with validation status that doesn't reflect the grid's data.

PR #12966, opened 2023-05-04
enhancement(moving): Move pointerDown and pointerMove logic to avoid needing additional custom attributes.

### LINKED ISSUE #12874: Web components and Blazor grids with moving and pinning throws error when unpinning a column
## Description  
Having web components or blazor grid with moving and pinning throws error when unpinning a column.
Open that sample https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/grid/column-moving

Note the header template: 
```
public webGridPinHeaderTemplate = (ctx: IgcColumnTemplateContext) => {

        const column = (ctx as any).column;
        return html`<div>
                     <span style="float:left">${column.field}</span>
                     <span style="float:right" @pointerdown=${(e: any) => this.toggleColumnPin(column.field)}>📌</span>
                   </div>`;
        };
```
See that the span has pointerdown event that pins/unpins a column. There is something with that pointerdown event and with drag-drop directive pointerdown.

NOTE: Note that if you change the event to click - the button stops working
 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open this sample https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/grid/column-moving
2. Unpin a column

## Result  
There is an error in the console

## Expected result  
There should be no error in the console.
If the event handler is set on click it should work like in Angular

skrustev:
Fix drag directive eating up event for children.

Closes #12874

 - [x] Bug fix
 - [ ] New functionality
 - [x] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


dkamburov:
@skrustev I've cherry-picked this commit to **angular-elements** branch and the issue started reproducing again. It was fixed beforehand and no error was thrown.

projects/igniteui-angular/src/lib/directives/drag-drop/drag-drop.directive.ts
```diff
@@ -909,7 +909,7 @@ export class IgxDragDirective implements AfterContentInit, OnDestroy {
 
         // Set pointer capture so we detect pointermove even if mouse is out of bounds until ghostElement is created.
         const handleFound = this.dragHandles.find(handle => handle.element.nativeElement === event.currentTarget);
-        const targetElement = handleFound ? handleFound.element.nativeElement : this.element.nativeElement;
+        const targetElement = handleFound ? handleFound.element.nativeElement : event.target || this.element.nativeElement;
```
dkamburov, line 912: @skrustev this is the line that causes the same error to appear again in Webcomponents - 
image

skrustev: @dkamburov This more or less caused the pinning happening on `pointerdown` instead of the usual `click` event. As per the sample, it can be seen that when the pin icon is clicked to pin the column, the focus moves the the `ProductName` column and the pointer down is also triggered for that column, but for some reason the `event.target` remains the icon of the previous column which has been recreated and the current `event.target` is no in the DOM.

I've added a check though in the IgxDragDirective to make sure the target is present in the DOM, so no error should be thrown now. 

But still we should probably fix that in the sample as well when this PR is merged, to pin/unpin the column on `click` instead of `pointerdown`

sdimchevski:
I like that you have to move at least a little for the "Drag a column header <here>" indicator to show. It was a little aggressive previously just on a click to show it as I would expect to select the cell this way.

I also think you shouldn’t be able to grab and drag the column by clicking on an internal element like sort arrows, 3-dot menus, etc.

dkamburov:
@skrustev please resolve conflicts

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row-dimension-content.component.html
```diff
@@ -13,16 +13,15 @@
 
 <ng-template #headerTemplate let-column>
     <div class='igx-grid__tr--header igx-grid__row-indentation--level-{{getLevel()}}'>
-        <igx-icon [attr.draggable]=" false" (click)="toggleRowDimension($event)">
+        <igx-icon [attr.draggable]="false" (click)="toggleRowDimension($event)">
```
MayaKirova, line 16: I assume here [attr.draggable] can also be removed.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row-dimension-content.component.html
```diff
@@ -13,16 +13,15 @@
...
             {{ getExpandState() ? 'expand_more' : 'chevron_right'}}</igx-icon>
         {{column.header}}
     </div>
 </ng-template>
 
 <ng-template #headerDefaultTemplate let-column>
-
     <div class='igx-grid__tr--header igx-grid__row-indentation--level-{{getLevel()}}'>
-        <igx-icon style='flex-shrink: 0;' [attr.draggable]=" false">
+        <igx-icon style='flex-shrink: 0;' [attr.draggable]="false">
```
MayaKirova, line 24: I assume here [attr.draggable] can also be removed.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row-dimension-content.component.html
skrustev: We've decided to not remove the `attr.draggable` from the grid yet, to be sure that the behavior remains the same.

MayaKirova (approved):

MayaKirova (approved):

PR #13023, opened 2023-05-17
Mass Merging 16.0.x to master

### LINKED ISSUE #12897: RadioGroup: falsy values always selected
## Description  
If you have a radio group where the formControlName or ngModel is set on each radio button (see https://github.com/IgniteUI/igniteui-angular/issues/11846) then falsy values are always initially selected.

 * igniteui-angular version: 15.1.3

## Attachments 
https://stackblitz.com/edit/angular-cfqezj

image

### LINKED ISSUE #12944: Igx-radio group requires 2x reset to change it's state (still dirty after 1st reset)
### Description  
Whenever i use a igx radio group with a formControl, the assigned formControl stays dirty after the reset operation

 * igniteui-angular version: 14.1.1 and 14.2.29
 * browser: Version 112.0.5615.138 (64-Bit)

### Steps to reproduce  

See this example
https://stackblitz.com/edit/angular-kzmx8y?file=src%2Fapp%2Fradio-group-sample%2Fradio-group-sample.component.ts,node_modules%2F%40angular%2Fforms%2Findex.d.ts&file=src%2Fapp%2Fradio-group-sample%2Fradio-group-sample.component.html

if it isn't working for you, see the official demo here 
https://www.infragistics.com/products/ignite-ui-angular/angular/components/radio-button#demo
and change the reset function to
`  public onReset() {
    this.fruitsForm.reset({
      favoriteFruit: this.model.favFruit,
      fullName: this.model.fullName,
    });

    console.log(this.fruitsForm.dirty);
    this.newModel = null;
  }`

1. click on any radio -> the form becomes dirty
2. click the reset button -> still dirty!
3. click the reset button again -> pristine (not dirty)

### Result  
Form state is dirty after calling the form reset function

### Expected result  
Form state should be pristine

### Attachments  
image

### LINKED ISSUE #12976: Theme for watermark is not being applied and/or is redundant
## Description  
Per the title. the watermark theme is targeting with `igx-` selector (name fallback) but it's a web component w/ `igc-`:
https://github.com/IgniteUI/igniteui-angular/blob/91669a690fdd952201d59fbd9eefacf5f2014fce/projects/igniteui-angular/src/lib/core/styles/components/watermark/_watermark-theme.scss#L35
So it needs it's selector adjusted at least or if the WC already responds to theme variables, watermark theme CSS vars might not be needed at all, considering they haven't been active for a while at least.
It also generates class selectors for internal structure, which are for sure not redundant with a web component.

### LINKED ISSUE #12980: [IgxDatePicker]: specialDates styles are not applied properly
## Description  
When setting the `specialDates` property, the background of the defined dates is not applied the same way as the ones in the `igx-calendar` although the `igx-date-picker` uses it internally.

* IgxCalendar
specialDates-calendar

* IgxDatePicker
specialDates-datePicker

 * igniteui-angular version: 15.1.x, 14.2.x
 * browser: any

## Steps to reproduce  

1. Use the following code and observe the defined special dates.
```html
  <igx-date-picker #datePicker [specialDates]="specialDates">
      <label igxLabel>Date</label>
  </igx-date-picker>
```
```ts
public today = new Date();
public range = [
    new Date(this.today.getFullYear(), this.today.getMonth(), 10),
    new Date(this.today.getFullYear(), this.today.getMonth(), 15),
    new Date(this.today.getFullYear(), this.today.getMonth(), 18)
];
public specialDates = [{ type: DateRangeType.Specific, dateRange: this.range }];
```

## Result  
The backgrounds are not the same.

## Expected result  
The backgrounds should be the same.

### LINKED ISSUE #12983: Error in TreeGridNormalizeRecordsPipe if state set too early
## Description  
When I use the layoutAPI on a TreeGrid before the grid actually has data the IgxTreeGridNormalizeRecordsPipe will give a fatal error because the flatData of the grid is still null, and there is no check for this.

 * igniteui-angular version: 15.0.15
 * browser: Chrome 113.0.5672.63

## Steps to reproduce  

Sadly don't have the time to make a repo. 

My proposed solution would be to check if flatData is not null in tree-grid.pipes.ts:314 before executing the .map() over said flatData.

### LINKED ISSUE #13017: avatar-theme doesn't build after update to 16
## Description  
Initialization code for `avatar-theme` now errors after an update to igniteui-angular `16.0.0`

 * igniteui-angular version: 16.0.0
 * browser: N/A

## Steps to reproduce  

1. I have the following code in my project:
```scss
  .empty-role {
    $empty-role-avatar: avatar-theme($color: #000, $size: 88px);
    @include css-vars($empty-role-avatar);
  }
```

## Result  
Build fails after an update to `16.0.0`

## Expected result  
Build should not fail, or my code should have been migrated.

## Attachments  
image

### LINKED ISSUE #13035: Switch error state colors
The switch in Angular should not have visual change when it's in an invalid state but the user should be able to achieve what we see in WebC.

WebC
bootstrap
Screenshot 2023-05-22 at 11 15 02

material
Screenshot 2023-05-22 at 11 04 41

fluent
Screenshot 2023-05-22 at 11 06 55

indigo
Screenshot 2023-05-22 at 11 15 15

### LINKED ISSUE #13039: IgxChip - The text content of a disabled chip is not visible in igx-grid when the theme is dark
## Description  
The text content of a disabled chip is not visible in igx-grid when theme is dark:

image

 * igniteui-angular version: 16.0
 * browser: any

## Steps to reproduce  

1. Download and run the following sample - [here](https://filetransfer.infragistics.com/public.php?service=files&t=830a82da214adc577e85c50591f935d8).
2. See the result.

## Result  
The text content of the disabled chips is not visible.

## Expected result  
The text content should be visible.

### LINKED ISSUE #13053: Consume schemas from igniteui-theming package
## Description  
Refactor all component themes to consume their schemas from the `igniteui-theming` package.

### LINKED ISSUE #13054: Grid no longer establishes a local stacking context, clips external items
## Description  
Per the title, the Grid no longer stablishes its own local stacking context (since #12850) leaves internal stacking order interacting with other elements in view and potentially clipping them.

 * igniteui-angular version: 14.2.x, 15.1.x, 16.0.x
 * browser: N/A

## Steps to reproduce  

1. Take the demo under https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/row-drag#templating-the-drag-ghost
2. Start dragging a row 
3. Move the drag ghost 'over' the origin grid's header/scroll

## Result  
The drag ghost clips under both:
image
That'd be because `z-index` of the the grid header (`2`) and especially the virtual scroll helper (`10001` ❗) are no longer contained and interact with other elements in the page. Without digging through the rest of the grid structure, IIRC there are other elements with pretty high indexes in use for internal ordering too.

@hanastasov @simeonoff This is just one example I could find, I'm sure there might be more and it's not hard to imagine other scenarios that'd break. Like the next dropdown-like component we don't control placed above the grid being clipped by scrollbar or something.
I see you did consider the implications of dropping the local context https://github.com/IgniteUI/igniteui-angular/pull/12994#discussion_r1189979310, though reading over the comments for #5674 I believe those still sand - and not about the particular issue, but in general.

## Expected result  
IMO, the loosing the local stacking context has the potential for more damage than the issue it resolves and should likely be restored. I still think my old comment about "no universal solution" is kinda valid, but so is the following part about containing the grid internals and [some conflicts being application-level scenarios](https://github.com/IgniteUI/igniteui-angular/issues/5674#issuecomment-525621994) - the issue in #12850 is exactly _that_ type of application scenario where the need to bunch some grids in a column layout is where you bump the `z-index` of the top grid to allow it to overlay on top of the others.

Not ideal, I realize. The only actual fix for this would be to do what was initially suggested back in the day - move the local context to another element, so the outlets for overlays can be outside while still in the host element for theming, if that's even feasible. Or move the entire overlay service to a `top-layer` once something like [`showPopover`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/showPopover) is widely available.

## Attachments  

### LINKED ISSUE #13075: [Angular Elements Grid] No style property to change grid's text on editing
## Description  
Cannot change text color of grid's cell during edit event. According to styling properties as per [Angular theming](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/data-grid/theming), --edit-mode-color should change the text color in edit mode, however it changes the cell border color instead. 

## Steps to reproduce  

https://stackblitz.com/edit/yynrwb?file=src%2Fapp%2Fgrid%2Fgrid-editing-events%2Fgrid-editing-events.component.scss,src%2Fapp%2Fgrid%2Fgrid-editing-events%2Fgrid-editing-events.component.html

### LINKED ISSUE #13084: igx-grid advanced filtering dialog action buttons margins/paddings do not match the design system definition
## Description  
The title says it all.

 * igniteui-angular version: 16.0.0
 * browser: all

## Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/advanced-filtering
2. Click the advanced filtering button

## Result  
image

## Expected result  
image

### LINKED ISSUE #13085: Grid Row Editing: displayDensity="compact" results in the igxRowEditingOverlayOutlet div not displayed correctly
## Description  
For a row-editable grid having its `displayDensity` set to "compact", the igxRowEditingOverlayOutlet div is placed above the other rows and its background is transparent, which makes the text in it not readable.

 * igniteui-angular version: 16.0.x
 * browser: any

## Steps to reproduce  

1. Attempt to edit a row in [this demo](https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/validation#cross-field-example).

## Result  
image

## Expected result  
The igxRowEditingOverlayOutlet div to look as when the `dsiplayDensity` is set to `comfortable` or `cozy`.
image

### LINKED ISSUE #13114: Grid: scrollbars on top of nav-drawer
## Description  
The grid's scrollbars have a higher z-index than the navigation-drawer so they are always on top.

 * igniteui-angular version: 16.0.5

## Attachments  
https://stackblitz.com/edit/angular-tc333n-fruz8v

image

ChronosSF:
Closes #12976
Closes #12980
Closes #12897
Closes #12944
Closes #13017
Closes #13035
Closes #13039
Closes #12983
Closes #13053
Closes #13075
Closes #13084
Closes #13085
Closes #13054
Closes #13114


radomirchev:
@ChronosSF what is the status of this PR?

ChronosSF:
> 

It's the usual mass merge PR. It can remain open for a long time.

PR #13071, opened 2023-05-31
feat(combo): Exported filter strategy for non-diacritic filter

### LINKED ISSUE #12876: Support for Non-Diacritics characters in combo boxes
Searching for `México` is quite hard. In spanish, diacritics are something that most people pays attention when they are filling "main content" but not when "searching for them", like when filling a catalog with the correct country strings, you'll find "Mexico" being written as "México", but when daily users want to select that value in a combo they'll probably write "Mexico". 

This issue occurs especially with two diacritics `´` and `¨`, being the most common the first one. As it's used A LOT in spanish, even people names have it, daily use objects, verbs, subjects, everything (!). For english, happens a bit because their keyboards dont have `ñ` but they interact with data written in spanish; So, they are unable to type "Niño" (means kid) they type "Nino" expecting to find the value.

The "México" example is the one issue I encounter the most when filling forms in the web. However, my main (real) issue are people names, some méxican examples: "José", "Óscar", "Ángel", "Germán" it gets even worse because there's names that can be written with the diacritics in two different ways, like "German" might be written without the diacritic and it's a valid name, but spelled differently. 

With the component current behaviour users claim their name (or country) is missing.

#### Filtering
I would like to forget about diacritics when searching for a value inside the `igx-simple-combo` or `igx-combo`, I want to type "Mexico" and being shown any of the following "México", "Méxícó", "mexicÓ". Diacritics shouldn't matter.
#### Sorting
I would like the component to automatically sort display values alphabetically considering diacritics expected order. Sorting without considering diacritics is mandatory, in spanish `o` and `ó` are the same, it's a know issue of all encodings to order them "alphabetically" with those characters really far away.

Specifically, every name starting with a diacritic character are the ones usually claimed as missing. That's because, users don't find it by using the filter, nor scrolling to its expected ordered character.

Currently I've made an implementation myself, it kind of does the following:
1. When `data` array is setted, Only then, and only once: I create a clone attribute called `escapedText` and fill it with the `displayKey` value
2. The `escapedText` attribute is filled with diacritics removed and lower-cased ("México" is saved as "mexico")
3. I use the `escapedText` value for two purposes, being the first one to SORT the values... this is specially needed for words starting with diacritics. Like the `Óscar` name. The second purpose is for filtering.
4. I implemented a pipe filter, the `transform` method kinda looks like this:
```typescript
/**
 * `filterTerm` must be parametrized without dielectrics 
 **/
public transform(collection: Array<{escapedText: string}>, filterTerm: string): any[] {
    const specialCharacterRegExp = /(\+|\*|\\|\^|\.|\[|\]|\$|\&|\||\?)/gm;
    const regexp = new RegExp('^.*?' + filterTerm.replace(specialCharacterRegExp, '\\$1').replace(/\s/gi, '.*?') + '.*?$', 'is');
    return collection.filter((item) => {
      return item.escapedText.toString().match(regexp) !== null;
    });
 }
```

## Additional context
The regexp replacing spaces `\s` to `.*?` is a plus, as I found incredibly useful to treat spaces as `*` wildcards (vscode does the same when using `cntrl+p`), I removed diacritics by using [this code](https://stackoverflow.com/a/5912746/1657465). Removing diacritics is a costly process, but it is not when you do it only once, being the right time when the component data is ready.

[Here's an example of the issue](https://stackblitz.com/edit/angular-tkmedg?file=src/app/combo-main/local-data.ts), there's `New Mexico` and `Old México` in the available data. But by typing "Mexico", you'll find only "New Mexico"

rkaraivanov:
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


rkaraivanov:
@damyanpetev 
I didn't add an entry to the changelog since I'm not 100% sure on the naming.
I'm open for suggestions.

Lipata (approved):

projects/igniteui-angular/src/lib/combo/combo.pipes.ts
```diff
@@ -70,18 +70,45 @@ export class IgxComboGroupingPipe implements PipeTransform {
...
+/**
+ * Combo filter function which does not distinguish between accented letters and their base letters.
+ * For example, when filtering for "resume", this function will match both "resume" and "résumé".
+ *
+ * @example
+ * ```html
+ * <igx-combo [filterFunction]="comboIgnoreDiacriticFilterFunction"></igx-combo>
+ * ```
+ */
+export function comboIgnoreDiacriticFilterFunction<T>(collection: T[], searchValue: string, filteringOptions: IComboFilteringOptions): T[] {
```
jackofdiamond5, line 102:
```suggestion
export function ignoreDiacriticsFilterFunction<T>(collection: T[], searchValue: string, filteringOptions: IComboFilteringOptions): T[] {
```

Since we already have a `defaultFilterFunction`, it makes sense to me to name the new filter function in a similar fashion.

rkaraivanov: It does but that is in the context of the combo internal imports. Since this is exported at the top-level public API the import will look like this:
```typescript
import { ignoreDiacriticsFilterFunction } from 'igniteui-angular';
```
which does not really convey the information that this is related to the combo component.

If it was something akin to:
```typescript
import { ignoreDiacriticsFilterFunction } from 'igniteui-angular/combo';
```
then that would be better but AFAIK we do support these kind of imports.

rkaraivanov: It does but that is in the context of the combo internal imports. Since this is exported at the top-level public API the import will look like this:
```typescript
import { ignoreDiacriticsFilterFunction } from 'igniteui-angular';
```
which does not really convey the information that this is related to the combo component.

If it was something akin to:
```typescript
import { ignoreDiacriticsFilterFunction } from 'igniteui-angular/combo';
```
then that would be better but AFAIK we do support these kind of imports.

jackofdiamond5 (approved):

projects/igniteui-angular/src/lib/combo/combo.pipes.ts
Lipata: @jackofdiamond5 `defaultFilterFunction` is not public and it should not decide the name for the public method. It should be either descriptive enough, like your initial suggestion, @rkaraivanov: `comboIgnoreDiacriticsFilterFunction` (with Diacritic**s**?), or short enough like `comboDiacriticsFunction`?

damyanpetev:
@rkaraivanov Since this only add the filtering capability, I wouldn't really say it completes the referenced issue. Still need to look at options for sorting, right?

Lipata (approved):

PR #13095, opened 2023-06-07
fix(drag): Add handling for ghostTemplates that have display contents.

### LINKED ISSUE #13033: [Angular elements] Grid's dragGhostCustomTemplate renders ghost with incorrect position.
## Description  
When setting the dragGhostCustomTemplate on the grid the ghost position is incorrect on drag.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Set `row-draggable="true"` for the grid.
2. Set a custom drag ghost template, for example:

```
   grid1.dragGhostCustomTemplate = () => html`Test`;
```
3. Start dragging a row.

## Result  
The template renders but somewhere on the bottom of the page.

## Expected result  
It should position correctly under the mouse pointer.

skrustev:
Closes #13033 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/directives/drag-drop/drag-drop.directive.ts
+++ b/projects/igniteui-angular/src/lib/directives/drag-drop/drag-drop.directive.ts
@@ -1194,6 +1194,10 @@ export class IgxDragDirective implements AfterContentInit, OnDestroy {
 
         if (this.ghostTemplate) {
             this._dynamicGhostRef = this.viewContainer.createEmbeddedView(this.ghostTemplate, this.ghostContext);
+            if (this._dynamicGhostRef.rootNodes[0].style.display === 'contents') {
+                // Change the display to default since display contents does not position the element absolutely.
+                this._dynamicGhostRef.rootNodes[0].style.display = 'block';
+            }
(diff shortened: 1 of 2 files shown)


damyanpetev:
@skrustev @MayaKirova The fix is working, however that's one small gotcha - by converting the non-layout container for dragging purposes, there's a slight difference with the setup in Elements and you can't just adjust the `z-index` alone on the actual provided template directly. Because the absolute position the drag applies does to the parent, you have to specify position different than static on the provided template for z-index to kick in.

So with Elements the setup will have one more line in the CSS:
```css
.z-index {
    z-index: 1;
    position: relative; /* need to add this */
}
```
```js
grid1.dragGhostCustomTemplate = () => html`<div class="z-index">Test</div>`;
```

I might be okay with that approach given the current options, if everyone else is too. I still think we should probably log some general enhancement for the drag&drop to manage it's own container **_and_** z-index (like overlays do). Both are also candidates for a CSS top-layer treatment down the line too

damyanpetev (approved):

PR #13135, opened 2023-06-19
fix(igxGrid): Fix issue with groupby UI not updating on column/column…

MayaKirova:
… prop changes.

Closes [#700](https://github.com/IgniteUI/igniteui-webcomponents/issues/700)

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -323,6 +323,7 @@ export interface GridType extends IGridDataBindable {
     renderedRowHeight: number;
     pipeTrigger: number;
     summaryPipeTrigger: number;
+    groupablePipeTrigger: number;
```
damyanpetev, line 326:
```suggestion
    /** @hidden @internal */
    groupablePipeTrigger: number;
```

damyanpetev (commented):
Ah, the fix works for Angular alright, but unfortunately doesn't quite fix the issue with Elements linked. The problem is with creating dynamic columns and the way Elements populates the child content query with some delay.

It seems like it's near impossible to repro in normal Angular use since it will always execute the template (which sets the inputs) and update the content query in a single view refresh, so with the current setup by the time the group area pipe is ran the grid columns are updated too. Basically can't hit the setter on a column and have it outside the grid's query (I tried). Perhaps with some `createComponent` use and manually pushing to the query a dynamic column can be set `groupable` and added to the grid after, but seems like a massive PITA to repro like that and it's basically hacking normal use just to break it.

Anyway, I've added another trigger on columns changed here #13163. It should be pretty benign for Angular, though I could simply move it to the Elements branch only and attempt to cover it there. Unless someone has ideas if this can be repro'd in Angular or an alternative caching setup.

PR #13164, opened 2023-06-26
fix(grid,kb-nav): prevent error on arrow key in filter UI row

### LINKED ISSUE #13155: Some keyboard operations on the filter UI of the Grid show errors in the browser's console
## Description  

Some keyboard operations on the filter UI of the Grid show errors in the browser's console.

## Steps to reproduce  

1. Open the filter UI of the igc-grid by clicking the "Filter" button. (The focus will be entered into the filter input box.)
  https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/cell-editing#angular-grid-cell-editing-and-edit-templates-example (Stackblitz snapshot: https://stackblitz.com/edit/dsjjm3)
2. Enter `Shift` + `Tab` key. (The filter input box lost focus but is still active.)
3. Enter the `Left` or `Right` key. (Important: This seems to be the trigger of the problem I'm reporting in this issue. Pressing any other keys doesn't cause the problem.)

## Result  

After doing that, you will find there is an error below in the browser's console:
```
ERROR TypeError: Cannot read properties of undefined (reading 'level')
```

After once happens that error, any user interaction events, such as keydown, mousemove, etc., cause the same error continuously.

## Expected result  

No error messages are in the browser's console.

## Attachments
- https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/cell-editing#angular-grid-cell-editing-and-edit-templates-example
- https://jp.infragistics.com/products/ignite-ui-blazor/blazor/components/grids/grid/cell-editing

movie-002

onlyexeption:
Closes #13155 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -110,7 +110,7 @@ export class IgxGridNavigationService {
 
     public headerNavigation(event: KeyboardEvent) {
         const key = event.key.toLowerCase();
-        if (!HEADER_KEYS.has(key)) {
+        if (!HEADER_KEYS.has(key) || Object.keys(this.activeNode).length === 0) {
```
damyanpetev, line 113: Okay, I initially thought this can't possibly be the check, but it seems like it sort of is as there's already a precedent in the service doing the same thing.

@onlyexeption You should also check if `this.activeNode` is defined first, because some parts of the service reset the object
https://github.com/IgniteUI/igniteui-angular/blob/dc33a249f7180d361d61dcb95486fe4529d494e9/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts#L3486
but some parts also set it to null:
https://github.com/IgniteUI/igniteui-angular/blob/dc33a249f7180d361d61dcb95486fe4529d494e9/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts#L136

damyanpetev: Actually a two-in-one check could be `this.activeNode?.row !== -1` - as far as I can tell header nav always sit on row -1.

damyanpetev (approved):

PR #13190, opened 2023-06-29
feat(forOf): Add data types for igxForOf and igxGridForOf.

### LINKED ISSUE #12182: [IgxForOfDirective] use generic types

When using igxForOf my items are inferred as type any.

e.g. `<div *igxFor="let item of [1,2,3]"></div>`

image

I'd like the directive to take advantage of generics to get proper types in my items.

None

skrustev:
Closes #12182

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -83,7 +84,7 @@ export class IgxForOfContext<T> {
     standalone: true
 })
 // eslint-disable @angular-eslint/no-conflicting-lifecycle
-export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestroy, AfterViewInit {
+export class IgxForOfDirective<T, U extends Array<T> = Array<T>> implements OnInit, OnChanges, DoCheck, OnDestroy, AfterViewInit {
```
MayaKirova, line 87: Unfortunately does not work with `Array<T>`, but works with `NgIterable<T>`. 
We do depend on the array's api a lot and changing it to use iterable would be a bit of a breaking change and a bit of a pain. Maybe we could use something like:
```
export declare type IgIterable<T> = Array<T> & Iterable<T>;
```

skrustev: @MayaKirova Can you check how it is now. For some reason when using the `Array<T>` type it does not recognize it, but using tuple `T[]` seems to do the trick. I guess because tuple recognizes the types the array contains while the other type does not.

MayaKirova: Looks ok with `T[]`

MayaKirova (approved):

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -92,7 +92,7 @@ export class IgxForOfDirective<T> implements OnInit, OnChanges, DoCheck, OnDestr
      * ```
      */
     @Input()
-    public igxForOf: any[] | null;
+    public igxForOf: U&T[] | null;
```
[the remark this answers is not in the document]
pmoleri: This is how Angular does it. I guess there's a reason:

https://github.com/angular/angular/blob/4d18ba1af78dae2ba14f45ed092a4c55335edd9c/packages/common/src/directives/ng_for_of.ts#L140-L149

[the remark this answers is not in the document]
skrustev: The intersection helps in the typing for the single item in the `for...of`, otherwise it does not appear typed.
Here's the data:
image

If we set only `U` type on the `igxForOf` input:
image

If we set it as `T[]` only, `U` ends up unused and appears to be used in the Angular ngFor, for the context that provides fallback to the ngFor data value where you can get it in the template as:
image

The last bit we currently do not support though, which we can log it as a feature request.

PR #13267, opened 2023-07-17
fix(mask): move the cursor to the next character if the same character is typed - master

### LINKED ISSUE #13260: [igxMask]: cursor does not move to next character if the same character is typed
## Description  
For example, consider the following:
```html
<igx-input-group>
    <label igxLabel>Date</label>
    <input igxInput
    type="text"
    [igxMask]="'00/00/0000'"
    [placeholder]="'dd/mm/yyyy'"/>
</igx-input-group>
```
I have typed `20/02/2000` and I want to change it to `21/02/2000` by typing `21` in the beginning. If I try typing `2`, the cursor remains in the same position since the character `2` is already present.

mask

 * igniteui-angular version: 16.1.x, 16.0.x, 15.1.x
 * browser: any

## Steps to reproduce  

1. Open this sample [here](https://www.infragistics.com/products/ignite-ui-angular/angular/components/mask#adding-a-placeholder).
2. Type `20/02/2000` and try changing it to `21/02/2000` by typing `21` in the beginning.

## Result  
The cursor remains in the same position. 

## Expected result  
The cursor should move to the next character.

georgianastasov:
Closes #13260   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/mask/mask-parsing.service.ts
```diff
@@ -156,8 +157,8 @@ export class MaskParsingService {
         }
 
         if (value.length <= 1) {
-            let isDelete = false;
```
jackofdiamond5, line 159: Let's try to simplify this code a little bit. We can get rid of this whole check (lines 159 through 179) and then use the above loop to remove characters step by step as soon as a `delete` key is pressed. To do that, we can check if the `value` that we've received has any length to it, since on a `delete` it should be empty (even if we have some text selection beforehand). Then we need to make sure that we are not on a mask literal, since we don't want these deleted and finally if both of these conditions are met we can set the `maskedValue` to be a `promptChar`, which we can take from the `maskOptions`. In all cases we, should update the `cursor`, to make sure that it's always at the end of the selection/that it goes forward when `delete` is pressed.

projects/igniteui-angular/src/lib/directives/mask/mask-parsing.service.ts
```diff
@@ -152,24 +151,11 @@ export class MaskParsingService {
...
-                if (value === '') {
-                    // on `delete` the cursor should move forward
-                    cursor = Math.max(cursor, end);
-                    isDelete = true;
-                } else if (cursor === literalsPositions[i]) {
-                    cursor = literalsPositions[i] + 1;
-                }
-            }
-            if (!isDelete && initialMaskedValue !== maskedValue) {
+            if(value.length < 1){
```
jackofdiamond5, line 154:
```suggestion
            if (value.length < 1){
```

jackofdiamond5 (approved):

PR #13288, opened 2023-07-19
Remove createComponentInstance method & fix overlay positioning

### LINKED ISSUE #12888: Grid ESF filter dropdown overlay handling cleanup
See https://github.com/IgniteUI/igniteui-angular/pull/12860#discussion_r1160549369 and https://github.com/IgniteUI/igniteui-angular/pull/12860#discussion_r1160549672
The Grid ESF has a reserved `createComponentInstance` helper method in the Grid that, while significantly simplified now, is not used/needed for anything else. What's more - the overlay service already has built-in functionality to create components that can be used directly. Also the ESF dropdown hooks to the app ref which should not be needed, yet as reported the positioning breaks if the grid's view ref is used.
In any case, this warrants a closer look - both to reduce the duplicated code if possible and uncover possible shortcomings in the overlay service.

mtsvyatkova:
Closes #12888

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


mtsvyatkova:
Note: Initially, the esf dropdown has the same width as the grid when the grid reference is used in the `createFilterDropdown` method. Though the correct width is calculated when the dropdown component is moved to the overlay, the overlay is using the cached width while positioning the element:
```typescript
private updateSize(info: OverlayInfo) {
     …
      // set content div width only if element to show has width
      if (info.initialSize.width !== 0) {
          info.elementRef.nativeElement.parentElement.style.width = info.initialSize.width + 'px';
      }
  }
```
This results in the wrong display position for the esf dropdown. Currently, I have excluded the esf filtering from the caching in the `updateSize` method but we may need a better approach.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -703,7 +703,8 @@ export class IgxOverlayService implements OnDestroy {
         }
 
         // set content div width only if element to show has width
-        if (info.initialSize.width !== 0) {
+        // excluding the esf filtering as its initial width is the same as grid's width and is resized after the element is moved to the overlay
+        if (info.initialSize.width !== 0  && info.elementRef.nativeElement.tagName?.toLowerCase() !== 'igx-grid-excel-style-filtering') {
```
wnvko, line 707: Please no. Overlay has no knowledge about any component and it should not have!
I see two ways to fix this:
- easy one. Handle `contentAppended` event of the overlay. Get the overlay info and set its `initialSize` to (0; 0).
- hard one. We can remove `initialSize` from overlay info. This will require removing of second parameter of `IPositionStrategy.position` method. Also fixing of drop down as, if I am not missing something, this is the only component (and all the ones inheriting it) that benefits from setting the `width` of the overlay content element.

wnvko (changes requested):

wnvko (dismissed):

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3752,19 +3751,13 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
             this.excelStyleFilteringComponent.overlayComponentId = id;
             return { id, ref: undefined };
         }
-        const ref = this.createComponentInstance(IgxGridExcelStyleFilteringComponent);
+        const ref = this.viewRef.createComponent(IgxGridExcelStyleFilteringComponent, { injector: this.injector } );
```
damyanpetev, line 3754: Okay, so I was thinking in the lines of that this use case is already covered by the overlay service too. As in something like `this.overlayService.attach(IgxGridExcelStyleFilteringComponent, this.viewRef);` so the overlay will manage the component ref - both create (available on opening's args) and will destroy it as well, so you won't have to.
And if there's some more handling done around this - even better.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3752,19 +3751,13 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
         ref.instance.initialize(column, this.overlayService);
         const id = this.overlayService.attach(ref.instance.element, options);
         ref.instance.overlayComponentId = id;
         return { ref, id };
     }
 
-    private createComponentInstance(component: any) {
-        const dynamicComponent: ComponentRef<any> = createComponent(component, { environmentInjector: this.envInjector, elementInjector: this.injector } );
-        this.appRef.attachView(dynamicComponent.hostView);
```
damyanpetev, line 3764: Also this is the only thing using `appRef` in the grids, which leaves it unused, so you'll get `Property 'appRef' is declared but its value is never read` warnings and should remove it.

damyanpetev: Note: That will obv drag all other Grid (and Row Island) constructors. Not entirely sure if we should mark that as breaking for people extending the grid. At least this is targeting a major version.

projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts
```diff
@@ -67,6 +67,16 @@ export class IgxFilteringService implements OnDestroy {
         const filterIcon = column.filteringExpressionsTree ? 'igx-excel-filter__icon--filtered' : 'igx-excel-filter__icon';
         const filterIconTarget = element.querySelector(`.${filterIcon}`) as HTMLElement || element;
 
+        this._overlayService.contentAppended
+            .pipe(
+                first(),
```
damyanpetev, line 72: Probably nearly impossible to break this, but the service being root singleton and with some concurency it is possible to pick up another overlay's event, so would be much happier with this checing like the other events below:
```suggestion
                first(overlay => overlay.id === id),
```

damyanpetev: Although, kind of interested if we can skip this extra event entirely or if the size adjust can be done either in order directly or during opening

mtsvyatkova: Removed the `contentAppended` as now there is not need to adjust the size

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
mtsvyatkova: @kdinev should we treat removing the `appRef` from the constructors in any special way? May be add it as a Breaking change in the Changelog.md?

kdinev: @mtsvyatkova No, I don't think we should treat it in any special way or add it in the changelog

damyanpetev (approved):

PR #13305, opened 2023-07-25
Grid: remove inline styles

### LINKED ISSUE #12713: Grids templates contain css attributes, which should be in css classes
## Description  
The grid templates have inline css attributes that should be in classes, like `float` on the scrollbars:

image

 * igniteui-angular version: 15.0.x
 * browser: N/A

## Result  
The template is polluted by attributes, and such attributes like `float` would probably be affected by directionality and wouldn't properly adjust in `rtl` scenario.

## Expected result  
All css attributes should be in classes, unless there's a specific reason for an attribute to have to be inline.

desig9stein:
Closes #12713  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-header-row.component.html
+++ b/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-header-row.component.html
@@ -12,7 +12,7 @@
                           *ngIf="grid.filterDimensions.length === 0"
                           class='igx-grid__pivot-empty-chip-area'>{{grid.resourceStrings.igx_grid_pivot_empty_filter_drop_area}}</span>
                         <ng-container *ngFor="let filter of this.filterAreaDimensions; let last = last;">
-                        <span [style.visibility]='"hidden"' class="igx-grid__tr-pivot--chip_drop_indicator"
+                        <span class="igx-grid__tr-pivot--chip_drop_indicator"
                               [style.height.px]='getAreaHeight(filterArea)'></span>
                             <igx-chip [id]="filter.memberName" [draggable]="true" [displayDensity]="grid.displayDensity" [data]="{ pivotArea: 'filter' }"
                                       [removable]="true" (remove)="filterRemoved($event)" (dragOver)="onDimDragOver($event, 2)"
@@ -23,7 +23,7 @@
                                 {{filter.memberName}}
                             </igx-chip>
                             <ng-container *ngIf='last'>
-                            <span [style.visibility]='"hidden"' class="igx-grid__tr-pivot--chip_drop_indicator"
+                            <span class="igx-grid__tr-pivot--chip_drop_indicator"
                                   [style.height.px]='getAreaHeight(filterArea)'></span>
                             </ng-container>
                         </ng-container>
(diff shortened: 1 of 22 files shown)


projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-header-row.component.html
```diff
@@ -226,6 +226,7 @@
     </igx-grid-excel-style-filtering>
 </div>
 
+<!-- TODO  Remove inline styles if possible -->
 <div style="visibility: hidden; position: absolute; top: -10000px">
```
desig9stein, line 230: @mddragnev you added those styles in [this commit](https://github.com/IgniteUI/igniteui-angular/commit/f97fe06c97be734c799c2b7514bff35cdfdb6db1). I want to move them to a class, but I am not sure if we need them at all, I remove them to test and I didn't see any changes.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row-dimension-content.component.html
```diff
@@ -21,6 +21,7 @@
 
 <ng-template #headerDefaultTemplate let-column>
     <div class='igx-grid__tr--header igx-grid__row-indentation--level-{{getLevel()}}'>
+        <!-- TODO remove inline style if possible -->
```
desig9stein, line 24: @MayaKirova I am trying to remove all inline styles from the grid templates and since the `flex-shrink: 0` was added by you in [this commit](https://github.com/IgniteUI/igniteui-angular/commit/a2defd78f8c6cb183a91e7fc1b6c1d62e442179e), I wanted to ask you why we need that style.

If the idea was to shrink the icon when it doesn't have value, `flex-shrink` doesn't work. If this is a bug I can log it as a separate issue, otherwise I can remove the inline style.

Screenshot 2023-07-25 at 14 27 09

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-header-row.component.html
mddragnev: As long as I remember, this is used in order to generate the chips for the filtering area of the pivot grid.  I was using that div (which is hidden and out of the view)  to calculate whether we should have a dropdown button or if we have enough space for all chips. I think that as long as this div stays hidden and out of view you can move them to a class

kdinev: @desig9stein I believe this one shouldn't be made part of the theme. It can stay inline, as it's a sizing container and it doesn't really belong anywhere in the theme.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-row-dimension-content.component.html
MayaKirova: @desig9stein It's actually for scenarios where the text in some cells is longer than the container, in which case if there's no flex-shrink:0 it will shrink the icon and the texts will no loner be aligned with the other cells. For example:
image

desig9stein: Ok, I am gonna leave this as an inline style

simeonoff (approved):

PR #13358, opened 2023-08-04
fix(splitter): improving sizing calculations in percent splitter #13252 - master

### LINKED ISSUE #13252: IgxSplitter:  when a pane is sized to minimun, sibling splitter panes width is not adjusted on window resize
## Description  
The layout includes a splitter with Vertical type having its style.width set to '100%' and containging two splitter panes. The left one has its `minSize` property set, for instance to '150px'. Then the splitter handle is dragged to the left, such that the left pane reaches its mimimum width and cannot be shrinked anymore (the same happens if no `minSize` is set and the pane is fully collapsed). After that the window is resized to be smaller in width. When the window is resized back to be wider, the right splitter panes does not fill the right are as it did before the resize.

 * igniteui-angular version: 16.1.x, 16.0.x, 15.1.x
 * browser: any

## Steps to reproduce  

1. Open [this sample](https://stackblitz.com/edit/igx-splitter-resize-to-min-size-of-pane-behavior?file=src%2Fapp%2Fsplitter-horizontal-sample%2Fsplitter-horizontal-sample.component.html).
2. Drag the splitter handle to the left until the left pane cannot be shrinked anymore.
3. Resize the window to a smaller width.
4. Resize the window back to a bigger width.
5. Observe the right splitter pane not taking the space before the resize.

## Result
splitter2

no `minSize`:
splitter3

## Expected result  
The splitter panes to adjust to take the remaining width of the splitter after the window resize.

When the left pane is not on its min size:
splitter4

_Note: Not sure if this is an issue, since nothing sets the right pane width to take up what space remains, however, such a result might not be considered ideal, also considering the latter gif and the different outcome in both scenarios. This behavior was reported in relation to #13249_

ChronosSF:
Closes #13252   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


mddragnev:
There is still ~1 px of orange background (as in the sample) when reproducing the issue. Probably because you are reducing the delta too much when calculating the maximum delta. 
Screenshot 2023-08-07 at 18 02 16

ChronosSF:
> There is still ~1 px of orange background (as in the sample) when reproducing the issue. Probably because you are reducing the delta too much when calculating the maximum delta. Screenshot 2023-08-07 at 18 02 16

Yeah I noticed but my belief is that it actually comes from imperfection of the percent calculations when the browser tries to render them with subpixel precision combined with the presence of the splitter element. Before the fix, what would happen is that the pane getting enlarged would receive additional size because it didn't take the minimum size of the shrinking panel into account when determining max size during the drag. This made the percentages exceed 100% and all kinds of weird issues coming from that. 

I'd say that if you can't reproduce this with pixel resizing, we can probably leave it as is unless you have other ideas.

ChronosSF:
As discussed with @MayaKirova , I added flex-grow/shrink for % panes to ensure this doesn't require resize observer.

projects/igniteui-angular/src/lib/splitter/splitter-pane/splitter-pane.component.ts
```diff
@@ -191,8 +191,8 @@ export class IgxSplitterPaneComponent {
     @HostBinding('style.flex')
     public get flex() {
         const isAuto = this.size === 'auto' && !this.dragSize;
-        const grow = !isAuto ? 0 : 1;
         const size = this.dragSize || this.size;
+        const grow = isAuto || (size as string).endsWith('%') ? 1 : 0;
```
MayaKirova, line 195: Could use isPercentageSize, also it should not grow/shrink while it's in the process of being dragged so:

```
const grow = this.isPercentageSize && !this.dragSize ? 1 : 0;
```

MayaKirova (changes requested):

dkamburov (approved):

PR #13384, opened 2023-08-17
fix(cellEditing): isCellInEditMode changed master

### LINKED ISSUE #13201: Two grid columns bound to one data fields leads to cell.editValue undefined upon editing
## Description  
cell.editValue is undefined when its column is bound to already existing column bound to the same Data Field.

 * igniteui-angular version: 16.0.10
 * browser: any

## Steps to reproduce  

1. Add two columns
3. Set them to one data field (example with UnitPrice)
4. Add custom editing template to the second column
5. Bound the cell to igx-input-group with ngModel
6. Run the sample and enter edit mode of the second column
7. Cell edit value is missing
8. If the first column is removed and the column with the edit template remains, everything is working as expected.

## Result  
Cell editing is not working when two grid columns bound to one data fields

## Expected result  
Cell editing to be working to the second column when two grid columns bound to one data fields 

## Attachments
https://stackblitz.com/edit/uvcpwp

GrblackStar:
Closes #13201

 - [X] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -4318,7 +4318,7 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
     public getColumnByName(name: string): IgxColumnComponent {
         return this._columns.find((col) => col.field === name);
     }
-
+
```
ddincheva, line 4321: please remove this file since there is no actual change in it

ddincheva (dismissed):

projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -255,7 +255,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
      * @memberof IgxGridCellComponent
      */
     public get context(): IgxCellTemplateContext {
-        const getCellType = () => this.getCellType(true);
+        const getCellType = () => this.editMode ? this.grid.crudService.cell : this.getCellType(true);
```
hanastasov, line 258: This looked initially to me like treating the symptom, rather than fixing the core issue. Debugging into `this.getCellType` I stumbled upon code with the same meaning:

image

It feels unnatural to introduce the second code here, when it is expected t get the correct info by using `this.getCellType`. So let;s change the fix to have `this.getCellType` return correct result.

hanastasov (changes requested):

projects/igniteui-angular/src/lib/grids/grid-public-cell.ts
```diff
@@ -60,6 +60,21 @@ export class IgxGridCell implements CellType {
...
     }
 
+    private searchColumnIndexByName(columnIndex: number, name: string): any {
+        const columns = this.grid.getAllColumnsByName(name);
+
+        if (columns.length === 0 || columnIndex < 0 || columnIndex >= columns.length) {
+            return -1;
+        }
+    
+        const columnAtIndex = columns[columnIndex];
```
hanastasov: columnIndex here is a "global" index, i.e. corresponding to column position among all grid columns. Using this index to query a column from a filtered columns collction (line 64), is not correct and will return an error when index exceeds collection length.

projects/igniteui-angular/src/lib/grids/grid-public-cell.ts
```diff
@@ -60,6 +60,21 @@ export class IgxGridCell implements CellType {
...
+        if (columnAtIndex.field === name) {
```
hanastasov: the columnAtIndex is initally retreieved by filtering columns by name (line 64). Then this columnn is compared again against the same `name`).

projects/igniteui-angular/src/lib/grids/grid-public-cell.ts
```diff
@@ -273,7 +288,14 @@ export class IgxGridCell implements CellType {
     }
 
     private isCellInEditMode(): boolean {
-        return this.grid.crudService.cellInEditMode;
+        if (this.grid.crudService.cellInEditMode) {
+            const cellInEditMode = this.grid.crudService.cell.id;
+            const isCurrentCell = cellInEditMode.rowID === this.id.rowID &&
+                cellInEditMode.rowIndex === this.id.rowIndex &&
+                cellInEditMode.columnID === this.searchColumnIndexByName(cellInEditMode.columnID, this._columnField);
```
hanastasov: I like this approach with introducing a method to look for a column by index(check visibleIndex too). But to make it more generic, it is the `get column` method to be fixed, so that correctresult is alwways returned, rather than fixed only for this case in the isCellinEditMode method

projects/igniteui-angular/src/lib/grids/cell.component.ts
ddincheva: Since the IgxGridCell constructor accept column field name or a column component 
image
 I can suggest just to pass the correct column when the grid cell type is created in the getCellType method... because we do not need to search for the column in the column list when the cell holds that reference already

hanastasov: > Since the IgxGridCell constructor accept column field name or a column component image I can suggest just to pass the correct column when the grid cell type is created in the getCellType method... because we do not need to search for the column in the column list when the cell holds that reference already

That's true and will solve it, but will still hide that the column getter may return incorrect info:
image

hanastasov: So I think one does not interfere with the other. Hrisi, I think you can go this route:
1) Fix the column getter to return correct info
2) When creating an instance of IgxGridCell, if possible pass the column object instead of the column field
3) Find other places in the code where column is searched by field name using find and see if it leads to errors

hanastasov (approved):

hanastasov (approved):

PR #13395, opened 2023-08-23
fix(grid: column-pinning): working pinColumn with indexes for the last position

### LINKED ISSUE #13391: pinColumn method does not work correctly with passed index of type number
## Description  
IgxGridComponents's [pinColumn](https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/IgxGridComponent.html#pinColumn) method does not work correctly with passed index of type number when we have already pinned columns and want to pin a column on the last place.

 * igniteui-angular version: ^15.1 ^16.0 ^16.1
 * browser: all

## Steps to reproduce  

1. Go to this StackBlitz [sample](https://stackblitz.com/edit/kyvgey?file=src%2Fapp%2Fgrid%2Fgrid-column-pinning%2Fgrid-column-pinning.component.html). 
2. Observe that there are already two pinned columns in the grid.
3. Press the first button that performs a function that pins a column with the pinColumn method, to which is passed the field of the column and the index 2 to pin the column on the last place, as the index is of type number.
 `this.grid.pinColumn("City", 2);`

## Result  
The column is pinned in the second place or in the middle between the other two columns, which is actually index 1 not 2.

## Expected result  
The column should be pinned on the third (last) place, i.e. at index 2.

## Additional information
If we press the second button, which performs a function that pins a column with the pinColumn method, but instead of passing a number-type index, a string-type index is passed, everything works as expected and the column is pinned on the last place after the other pinned columns.
`this.grid.pinColumn("City", "2");`

The same thing happens if we don't pass any index at all but only a column field to the pinColumn method everything works as expected and the column is pinned on the last place.
`this.grid.pinColumn("City");`

## Attachments  
issue

georgianastasov:
Closes #13391   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -2118,7 +2118,7 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
         }
 
         if (hasIndex) {
-            grid._moveColumns(this, targetColumn);
+            index === grid._pinnedColumns.length - 1 ? grid._moveColumns(this, targetColumn, 1) : grid._moveColumns(this, targetColumn, 0);
```
mddragnev, line 2121: You should use `DropPosition` enum when setting the drop position inside the `_moveColumns` method instead of plain numbers

mddragnev, line 2121: You should use DropPosition enum when setting the drop position inside the _moveColumns method instead of plain numbers

mddragnev (changes requested):

mddragnev:
Also, one thing made me look more closely into how this method works : `instead of passing a number-type index, a string-type index is passed, everything works as expected and the column is pinned on the last place after the other pinned columns.` This was you in the issue. I think that even if the index is passed as string it should fail in master (without your fix) but it does not fail because it cannot find a targetColumn and therefore it works exactly like if there is no index provided. There is a check `args.insertAtIndex === grid._pinnedColumns.length` which is false because you check whether `"2" === 2` resulting in undefined target column . I think that you should parse that string to a number

mddragnev (approved):

PR #13412, opened 2023-08-30
fix(*): multiple level directive spreads break tree-shaking

damyanpetev:
As far as I can tell, multiple levels of spreading directive arrays break the packaging of the library in such a way that the whole grid bundle (including all component dependencies) cannot be tree-shaken. From what I can see the one level produced by the user-facing per-grid directives arrays that still spread the common, paginator, etc seem to be doing fine. Unfortunately, that means we can't neatly bundle directives from separate modules into common and spread that again and I had to inline those and since it's related to the build tooling it can also be somewhat unpredictable still in the future.
I've left the unused directive arrays commented out for now, as I believe this could be an issue that might be resolved and allow for those to come back.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


kdinev:
@damyanpetev Should the same fix also be applied to `igx-accordion` and `igx-action-strip` public_api?

damyanpetev:
> @damyanpetev Should the same fix also be applied to `igx-accordion` and `igx-action-strip` public_api?

Those _should_ be fine, so far the issues I see are only with multiple levels of those const array spreads.

projects/bundle-test/tsconfig.app.json
```diff
@@ -6,7 +6,7 @@
     "types": [],
     "paths": {
         "igniteui-angular": [
-          "projects/igniteui-angular/src/public_api.ts"
+          "dist/igniteui-angular"
```
kdinev, line 9: Should this path change to `dist/`? I don't think we have to build the product in order to test.

kdinev (changes requested):

projects/bundle-test/tsconfig.app.json
damyanpetev: See last failed build, didn't quite take. Those internal refs don't get treated the same as a package, more like parts of the same app source. Same already happened with the i18n package where it _was_ tree-shaking without the fix when linking the source.
Also makes sense - we're interested in the shipped product and that is the dist output.

kdinev (approved):

PR #13474, opened 2023-09-19
feat(button-group): adding selectionMode property

### LINKED ISSUE #13372: [IgxButtonGroup] Deselection can't be disabled or prevented
## Description  
It seems like disabling the deselection of items is not possible.
One of the usual scenarios for buttongroups is using them as a substitution of a radio button list (but with a different visual pattern). In that case we'd need to have always one item selected. If you click a selected item, the selected item should stay selected, instead of switching back to its normal (not selected) state.

 * igniteui-angular version: 15.11.9

## Steps to reproduce  

1. Navigate to https://www.infragistics.com/products/ignite-ui-angular/angular/components/button-group
2. Click the selected item of the first example (text alignment items)

## Result  
After step 2, there's no selected item, so how should be the text aligned in that state?

## Expected result  
We could have a way to prevent the deselection of a selected item by clicking it, probably by setting a property (for example, [allowDeselection]="false", or something like that, I'm just thinking out loud).

I logged it as a bug and not as a feature request, because I consider this a basic feature of a buttongroup component. Please, feel free to move it as a feature request if you this that's better. 

## Workarounds
I tried some ways to prevent the deselection, but none of them was succeeded.

1. Deselected event can't be prevented or stopped.
I tried to prevent the deselected event (It's not a good approach, because we should prevent it again on every buttongroup instance), besides that, the event prevention is not working due to the way the component is being subscribed (on its internal code) to buttons click events.

2. Avoiding the use of selected and deselected events doesn't help.
When I don't use the selected and deselected events to avoid the deselection and I use the inner buttons click event instead, I found a different issue, the unselected state is visually set by the internal code, by ignoring the current state of the bound model.

georgianastasov:
Closes #13372   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
+++ b/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
@@ -42,12 +42,12 @@ let NEXT_ID = 0;
  * **Ignite UI for Angular Button Group** -
  * [Documentation](https://www.infragistics.com/products/ignite-ui-angular/angular/components/buttongroup.html)
  *
- * The Ignite UI Button Group displays a group of buttons either vertically or horizontally.  The group supports
- * single, multiple and toggle selection.
+ * The Ignite UI Button Group displays a group of buttons either vertically or horizontally. The group supports
+ * single, multi and singleRequired selection.
  *
  * Example:
  * ```html
- * <igx-buttongroup multiSelection="true" [values]="fontOptions">
+ * <igx-buttongroup selectionMode="multi" [values]="fontOptions">
  * </igx-buttongroup>
  * ```
  * The `fontOptions` value shown above is defined as:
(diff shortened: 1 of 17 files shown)


rkaraivanov (approved):

projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
```diff
@@ -119,12 +119,12 @@ export class IgxButtonGroupComponent extends DisplayDensityBase implements After
...
     /**
-     * An @Input property that enables selecting multiple buttons. By default, multi-selection is false.
+     * An @Input property that sets the selection mode of the buttons. By default, the selection mode is single.
      * ```html
-     * <igx-buttongroup [multiSelection]="false" [alignment]="alignment"></igx-buttongroup>
+     * <igx-buttongroup [selectionMode]="'multi'" [alignment]="alignment"></igx-buttongroup>
      * ```
      */
-    @Input() public multiSelection = false;
+    @Input() public selectionMode: 'single' | 'singleRequired' | 'multi' = 'single';
```
damyanpetev, line 127: @georgianastasov @rkaraivanov 
The PR says `multiSelection` is "deprecated", but this essentially replaces it with the new prop, which results in an actual breaking change, which I don't think was intended. Can we keep the old prop as deprecated and proxy get/set of the new `selectionMode`?

georgianastasov: There is a new PR [here](https://github.com/IgniteUI/igniteui-angular/pull/13521) with the new required changes.

PR #13482, opened 2023-09-20
fix(grid): slightly different attr for formatters

ig-robstoffers:
We should use `blazorAlternateType` here since we only want the C# output to use these formatter types. The typescript declarations that we generate from the api analyzer should continue using the original function type. This will fix the web components typedoc build.


kdinev:
@damyanpetev Pls review this

damyanpetev (approved):

Issue #13467, opened 2023-09-16
ng update @infragistics/igniteui-angular doesn't update igniteui-angular-i18n


kdinev:
## Description  
When I perform `ng update` on `igniteui-angular` the version of the `igniteui-angular-i18n` package stays the same, despite having a new version available.

 * igniteui-angular version: 16.0.x
 * browser: N/A

## Steps to reproduce  

1. Create a new project with `igniteui-angular` and `igniteui-angular-i18n`
2. Use a version for both different from the latest
3. Perform `ng update`

## Result  
Only the version of `igniteui-angular` gets updated

## Expected result  
The version of the i18n package should also get bumped.

PR #13513, opened 2023-09-29
fix(migrations): remove licensed package from group

Lipata:
#13467
The fix comes with a task in the licensed release pipeline, which replaces "igniteui-angular" with "@infragistics/igniteui-angular"

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular-i18n/package.json
+++ b/projects/igniteui-angular-i18n/package.json
@@ -26,7 +26,6 @@
     "packageGroup": [
         "igniteui-angular-i18n",
         "igniteui-angular",
-        "@infragistics/igniteui-angular",
         "@igniteui/angular-schematics"
     ]
   }
--- a/projects/igniteui-angular/package.json
+++ b/projects/igniteui-angular/package.json
@@ -92,7 +92,6 @@
         "migrations": "./migrations/migration-collection.json",
         "packageGroup": [
             "igniteui-angular",
-            "@infragistics/igniteui-angular",
             "igniteui-angular-i18n",
             "@igniteui/angular-schematics"
         ]


kdinev (dismissed):

Lipata:
@kdinev, @valadzhov, @damyanpetev it seems that when two packages are in a group only one is visible when `ng update` is called. Of course, both are updated when the `ng update igniteui-angular` is called.

When the group config in the `igniteui-angular-i18n` is the same as in the `igniteui-angular`:

```
        "packageGroup": [
            "igniteui-angular",
            "igniteui-angular-i18n"
        ]
```

then `igniteui-angular` is visible on `ng update`:

image

If config is empty them again only `igniteui-angular-i18n` is visible on `ng update`:

image

I suggest using the first approach

kdinev:
> @kdinev, @valadzhov, @damyanpetev it seems that when two packages are in a group only one is visible when `ng update` is called. Of course, both are updated when the `ng update igniteui-angular` is called.
> 
> When the group config in the `igniteui-angular-i18n` is the same as in the `igniteui-angular`:
> 
> ```
>     "ng-update": {
>         "packageGroup": [
>             "igniteui-angular",
>             "igniteui-angular-i18n"
>         ]
>     },
> ```
> 
> then `igniteui-angular` is visible on `ng update`:
> 
> image
> 
> If config is empty:
> 
> ```
>   "ng-update": {
>   }
> ```
> 
> then again only `igniteui-angular-i18n` is visible on `ng update`:
> 
> image
> 
> I suggest using the first approach

`igniteui-angular-i18n` is not usable on its own, so the second approach doesn't really make sense.

damyanpetev:
Bit of an issue with that one - i18n is just one package, so can't name its group the trial or licensed product package :/
We might need to be satisfied with it being updated by the product package and still being visible in the ng update listing regardlessu (i.e. empty group)

kdinev:
> Bit of an issue with that one - i18n is just one package, so can't name its group the trial or licensed product package :/ We might need to be satisfied with it being updated by the product package and still being visible in the ng update listing regardlessu (i.e. empty group)

Apparently I didn't understand the previous comment. What does the `package.json` in the project running `ng update` look like? What is the `first` and `second` approach?

damyanpetev (approved):

PR #13550, opened 2023-10-13
Unsuppress col group collapsible prop.

MayaKirova:
Related to: https://infragistics.visualstudio.com/NetAdvantage/_workitems/edit/27337


gmurray81:
I can't really think of why we had this suppressed in blazor... it could be that its because its an override an it was already exposed via the base class? Or maybe it didn't like if it shifts from being read only to writeable? Anyhow, we should make sure its exposed for both react and blazor, but if this was suppressed, it was probably causing a compilation issue, I'd guess.

MayaKirova:
@gmurray81 
Not sure how to test it for blazor. At a glance it seems to generate correctly there as well:

```
	[Parameter]
	public bool Collapsible 
	{
	get { return this._collapsible; }
	set { if (this._collapsible != value || !IsPropDirty("Collapsible")) { MarkPropDirty("Collapsible"); } this._collapsible = value; }
	}
```

So I could try to enable it for all and see if any builds fail.

skrustev (approved):

gmurray81 (approved):

PR #13558, opened 2023-10-17
feat(esf): improve search results accessibility - 16.1.x

### LINKED ISSUE #13524: [Screen Reader] Screen reader does not announce search results information by entering 'valid/invalid' data under 'Name' dialog.
Screen reader users rely on auditory feedback to understand the outcome of their actions. If search results are not announced, users won't know whether their search query was successful or if there were any relevant matches.

Screen reader does not announce search results information by entering 'valid/invalid' data under 'Name' dialog.
NVDA remain silent.

Expected Result:

Screen reader should announce search results information by entering 'valid/invalid' data under 'Name' dialog.

source link: [Excel Style Filtering in Angular Data Grid- Ignite UI for Angular (infragistics.com)](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/excel-style-filtering)

onlyexeption:
Closes #13524 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


ChronosSF:
LGTM, though since the previous implementation is very non-accessible, I think we can safely add this to all supported versions, not just 17.0

I am also unsure... maybe when we make NVDA say how many matches there are, we should only include the actual matches and not the Select All / Add to Selection checkboxes as while they are in the UI, they are not exactly matching the search.

@sdimchevski , @kdinev , do you have comments about that?

ChronosSF:
@onlyexeption , speaking about this with @rkaraivanov and @damyanpetev , we decided that the choice of approach would need some further research. I'll freeze the PR for the time being and revisit it next sprint most likely.

ChronosSF:
@onlyexeption , @igdmdimitrov , to continue with this item please research for accessibility patterns and specs around search fields.

onlyexeption:
@damyanpetev, @rkaraivanov, after conducting a little research going through popular websites and applications with the NVDA reader, I've found out that the only websites (out of many) where search results were announced, are using the same approach with a hidden div that updates dynamically. Screenshots are attached for reference. LMK what you think.

image
⠀
⠀
⠀
image
⠀
⠀
⠀
image
⠀
⠀
⠀
image

ChronosSF:
We discussed this today again with @damyanpetev and decided that the approach is okay, as long as you change it to polite from assertive and localize the message. The localization should also allow for changing the position of the number and other words within. So basically for English it should be something like ```{num} matches.``` and then you will replace the result in the template.

ChronosSF (dismissed):

ChronosSF (approved):

PR #13584, opened 2023-10-23
fix(snackbar): use igxButton

### LINKED ISSUE #13574: igx-snackbar doesn't use igxButton
## Description  
The `igx-snackbar` doesn't use the `igxButton`. Instead it has it's own button class: `igx-snackbar__button`.

 * browser: n/a

Also, the standalone definition in 16.x+ doesn't import `IgxRippleDirective` even though the ripple is used in the template.

SisIvanova:
Closes #13574 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
@@ -108,6 +108,10 @@
             -webkit-tap-highlight-color: transparent;
             box-shadow: none;
         }
+
+        .custom-btn {
+            display: contents;
+        }
     }
 
     %igx-snackbar-message {
--- a/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
@@ -108,10 +108,10 @@
             -webkit-tap-highlight-color: transparent;
             box-shadow: none;
         }
+    }
 
-        .custom-btn {
-            display: contents;
-        }
+    %igx-snackbar-button {
+        display: contents;
     }
 
     %igx-snackbar-message {
--- a/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
@@ -101,22 +101,12 @@
         backdrop-filter: blur(8px);
 
         [igxButton] {
+            @include animation(fade-in .35s ease-out);
             --ig-size: 1;
             background: transparent;
             color: var-get($theme, 'button-color');
-            text-transform: uppercase;
-            user-select: none;
-            font-weight: $snackbar-button-font-weight;
             -webkit-tap-highlight-color: transparent;
-            outline: none;
             box-shadow: none;
-            transition: color .2s ease;
-            font-family: inherit;
-            @include animation(fade-in .35s ease-out);
-
-            &:hover {
-                color: var-get($theme, 'button-color');
-            }
         }
     }
 
--- a/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
@@ -81,8 +81,6 @@
 
     $snackbar-min-height: rem(48px);
     $snackbar-padding: rem(7px) rem(24px);
-    $snackbar-button-left-margin: rem(24px);
-    $snackbar-button-line-height: rem(16px);
     $snackbar-button-font-weight: 600;
 
     %igx-snackbar-display {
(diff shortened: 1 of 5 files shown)


projects/igniteui-angular/src/lib/snackbar/snackbar.component.html
```diff
@@ -2,6 +2,6 @@
     {{ textMessage }}
     <ng-content></ng-content>
 </div>
-<button class="igx-snackbar__button" igxRipple="white" *ngIf="actionText" (click)="triggerAction()">
+<button igxButton *ngIf="actionText" (click)="triggerAction()">
```
kdinev, line 5: @SisIvanova I see you're getting rid of the default ripple? Btw, it doesn't show because the import of the ripple is not included in the component, but if you add it, it would show. Does this button have a ripple by default according to spec? @simeonoff

projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
```diff
@@ -94,39 +92,37 @@
...
-        -webkit-tap-highlight-color: transparent;
-        outline: none;
-        transition: color .2s ease;
-        font-size: inherit;
-        font-family: inherit;
-        cursor: pointer;
-        @include animation('fade-in' .35s ease-out);
-
-        &:hover {
+        [igxButton] {
```
kdinev, line 103: @simeonoff You will review the actual styles but are all of the overrides included here actually needed?

projects/igniteui-angular/src/lib/snackbar/snackbar.component.html
simeonoff: Since the configuration of the ripple is internal to the template, i.e. not configurable, we decided to get rid of it. However, we can make it configurable by providing in input, say `buttonRippleColor`, or by making the button templateable. In that way we can either assign the default value of the `buttonRippleColor` to be `transparent`, while still retaining the functionality or the users can just pass in their own button to which they can attach a ripple. What do you think?

projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
simeonoff: We are checking. Will compare to the UI kit and web components framework and remove any unnecessary styles.

projects/igniteui-angular/src/lib/snackbar/snackbar.component.html
kdinev: Let's get rid of it in the default template. Not to mention it's not working since `16.0.0` :). Being able to provide your own button covers the scenario of being able to also add `igxRipple` to it.

kdinev: It can be something of the sort:

```html
<ng-content #customButton select="[igxButton]"></ng-content>
<button *ngIf="!customButton" igxButton *ngIf="actionText" (click)="triggerAction()">{{actionText}}</button>
```

projects/igniteui-angular/src/lib/snackbar/snackbar.component.ts
```diff
@@ -102,6 +104,22 @@ export class IgxSnackbarComponent extends IgxNotificationsDirective
      */
     @Output() public animationDone = new EventEmitter<ToggleViewEventArgs>();
 
+    /**
+     * @hidden
+     */
+    public get buttonTemplate(): boolean {
```
kdinev, line 110: I don't think you need this. Did it now work by simply putting an identifier on the ng-content and then checking that that element exists?

projects/igniteui-angular/src/lib/snackbar/snackbar.component.spec.ts
```diff
@@ -221,7 +221,7 @@ describe('IgxSnackbar with custom content', () => {
         expect(messageElRect.left <= customContentRect.left).toBe(true, 'The message is not on the left of the custom content');
 
         // Verify the custom content element is on the left side of the button
-        const button = fixture.debugElement.query(By.css('.igx-snackbar__button'));
+        const button = fixture.debugElement.query(By.css('.igx-button'));
```
kdinev, line 224: Also add a new test checking that adding a custom `igxButton` works.

projects/igniteui-angular/src/lib/snackbar/snackbar.component.html
```diff
@@ -2,7 +2,9 @@
     {{ textMessage }}
     <ng-content></ng-content>
 </div>
-<ng-content *ngIf="buttonTemplate" select="[igxButton]"></ng-content>
-<button *ngIf="!buttonTemplate && actionText" igxButton (click)="triggerAction()">
+<div #customButton class="custom-btn">
```
kdinev, line 5: Have you tried using `ng-container` instead of `div`? Then it shouldn't interfere with the DOM and the result would be identical with default and custom button. You would also not need the class.

kdinev (approved):

projects/igniteui-angular/src/lib/core/styles/components/snackbar/_snackbar-theme.scss
```diff
@@ -108,6 +108,10 @@
             -webkit-tap-highlight-color: transparent;
             box-shadow: none;
         }
+
+        .custom-btn {
+            display: contents;
+        }
```
simeonoff, line 114: Use BEM, please. Let's keep it consistent across.

simeonoff (dismissed):

simeonoff (approved):

Issue #13568, opened 2023-10-18
Pressing Enter in IgxSimpleCombo throws error


Timmeeeey:
## Description  
When you select a value in IgxSimpleCombo and press Enter, then an error is thrown.
Sometimes another item is selected and the error is thrown after pressing Enter multiple times.

`TypeError: Cannot read properties of null (reading 'itemID')`

In following code line:
https://github.com/IgniteUI/igniteui-angular/blob/913905bca392b01f5c0b6c74df920549abec89dd/projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts#L306

 * igniteui-angular version: 16.1.2
 * browser: Chrome

## Steps to reproduce  

1. Open https://stackblitz.com/edit/ezpv2p
2. Select Item 3
3. Press Enter multiple times

## Result  
First Item 31 gets selected then Item 311 and then the error is thrown.

## Expected result  
Pressing Enter should not throw an error or select a different item.

## Attachments  
https://stackblitz.com/edit/ezpv2p

![2023-10-19_09h09_11](https://github.com/IgniteUI/igniteui-angular/assets/46870283/94de15c9-8032-4267-859e-710e8368d3cd)

ChronosSF:
@Lipata , we'll have to look into these simple combo issues next sprint.

PR #13618, opened 2023-10-31
fix(simple-combo): only select with enter when dropdown is open - 17.0

ChronosSF:
Closes https://github.com/IgniteUI/igniteui-angular/issues/13568 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


wnvko:
It is look like the `dropdown.focusedItem` for some reason is `null`. This should not happen. The fix now workarounds this issue but does not fix it. Try to find out why the `focusedItem` is `null` and fix this.

jackofdiamond5:
I don't think the reason for this error is because the `focusItem` is `null` when the combo is closed. It kind of makes sense to not have a `focusItem` as long as the dropdown is closed, however, it seems to me like the real issue here is caused by something that is moving the `focusItem` every time the `ENTER` key is pressed, If we notice the gif that's provided in the issue, we will see that, at first, an item is selected from the list, then the `ENTER` key is pressed which causes the selected item to change, which **should never happen**, pressing the `ENTER` key again causes the item to shift once more, and it always seems to go to an item after the currently selected one. And after there are no more items, then `focusItem` becomes `null` and the error is thrown. So imho the problem here is not that the `focusitem` is `null` but that something is modifying it every time `ENTER` is pressed. Will need to investigate further to figure out what is causing this though.

wnvko (approved):

PR #13632, opened 2023-11-03
fix(button-group): reverted cancellable on selected/deselected events and adding selecting/deselecting cancellable events

### LINKED ISSUE #13580: Button group selected and deselected events are fired before the buttons are selected and deselected
## Description  
The `selected` and `deselected` events of the `IgxButtonGroup` component are fired before the buttons are selected and deselected.

 * igniteui-angular version: 15.1 16.1

## Steps to reproduce  

1. Go to this [sample](https://stackblitz.com/edit/pdpfrn-8aayy3?file=src%2Fapp%2Fbutton-group-sample-1%2Fbutton-group-sample-1.component.html). 
2. Open the dev console.  
3. Select one button from the button group component. 
4. Observe that the `selected` event is fired but the selected buttons `length` is **0**. 
5. Deselect the button from the button group component.
6. Observe that the `deselected` event is fired but the selected buttons `length` is **1**. 

## Result  
When a button is selected from the button group component, the `selected` event is fired before the button is selected and the `length` of the selected buttons is **0**. When a button is deselected from the button group component, the `deselected` event is fired before the button is deselected and the `length` of the selected buttons is **1**.

## Expected result  
When a button is selected from the button group component, the `selected` event should fire after the button is selected and the `length` of the selected buttons should be **1**. When a button is deselected from the button group component, the `deselected` event should fire after the button is deselected and the `length` of the selected buttons should be **0**.

## Attachments  
issue

georgianastasov:
Closes #13580 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
+++ b/projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
@@ -23,7 +23,7 @@ import { IgxRippleDirective } from '../directives/ripple/ripple.directive';
 
 import { takeUntil } from 'rxjs/operators';
 import { DisplayDensityBase, DisplayDensityToken, IDisplayDensityOptions } from '../core/density';
-import { CancelableEventArgs, IBaseEventArgs } from '../core/utils';
+import { IBaseEventArgs } from '../core/utils';
 import { mkenum } from '../core/utils';
 import { IgxIconComponent } from '../icon/icon.component';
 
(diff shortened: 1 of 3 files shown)


ChronosSF (changes requested):

ChronosSF (dismissed):

ChronosSF:
So testing this I came by an interesting issue - @damyanpetev , @georgianastasov - in an exclusive button group, cancelling `deselecting` should also cancel the subsequent selection, right? Currently cancelling `deselecting` in this scenario has no effect.

damyanpetev:
> So testing this I came by an interesting issue - @damyanpetev , @georgianastasov - in an exclusive button group, cancelling `deselecting` should also cancel the subsequent selection, right? Currently cancelling `deselecting` in this scenario has no effect.

Haven't gotten around to reviewing this one, though thought we were already aware of it. But yeah - the `-ing` events are not set up as they should be, will leave review comments in a bit

projects/igniteui-angular/src/lib/buttonGroup/buttonGroup.component.ts
```diff
@@ -482,34 +516,40 @@ export class IgxButtonGroupComponent extends DisplayDensityBase implements After
...
         const button = this.buttons[index];
-        const args: IButtonGroupEventArgs = { cancel: false, owner: this, button, index };
+        const args: IButtonGroupEventArgs = { owner: this, button, index };
+        const cancellableArgs: IButtonGroupCancellableEventArgs = { cancel: false, owner: this, button, index };
 
         if (this.selectionMode !== 'multi') {
             this.buttons.forEach((b, i) => {
                 if (i !== index && this.selectedIndexes.indexOf(i) !== -1) {
-                    this.deselected.emit({ cancel: false, owner: this, button: b, index: i });
+                    this.deselecting.emit({ cancel: false, owner: this, button: b, index: i });
```
damyanpetev, line 525: Yeah, this won't do the trick - this was an issue with the previous attempt for cancellable as well:
This deselection is a side effect of the actual selection of something else, right? Although it's a bit odd in a loop, we're talking single selection and one button active that needs to be deselected because another button is selected.
Anyway, in this case I'd expect you to cancel the primary action (selecting) and this portion only emits the old `deselected` event as a by-product that can't be avoided, otherwise you end up in an invalid state against the more, which I don't think we want to do.
In other words - this entire operation (deselecting an active button) cannot happen before `this.selecting.emit(cancellableArgs);` is emitted and not cancelled. Otherwise, this shouldn't happen at all.

ChronosSF: If we are not firing deselecting for exclusive selection, you should also add this to the API docs.

damyanpetev (approved):

PR #13649, opened 2023-11-08
fix(lint): speeding up lint a bit

kdinev:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [x] CI/CD

### Checklist:


.eslintignore
```diff
@@ -0,0 +1,6 @@
+dist/
+projects/igniteui-angular/migrations/
+projects/igniteui-angular/schematics/
```
damyanpetev, line 3: I'd actually argue that I'd like for the code in migrations/schematics to still be linted though. As I understand it those don't work ATM because they build separately and thus use a different TS project/config and that's part of the slow-down.

@Lipata Can you log a task for this to be investigated, assuming a separate lintconfig might be needed for just schematics/migrations to configure it correctly, adjust rules if needed and verify it doesn't break run times as well.

damyanpetev (approved):

PR #13678, opened 2023-11-13
fix(dropdown,select): add igxFor DI alias for tree-shaking

### LINKED ISSUE #13677: select and drop down don't tree-shake `igxForOf`

damyanpetev:
Closes #13677
Note I've intentionally kept the token internal and kept it out of the public export.
Also didn't add the full public API, just the bits needed internally.

Before in issue above and after:
image
image

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -343,6 +343,7 @@ export class IgxDropDownComponent extends IgxDropDownBaseDirective implements ID
             return;
         }
         let targetScroll = this.virtDir.getScrollForIndex(this.selectedItem.index);
+        // TODO: This logic _cannot_ be right, those are optional user-provided inputs that can be strings with units, refactor:
         const itemsInView = this.virtDir.igxForContainerSize / this.virtDir.igxForItemSize;
```
damyanpetev, line 347: @Lipata Added this comment since there's no way this logic is correct:
`igxForItemSize` and `igxForContainerSize` are documented accepting strings with units (like `'500px'`) and as such this is only written for/tested with a narrow scenario. Evaluate the need for this logic entirely, I'm assuming the `igxFor` should already have a method to scroll to a specific item so all of those calculations could be replaced with `scrollTo(this.selectedItem.index)` ideally and the props can be removed from the token. Can add in this PR if you're up for testing

projects/igniteui-angular/src/public_api.ts
```diff
@@ -19,7 +19,9 @@ export * from './lib/directives/drag-drop/public_api';
 export * from './lib/directives/filter/filter.directive';
 export * from './lib/directives/focus/focus.directive';
 export * from './lib/directives/focus-trap/focus-trap.directive';
-export * from './lib/directives/for-of/for_of.directive';
+export {
+    IgxForOfContext, IgxForOfDirective, IForOfState, IgxGridForOfContext, IgxGridForOfDirective
+} from './lib/directives/for-of/for_of.directive';
```
damyanpetev, line 24: The star export also included a random `getTypeNameForDebugging` function and `IForOfDataChangingEventArgs` for an internal event, which I skipped along with the new token
// CC @kdinev

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
damyanpetev: Okay, tested with `scrollTo(this.selectedItem.index)` which works but just barely gets the item in view, often at the bottom of the list, which I assume is not the goal. However, the Combo overrides this method with 
```ts
this.virtDir.getScroll().scrollTop = this._scrollPosition;
```
And `this._scrollPosition` is something the dropdown maintains and that seems to work just fine. Except it no longer tries to center, but I think we can reduce this code a bit if we're okay with a minor behavior change that will have the combo and virtual dd behaving the same way. Might be worth it.

kdinev (approved):

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
Lipata: As we talked - when re-opening a combo, it positiones the selected item in the middle and this is for both - local and remote binding. You are right that we can reduce that amount of code, but I prefer not to change that functionality - because of the slight UI behavior change and the need to change it for local and remote in order to keep the consistency.

Lipata (approved):

PR #13706, opened 2023-11-21
samples(grid): update cell editing sample

simeonoff:
Update the size of the grid using `--ig-size` to showcase how the size of the grid can be updated without binding to `displayDensity`. Useful for testing and validating `--ig-size` with the grid component.


ChronosSF:
This PR is no longer needed because 17.0.x will not receive future patches.

Issue #13668, opened 2023-11-10
Migrations are failing, when @angular-devkit/core is not on root level in the node_modules


Lipata:
## Description  
It is possible that the @angular-devkit/core is not at the root level in the node_modules. Migrations are developed so that they require that package on the root level, however, we should revisit that.

 * igniteui-angular version: 17.0.x
 * browser: N/A
 
## Steps to reproduce  

1. Have a project with @angular/core & @angular/cli version 16.2
2. ng update @angular-eslint/schematics @angular/cli @angular/core @infragistics/igniteui-angular
or ng update @angular-eslint/schematics @angular/cli @angular/core @infragistics/igniteui-angular @igniteui-cli

## Result  
Migrations for igniteui-angular or igniteui-cli fails

## Expected result  
Migrations are executed

Lipata:
We will do the following:
 - dig deeper to understand the reason behind the missing @angular-devkit/core package on the root level.
 - in any case we should: 
     - see if we can get rid of the @angular-devkit/core dependency in the migrations
     - or see if we can load @angular-devkit/core dynamically, like we are doing with the language service

kdinev:
@Lipata We can't rely on a package we depend on to just be present. We have to make sure we load the package regardless of whether the project already has it or not.

jackofdiamond5:
@kdinev It seems to me like during an `ng update` to `17` the package `@nguniversal/builders` gets replaced with the package `@angular-devkit/build-angular` so it is removed from the `package.json` file. 

Screenshot during `ng update`:
![image](https://github.com/IgniteUI/igniteui-angular/assets/16020256/0c957371-39e3-4c54-ba7d-b9fd83f5e9eb)

However, it is still present in the lock file which causes a peer dependency conflict, so the `update` schematic enforces an `npm install` which messes up the dependency tree and the resulting lock file does not have an `@angular-devkit/core` package at the root level, instead, it is only present for packages that explicitly list it as a dependency. This in turn causes the error in our migrations.

I managed to reproduce this behavior on a clean Angular project (without using `ng update`) using only dependencies and dev-dependencies from the `bellum-gens` repo. One solution for this problem is for the project to do a clean install or to remove the lock file before updating. As for `igniteui-angular` we may be able to work around this by getting rid of all the code that we use from the `@angular-devkit/core` package. This is far from ideal though as a much better solution (as proposed by @damyanpetev) could be to separate the migrations logic to a different package and install that during an update.

[...]

PR #13712, opened 2023-11-27
fix(migrations,ng-add): turn on encapsulation for devkit/schematics deps

damyanpetev:
Related to #13668

As suggested by the Angular Team - using `"encapsulation": true` for the migration/schematics collections. This allows imports (well, `require`-s really in the node context), to pass through the Angular CLI-s context - i.e. resolve that package's deps. While useful if the tree messes up and our imports won't fail, it should in theory also more closely guarantee the version used, assuming the ng CLI is updated first.

The related code that handles the `encapsulation` option mostly lives here: https://github.com/angular/angular-cli/blob/4da732ddf686198ae598698dc4c74515d46624f0/packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts

This is based on Node's [`module.createRequire`](https://nodejs.org/docs/latest-v18.x/api/module.html#modulecreaterequirefilename) to create a context-bound require from the Angular CLI and override in consumed schematics using the VM context API to run said script with the overrides via [`vm.runInNewContext`](https://nodejs.org/docs/latest-v20.x/api/vm.html#scriptruninnewcontextcontextobject-options).
Note that this _restricts_ the global API surface schematics/migrations can access as from what I can tell the new context is final in the sense that it doesn't backfill global API if not provided and so the list is fairly short:
https://github.com/angular/angular-cli/blob/4da732ddf686198ae598698dc4c74515d46624f0/packages/angular/cli/src/command-builder/utilities/schematic-engine-host.ts#L218-L234
The basics (`__dirname`, `__filename`, `console`, `process`) are available, but we should be really careful not to step outside of those in the direct source (supporting packages should still be okay).

Edit: Apparently will also need to be strict about circular refs, since that wrap is recursive and doesn't handle circular refs like node would, leading to 'Maximum call stack size exceeded' error.
The other part we were warned about are dynamic imports, which are not supported. The one we use atm was a workaround, will investigate if possible to remove.

Update:
Okay, so for the most part it seems we're not really going too far outside of the supported API. However, the dynamic import is still needed for `@angular/compiler` since that's ESM and schematics are strictly CommonJS still and there's no alternative, regardless if we use the pre-compiled js or not. The warning turned out true though, see the [`new vm.Script`](https://nodejs.org/docs/latest-v20.x/api/vm.html#new-vmscriptcode-options) notes under `importModuleDynamically`. So that's not supported yet at all (behind experimental flag) and I get that error if a migration uses that.
I do have a very cheeky workaround for that - if we require the specific file that only does the dynamic import as non-relative, it's a way to escape the wrap functionality and have that file run in the normal non-encapsulated context:
image
https://github.com/damyanpetev/bellum-gens/commits/encapsulation-test

Note: This will require us to use that in the source and some work for the licensed pipelines to edit it.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


damyanpetev:
Okay, I've pushed the change to use the bare specifier to escape the encapsulation on that specific helper so we can still use the dynamic import and branch should be working/testable now.
Not a whole lot I can find regarding this specific use case, but the docs on https://nodejs.org/docs/latest-v18.x/api/esm.html#import-specifiers don't explicitly forbid it and without going too deep in checks the https://nodejs.org/docs/latest-v18.x/api/esm.html#resolution-algorithm does mention even selfUrl so for the time being I'm leaning towards this being valid use.

There are path maps for the build and a task in the licensed pipe to update the package name in place already. I suspect debugging locally might be impacted, yet to try.
Note this is still a bit fragile (like the rest in there really) as it relies on the current encapsulation check mechanism in the schematics engine.

Lipata (approved):

Lipata:
Tested with a very old version of `igniteui-angular-samples` browser and I've run all the existing 32 migrations with success. The 'ng add' schematic is also running with no problems.

@damyanpetev you can create a branch to master.

PR #13774, opened 2024-01-03
chore(*): Move event handlers from the base directive. Remove suppress…

MayaKirova:
…es. Add also some marshalByValue attributes.

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/state-base.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/state-base.directive.ts
@@ -500,11 +500,7 @@ export class IgxGridStateBaseDirective {
      * this.state.setState(gridState);
      * ```
      */
-    protected setStateInternal(state: IGridState | string, features?: GridFeatures | GridFeatures[]) {
-        if (typeof state === 'string') {
-            state = JSON.parse(state) as IGridState;
-            this.stateParsed.emit(state)
-        }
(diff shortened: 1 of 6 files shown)


projects/igniteui-angular/src/lib/grids/state-base.directive.ts
```diff
@@ -429,16 +431,6 @@ export class IgxGridStateBaseDirective {
...
 
-    /**
-     *  Event emitted when set state is called with a string.
-     * Returns the parsed state object so that it can be further modified before applying to the grid.
-     * ```typescript
-     * this.state.stateParsed.subscribe(parsedState => parsedState.sorting.forEach(x => x.strategy = NoopSortingStrategy.instance()});
-     * ```
-     */
-    @Output()
-    public stateParsed = new EventEmitter<IGridState>();
```
skrustev, line 440: I am not sure why the move the event from base directive.
But anyway the event is emitted in `setStateInternal`, so emit should be moved outside in the inheriting directive and component as well.

skrustev (changes requested):

dkamburov (approved):

projects/igniteui-angular/src/lib/grids/state-base.directive.ts
dkamburov: @skrustev can you please review it again?

skrustev (approved):

PR #13785, opened 2024-01-08
fix(grid): Correct shift-click selection in grouped IgxGrid #13757

### LINKED ISSUE #13757: IgxGrid: Shift click not working correctly when grouping is activated
## Description  
Shift click doesn't work correctly when grouping is activated.

 * igniteui-angular version: 17.0.6
 * browser: any

## Steps to reproduce  
https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/groupby#angular-grid-group-by-example

2023-12-13_10h44_58

## Result  
7 rows are selected.

## Expected result  
Only 3 rows should be selected.

valeriatoneva:
Closes #13757

This PR addresses the issue where shift-click was not functioning correctly in an IgxGrid with grouped rows. The selection logic has been updated to properly handle multi-level group comparisons, ensuring that the shift-click selection is consistent and intuitive in grouped grids.

Additional information:
- [x] Bug fix
- [ ] New functionality
- [ ] Documentation
- [ ] Demos
- [ ] CI/CD

Checklist:
- [ ] This PR includes feature/README.MD updates for the feature docs
- [ ] This PR includes general feature table updates in the root README.MD
- [ ] This PR includes CHANGELOG.MD updates for newly added functionality
- [ ] This PR includes ng update migrations for the breaking changes


projects/igniteui-angular/src/lib/grids/selection/selection.service.ts
```diff
@@ -586,24 +586,36 @@ export class IgxGridSelectionService {
...
+        const lastSelectedGroup = this.getGroup(this.getRowDataById(lastSelectedRowID));
+    
+        if (!this.areGroupsEqual(currentGroup, lastSelectedGroup)) {
+            this.selectRowById(rowID, false, event);
+            return;
+        }
+    
+        const rangeStart = Math.min(currentRowIndex, lastSelectedRowIndex);
+        const rangeEnd = Math.max(currentRowIndex, lastSelectedRowIndex) + 1;
+        const rowsToSelect = gridData.slice(rangeStart, rangeEnd).filter(row =>
```
hanastasov, line 611: Generally It seems you grasped what the issue is - the grouped data displayed in the grid does not represent what each row index is in the gridData collection, i.e gridData collection is not the grouped data. The previous code is wrongly assuming `gridData` is what is displayed and is calculating rows to be selected based off on the indexes found in gridData.

I think your fix conception is right - after getting all potential selected rows, find the group they exist in and only select rows from the current group.

I think the code you have written to do this might be optimized. I am pretty sure what you do should exist somewhere as code ( the fact that the group you are trying to define is already displayed, means it has already been resolved/calculated somewhere(). 

Same goes for the `areGroupsEqual` function - most probably we will be reusing some code for that. Still I can give you a feedback - this function is good but it seems to do only a root level comparison of the object - if it has hierarchy in it, this methoud does not considers it.

We will discuss those tomorrow :) Good job so far!

hanastasov:
Please remove the changes from package, package-lock files, the forgotten console.logs, and formattings that appear in the file. Also, add unit test covering the scenario reported. The unit test should fail without the fix and pass with the fix.

PR #13883, opened 2024-02-08
ci(schematics): skipLibCheck on compile for schematics

Lipata:
Enable `skipLibCheck` for schematics compilation, because `node_modules/@types/node` contains multiple versions and leads to the following [problem](https://github.com/IgniteUI/igniteui-angular/actions/runs/7831251859/job/21367400835). The solution was taken from: https://bobbyhadz.com/blog/typescript-duplicate-identifier#pulling-in-typings-from-multiple-versions-of-a-package.

I've tried to update the @types/node package, but even with the latest, it contains multiple versions and the tsconfig setting is needed. ~Anyway, all the types are updated along with the typescript package.~ Typescript was initially updated, along with @types, but we prefer to be automatically updated by a major/minor ng-update.

The `build:schematics` script can be used for verification.


damyanpetev:
Correction on the description: `@types/node` has had this split by `typesVersions` folders since at least v16 (likely earlier) and moving through supported TypeScript versions. At least in the previous version we had on this duplicate for URL _was_ present https://unpkg.com/browse/@types/node@18.11.19/ts4.8/url.d.ts but not double-included in the build process as intended.

What has changed though:
v16 of the DevKit schematics did not have explicit references https://unpkg.com/browse/@angular-devkit/schematics@16.2.12/src/engine/engine.d.ts
vs v17 https://unpkg.com/browse/@angular-devkit/schematics@17.1.3/src/engine/engine.d.ts:
```ts
/// <reference types="@types/node/url" />
/// <reference types="@types/node/ts4.8/url" />
```
Which causes this:
```text
node_modules/@types/node/url.d.ts
  Type library referenced via '@types/node/url' from file 'node_modules/@angular-devkit/schematics/src/engine/engine.d.ts' with packageId '@types/node/url.d.ts@20.8.10'
  Referenced via 'url.d.ts' from file 'node_modules/@types/node/index.d.ts'
node_modules/@types/node/ts4.8/url.d.ts
  Type library referenced via '@types/node/ts4.8/url' from file 'node_modules/@angular-devkit/schematics/src/engine/engine.d.ts' with packageId '@types/node/ts4.8/url.d.ts@20.8.10'
```
and then the build error:
```text
node_modules/@types/node/ts4.8/url.d.ts:365:11 - error TS2300: Duplicate identifier 'URL'.
365     class URL {
              ~~~
  node_modules/@types/node/url.d.ts:365:11
    365     class URL {
                  ~~~
    'URL' was also declared here.
```
Not exactly sure what the compile reason is for those references (I believe those are auto-emitted) and it might be a third-party issue to begin with.

However, the default empty schematic project as provided by the Schematics CLI via `schematics blank` also sets `skipLibCheck` by default (https://github.com/angular/angular-cli/blob/860b93a25bd7a013e1542e09f96766bb7f384e39/packages/angular_devkit/schematics_cli/blank/project-files/tsconfig.json#L16) and also fails similarly if disabled for the same reason, so I guess running without the check is not uncommon and we could do that.

PR #13896, opened 2024-02-12
fix(grid,tree-shake): use lightweight token for paginator

damyanpetev:
Added a DI alias for the paginator using a lightweight token and swapped the content queries in the grids to use it instead so the component can be tree-shaken when not in use.
Note, it's not a "by-the-book" lightweight token implementation since the `IgxPaginatorComponent` still extends `DisplayDensityBase` and I didn't want to mix that with the token. Instead, left a comment since the density is deprecated and will be dropped eventually. Also, the `paginator` prop on the grids is still part of the public interface (even as hidden) and I left that with just type imports to avoid propagating the change further and will leave that again for later once the public interface is cleaned up.

Results:
Given the app imports just `IgxGridComponent, IgxColumnComponent` before and after:
image
Small win, but the correct approach for this type of child component relations nonetheless :)

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/bundle-test/src/app/app-routing.module.ts
+++ b/projects/bundle-test/src/app/app-routing.module.ts
@@ -6,7 +6,8 @@ const routes: Routes = [
     { path: '', redirectTo: '/chip', pathMatch: 'full' },
     { path: 'chip', component: ChipComponent },
     // { path: 'form', loadComponent: () => import('./form/form.component').then(m => m.FormComponent) },
-    // { path: 'stepper', loadComponent: () => import('./stepper/stepper.component').then(m => m.StepperComponent) }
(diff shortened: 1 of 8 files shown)


kdinev (approved):

kdinev (changes requested):
Please include an `CHANGELOG.MD` update.

damyanpetev:
> Please include an `CHANGELOG.MD` update.

Will make a 3rd PR for the action strip and will add the changelog there to avoid conflicting with the current 2

kdinev (approved):

PR #13905, opened 2024-02-14
Make HammerJs optional.

MayaKirova:
Closes #  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


kdinev:
@MayaKirova Please resolve the `package.json` conflicts.

projects/igniteui-angular/schematics/ng-add/index.spec.ts
```diff
@@ -90,7 +90,8 @@ describe('ng-add schematics', () => {
     await runner.runSchematic('ng-add', { normalizeCss: false }, tree);
     const pkgJsonData = JSON.parse(tree.readContent('/package.json'));
     expect(pkgJsonData.dependencies['fflate']).toBeTruthy();
-    expect(pkgJsonData.dependencies['hammerjs']).toBeTruthy();
+    // hammer is optional now.
+    //expect(pkgJsonData.dependencies['hammerjs']).toBeTruthy();
```
kdinev, line 94: Can we make this check for optional deps?

MayaKirova: I don't think ng-add adds the optional dependencies by default. I could take a look and see if it can be configured to add them in some way for this test in particular, but out of the box it would not.

kdinev: > I don't think ng-add adds the optional dependencies by default. I could take a look and see if it can be configured to add them in some way for this test in particular, but out of the box it would not.

That is true. Ignore my comment and just fix the conflict :)

projects/igniteui-angular/schematics/ng-add/index.spec.ts
```diff
@@ -105,13 +106,15 @@ describe('ng-add schematics', () => {
     expect(testTs).not.toContain('import \'hammerjs\';');
   });
 
-  it('should add hammer.js in angular.json build options under scripts', async () => {
+  // Hammer is optional now.
+  xit('should add hammer.js in angular.json build options under scripts', async () => {
```
damyanpetev, line 110: Might as well delete those test, doubt we'll ever again make it a requirement so the functionality to add it with ng-add won't be needed again

projects/igniteui-angular/src/lib/core/touch-annotations.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+/**
+* @hidden
+* @internal
+*/
+  export interface HammerInstance {
+    on(eventName: string, callback: Function): void;
+    off(eventName: string, callback: Function): void;
+  }
```
damyanpetev, line 58: These are entirely unused and the rest of the declarations can be reduced quite a bit as well.
These are only so our internal build passes more/less since nothing from this is exposed public.
Well, except `CarouselHammerConfig` for some reason, and even that compiles down to number in the dist declarations:
```ts
export declare class CarouselHammerConfig extends HammerGestureConfig {
    overrides: {
        pan: {
            direction: number;
        };
    };
    static ɵfac: i0.ɵɵFactoryDeclaration<CarouselHammerConfig, never>;
    static ɵprov: i0.ɵɵInjectableDeclaration<CarouselHammerConfig>;
}
```
Though that's not super relevant

projects/igniteui-angular/package.json
```diff
@@ -79,6 +77,10 @@
         "igniteui-theming": "^4.4.0",
         "@igniteui/material-icons-extended": "^3.0.0"
     },
+    "optionalDependencies": {
+        "@types/hammerjs": "^2.0.40",
+        "hammerjs": "^2.0.8"
+    },
```
damyanpetev, line 83: Hm, oddly enough [`optionalDependencies`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies) seems to still be installed by default. Perhaps that's intended or it matters much, but perhaps we could even not list hammer at all in that case? 
//CC @kdinev

package.json
```diff
@@ -138,6 +140,8 @@
     "ts-node": "^10.8.1",
     "typedoc": "^0.25.3",
     "typedoc-plugin-localization": "^3.0.3",
-    "typescript": "5.2.2"
+    "typescript": "5.2.2",
+    "hammerjs": "^2.0.8",
+    "@types/hammerjs": "^2.0.40"
```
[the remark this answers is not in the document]
damyanpetev: And more to the point, don't edit deps manually :X leaves the lock out of sync

projects/igniteui-angular/schematics/ng-add/index.spec.ts
```diff
@@ -105,13 +106,15 @@ describe('ng-add schematics', () => {
...
     await runner.runSchematic('ng-add', { normalizeCss: false }, tree);
     const ngJsonConfigResult = JSON.parse(tree.read('/angular.json').toString());
     expect(ngJsonConfigResult.projects.testProj.architect.build.options.scripts).toContain('./node_modules/hammerjs/hammer.min.js');
   });
 
-  it('should add hammer.js in angular.json test options under scripts', async () => {
+  // Hammer is optional now.
+  xit('should add hammer.js in angular.json test options under scripts', async () => {
```
damyanpetev, line 117: More importantly, we _should_ drop actual the functionality that adds hammer to the angular.json scripts collections. Commenting out the tests doesn't disable the actual logic

damyanpetev, line 117: We _should_ drop actual the functionality that adds hammer to the angular.json scripts collections. Commenting out the tests doesn't disable the actual logic and I see no changes in the ng-add schematic

Otherwise, this test should stay active

damyanpetev (changes requested):

projects/igniteui-angular/package.json
kdinev: OK, so from the documentation, I see that the correct way to specify the hammer dependency is as a `peerDependency` with `peerDependencyMeta` as optional: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta

damyanpetev (approved):

kdinev (approved):

PR #13906, opened 2024-02-14
Make HammerJs optional. 

MayaKirova:
Closes #  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes for each version of this project will be documented in this file.
 
+## 17.1.0
+### General
```
damyanpetev, line 6: Heheh, might've been too diligent transferring changes. Don't think we want the changelog at all since this might not end up in 17.1 directly, plus that's already there. Might also want to drop the schematics stuff, I think we're mostly interested in the commits that modify the actual source only. Will see if that can be done by skipping other commits only

projects/igniteui-angular-elements/src/polyfills.ts
```diff
@@ -51,7 +51,6 @@ import 'zone.js';  // Included with Angular CLI.
 /***************************************************************************************************
  * APPLICATION IMPORTS
  */
-import 'hammerjs';
```
damyanpetev, line 54: Hm, we'll need to likely update docs for WC/Blazor/React due to this. Don't know if it currently mentions Hammer, but now it'd be up to the app to include it.

damyanpetev (approved):
Other than the docs comment, LGTM

projects/igniteui-angular-elements/src/polyfills.ts
damyanpetev: Okay, did a bit of digging and the list of Grid-related functionality that can rely on Hammer:
- The cell doubletap workaround for iOS (that also includes a check to prevent doubletap zoom it seems). This one we'll need to check out how much is still valid
- The list pan gestures, which the grid lists do not use, so we can safely ignore
- Time picker pan to scroll through the list of hours/minutes
- Calendar (Date picker) year list in the same style, which won't matter once the calendar redesign lands

So it's almost entirely restricted to Grid Editing and components used by the feature and 2 out of 3 of those might be resolvable with refactoring/updates. So we could almost wait depending on the release targets or only update the editing topic.

MayaKirova: @dkamburov @damyanpetev 

I don't think we have any mentions in the wc/react/blazor topics around the grid related to touch interactions right now.

Not sure what the time frame for fully phasing out hammerjs is. Maybe @dkamburov will have more information? 
If it's in the near future I'm more in favor of waiting and then just adding an actual Touch support section for wc/react/blazor. Might not be a bad idea to add something like that to Angular as well, since it's a bit difficult to track what interactions we actually support for touch right now in each component.

PR #13907, opened 2024-02-15
feat(h-grid): provide injector to simplify child toolbar template setup

damyanpetev:
This allows the definition of child grid toolbar templates without the explicit need to get the grid ref and assign it to the toolbar.
So from:
```html
<igx-row-island>
  <igx-grid-toolbar [grid]="childGrid" *igxGridToolbar="let childGrid">
```
to
```html
<igx-row-island>
  <igx-grid-toolbar *igxGridToolbar>
```

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


kdinev (dismissed):

MayaKirova:
@damyanpetev LGTM. 
I'm just wondering if maybe we should also deprecate/remove the `grid` property on to the toolbar as a whole. It seems to have been used only for child hierarchical grid, so if we no longer use it might be better to just remove it.
@dkamburov We'll need to update the docs for the next release for the hgrid toolbar for this change: https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/toolbar

damyanpetev:
> @damyanpetev LGTM. I'm just wondering if maybe we should also deprecate/remove the `grid` property on to the toolbar as a whole. It seems to have been used only for child hierarchical grid, so if we no longer use it might be better to just remove it. @dkamburov We'll need to update the docs for the next release for the hgrid toolbar for this change: https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/toolbar

I think the grid reference can still be useful for other use cases (perhaps custom buttons that call some grid API), so it could be left exposed. Wondering if we should expose it the same way for the paginator, but I digress.
And already on the docs change https://github.com/IgniteUI/igniteui-docfx/pull/5713 :)

dkamburov (dismissed):

MayaKirova:
> > @damyanpetev LGTM. I'm just wondering if maybe we should also deprecate/remove the `grid` property on to the toolbar as a whole. It seems to have been used only for child hierarchical grid, so if we no longer use it might be better to just remove it. @dkamburov We'll need to update the docs for the next release for the hgrid toolbar for this change: https://www.infragistics.com/products/ignite-ui-angular/angular/components/hierarchicalgrid/toolbar
> 
> I think the grid reference can still be useful for other use cases (perhaps custom buttons that call some grid API), so it could be left exposed. Wondering if we should expose it the same way for the paginator, but I digress. And already on the docs change [IgniteUI/igniteui-docfx#5713](https://github.com/IgniteUI/igniteui-docfx/pull/5713) :)

The directive that holds the grid context can stay. I'm talking about this `grid` input that is no longer needed: https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar.component.ts#L58

I don't think the toolbar can be used outside of the grid. Since something like this:
```
    <igx-grid-toolbar [grid]="grid1">
        <igx-grid-toolbar-title>{{ toolbarTitle }}</igx-grid-toolbar-title>
    </igx-grid-toolbar>
    <igx-hierarchical-grid #grid1 ...
```
Thrown an injection error.

And judging by the api docs around the prop, it was used for the hierarchical grid children only, so maybe it can be removed now.
It can be logged separately of course, it just bugs me that it's there and seems to be useless now.

dkamburov (approved):

PR #13950, opened 2024-02-23
fix(igxGrid): When autosizing something with template that may need t…

MayaKirova:
…ime post injection in DOM to fully initialize, skip initial renders when the cell is still with no content.

For https://github.com/IgniteUI/igniteui-webcomponents/issues/1046

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
@@ -7214,8 +7214,17 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
                 const cells = this._dataRowList.map(x => x.cells.find(c => c.column === col));
                 cells.forEach((cell) => cellsContentWidths.push(cell?.nativeElement?.offsetWidth || 0));
                 const max = Math.max(...cellsContentWidths);
-                if (max === 0) {
-                    // cells not in DOM yet...
+                // in cases with template contains something, like a webcomponent,
+                // that renders fully only after it is already injected in the DOM,
+                // and initially renders as empty, skip measuring it.
+                let emptyCellWithPaddingOnly = 0;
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -7214,8 +7214,17 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
+                // in cases with template contains something, like a webcomponent,
+                // that renders fully only after it is already injected in the DOM,
+                // and initially renders as empty, skip measuring it.
+                let emptyCellWithPaddingOnly = 0;
+                if (cells.length > 0 && !!col.bodyTemplate) {
+                    const cellStyle = this.document.defaultView.getComputedStyle(cells[0].nativeElement);
+                    emptyCellWithPaddingOnly = parseFloat(cellStyle.paddingLeft) + parseFloat(cellStyle.paddingRight);
+                }
+
+                if (max === 0 || max <= emptyCellWithPaddingOnly) {
```
damyanpetev, line 7226: Um, is it me or does this have the potential for looping hard and also possibly affecting the Angular product as well? Assuming there is a template that literally renders nothing on purpose for whatever reason (for example conditionally shown button), would this not get stuck trying to measure that empty column until something _is_ rendered?

damyanpetev (approved):
Outside of my previous comment about the possibility of the column remaining in `content-fit` mode if the template is intentionally empty which is pretty unlikely, I think we can accept that fix initially.
Still think we can improve, since it's a targeted scenario and we can only re-test the column one more time after the DOM is ready, though that might require some more digging to figure out the exact order of events

PR #14042, opened 2024-04-02
fix(igxHierarchicalGrid): Make sure state component gets projected in…

MayaKirova:
…side the hierarchical/pivot grid's templates as well.

Closes https://github.com/IgniteUI/igniteui-webcomponents/issues/1120  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
+++ b/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
@@ -1,3 +1,4 @@
+<ng-content select="igx-grid-state,igc-grid-state"></ng-content>
 <ng-content select="igx-grid-toolbar,igc-grid-toolbar"></ng-content>
 <ng-container #toolbarOutlet></ng-container>
 
--- a/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
+++ b/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
@@ -1,4 +1,3 @@
-<ng-content select="igx-grid-state,igc-grid-state"></ng-content>
 <ng-content select="igx-grid-toolbar,igc-grid-toolbar"></ng-content>
 <ng-container #toolbarOutlet></ng-container>
 
@@ -226,6 +225,7 @@
 <div class="igx-grid__loading-outlet" #igxLoadingOverlayOutlet igxOverlayOutlet></div>
 <div class="igx-grid__outlet" #igxFilteringOverlayOutlet igxOverlayOutlet></div>
 <div #sink style="display: none;">
+    <ng-content select="igx-grid-state,igc-grid-state"></ng-content>
     <ng-content select="igx-column,igc-column,igx-column-group,igc-column-group,igx-action-strip,igc-action-strip"></ng-content>
     <ng-content select="igx-row-island,igc-row-island"></ng-content>
 </div>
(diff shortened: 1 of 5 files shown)


projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
```diff
@@ -1,3 +1,4 @@
+<ng-content select="igx-grid-state,igc-grid-state"></ng-content>
```
damyanpetev, line 1: There should be a `<div #sink style="display: none;">` at the end of the template that generally collects all components that don't sit in the actual UI layout and I think the state belongs there.
I guess it doesn't mess with anything right now since the state component host elements probably doesn't have meaningful content, but CSS rules of the grid can affect it and it could in theory mess with the layout unexpectedly in the future.

damyanpetev: I see this doesn't include the Grid component. so assuming that one is already there and if it's also not in the sink - more it for the grid too please.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.html
```diff
@@ -1,3 +1,4 @@
+<ng-content select="igx-grid-state,igc-grid-state"></ng-content>
```
damyanpetev, line 1: Don't think the pivot had an actual sink container till now (one of the outlets is marked to keep the query working) since it didn't really have columns or other elements to project. So might be time to create that hidden div here too

damyanpetev (changes requested):

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.html
MayaKirova: Is that really necessary? The state does not render anything. Moving it in another place might break React/Blazor, since it will have to search for the ref deeper in the hierarchy and we've already released it for the base grid as is. This will require a full re-testing for the state in all platforms.

damyanpetev: I think it is, anything that doesn't normally render under Angular runtime should go in the sink for Elements, because like I said this present in the main layout has the possibility to interact with it, depending on styling and that won't be immediately obvious when running Angular alone.

Shouldn't affect Blazor much/at all, since the .NET side of the query is not ran in the DOM and already works for other components. For React, really not sure what client runtime functionality there is, in theory there should be none (I hope).
As for Elements, that query in again in the DOM and shouldn't be affected at all (again, already running for others)

MayaKirova: The element's part should be fine, however the wrappers for React/Blazor do need reference to the original wc element to be able to call methods on it, and since the state functionality is all via methods that could potentially break it, since there is custom handling that searches the dom hierarchy to find the target element. We've had similar bugs before where elements that get moved in the sink containers cannot be found for methods calls in React/Blazor (like columns, column groups etc.) so I'm just saying that it should be thoroughly re-tested so we do not regress it for the grid when we change the structure.

MayaKirova:
@dkamburov Note that this changes the structure in the Grid as well, so it this PR should be tested for WC, React, Blazor for Grid, TreeGrid, HierarchicalGrid and PivotGrid.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid.component.html
```diff
@@ -171,3 +171,7 @@
         </igx-excel-style-filter-operations>
     </igx-grid-excel-style-filtering>
 </div>
+
+<div #sink style="display: none;">
```
damyanpetev, line 175: @MayaKirova Minor leftover from the old comment about the pivot was that "one of the outlets is marked to keep the query working" so there are now two `#sink` items in the template. Should be fine to remove it from the outlet container.

damyanpetev (approved):

dkamburov (approved):

Issue #13688, opened 2023-11-14
IgxGrid: wrong order and duplication of columns after resizing and scrolling the grid


ddaribo:
## Description  
An `IgxGrid`, where the columns widths are defined, such that they take less than the available grid horizontal space (for example column width for all is set to `100px` and the grid's width resolves to > `columns.length * 100`. The width of the columns could also be set to `"auto"`, provided that the first condition is met and there is empty space after the last column.

When resizing the screen horizontally, such that a scrollbar appears in the grid, scrolling it to the end horizontally and then going back to full width of the screen, the columns order is not as defined and there are some duplicated ones.

 * igniteui-angular version: 17.0.x, 16.1.x, 15.1.x and earlier
 * browser: any

## Steps to reproduce  

1. Open [this sample](https://stackblitz.com/edit/igx-grid-cols-order-issue-on-resize-window-and-scroll?file=src%2Fapp%2Fgrid%2Fgrid-moving-sample%2Fgrid-moving-sample.component.html) and click "Open in New Tab" in StackBlitz
2. Open the Dev Tools
3. Resize the Dev Tools panel so that the grid is significantly shrinked, for example only the first three columns remain visible
4. Scroll the grid hroizontally to the end
5. Close the Dev Tools

6. Alternatively, perform the same by minimizing/maximizing the browser window.

## Result  
The columns are not rendered in the correct order and some of them ar duplicated. When clicking on any of the duplicated columns' headers, both instances get highlighted as selected.

## Expected result  
Keep columns as defined, no duplicates.

## Attachments  

![igx-grid-issue](https://github.com/IgniteUI/igniteui-angular/assets/35433937/c3d6762c-139f-44f0-8fac-9e3629ec9045)

PR #14079, opened 2024-04-08
fix(igxGrid): Fix timing issue between forof resetting and the grid h…

MayaKirova:
…iding its scrollbar on resize.

Note: Issue is that the grid's `isHorizontalScrollHidden` getter got resolved first, hence the scrollbar gets hidden before the forof's resize observer is called, so at that point the for-of cannot reset the scrollbar because it is already hidden from the grid. After that once it is resized again to have a scrollbar, the scrollbar resets to its last known scroll position, hence the desync.  

Closes #https://github.com/IgniteUI/igniteui-webcomponents/issues/1116  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
@@ -3605,6 +3605,14 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
             this.cdr.detectChanges();
         });
 
+
+        this.headerContainer.scrollbarVisibilityChanged.pipe(filter(() => !this._init), destructor).subscribe(() => {
+            // the horizontal scrollbar showing/hiding
+            // update scrollbar visibility and recalc heights
+            this.notifyChanges(true);
+            this.cdr.detectChanges();
+        });
+
         this.verticalScrollContainer.contentSizeChange.pipe(filter(() => !this._init), throttleTime(30), destructor).subscribe(() => {
             this.notifyChanges(true);
         });
@@ -6056,7 +6064,7 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
      * @hidden @internal
      */
     public hasHorizontalScroll() {
-        return this.totalWidth - this.unpinnedWidth > 0;
+        return this.totalWidth - this.unpinnedWidth > 0 && this.width !== null;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6670,6 +6676,14 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
             this.calculateGridWidth();
             this.cdr.detectChanges();
         }
+
+        // in case horizontal scrollbar has appeared recalc to size correctly.
+        if (hasHScroll !== this.hasHorizontalScroll()) {
+            this.isHorizontalScrollHidden = !this.hasHorizontalScroll();
+            this.cdr.detectChanges();
+            this.calculateGridHeight();
```
mddragnev, line 6684: Why do we need to calculate the grid height again when the horizontal scroll appears?

MayaKirova: Because the horizontal scrollbar takes up some of the vertical space when it show/hides.

mddragnev: Is this fixing a specific issue because when resizing the resize observer triggers the calculation of the grid height couple of times and it seems ok to me?

MayaKirova: Yes. If you go from a grid that has horizontal scrollbar to a grid that should have no horizontal scrollbar (for example going from a minimized window and them maximizing it so that all columns fit) then the horizontal scrollbar container will not hide without the additional handling. There were some tests that were failing due to this.

Wrong behavior, can be observed for example here:
horizontalNoHide

Notice that the horizontal scrollbar is no longer needed when maximized, but still shows.

Correct behavior would be for it to hide fully:
 
horizontalHide

MayaKirova:
> There is an issue which is somehow related to this PR but also with #13688 . If you open column actions dev sample and then set the columns collection to be
> 
> ```
> this.columns = [
>             { field: 'ID', width: '100px', hidden: false },
>             { field: 'CompanyName', width: '100px' },
>             { field: 'ContactName', width: '100px', pinned: false },
>             { field: 'ContactTitle', width: '100px', pinned: false },
>             { field: 'Address', width: '100px' },
>             { field: 'City', width: '100px' },
>         ];
> ```
> 
> then set couple of more empty columns to the template
> 
> ```
>         <igx-column field="test" width="100px"></igx-column>
>         <igx-column field="test" width="100px"></igx-column>
>         <igx-column field="test" width="100px"></igx-column>
>         <igx-column field="test" width="100px"></igx-column>
>         <igx-column field="test" width="100px"></igx-column>
> ```
> 
> and set grid width to 100% Then if you resize the window to be smaller, scroll right on the columns then resize the window to take the full screen => some columns are reordered and some columns are not visible in the grid
> 
>  test420p.mov

Issue should be fixed now.

mddragnev (approved):

PR #14141, opened 2024-04-22
fix(igxGrid): Enable check for whether container size changes when gr…

### LINKED ISSUE #14138: When grid is inside a container with `display: contents` it does not span the full 100% of the height.
## Description  
When grid is inside a container with `display: contents` it does not span the full 100% of the height.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open sample: https://stackblitz.com/edit/tchxxa

## Result  
There is white space below the grid.

## Expected result  
It should expand to 100% height.

MayaKirova:
…id is rendered. If not do not autosize the grid.

Closes #14138  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
```diff
@@ -1617,6 +1617,11 @@ export class IgxGridForOfDirective<T, U extends T[] = T[]> extends IgxForOfDirec
         const containerSize = 'igxForContainerSize';
         if (containerSize in changes && !changes[containerSize].firstChange && this.igxForOf) {
             this._recalcOnContainerChange();
+            if (changes[containerSize].previousValue && this.igxForScrollOrientation === "horizontal") {
+                const scrollAmount = this.scrollComponent.nativeElement["scrollLeft"];
+                this.scrollComponent.scrollAmount = scrollAmount;
+                this._updateScrollOffset();
```
skrustev, line 1623: I think I understand correctly, but scrollPosition shouldn't in general shouldn't change on window resize unless at the end of the scroll but it appears to be doing just that? Because this triggers change chain reaction for the grid scrollbar that is common to all rows, when the sizeChaches change due to window resize..
There is a slight change in behavior because of that and the use of `_updateScrollOffset` when there are columns with width in percentages. The scrollbar will reset to the start now when window is resized. Here's a gif of that:

grid_resize_scroll

PS: I use break points to make it more clear, without them it also acts the same but faster.

MayaKirova: That one was for a specific issue reported in one of the other PRs: https://github.com/IgniteUI/igniteui-angular/pull/14224#issuecomment-2111758118

MayaKirova: @skrustev Both should be fixed now.

skrustev (approved):

projects/igniteui-angular/src/lib/directives/for-of/for_of.directive.ts
skrustev: Hmm ok, maybe previously was better because now when resizing the window larger, if the horizontal scroll is at the end, the rows do not update.
grid_resize_scroll2

skrustev: I've logged it separately #14278

damyanpetev: @MayaKirova @skrustev This branch is quite behind and doesn't include the density remove that also shifts resize observers around IIRC that should in theory be updating the virtualization. Might be worth to re-check behaviors after it's synced.

PR #13684, opened 2023-11-14
fix(grid): rem throtttime lead of event

tishko0:
Closes #13000

### Additional information (check all that apply):
 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [ ] All relevant tags have been applied to this PR
 - [ ] This PR includes unit tests covering all the new code ([test guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Test-implementation-guidelines-for-Ignite-UI-for-Angular))
 - [ ] This PR includes API docs for newly added methods/properties ([api docs guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines))
 - [ ] This PR includes `feature/README.MD` updates for the feature docs
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [ ] This PR includes `CHANGELOG.MD` updates for newly added functionality
 - [ ] This PR contains breaking changes
 - [ ] This PR includes `ng update` migrations for the breaking changes ([migrations guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Update-Migrations))
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

Issue #13000, opened 2023-05-11
Error thrown by ResizeObserver


hanastasov:
* igniteui-angular version: master
 *** browser:**

  -- Chrome 113.0.5672.92 - reproduce
  -- Edge 113.0.5672.92 - reproduce
  -- Firefox 110.0.1 - **does not reproduce**

## Steps to reproduce  

1. Run dev demos, go to grid sample http://localhost:4200/grid

## Result  
![image](https://github.com/IgniteUI/igniteui-angular/assets/14248932/1c60b2a8-35d5-4a57-a9a2-0a4f43c63123)

hanastasov:
Was reported in TaskPlanner app also, after upgrade to version 16, see https://github.com/IgniteUI/TaskPlanner/pull/65#issuecomment-1540227450

MayaKirova:
@tishko0 An easier way to reproduce in master is to attach to the window's error event, since the error no longer pops in a overlay. For example, for the dev demos:
1. Open the grid-auto-size.sample.ts sample and attach the window error event:

```
    public ngOnInit(): void {
        window.addEventListener("error", (e) => {
            console.log(e)
        });
...
}
```

2. Run the samples and open the related dev sample at: http://localhost:4200/gridAutoSize
3. Click the "Toggle Container Visibility" button twice to hide/show the container.
4. Open the console and observe the error and details:
![image](https://github.com/IgniteUI/igniteui-angular/assets/10397980/7e0e5e6b-c950-40d6-8daa-62c344497a46)

PR #14163, opened 2024-04-24
refactor(grid): Remove deprecated usage of display density

### LINKED ISSUE #14015: Grid: Remove deprecated usage of `displayDensity`
The grid component relies on the value of the deprecated `displayDensity` token when calculating the heights of various elements in the grid. These values should rather come from the styles and be read by the grid's via computedStyles if they need to be factored in the sizing of the grid. The reason being is that the grid doesn't fully work with the `--ig-size` property right now (runtime) because it heavily depends on hardcoded values in the business logic.

mddragnev:
Closes #14015   
Updated the update guide in the docfx: https://github.com/IgniteUI/igniteui-docfx/pull/5807

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3539,17 +3540,29 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
 
         this.resizeNotify.pipe(
             filter(() => !this._init),
-            throttleTime(40, animationFrameScheduler, { leading: false, trailing: true }),
+            throttleTime(40, animationFrameScheduler, { leading: true, trailing: true }),
```
MayaKirova, line 3543: Is the `leading:true` necessary?  I think we removed it intentionally as part of a fix for the resize observer issue: https://github.com/IgniteUI/igniteui-angular/issues/13000
In PR: https://github.com/IgniteUI/igniteui-angular/pull/13684

projects/igniteui-angular/src/lib/grids/grid/grid-row.component.html
```diff
@@ -107,7 +107,6 @@
         [intRow]="this"
         [lastPinned]="col.columnLayoutChild ? null : col.isLastPinned"
         [firstPinned]="col.columnLayoutChild ? null : col.isFirstPinned"
-        [style.min-height.px]="cellHeight"
```
MayaKirova, line 110: Is min-height no longer needed?

mddragnev: Yes, I changed the way the row gets its min-height. I have included it in the theme and it works with css now: https://github.com/IgniteUI/igniteui-angular/pull/14163/commits/8be56af333b6fb0a731190b13422f069cf857b56#diff-29930daadb3a8deb436ab411208d6c7f5008e2239dd5ecb3e6f3669b36d02e22R1434 . I have also removed the defaultRowHeight from the grid.

MayaKirova (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes for each version of this project will be documented in this file.
 
+## 18.0.0
+### General
+- `IgxGrid`, `IgxTreeGrid`, `IgxHierarchicalGrid`, `IgxRowIsland`
+  - Removed deprecated property `displayDensity`.
```
damyanpetev, line 8: This will likely become a general bullet rather than just the grid, once the other components follow, but still:
```suggestion
  - Removed deprecated property `displayDensity`. Size is now controlled only though the custom CSS property `--ig-size`. Refer to the [Update Guide](https://www.infragistics.com/products/ignite-ui-angular/angular/components/general/update-guide) and components documentation for usage details.
```

projects/igniteui-angular/src/lib/core/styles/components/grid/_excel-filtering-component.scss
```diff
@@ -10,6 +10,10 @@
             @extend %igx-excel-filter__loading !optional;
         }
 
+        @include e(sizing) {
+            @extend %igx-excel-filter__sizing !optional;
+        }
```
simeonoff, line 15: Why do we need a new class selector? Can't we simply use the existing %grid-excel-filter?

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-filtering.component.ts
```diff
@@ -73,20 +72,33 @@ export class IgxExcelStyleFilterOperationsTemplateDirective { }
...
      */
     @HostBinding('class.igx-excel-filter--inline')
     public inline = true;
 
+    /**
+     * @hidden @internal
+     */
+    @HostBinding('style.--component-size')
+    protected get componentSize(): string {
+        return getComponentCssSizeVar(this.size);
```
damyanpetev, line 99: This isn't conditional, so it may very well be directly in CSS, no?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3539,17 +3540,29 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
...
         )
-            .subscribe(() => {
-                this.zone.run(() => {
-                    // do not trigger reflow if element is detached.
-                    if (this.document.contains(this.nativeElement)) {
-                        this.notifyChanges(true);
+        .subscribe(() => {
+            this.zone.run(() => {
+                // do not trigger reflow if element is detached.
+                if (this.document.contains(this.nativeElement)) {
```
damyanpetev, line 3549: Technically, `this.nativeElement.isConnected` should give you the same thing, right?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -5237,6 +5238,13 @@ export abstract class IgxGridBaseDirective extends DisplayDensityBase implements
         return this.hasVerticalScroll() ? this.calcWidth + this.scrollSize : this.calcWidth;
     }
 
+    /**
+     * Gets the size of the grid
+     */
+    public get gridSize(): Size {
```
damyanpetev, line 5244: `@hidden @internal`

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -622,3 +622,26 @@ export function* intoChunks<T>(arr: T[], size: number) {
...
+        case "3":
+        default:
+            return 'var(--ig-size, var(--ig-size-large))';
+    }
+}
+
+/**
+ * Sets element size as a inline style
+ */
+export function setElementSize(element: HTMLElement, size: string) {
```
damyanpetev, line 645: If that's used by tests only, it could and should live in the test utils? See projects/igniteui-angular/src/lib/test-utils/

src/app/grid-auto-size/grid-auto-size.sample.ts
```diff
@@ -1,8 +1,8 @@
-import { Component, ViewChild, OnInit } from '@angular/core';
+import { Component, ViewChild, OnInit, HostBinding } from '@angular/core';
 import { NgFor } from '@angular/common';
 
 import { SAMPLE_DATA } from '../shared/sample-data';
-import { DisplayDensity, GridSelectionMode, IgxGridComponent, IGX_BUTTON_GROUP_DIRECTIVES, IGX_GRID_DIRECTIVES } from 'igniteui-angular';
+import { GridSelectionMode, IgxGridComponent, IGX_BUTTON_GROUP_DIRECTIVES, IGX_GRID_DIRECTIVES, Size } from 'igniteui-angular';
```
damyanpetev, line 5: Oh, do we really intend to still keep exposing an enum for this? I sort of expected that to be internal, so it would have no place in any sample. Below sizes can be typed with an union like `'small' | 'medium' | 'large'` just fine

projects/igniteui-angular/src/lib/core/styles/components/grid/_excel-filtering-component.scss
mddragnev: If you take a look at the usage of the newly introduced class:
```
    @HostBinding('class.igx-excel-filter__sizing')
    protected get shouldApplySizes(): boolean {
        return !(this._minHeight || this._maxHeight);
    }
```
we set this only if we do not have neither minHeight nor maxHeight set from the inputs of the ESF. If we use the exising main css class %grid-excel-filter we cannot do this unless we set min/maxHeight as inline styles that override the default behaviour of the sizing mixin which I don't really like

src/app/grid-column-resizing/grid-column-resizing.sample.html
```diff
@@ -1,5 +1,5 @@
 <div class="wrapper">
-    <igx-grid #grid1 [data]="data" [width]="'100%'" [height]="'800px'" displayDensity="compact"  [rowSelection]="'multiple'" >
+    <igx-grid #grid1 [data]="data" [width]="'100%'" [height]="'800px'" [style.--ig-size]="1" [rowSelection]="'multiple'" >
```
damyanpetev, line 2: PS: Those static values can just be moved to styles instead.

damyanpetev (approved):
Not sure if the styling still needs more discussing and the dev demos can be cleaned up some more even separately, the rest LGTM.

PR #14168, opened 2024-04-25
feat(combo, simple-combo): hide clear icon when custom template is empty

### LINKED ISSUE #14104: Allow removing of combo's clear icon
I have a combo where I can select items. In most of the cases only single item will be selected. However, there are cases where multiple items need to be selected. All this is achievable now. However, I do not need the clear icon in the combo.

Allow hiding of the clear icon. I tried with empty template like this:
```html
<ng-template igxComboClearIcon></ng-template>
```
but as there is a padding in the suffix the clear icon is still clickable
image

If clicking on the clear icon is firing some event this will also help me solve my issue.

Either expose a property `showClearIcon` or event `onClear`.

georgianastasov:
Closes #14104   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ All notable changes for each version of this project will be documented in this 
 - `IgxGrid`, `IgxTreeGrid`, `IgxHierarchicalGrid`
     - Enhanced the advanced filtering to emit the `filtering` event when filters are applied.
 - `IgxCombo`, `IgxSimpleCombo`:
-    - Introduced support for hiding the clear icon button when the custom clear icon template is empty.
+    - Introduced abillity for hiding the clear icon button when the custom clear icon template is empty.
 
 ## 17.1.0
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ All notable changes for each version of this project will be documented in this 
     - Removed `leftButtonColor`, `leftButtonBackgroundColor` `rightButtonColor`, and `rightButtonBackgroundColor`  properties.
 - `IgxGrid`, `IgxTreeGrid`, `IgxHierarchicalGrid`
     - Enhanced the advanced filtering to emit the `filtering` event when filters are applied.
+- `IgxCombo`, `IgxSimpleCombo`:
+    - Introduced support for hiding the clear icon button when the custom clear icon template is empty.
 
 ## 17.1.0
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes for each version of this project will be documented in this file.
 
+## 18.0.0
+### New Features
+- `IgxCombo`, `IgxSimpleCombo`:
+    - Introduced abillity for hiding the clear icon button when the custom clear icon template is empty.
+
 ## 17.2.0
 ### New Features
 - `IgxAvatar`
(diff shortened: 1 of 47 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ All notable changes for each version of this project will be documented in this
     - Removed `leftButtonColor`, `leftButtonBackgroundColor` `rightButtonColor`, and `rightButtonBackgroundColor`  properties.
 - `IgxGrid`, `IgxTreeGrid`, `IgxHierarchicalGrid`
     - Enhanced the advanced filtering to emit the `filtering` event when filters are applied.
+- `IgxCombo`, `IgxSimpleCombo`:
+    - Introduced support for hiding the clear icon button when the custom clear icon template is empty.
```
damyanpetev, line 28: That's an actual fix though :) Not sure it needs a changelog entry

damyanpetev (commented):
Alternative fix suggestion - use the `:empty` state of the suffix:
```css
igx-suffix.igx-combo__clear-button:empty {
    padding: 0;
}
```
That should drop the paddings and let it collapse correctly. Try it

CHANGELOG.md
georgianastasov: Could you confirm whether this adjustment should be considered a feature or a fix for versions 16.1, 17.1, and master?

Regarding the approach, I'd like to understand your preference: should I prioritize this solution proposed in the PR, or do you recommend implementing the suggested alternative?

damyanpetev: Hm, might've confused that with the somewhat similar Data Picker toggle/clear icons, because I remember discussing with the team that the default could be removed by providing an empty template for those for sure. Sort of assumed it's the same for the Combo, but looks to be a bit older and doesn't seem to be documented either, so it might as well be new feature :)

As for the implementation: If the main issue is the remaining padding that needs to be removed, I tested this just briefly my suggestion in dev tools on a live sample but it should do the trick. The current version that creates an embedded view on the fly just to check its contents and discards it, while fun, could be a non-trivial amount of work for such an issue depending on the template. Also consider that being a getter will run on every change detection cycle, so would realistically leave that a last resort if nothing else works. I think we even have precedent with taking a ref and checking childNodes in the toolbar that can do the same check, but if a simple style would do too - I think that's our best option.

damyanpetev:
LGTM, though this missed the window for 17.2 and will likely land in 18.0 so the changelog will need an update. No rush though.
Also, @simeonoff If you can quickly take a quick look to make sure we're not doing something silly, wouldn't mind :)

wnvko:
Having combo set like this:
```html
<igx-combo [data]="uniqueFalsyData" [displayKey]="'field'" [valueKey]="'value'">
    <ng-template igxComboClearIcon>
        <igx-icon>home</igx-icon>
    </ng-template>
</igx-combo>

<igx-combo [data]="uniqueFalsyData" [displayKey]="'field'" [valueKey]="'value'">
    <ng-template igxComboClearIcon>
    </ng-template>
</igx-combo>
```
I am ending up with clear icon like this:
image

I am not even able now to change the clear icon!

wnvko (changes requested):
For some reason clear icon template do not work at all now.

georgianastasov:
@wnvko I've investigated further, and it seems like the issue persists even when I comment out my changes and also when i test it on the master branch. I think that the problem might not be specific to this style changes but could be a separate issue.

georgianastasov:
@damyanpetev Yes, indeed, the IgxComboClearIconDirective is not imported by default in the combo.sample.ts file and also we missed importing the IGX_COMBO_DIRECTIVES, which caused the described behavior.

wnvko (approved):

PR #14208, opened 2024-05-10
fix(simple-combo): filter on typing when combo is focused but not opened - master

### LINKED ISSUE #14187: Simple ComboBox: Filter is sometimes not applied
## Description  
If the dropdown of the Simple ComboBox component has not been opened before, the filter is not applied when typing the first character.
When entering the first character, the dropdown opens and after typing the second character, the filter works as expected again.

 * igniteui-angular version: tested with 16.0.13 and latest version of demo page
 * browser: any

## Steps to reproduce  

1. Go to the demo page of the Simple ComboBox component: https://www.infragistics.com/products/ignite-ui-angular/angular/components/simple-combo
2. Clear the current pre-selected value "CubicFit" (x-icon)
3. Filter by typing any character, e.g. a "c"

## Result  
The dropdown opens but the filter is not applied. All values are shown.

## Expected result  
The dropdown opens but the filter with the value "c" should be applied so that in this example only 5 values containg a "c" are selectable.

## Attachments  
Wrong state:
image

georgianastasov:
Closes #14187   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts
```diff
@@ -327,15 +327,23 @@ export class IgxSimpleComboComponent extends IgxComboBaseDirective implements Co
...
             || event.key === this.platformUtil.KEYMAP.DELETE) {
             this._updateInput = false;
             this.clearSelection(true);
-        }
-        if (!this.collapsed && event.key === this.platformUtil.KEYMAP.TAB) {
+        } else if (event.key === this.platformUtil.KEYMAP.TAB) {
             this.clearOnBlur();
             this.close();
+        } else {
+            if (event.key !== this.platformUtil.KEYMAP.ARROW_DOWN && event.key !== this.platformUtil.KEYMAP.SPACE) {
```
wnvko, line 338: should not filter out `SPACE` here. I can filter values like `New Mexico`, `New York`, `New Jersey` and so on by typing `SPACE` + `M` and combo should show first all the items containing space, then items containing space and m charecter.

projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts
```diff
@@ -327,15 +327,23 @@ export class IgxSimpleComboComponent extends IgxComboBaseDirective implements Co
...
+                const newFilterValue = this.filterValue + event.key.toLowerCase();
+                this.filterValue = newFilterValue;
```
wnvko, line 340: clicking on `control`, `alt` or any other special key and you will end up with `filtervalue` equal to `arrowupshiftaltcontrol` - this is not what we need here. Also, not sure how this will work with Japanese input !

wnvko (changes requested):

wnvko:
@georgianastasov 
The issue comes from the wrong filtering. The filtering happens in the `IgxComboFilteringPipe` depending on the `searchValue`. What happens here is when you type the first character data pass trough the filtering pipe and is filtered correctly. Then combo's drop-down start to open and in `opening` event we are clearing the `searchValue` - filtering pipe is triggered again and no filtering is applied. Please check why we are clearing the `searchValue` and try to fix this.

projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts
```diff
@@ -240,7 +240,9 @@ export class IgxSimpleComboComponent extends IgxComboBaseDirective implements Co
                 this.filterValue = this.searchValue = this.comboInput.value;
                 return;
             }
-            this.filterValue = this.searchValue = '';
+            if (!this.comboInput.value) {
```
wnvko, line 243: now if `this.comboInput.Value` is set, e.g. combo has initial value, same issue is reproducible.

wnvko (changes requested):

projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts
```diff
@@ -285,14 +287,15 @@ export class IgxSimpleComboComponent extends IgxComboBaseDirective implements Co
         if (this.collapsed && this.comboInput.focused) {
             this.open();
         }
-        if (!this.comboInput.value.trim() && super.selection.length) {
+        if (!this.comboInput.value.trim() && super.selection.length && super.selection[0][this.valueKey] !== "") {
```
wnvko, line 290: This
```ts
super.selection[0][this.valueKey] !== ""
```
will return always `true` if there is no `valueKey`. Also this will return `true` is the selected item value is not empty string. This will return `false` only if there is `valueKey` and the value of the selected item is empty string. Do we need this change?

wnvko: one more thing. This entire `if` statement is trying to check if user initially pressed space key. If so, it clears the and actually suppress the filtering. If user needs to filter all the items containing empty space, he should press the space bar twice - variation of the issue related to this PR.

wnvko: @jackofdiamond5 
Please take a look at this PR.

jackofdiamond5: The way I see it, we don't really need this line: https://github.com/IgniteUI/igniteui-angular/blob/668f38b2345e978e709d36bd1099d0d43c6f2e21/projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts#L243

Since on opening we need to know what is being filtered and what the current search term is in order to show the respective elements in the dropdown. So imo, just get rid of that line altogether and this should fix the issue. The reason why it was present to begin with was because it was a partial fix of an issue related to entering partial matches and making sure that the combo clears on blur if no match is found. With the current implementation, the combo does not need that assignment anymore.

On a side note, I found another thing which is kind of related to filtering - https://github.com/IgniteUI/igniteui-angular/issues/14352 which can be fixed relatively easy, if we assume that we don't want to apply selection and we just want to close the combo on  `ArrowUp`. But this can be considered a UX question, so I've logged a separate issue just in case. P.S. I've tested with the WC combo and it immediately closes on `ArrowUp` sooo yeah.

projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts
```diff
@@ -293,6 +292,7 @@ export class IgxSimpleComboComponent extends IgxComboBaseDirective implements Co
         }
         if (super.selection.length) {
             this.selectionService.clear(this.id);
+            this.filterValue = this.searchValue = typeof event === 'string' ? event : event.target.value;
```
georgianastasov, line 295: I left this line so that as there is a set value in the combo and we try to filter the combo by writing in the input, it should be filtered.

jackofdiamond5 (approved):

PR #14328, opened 2024-06-04
ESF on igx-grid does not reflect current filters in esf dialog

### LINKED ISSUE #14188: Data grid filtering:Programatic filtering on data-grid does not reflect selection in filter dialog
## Description 
When data filtered programatically with [Data-grid-excel-style-filtering](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/excel-style-filtering) although the actual filtering operation performs as expected, however the filter representation/selection in filter dialog is absent 

 * igniteui-angular version: "^17.0.11"
 * browser: Edge

## Steps to reproduce  
Here is stackblitz where the issue can be reproducible: [stackblitz](https://stackblitz.com/edit/zkcdsk?file=src%2Findex.html)

1. Step 1: Click any link displayed in first column of the grid => '**ProductName**'. This should filter the grid by whichever value the user clicked on.
2. Step 2: The filtering indicator(dots) for the productName column will be highlighted to indicate that this column has same filter present.
3. Step 3: Click on the indicated filter to check if the dialog reflects the applied filtered condition.

## Result  
Once the filter dialog is opened although the actual filtering operation performs as expected however there is no indication/selection in the filter dialog to indicate that a programatic filter was applied to the grid

## Expected result  
There should be some indication/selection in the filter dialog that displays that a filter was applied to this grid programatically.

## Attachments  
Expected"
image

Result:
image

ddincheva:
Closes #14188  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


deyvidnenchev:
https://github.com/IgniteUI/igniteui-angular/assets/60451908/b64961b9-4362-48cc-af6b-6576b159f184

deyvidnenchev:
According to [this](https://github.com/IgniteUI/igniteui-angular/issues/14188#issuecomment-2149881375) comment we must show the number of column filtered not the number of filtering operands (in advanced filtering we can apply multiple operands to one column)
ec1

deyvidnenchev (approved):

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -150,11 +150,10 @@ export class FilteringExpressionsTree implements IFilteringExpressionsTree {
         for (const expr of expressionsTree.filteringOperands) {
             if ((expr instanceof FilteringExpressionsTree)) {
                 return this.isFilteringExpressionsTreeForColumn(expr, fieldName);
-            } else {
-                return (expr as IFilteringExpression).fieldName === fieldName;
+            } else if ((expr as IFilteringExpression).fieldName === fieldName) {
```
hanastasov, line 153: Do we need to introduce this change and make the code more verbose

projects/igniteui-angular/src/lib/grids/headers/grid-header.component.ts
```diff
@@ -276,6 +276,13 @@ export class IgxGridHeaderComponent implements DoCheck, OnDestroy {
         this.sortDirection = expr ? expr.dir : SortingDirection.None;
     }
 
+    protected isAdvancedFilterApplied() {
+        if(!this.grid.advancedFilteringExpressionsTree) {
+            return false;
+        }
+        return !!this.grid.advancedFilteringExpressionsTree.find(this.column.field);
```
hanastasov, line 283: `some` can be used instead of `find` and this way the value would not need to be converted to a boolean

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -547,6 +547,38 @@ export class IgxDropDownComponent extends IgxDropDownBaseDirective implements ID
         }
     }
 
+     /**
+     *
+     * @hidden
+     * @internal
+     * @param newSelection
+     */
+     public selectItemWithoutEvent(item: IgxDropDownItemBaseDirective) {
```
damyanpetev, line 556: This seems to be a complete repeat of the `selectItem` method just without the emit, which seems redundant to me. If `selectItem` is already internal, how about we simply add an optional `emit = true` param to it and call it with false for the EFS case?
Also kind of curios what trouble the event causes, though I can sort of imagine :)

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-conditional-filter.component.ts
```diff
@@ -40,6 +40,10 @@ export class IgxExcelStyleConditionalFilterComponent implements OnDestroy {
     @ViewChild('subMenu', { read: IgxDropDownComponent })
     public subMenu: IgxDropDownComponent;
 
+    public get filterNumber() {
```
damyanpetev, line 43: For some inexplicable reason `IgxExcelStyleConditionalFilterComponent` is exported in the public API https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/IgxExcelStyleConditionalFilterComponent.html
So for such component, the general rule - either document API meant to be public _or_ in this case, unless I'm missing a need for this to be used externally, just mark it `protected` which has been usable in templates for a while now.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-conditional-filter.component.ts
```diff
@@ -117,6 +121,17 @@ export class IgxExcelStyleConditionalFilterComponent implements OnDestroy {
         return this.esf.column.filters.condition(value);
     }
 
+    /**
+     * @hidden @internal
+     */
+    public getSelectedCondition(condition: string): boolean {
```
damyanpetev, line 127: same reasoning here - can just be `protected` if not needed elsewhere

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
```diff
@@ -26,8 +28,11 @@ import { IgxButtonDirective } from '../../directives/button/button.directive';
     standalone: true,
     imports: [IgxButtonDirective, IgxRippleDirective, NgClass, IgxIconComponent, NgIf]
 })
-export class IgxGridToolbarAdvancedFilteringComponent {
-
+export class IgxGridToolbarAdvancedFilteringComponent implements AfterViewInit {
+    /**
+     * @hidden @internal
+     */
+    public numberOfColumns: number;
```
damyanpetev, line 35: And same here :)

damyanpetev (changes requested):

hanastasov:
I opened gridFiltering sample and applied two filter conditions for the same column - CompanyName. Still, the button shows that 2 columns have applied filters:

image

projects/igniteui-angular/src/lib/grids/headers/grid-header.component.ts
ddincheva: this is not an array, advancedFilteringExpressionsTree is of type IFilteringExpressionsTree so it exposes only `find` and `findIndex` methods

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -505,9 +505,10 @@ export class IgxDropDownComponent extends IgxDropDownBaseDirective implements ID
      * @hidden
      * @internal
      * @param newSelection
+     * @param emit
      * @param event
      */
-    public override selectItem(newSelection?: IgxDropDownItemBaseDirective, event?: Event) {
+    public override selectItem(newSelection?: IgxDropDownItemBaseDirective, emit = true, event?: Event) {
```
damyanpetev, line 511: `emit` could also be last if you don't want to alter any of the existing dropdown code and tests :)
Also there's now an override in the Select for example that's doesn't match the base signature and can in theory be called with the wrong order of params

projects/igniteui-angular/src/lib/drop-down/drop-down.component.ts
```diff
@@ -524,18 +525,22 @@ export class IgxDropDownComponent extends IgxDropDownBaseDirective implements ID
                 index: newSelection.index
             } as IgxDropDownItemBaseDirective;
         }
-        const args: ISelectionEventArgs = { oldSelection, newSelection, cancel: false, owner: this };
-        this.selectionChanging.emit(args);
+        let args: ISelectionEventArgs;
+        if (emit) {
+            args = { oldSelection, newSelection, cancel: false, owner: this };
+            this.selectionChanging.emit(args);
+        }
```
damyanpetev, line 532: Honestly, not much to be gained in perf by skipping the args creation IMO, the code was fine as it was, just a single if around the actual emit should do:
```suggestion
        const args: ISelectionEventArgs = { oldSelection, newSelection, cancel: false, owner: this };

        if (emit) {
            this.selectionChanging.emit(args);
        }
```
and then `if (!args.cancel)` and the rest of the logic can flow as it was :)

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
```diff
@@ -26,8 +28,11 @@ import { IgxButtonDirective } from '../../directives/button/button.directive';
     standalone: true,
     imports: [IgxButtonDirective, IgxRippleDirective, NgClass, IgxIconComponent, NgIf]
 })
-export class IgxGridToolbarAdvancedFilteringComponent {
-
+export class IgxGridToolbarAdvancedFilteringComponent implements AfterViewInit {
+    /**
+     * @hidden @internal
+     */
```
damyanpetev, line 34:
```suggestion
```
same here

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
```diff
@@ -38,5 +40,26 @@ export class IgxGridToolbarAdvancedFilteringComponent {
...
+        const columnNames = [];
+        if (!filteringTree) return columnNames;
+        filteringTree.filteringOperands.forEach((expr) => {
+            if (expr instanceof FilteringExpressionsTree) {
+                columnNames.push(...this.extractUniqueFieldNamesFromFilterTree(expr));
+            } else {
+                columnNames.push((expr as IFilteringExpression).fieldName);
+            }
+        });
+        return columnNames.filter((value, index, self) => self.indexOf(value) === index);
```
damyanpetev, line 63: PS: Though I doubt those arrays will be big enough to make a difference but `[...new Set(columnNames)]` is shorter and should be faster :)

damyanpetev (commented):
Is it me or this fix actually adds bits of functionality is a few places, but I only see a test for the unique filtered fields method, but nothing tests that's reflected in the UI or the initial fix that shows the count of active filters in the ESF and the active state of those in the dropdown

damyanpetev (dismissed):

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-custom-dialog.component.ts
```diff
@@ -245,6 +245,23 @@ export class IgxExcelStyleCustomDialogComponent implements AfterViewInit {
     }
 
     private createInitialExpressionUIElement() {
+        if (this.expressionsList.length == 1) {
+            const currentExprUI = this.expressionsList.pop();
+            if (currentExprUI.expression.condition.name === this.selectedOperator) {
+                currentExprUI.beforeOperator = FilteringLogic.And;
```
hanastasov, line 251: I think that setting the beforeOperator for the first condition is not needed and this line can be safely removed.

Then, the following code is repeated and can be extracted in a method, I mean the lines that create the secondExpressionUI.

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-custom-dialog.component.ts
```diff
@@ -245,6 +245,23 @@ export class IgxExcelStyleCustomDialogComponent implements AfterViewInit {
...
+                const secondExprUI = new ExpressionUI();
```
hanastasov, line 252: Then, this code is repeated and can be extracted in a method, I mean the lines that create the secondExpressionUI.

hanastasov (approved):

AnjiManova:
[Figma Design File and references related to the issue](https://www.figma.com/design/mbhx5nhl5LItmiiRcnbyQz/14188-Issue-%7C-Grids%3A-Advanced-Filtering?m=auto&node-id=0-1&t=UUKt6urjGbV47Tfy-1)

We suggest removing the custom button border style after adding a span with a number of the filtered columns. There is no need to add additional visual indication and customize the buttons this way.

image

image

AnjiManova:
### **Create a new issue from this comment because it is not related to this PR**

I noticed something else, and we talked with Desi: This is an additional issue that should be logged, but we can create such from this comment here. 

When I select one list item, the Text filter is changed to Text filter (1), and "Equal" is selected from the filters.
When I select two list items, the Text filter is changed to Text filter (2), and the "Custom filter" is selected from the filters. However, when I select three items, the Text filter is changed to Text filter (1), and there is no selection in the Text filters. Shouldn't it be showing the "Custom filter" again? This is replicated from the Master, and it is an issue there as well.

https://github.com/IgniteUI/igniteui-angular/assets/127843837/eb651517-be31-468c-81a7-98dc938d5417



I believe the logic should be: when two or more items are selected to apply the Custom filter, the filters should be indicated with (n) next to the Text filter.

Something interesting here is that everything looks fine if I select two items and then add more from the Custom filter.

https://github.com/IgniteUI/igniteui-angular/assets/127843837/73b7640c-7f0d-4ea9-b953-78786037dfbb

hanastasov:
Please add one more test, similar to `Should show the previously entered filter value when reopen esf dialog`, but one that actually selects a different condition from the dropdown, thus should reopen the dialog with empty conditions. This is one of my the main scenarios that we determined as being buggy from very beginning, so it should have a test

projects/igniteui-angular/src/lib/grids/grid/grid-filtering-ui.spec.ts
```diff
@@ -3322,6 +3322,37 @@ describe('IgxGrid - Filtering actions - Excel style filtering #grid', () => {
...
+
+            const excelMenu = GridFunctions.getExcelStyleFilteringComponent(fix);
+            const checkboxes: any[] = Array.from(GridFunctions.getExcelStyleFilteringCheckboxes(fix, excelMenu));
+            checkboxes.forEach(c => expect(c.checked).toBeFalsy());
+
+            GridFunctions.clickExcelFilterCascadeButton(fix);
+            tick(30);
+            fix.detectChanges();
+
+            expect(GridFunctions.getExcelStyleFilteringComponent(fix).querySelector('.igx-drop-down__item--selected')).toBeDefined();
```
hanastasov, line 3348: This does not ensure that the corresponding item in the dropdown is selected. Enhace it to actually determines if the expected item from the list is selected.

hanastasov (changes requested):

projects/igniteui-angular/src/lib/grids/grid/grid-filtering-ui.spec.ts
```diff
@@ -3322,6 +3322,37 @@ describe('IgxGrid - Filtering actions - Excel style filtering #grid', () => {
             checkboxes.forEach(c => expect(c.checked).toBeFalsy());
         }));
 
+        it('Should show the previously entered filter value when reopen esf dialog.', fakeAsync(() => {
```
hanastasov, line 3325: I think this test verifies the single case when clicking the Custom filter, does it. Since opening the Custom filter works differently than all other conditions (Custom loads the dialog with the current conditions, while clicking any other condition resets the expressions (unless clicking on the condition that is already applied). Let's have the test name be more descriptive about it
```suggestion
        it('Should show the previously entered filter value when reopen esf dialog from Custom filter menu.', fakeAsync(() => {
```

hanastasov (changes requested):

projects/igniteui-angular/src/lib/grids/grid/grid-filtering-ui.spec.ts
hanastasov: Let's use the same bed to verify that the excel filter icon has the filtered class applied:

            const activeFilterIcon = header.nativeElement.querySelector('.igx-excel-filter__icon--filtered');
            expect(activeFilterIcon).toBeDefined('no active filter icon was found');
            
            
image

hanastasov (changes requested):

hanastasov:
> ### **Create a new issue from this comment because it is not related to this PR**
> I noticed something else, and we talked with Desi: This is an additional issue that should be logged, but we can create such from this comment here.
> 
> When I select one list item, the Text filter is changed to Text filter (1), and "Equal" is selected from the filters. When I select two list items, the Text filter is changed to Text filter (2), and the "Custom filter" is selected from the filters. However, when I select three items, the Text filter is changed to Text filter (1), and there is no selection in the Text filters. Shouldn't it be showing the "Custom filter" again? This is replicated from the Master, and it is an issue there as well.
> 
>  Screen.Recording.2024-06-28.at.11.12.37.mov 
> I believe the logic should be: when two or more items are selected to apply the Custom filter, the filters should be indicated with (n) next to the Text filter.
> 
> Something interesting here is that everything looks fine if I select two items and then add more from the Custom filter. After manually adding an additional filter, there is no indication of which items/filters are selected.
> 
>  Screen.Recording.2024-06-28.at.11.12.58.mov

Nice catch, @AnjiManova !

This is due to the fact how ESF (Excel Style Filtering) works:
For relatively small number of items selected through the checkboxes - an "equals" filtering condition is created for each item. So, for 2 items selected - two filtering conditions are created, therefor you see Text filter (2) displayed.
When the number of items pass a certain threshold - like when you selected the third item, a single condition of different type is created (IIRC the condition is called "in" - like the filered records should be "in" the collection of selected items, smth like that) - therefore now you see Text filter (1)

I feel this is kind of an edge case and not even sure if a bug, so yes - please log it as a separate issue, tagged with `question` label

hanastasov (approved):

hanastasov (approved):
Since the comment about extracting the code to initialize expression in a new method was resolved without the comment being addressed, I am giving an example what I imagined for a method without this repetitive code.:

```
    private createInitialExpressionUIElement() {
        const firstExprUI =  this.initializeExpressionUI();
        const secondExprUI = this.initializeExpressionUI();
 
        if (this.expressionsList.length === 1 && this.expressionsList[0]?.expression?.condition?.name === this.selectedOperator) {
            ...
        } else {
            ...
        }

        firstExprUI.afterOperator = FilteringLogic.And;
        secondExprUI.beforeOperator = FilteringLogic.And;
        this.expressionsList = [firstExprUI ,secondExprUI];
    }
```

The other changes in the method are a custom preference of course, but it makes it look more readable.

Issue #14249, opened 2024-05-21
Headers focus and keyboard navigation for Pivot Grid Row Dimension Headers. 


skrustev:
## Is your feature request related to a problem? Please describe.
Add focus/activation for igxPivotGrid row dimension headers on click. 

## Describe the solution you'd like
Using keyboard navigation the user should be able to go from row headers to row dimension header to column headers and back. Pretty much go to any pivot grid headers when the row dimension headers are enabled.

Sorting should be possible throw keyboard interaction as well.

Currently dimension expand/collapse through keyboard interaction is not possible, so maybe include that as well?

PR #14342, opened 2024-06-07
feat(pivotGrid): row dimension expand/collapse with kb interaction

IMinchev64:
Part of #14249  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-navigation.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-navigation.service.ts
@@ -1,7 +1,8 @@
 import { IgxGridNavigationService } from '../grid-navigation.service';
 import { Injectable } from '@angular/core';
 import { IgxPivotGridComponent } from './pivot-grid.component';
-import { HEADER_KEYS } from '../../core/utils';
+import { HEADER_KEYS, ROW_COLLAPSE_KEYS, ROW_EXPAND_KEYS } from '../../core/utils';
+import { PivotUtil } from './pivot-util';
 
 @Injectable()
 export class IgxPivotGridNavigationService extends IgxGridNavigationService {
@@ -32,6 +33,10 @@ export class IgxPivotGridNavigationService extends IgxGridNavigationService {
                 layout: null
             }
 
+            if (event.altKey) {
+                this.handleAlt(key, event);
+                return;
+            }
             if ((key.includes('left') || key === 'home') && this.activeNode.column > 0) {
                 newActiveNode.column = ctrl || key === 'home' ? 0 : this.activeNode.column - 1;
             }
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-navigation.service.ts
```diff
@@ -67,6 +72,35 @@ export class IgxPivotGridNavigationService extends IgxGridNavigationService {
         }
     }
 
+    public override handleAlt(key: string, event: KeyboardEvent): void {
+        event.preventDefault();
+
+        let row = this.grid.gridAPI.get_row_by_index(this.activeNode.row);
+        let expansionRowKey = PivotUtil.getRecordKey(row.data, row.data.dimensions[0])
```
MayaKirova, line 79: `row.data.dimensions[0]` will only ever get the first row dimension and its hierarchy. If you have a second dimension with hierarchy you would not be able to toggle it. You can find the correct dimension based on the `activeNode`'s column.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-navigation.service.ts
```diff
@@ -67,6 +72,35 @@ export class IgxPivotGridNavigationService extends IgxGridNavigationService {
...
+        let isExpanded = this.grid.expansionStates.has(expansionRowKey) ? this.grid.expansionStates.get(expansionRowKey) : true;
+
+        if (!isExpanded && ROW_EXPAND_KEYS.has(key)) {
+            if (row.key === undefined) {
```
MayaKirova, line 83: Is there instances where there's no key for the row? I think that might be true for the base grid, but the pivot always generates keys so we probably will never have such a scenario. So this can be simplified.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-navigation.service.ts
```diff
@@ -67,6 +72,35 @@ export class IgxPivotGridNavigationService extends IgxGridNavigationService {
...
+                // TODO use expanded row.expanded = !row.expanded;
+                (row as any).toggle();
+            } else {
+                const newExpansionState = new Map<any, boolean>(this.grid.expansionStates);
+                newExpansionState.set(expansionRowKey, true);
+                this.grid.expansionStates = newExpansionState;
```
MayaKirova, line 89: This is fine, but it probably won't trigger any events, like the `rowToggle` event. Would be better to use the existing api:
`this.grid.gridAPI.set_row_expansion_state`

MayaKirova (changes requested):

MayaKirova (approved):

PR #14408, opened 2024-06-18
Change searchPlaceholder depending on the filterable property

### LINKED ISSUE #14340: Choose proper combo filtering options
## Description  
The removal of `filterable` property and integration inside the `filteringOptions` needs to be reconsidered. The main reason is that in `igniteui-webcomponents` combo they are separated. If this is a better approach, we should apply it back in the `igniteui-angular` combo, accompanied by the proper migration. 

More info: https://github.com/IgniteUI/igniteui-angular/pull/14339#issue-2338455618

MonikaKirkova:
Closes #14340 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


Lipata:
@sdimchevski With this PR in `igniteui-angular` we are always making filtering available and removing the option to enable/disable it. This is a little bit different from what is in `igniteui-webcomponents`, where we have the `disableFiltering` option available.

You can see that in angular [sample](https://www.infragistics.com/products/ignite-ui-angular/angular/components/combo-features), disabling the filter will not remove the search input, just will stop the filtering and reduce the list. While in the webcomponent [sample](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/combo/features) when filtering is disabled, then the search input is removed.

On one hand, we want to have consistency between different platforms. On the other hand disabling the filtering may not be a common scenario, but would appreciate any design input from you on the matter.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes for each version of this project will be documented in this file.
 
+## 18.1.0
+### General
+- `IgxCombo` and  `IgxSimpleComboComponent`
+    - **Breaking Change** `filterable` is removed from the `filteringOptions`.
```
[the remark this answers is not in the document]
MonikaKirkova: There is a custom migration for the filteringOptions.

Lipata (approved):

sdimchevski:
> @sdimchevski With this PR in `igniteui-angular` we are always making filtering available and removing the option to enable/disable it. This is a little bit different from what is in `igniteui-webcomponents`, where we have the `disableFiltering` option available.
> 
> You can see that in angular [sample](https://www.infragistics.com/products/ignite-ui-angular/angular/components/combo-features), disabling the filter will not remove the search input, just will stop the filtering and reduce the list. While in the webcomponent [sample](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/combo/features) when filtering is disabled, then the search input is removed.
> 
> On one hand, we want to have consistency between different platforms. On the other hand disabling the filtering may not be a common scenario, but would appreciate any design input from you on the matter.

@Lipata @kdinev From a UX point of view, right now in Angular it’s not OK and is confusing to set filtering disabled and still have the search option. The behavior in Angular is actually the same as in Web components, once you switch off the Allow Custom Values feature (there is no such feature in WebC). So instead of removing the option to enable/disable the filtering in Angular, when we have the filtering disabled and the Allow Custom Values enabled, why don't we change the Placeholder from Search to Add Item?

kdinev:
> > @sdimchevski With this PR in `igniteui-angular` we are always making filtering available and removing the option to enable/disable it. This is a little bit different from what is in `igniteui-webcomponents`, where we have the `disableFiltering` option available.
> > You can see that in angular [sample](https://www.infragistics.com/products/ignite-ui-angular/angular/components/combo-features), disabling the filter will not remove the search input, just will stop the filtering and reduce the list. While in the webcomponent [sample](https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/combo/features) when filtering is disabled, then the search input is removed.
> > On one hand, we want to have consistency between different platforms. On the other hand disabling the filtering may not be a common scenario, but would appreciate any design input from you on the matter.
> 
> @Lipata @kdinev From a UX point of view, right now in Angular it’s not OK and is confusing to set filtering disabled and still have the search option. The behavior in Angular is actually the same as in Web components, once you switch off the Allow Custom Values feature (there is no such feature in WebC at the moment). So instead of removing the option to enable/disable the filtering in Angular, when we have the filtering disabled and the Allow Custom Values enabled, why don't we change the Placeholder from Search to Add Item?

I agree with your suggestion and if the placeholder displays the correct prompt, then I'm OK.

MonikaKirkova:
@sdimchevski, @kdinev Except changing the placeholder, should we remove the `filterable` from the `filteringOptions` and rename the `filterable` property to `disableFiltering` as in the web components.

kdinev:
> @sdimchevski, @kdinev Except changing the placeholder, should we remove the `filterable` from the `filteringOptions` and rename the `filterable` property to `disableFiltering` as in the web components.

Yes, please do that!

damyanpetev:
@MonikaKirkova Since we want to have the search placeholder change between filtering & custom value state being enabled, it might not be bad idea to have those placeholders in the `resourceStrings` for the Combo. That's now hardcoded in English in the `@Input() searchPlaceholder`, but since it'll be dynamic I think we should deprecate the input and leave it empty by default.

PS: Talking of localization, there are also string of hardcoded English in the markup like `"Clear Selection"` and `Add item` which should all be localizable and in the resource strings. Could be a separate task ofc for any such strings you find.

Lipata (approved):

Lipata (approved):

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ All notable changes for each version of this project will be documented in this
     - The `isFirstPageDisabled` and `isLastPageDisabled` have been deprecated in favor of the identical `isFirstPage` and `isLastPage` getter.
 - `IgxOverlayService`
     - The `attach` method overload accepting `Type` and `OverlaySettings` now accepts `OverlayCreateSettings` as second parameter. This interface extends `OverlaySettings` with an additional `injector` property used as `ElementInjector` when creating the dynamic component.
+- `IgxCombo`:
+    - **Breaking Change** The `filterable` property is renamed to `disableFiltering`.
```
damyanpetev, line 32: Um, the main breaking is the `filterable` in the `filteringOptions` being removed?
Might not be an easy way to migrate that - depending on usage, should we consider deprecating it instead? ( I know 😑)

IIRC `filterable` was marked as deprecated, so it should have no active usage (hopefully), though that's also breaking, sure.

damyanpetev: Also, also this should be for 18.2 now?

Lipata (approved):

projects/igniteui-angular/src/lib/core/i18n/combo-resources.ts
```diff
@@ -1,11 +1,13 @@
 export interface IComboResourceStrings {
     igx_combo_empty_message?: string;
+    igx_combo_enabledFilter_placeholder?: string;
     igx_combo_addCustomValues_placeholder?: string;
     igx_combo_clearItems_placeholder?: string;
```
damyanpetev, line 5: Eh, somewhat assumed this was going to be added with this PR, but instead it went in with #14620 (and propagated back too), so a bit late for this comment:
Do be careful with naming public API - it's not a `addCustomValues_placeholder` if it's also the text of the button, no? So the name should've been more generic or reflective of the use. I do hope that's been communicated so the translation makes sense in both places, though we might very well consider 
Similar for `clearItems_placeholder`.. that's a tooltip or aria label.

---

As for the actual `enabledFilter_placeholder` - could be more descriptive, either filter field or filter search placeholder (to go with the old prop I guess).

damyanpetev (changes requested):
Mostly the changelog/public API and some parts of the migrations that might need addressing. The rest are comments for now or at least something to discuss.

projects/igniteui-angular/migrations/update-18_2_0/index.ts
```diff
@@ -57,8 +64,76 @@ export default (): Rule => async (host: Tree, context: SchematicContext) => {
...
+            let newMatch = match[0].replace(/(\{|\s*,)\s*filterable:\s*true(?=\s*,|\s*\})/g, (match, p1) => p1 === '{' ? '{' : '');
+            newMatch = newMatch.replace(/\{\s*,/, '{').replace(/,\s*\}/, '}');
+            content = content.replace(match[0], newMatch);
+        }
+
+        const filterableMatch = /(filterable\s*(?::|=)\s*false)[^\n]*/g;
+        content = content.replace(filterableMatch, match => {
+            return `${match} // TODO: Replace with disableFiltering`;
+        });
+        host.overwrite(path, content);
```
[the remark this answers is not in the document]
damyanpetev: Okay, the extra changes from visitor will drop out with #14787 so at last the potential for conflicting changes and will be left to just the remaining combo stuff.

damyanpetev (approved):

PR #14414, opened 2024-06-19
fix(combo): Handle selection when placed in an IgxGrid cell template and grid is virtualized -  master

### LINKED ISSUE #14305: When change in first row of dropdown then 13 th row dropdown changes automatically
After an investigation, I assume that this behavior is observed because of the virtualization of the igx-grid and because only the first 12 rows are displayed initially.

 * igniteui-angular version:  17
 * browser:  Edge

## Result  
When change in first row of dropdown then 13 th row dropdown changes automatically

## Expected result  
When change in first row of dropdown then 13 th row should not show data from first row dropdown

## Attachments  
Attached is the sample code

[s6z9dg.zip](https://github.com/user-attachments/files/15516554/s6z9dg.zip)

ddaribo:
Closes #14305

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/combo/combo.common.ts
+++ b/projects/igniteui-angular/src/lib/combo/combo.common.ts
@@ -180,6 +180,9 @@ export abstract class IgxComboBaseDirective implements IgxComboBase, AfterViewCh
         if (selection) {
             this.selectionService.set(this._id, selection);
         }
+        if (this.dropdown.open) {
+            this.dropdown.close();
+        }
     }
 
     /**
--- a/projects/igniteui-angular/src/lib/combo/combo.common.ts
+++ b/projects/igniteui-angular/src/lib/combo/combo.common.ts
@@ -167,7 +167,20 @@ export abstract class IgxComboBaseDirective implements IgxComboBase, AfterViewCh
      */
     @HostBinding('attr.id')
     @Input()
-    public id = `igx-combo-${NEXT_ID++}`;
+    public get id(): string {
+        return this._id;
+    }
+
+    public set id(value: string) {
+        if (!value) {
+            return;
+        }
+        const selection = this.selectionService.get(this._id);
+        this._id = value;
+        if (selection) {
+            this.selectionService.set(this._id, selection);
+        }
+    }
 
     /**
      * Sets the style width of the element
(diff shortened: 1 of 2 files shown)


jackofdiamond5 (dismissed):

RivaIvanova:
There are two strange behaviors:

1. Usually when opening the combo without a selected item, the first item is focused, however, when opening the combo on the 11th row (which let's say corresponds to the combo in the 1st row), the item that was selected in the 1st combo is focused in the 11th combo.
2. Once I select a value, I cannot deselect it. If I click the clear button, scroll down/up, and return to the combo the previous value is selected. 

igx-simple-combo-in-grid

ddaribo:
@RivaIvanova,

After further investigation, I believe these behaviors have to be addressed on the sample level. I’ve already logged [an issue](https://github.com/IgniteUI/igniteui-angular-samples/issues/3499) in igniteui-angular-samples (it is linked in the current PR’s issue also). More details are listed in its description. There is a pending PR for that as well.

To sum up, IMO the sample has to be adapted to support a virtualized scenario and the introduced changes will address the behaviors you have observed. Please, do test this branch along with the refactored sample (and by adding more records). This leaves this PR’s main task to resolve the error that was raised when scrolling a grid with templated combos with changing IDs, whose selection sets were not properly initialized in that case. I have also added commits that ensures closing the combo and losing its input focus when the ID is changing, for example while scorlling the grid.

projects/igniteui-angular/src/lib/combo/combo.common.ts
```diff
@@ -167,7 +167,27 @@ export abstract class IgxComboBaseDirective implements IgxComboBase, AfterViewCh
...
+        }
+        const selection = this.selectionService.get(this._id);
+        this._id = value;
+        if (selection) {
+            this.selectionService.set(this._id, selection);
+        }
+        if (this.dropdown.open) {
+            this.dropdown.close();
+        }
+        if(this.inputGroup?.isFocused) {
```
jackofdiamond5, line 186:
```suggestion
        if (this.inputGroup?.isFocused) {
```

jackofdiamond5 (approved):

PR #14463, opened 2024-07-03
fix(grid): Fix filtering UI mousedown event handling

### LINKED ISSUE #14322: Cannot move the cursor with the mouse after using igxFilterTemplate
## Description  
While using data-grid, I wanted to customize the filter function, so I read the documentation and found the igxFilterTemplate directive. I successfully implemented it by following the documentation steps, but I found that when I focused the mouse over the input element, The cursor position cannot be selected by moving the mouse; it can only be selected by the keyboard

 * igniteui-angular version: 13.2.24
 * browser: Chrome 125.0.6422.113

## Steps to reproduce  

1.Go to the official website for an example of using custom filtering  [here](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/filtering#re-templating-filter-cell)
2. Type any number of characters in the input field
3. When the cursor is at the end of the text, use the left mouse button to select any preceding character

## Result  
When clicking with the mouse, the cursor remains at the end of the text.

## Expected result  
The cursor should be placed at the desired position.

## Attachments  
275473429-62adcce8-4666-432c-b6e8-453053902b5a

Zneeky:
Closes #14322

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/headers/grid-header-group.component.ts
```diff
@@ -262,8 +262,12 @@ export class IgxGridHeaderGroupComponent implements DoCheck {
      */
     @HostListener('mousedown', ['$event'])
     public onMouseDown(event: MouseEvent): void {
-        // hack for preventing text selection in IE and Edge while dragging the resize element
-        event.preventDefault();
+        if (!this.grid.allowFiltering || 
+            (event.composedPath().findIndex(el =>
+                (el as Element).tagName?.toLowerCase() === 'igx-grid-filtering-cell') < 1)) {
+                // Hack for preventing text selection in IE and Edge while dragging the resize element
+                event.preventDefault();
```
rkaraivanov, line 269: Since this is a workaround, targeting IE and the old Edge browser engine, I wonder if it is needed. Both of those browsers are not supported anymore. Why don't you try to delete the hack and see what the behavior of the grid is then?

rkaraivanov (changes requested):

projects/igniteui-angular/src/lib/grids/headers/grid-header-group.component.ts
Zneeky: I've taken the time to test the behavior of the grid without the workaround in both Chrome and Firefox. In Chrome, the first header group is being activated and all text is being selected. However, in Firefox, only the first header group is activated without any text selection.

showcaseOfOtherBorwsers



showcaseOfOtherBorwsersF

rkaraivanov: 😞 
I guess an issue to refactor the resizing stack to use pointer events and setPointerCapture/releasePointerCapture is in order then. This will be out of the scope of this fix which does it job.

rkaraivanov (approved):

Issue #14010, opened 2024-03-22
Infer as many editor options from column options as possible


ChronosSF:
## Is your feature request related to a problem? Please describe.
Users expect to set something once. There are cases where column options are not transferred to editor options and have to be set separately or even worse - call for entirely custom templates. One example is the date-time formatting. For cells you set it using pipeArgs to the Angular date pipe but you cannot set it for its corresponding date-time input. Its format is currently hardcoded through a hidden property.

## Describe the solution you'd like
The example above is just one of possibly many. The FR requires a research in what we use to create the underlying editors and how options for columns affect those and how these can  be transferred so the editors work seamlessly in the same way the user expects based on the columns they set.

ddaribo:
For columns of dataType `date`, `time` and `dateTime` and their corresponding editors - IgxDatePicker, IgxTimePicker, igxDateTimeEditor), while in theory their `inputFormat` can be inferred from the column’s `pipeArgs.format` property, I am not convinced this would be the best approach. Since the `pipeArgs.format` could be any among the predefined/custom ones, in case it includes non-numeric parts, it would not be applicable as `inputFormat`. Users would still need to override it by declaring a custom edit template.

Probably a separate property could be introduced for this. For example an `inputFormat` on the pipeArgs or something like `editorInputFormat`on the column, or else. This would rather be covered by the other logged issue on the topic - https://github.com/IgniteUI/igniteui-angular/issues/14009, as it is about providing additional properties.

Regarding the `pipeArgs.timezone` option, I am not sure if it is applicable in the context of editing, since the original value is edited?

For other data types – `number`, `currency` and `percent`, I believe what is lacking is the ability to set a mask on the editor. This could be another candidate for additional property on the pipeArgs or the column. I am not convinced if this should be related to other pipeArgs options, such as `digitsInfo`, or if it could and should be inferred from it. In any case, the user has to be able to opt in or out of applying a mask on the default editor.
For  `string` dataType a mask may still be applicable to it.

ChronosSF:
Thank you for looking into this. I agree it's probably the best to just allow editor options on a column level. I am closing this and leaving the improvements to #14009

Issue #14009, opened 2024-03-22
Provide a set of properties to be settable to default grid CRUD editors without the need for templates


ChronosSF:
## Is your feature request related to a problem? Please describe.
Currently every customization for the grid's default editors requires fully custom templates which however sometimes add complexity far beyond what the change calls for (e.g. having to handle change events and value assignments for simply wanting to change some display format)

## Describe the solution you'd like
The column component can accept some objects based on pre-defined interfaces that it then uses to init its editors with the default templates.

## Describe alternatives you've considered
n/a

## Additional context
n/a

ddaribo:
In continuation of [this comment](https://github.com/IgniteUI/igniteui-angular/issues/14010#issuecomment-2160775400), the currently identified candidate properties are:

- `inputFormat` - could be part of the `IFieldPipeArgs` interface; alternatively - defined on the column (i.e. `editorInputFormat` or else); applicable for the `date`, `dateTime` and `time` column types.
- `inputMask` - also could be part of the `IFieldPipeArgs` or else; applicable for `string` and `image` column data types as they both use the same template - a text input group. The `igxMask` cannot be applied for other types of inputs, so not applicable for `number`, `currency` and `percent` types.

Let me know what you think @ChronosSF, @rkaraivanov. @kacheshmarova could schedule a meeting to further discuss the specs.

ddaribo:
To sum up the meeting discussion:

-	An `editorOptions` object would be exposed on the column and for now it would contain a single property - `dateTimeFormat` (names to be further polished). This will be applied to the default editors of `date`, `time` and `dateTime` columns.
-	In case such an `editorOptions.dateTimeFormat` config is not provided, we would try to infer the date/time inputFormat from the set `pipeArgs.displayFormat`, if any. The condition for the `displayFormat` to be applied as `inputFormat` is that its date-time parts can be parsed as all numeric. If not, the editor’s inputFormat will fallback as per the grid’s locale (Note: currently it is a hardcoded default format for the `dateTime` column type). 
-	The current feature would not cover properties for numeric column types. They would be subject to future discussions about creating separate numeric inputs.

PR #14465, opened 2024-07-03
feat(grid): expose editorOptions for the default editors and improve built-in editing for date/time cols (input formats)

ddaribo:
Related to #14009 and #14010

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 -  [x] This PR includes API docs for newly added methods/properties ([api docs guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines))
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them - [compare revisions](https://github.com/IgniteUI/igniteui-angular/wiki/igxGrid-Column-Data-Types-and-Templates-Specification/_compare/3e68d657f4ddc0bc585642e0241b86ec0e69f44c)
 TODO: refactor the column datat types spec after comments, update editors spec


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
ChronosSF, line 1: Looks good. Just unsure about this file as the whole setByUser thing and checks around seem to be done just to allow the user to set null/empty/invalid objects which would result in the editors receiving defaultInputFormat instead of the format coming from pipeArgs. I have the feeling this could be done with less code overall. If your default is getting the pipeArgs, you could simply init `_editorOptions` as `{ inputFormat: this._columnPipeArgs.format }` and let the setter simply override it.

I am approving this anyway as I am not sure if I am considering all possible outcomes from the change. Just something to try and simplify if you agree.

ChronosSF (dismissed):

ChronosSF (approved):

ddaribo:
@ChronosSF, with the introduced changes,  the two properties hardcoding default date/time fromats are left unused: `defaultTimeFormat` and `defaultDateTimeFormat` ([ref](https://github.com/IgniteUI/igniteui-angular/blob/bpachilova/grid-editor-options-feat-14009/projects/igniteui-angular/src/lib/grids/columns/column.component.ts#L1662) in _column.component.ts_ and [ref](https://github.com/IgniteUI/igniteui-angular/blob/8ce97faf62ba2894a2ce111fd87c6c5dca6bd037/projects/igniteui-angular/src/lib/grids/common/grid.interface.ts#L308) in _grid.interface.ts_). Since we discussed these are rather controversial as part of the public API, should they be removed? If so, should this be done in the current PR, or to log a separate issue?

ChronosSF:
> @ChronosSF, with the introduced changes, the two properties hardcoding default date/time fromats are left unused: `defaultTimeFormat` and `defaultDateTimeFormat` ([ref](https://github.com/IgniteUI/igniteui-angular/blob/bpachilova/grid-editor-options-feat-14009/projects/igniteui-angular/src/lib/grids/columns/column.component.ts#L1662) in _column.component.ts_ and [ref](https://github.com/IgniteUI/igniteui-angular/blob/8ce97faf62ba2894a2ce111fd87c6c5dca6bd037/projects/igniteui-angular/src/lib/grids/common/grid.interface.ts#L308) in _grid.interface.ts_). Since we discussed these are rather controversial as part of the public API, should they be removed? If so, should this be done in the current PR, or to log a separate issue?


So normally `defaultTimeFormat` and `defaultDateTimeFormat` being hidden/internal should be safe to remove, but why and who sneaked them in a public interface makes it more difficult. Can you check where this FieldType is used? I can't remember using this interface. I hope it's not something people can use to represent a column. Check the samples browser / docs for it. If it's used we may need to add a migration and deprecate the props first.

ddaribo:
@ChronosSF, It is used in the Query Builder, so indirectly in the grid advanced filtering dialog as well.
_| `fields`  | FieldType[]  | An array of fields to be filtered. Contains information about  label, field, type, operands. |_
The `ColumnType` extends it.
`FieldType` is only mentioned in the Query Builder's topic code-snippets [here](https://www.infragistics.com/products/ignite-ui-angular/angular/components/query-builder#getting-started-with-ignite-ui-for-angular-query-builder) and is not imported in the [demo](https://www.infragistics.com/products/ignite-ui-angular/angular/components/query-builder#getting-started-with-ignite-ui-for-angular-query-builder)s since the `fields` are declared as `any[]`.
The two properties are not referenced anywhere in the topics/samples and `FieldType` is not mentioned in any other topic.

jackofdiamond5:
The changes look good to, though I have a little question. With the introduction of the `editorOptions` property on the column, I can now set the `dateTimeFormat` of the column which properly applies the mask/date string I've set while in edit mode. However, this affects only edit mode and I still have to use the `pipeArgs` to make any changes to the display format (using NG's DatePipe). To me this is an unneeded separation, as if I want to have both display and input/edit formats modified, I will have to set both the property and add a template/modify the `pipeArgs`. I think it would be more user friendly to do this in one place only - like for example through the exposed `editorOptions` - set the display and input formats there, ditch the `pipeArgs` for dates entirely and rely on the editors to handle everything afterwards. As the editors do use the NG's DatePipe internally when they build the display format, so no need to do it on the column level imo. Regarding the input format, with the newest implementation that's been added in the `DateTimeUtils` we should be able to handle NG's shorthand formats in the editors, like we've mapped them to whatever formats we've deemed appropriate. So atm at least I cannot see a reason why we don't leave the editors to handle both input and display formats, providing they've been given such throgh the `editorOptions`.

One little thing that I saw, which may have been discussed previously is that through the `editorOptions` I can set a `shortTime` string to a column with a data type set to `date`. Which results in a rather funny behavior where during editing you see a time format and while the cell is outside of editing you see the display format of the date portion. Additionally, it is a date picker, so yeah - `DateTimePicker` confirmed I guess 😆? 
image


Anyway, the same is true with a `time` column where we set the `dateTimeFormat` to `shortDate`. And while I agree that it's a niche case, it doesn't hurt to guard against it and to default to a format that will actually make sense in the given context and be useful rather than looking broken.

ChronosSF:
> The changes look good, though I have a little question. With the introduction of the `editorOptions` property on the column, I can now set the `dateTimeFormat` of the column which properly applies the mask/date string I've set while in edit mode. However, this affects only edit mode and I still have to use the `pipeArgs` to make any changes to the display format (using NG's DatePipe). To me this is an unneeded separation, as if I want to have both display and input/edit formats modified, I will have to set both the property and add a template/modify the `pipeArgs`. I think it would be more user friendly to do this in one place only - like for example through the exposed `editorOptions` - set the display and input formats there, ditch the `pipeArgs` for dates entirely and rely on the editors to handle everything afterwards. As the editors do use the NG's DatePipe internally when they build the display format, so no need to do it on the column level imo. Regarding the input format, with the newest implementation that's been added in the `DateTimeUtils` we should be able to handle NG's shorthand formats in the editors, like we've mapped them to whatever formats we've deemed appropriate. So atm at least I cannot see a reason why we don't leave the editors to handle both input and display formats, providing they've been given such throgh the `editorOptions`.
> 
> One little thing that I saw, which may have been discussed previously is that through the `editorOptions` I can set a `shortTime` string to a column with a data type set to `date`. Which results in a rather funny behavior where during editing you see a time format and while the cell is outside of editing you see the display format of the date portion. Additionally, it is a date picker, so yeah - `DateTimePicker` confirmed I guess 😆? image
> 
> Anyway, the same is true with a `time` column where we set the `dateTimeFormat` to `shortDate`. And while I agree that it's a niche case, it doesn't hurt to guard against it and to default to a format that will actually make sense in the given context and be useful rather than looking broken.

I don't necessarily agree about the guards and the implementation of those can be quite an endeavor . As long as we don't infer the options incorrectly we can let users make mistakes if they choose to. I think it would be quite obvious for everyone using this option why this happened.

As for the disjointed options, I think you have a point there and don't forget that we also have a `formatter` that also fights for precedence in how the cell is rendered. We'll have to discuss this more with @damyanpetev

ddaribo:
> With the introduction of the `editorOptions` property on the column, I can now set the `dateTimeFormat` of the column which properly applies the mask/date string I've set while in edit mode. However, this affects only edit mode and I still have to use the `pipeArgs` to make any changes to the display format (using NG's DatePipe). To me this is an unneeded separation, as if I want to have both display and input/edit formats modified, I will have to set both the property and add a template/modify the `pipeArgs`. 

@jackofdiamond5, If I understand you correctly, what I can say is that the agreed implementation does this, only the other way around - if the `editorOptions.dateTimeFormat` is not set, the `pipeArgs.format` (if set) is applied as input format to the built-in editors, as well as obviously display format for the cell. 
The  `editorOptions.dateTimeFormat` would be handy if we need a specific input format, different from the default locale one and from the display format set via `pipeArgs`.

> ... ditch the `pipeArgs` for dates entirely and rely on the editors to handle everything afterwards. As the editors do use the NG's DatePipe internally when they build the display format, so no need to do it on the column level imo. Regarding the input format, with the newest implementation that's been added in the `DateTimeUtils` we should be able to handle NG's shorthand formats in the editors, like we've mapped them to whatever formats we've deemed appropriate. So atm at least I cannot see a reason why we don't leave the editors to handle both input and display formats, providing they've been given such throgh the `editorOptions`.

I may not fully understand this suggestion yet, but isn't the `pipeArgs.format` rather applied as format on the cell display value and not the editor? When not in edit mode, this is simply a div with formatted text, so this is handled by the column indeed.

jackofdiamond5:
> > With the introduction of the `editorOptions` property on the column, I can now set the `dateTimeFormat` of the column which properly applies the mask/date string I've set while in edit mode. However, this affects only edit mode and I still have to use the `pipeArgs` to make any changes to the display format (using NG's DatePipe). To me this is an unneeded separation, as if I want to have both display and input/edit formats modified, I will have to set both the property and add a template/modify the `pipeArgs`.
> 
> @jackofdiamond5, If I understand you correctly, what I can say is that the agreed implementation does this, only the other way around - if the `editorOptions.dateTimeFormat` is not set, the `pipeArgs.format` (if set) is applied as input format to the built-in editors, as well as obviously display format for the cell. The `editorOptions.dateTimeFormat` would be handy if we need a specific input format, different from the default locale one and from the display format set via `pipeArgs`.
> 
> > ... ditch the `pipeArgs` for dates entirely and rely on the editors to handle everything afterwards. As the editors do use the NG's DatePipe internally when they build the display format, so no need to do it on the column level imo. Regarding the input format, with the newest implementation that's been added in the `DateTimeUtils` we should be able to handle NG's shorthand formats in the editors, like we've mapped them to whatever formats we've deemed appropriate. So atm at least I cannot see a reason why we don't leave the editors to handle both input and display formats, providing they've been given such throgh the `editorOptions`.
> 
> I may not fully understand this suggestion yet, but isn't the `pipeArgs.format` rather applied as format on the cell display value and not the editor? When not in edit mode, this is simply a div with formatted text, so this is handled by the column indeed.

I see what you mean and yeah I've missed the part where outside of edit mode the cell is just a div with formatting. So, yeah this pretty much answered my question - the formatting will be applied as expected on the editor as long as `pipeArgs.format` is present. In this case it makes sense for the `dateTimeFormat` to be applied only while editing, to provide additional customization options. Though, we may want to change the name to something that reflects that it is to be used specifically when the cell is in edit mode like `dateEditFormat` or something. Also, I like that this works well with a templated cell, since you can provide whatever to the `DatePipe` for a display value and then set the `pipeArgs.format` which will reflect only onto the edit mode format. And I don't know if this has been discussed or not but since this is yet another way to apply formatting to a cell, we may want to revisit the `formatter` and the cell templating options because one is bound to override the other and we should add this to the docs. Like in the aforementioned example - the templated cell will override the display format set by `pipeArgs.format`.

damyanpetev:
Hopefully the name of the main property being `editorOptions` makes it pretty clear it's targeting edit inputs (though we're discussing other interactions), so I'm thinking `editorOptions.dateTimeFormat` should be clear enough. If we include filtering in the mix, it might make more sense as `inputOptions` or something else, could still hopefully make it clear at the root, the format property is long enough already 🥴

And true the options are split, if anything `pipeArgs` is a bit too direct with the inner workings for my taste and should've probably been `display/formatOptions` or something, but that ship has sailed long ago. Similarly for the order of competing pipe options, column formatters and templates - those being a bit out of scope for this change, besides making sure those sync where meaningful.

onlyexeption (dismissed):

projects/igniteui-angular/src/lib/date-common/picker-base.directive.ts
```diff
@@ -29,15 +31,23 @@ export abstract class PickerBaseDirective implements IToggleView, EditorProvider
...
      * ```html
      * <igx-date-picker inputFormat="dd/MM/yy"></igx-date-picker>
      * ```
      */
     @Input()
-    public inputFormat: string;
+    public set inputFormat(value: string) {
+        if (value) {
+            this._userSetFormat = value;
+            this._inputFormat = DateTimeUtil.getNumericInputFormat(this.locale, value);
```
damyanpetev, line 44: Okay, yeah I'll echo another comment from Stamen - not liking the "user set" props. Just recalculate the `_defaultInputFormat` that's already internal and keep the public API property clean and assign either one or the other to the inner input if needed.
Not super expected for the user to set "short" and read the prop and get something else. Though I'm not entirely sure the editors/pickers by design are supposed to accept that at all, in which case this functionality would not have a reason to be here. In which case this would be just a call to `updateDefaultFormat()`

projects/igniteui-angular/src/lib/date-common/picker-base.directive.ts
```diff
@@ -128,6 +138,9 @@ export abstract class PickerBaseDirective implements IToggleView, EditorProvider
         } catch (e) {
             this._locale = this._localeId;
         }
+        // reassign the inputFormat in case the original one needs to be localized
+        this.inputFormat = this._userSetFormat;
```
damyanpetev, line 142: Here too.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -180,16 +181,24 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
...
      * @example
      * ```html
      * <igx-date-range-picker inputFormat="dd/MM/yy"></igx-date-range-picker>
      * ```
      */
     @Input()
-    public override inputFormat: string;
+    public override set inputFormat(value: string) {
+        if (value) {
+            this._inputFormat = DateTimeUtil.getNumericInputFormat(this.locale, value);
```
damyanpetev, line 192: Similar to the base - there's no need for the `getNumericInputFormat` here, just get the default by locale and handle accordingly.

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.spec.ts
```diff
@@ -938,12 +941,67 @@ describe('IgxDateRangePicker', () => {
                     fixture.componentInstance.displayFormat = displayFormat;
                     fixture.detectChanges();
 
-                    expect(startInputEditor.inputFormat).toEqual(inputFormat);
+                    expect(startInputEditor.inputFormat).toEqual('MM/dd/yyyy');
```
damyanpetev, line 944: Erm, that's actually wrong or am I missing something? If the component has `inputFormat = 'MMM, yy'` assigned, it should not be ignored/transformed in any way

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.spec.ts
```diff
@@ -938,12 +941,67 @@ describe('IgxDateRangePicker', () => {
...
+
+                    expect(startInputEditor.inputFormat).toEqual('yyyy/MM/dd');
+                    expect(startInputEditor.nativeElement.placeholder).toEqual('yyyy/MM/dd');
+                    expect(endInputEditor.inputFormat).toEqual('yyyy/MM/dd');
+
+                    dateRange.locale = 'bg-BG';
+                    fixture.detectChanges();
+                    tick();
+
+                    expect(startInputEditor.inputFormat.normalize('NFKC')).toEqual('dd.MM.yyyy г.');
```
damyanpetev, line 974: So this is basically the default functionality of the date time editor directive? IMO, unless the default doesn't match the default type the directive uses (date instead of date-time) I don't even see a reason for the picker to re-implement this at all and the only expectation here would be locale is correctly passed to the inputs and the rest is covered in the directive already?

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.spec.ts
```diff
@@ -938,12 +941,67 @@ describe('IgxDateRangePicker', () => {
...
+                it('should resolve to the default locale-based inputFormat for the start/end editors in case the one set contains non-numeric time parts', fakeAsync(() => {
+                    registerLocaleData(localeBg);
+                    const startInputEditor = startInput.injector.get(IgxDateTimeEditorDirective);
+                    const endInputEditor = endInput.injector.get(IgxDateTimeEditorDirective);
+
+                    dateRange.locale = 'bg-BG';
+                    fixture.detectChanges();
+                    tick();
+
+                    dateRange.inputFormat = 'full';
```
damyanpetev, line 988: Still don't think this is intended at all. What this could do would be to allow the display format to affect to input format if not provided just like the grid, but that could be a separate target.

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
```diff
@@ -148,8 +149,10 @@ export class IgxDateTimeEditorDirective extends IgxMaskDirective implements OnCh
     @Input(`igxDateTimeEditor`)
     public set inputFormat(value: string) {
         if (value) {
-            this.setMask(value);
-            this._inputFormat = value;
+            this._userSetFormat = value;
+            this._inputFormat = DateTimeUtil.getNumericInputFormat(this.locale, value);
```
damyanpetev, line 153: Add to the list with user set and changing the actual input format instead of the default and supporting non-numeric formats and so on :)

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1592,6 +1592,36 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
...
+    @notifyChanges()
+    @WatchColumnChanges()
+    @Input()
+    public set editorOptions(value: IColumnEditorOptions) {
+        this._editorOptions = value;
+        this.grid.pipeTrigger++;
+    }
+    public get editorOptions(): IColumnEditorOptions {
+        if (!this._editorOptions?.dateTimeFormat) {
+            return { ...this._editorOptions, dateTimeFormat: this.pipeArgs.format };
```
damyanpetev, line 1620: Any chance this can be centralized somewhere as a getter or check in the logic that's about to pass the format to the UI components?

projects/igniteui-angular/src/lib/grids/columns/interfaces.ts
```diff
@@ -51,6 +51,11 @@ export interface IFieldPipeArgs {
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface IColumnPipeArgs extends IFieldPipeArgs {}
 
+export interface IColumnEditorOptions {
```
damyanpetev, line 54: Assuming this is supposed to be used in the Query builder as well, there seems to be a considerable effort to not use column wording there, so we should consider having FieldEditorOptions and if needed alias those for the column like the pipe args

projects/igniteui-angular/src/lib/query-builder/query-builder.component.ts
```diff
@@ -911,12 +912,15 @@ export class IgxQueryBuilderComponent implements AfterViewInit, OnDestroy {
...
+    }
 
-        if (!field.defaultTimeFormat) {
-            field.defaultTimeFormat = DEFAULT_TIME_FORMAT;
+    private setDateTimeEditorFormat(field: FieldType, format: string) {
+        if (!DateTimeUtil.isDateTimeDataType(field.dataType)) {
+            return;
+        }
+        if (!field.editorOptions?.dateTimeFormat) {
+            field.editorOptions.dateTimeFormat = format;
```
damyanpetev, line 923: Oh, also not a fan of modifying these field options if those are directly the user-provided ones, but that seems to be a general issue with the QB handling looking around

CHANGELOG.md
```diff
@@ -28,6 +28,20 @@ For Firefox users, we provide limited scrollbar styling options through the foll
 - `--sb-thumb-bg-color`: Sets the background color of the scrollbar thumb.
 - `--sb-track-bg-color`: Sets the background color of the scrollbar track.
 
+#### Column editor options
+- `ColumnType`, `IgxColumn`
```
damyanpetev, line 32: There are at least two new features here, if not 3.
You have a general grid localiation of the input format used in editors (Edit, Filtering, etc UI), auto sync from the display format (pipeArgs.format) where applicable and then the new option to override the inferred format from either of the previous two.
Oh, and technically, those seem to be features baked into all date time editors and pickers too, right? (though how we describe that based on the current comments is TBD)

damyanpetev (changes requested):

CHANGELOG.md
damyanpetev: Oh and I guess we should include the Query Builder in the components getting the new functionality

projects/igniteui-angular/src/lib/date-common/util/date-time.util.ts
```diff
@@ -506,6 +529,58 @@ export abstract class DateTimeUtil {
...
+    }
+
+    /**
+     * Returns an input format that can be used by the date-time editors, as
+     * - if the input format is already numeric, return it as is
+     * - if it is among the predefined numeric ones, return it as the equivalent locale-based format
+     *   for the corresponding numeric date parts
+     * - otherwise, return an empty string
+     */
+    public static getNumericInputFormat(locale: string, inputFormat: string): string {
```
damyanpetev, line 557: Would actually rename the `inputFormat` param to avoid confusing it with the return :) Perhaps just `format`?
And the doc comment might want some rewording following the same logic.

projects/igniteui-angular/src/lib/date-common/util/date-time.util.ts
```diff
@@ -506,6 +529,58 @@ export abstract class DateTimeUtil {
...
+        let resultFormat = '';
+        if (predefinedNumericFormats.has(inputFormat)) {
```
damyanpetev, line 559: Was going to leave a comment where this was used with `getNumericInputFormat() || getDefaultInputFormat()`, but actually it should be here:
Do check if there's even a format to work with and add an early return to avoid doing extra work (attemting to parse, etc)

projects/igniteui-angular/src/lib/date-common/util/date-time.util.spec.ts
```diff
@@ -233,6 +236,30 @@ describe(`DateTimeUtil Unit tests`, () => {
         expect(result).toEqual('MM/dd/yyyy');
     });
 
+    it('should properly build input formats based on locale for dateTime data type ', () => {
+        spyOn(DateTimeUtil, 'getDefaultInputFormat').and.callThrough();
```
damyanpetev, line 240: PS: This spy is a bit useless :) Noticed it came from the prev test, but still :)

projects/igniteui-angular/src/lib/date-common/util/date-time.util.spec.ts
```diff
@@ -628,4 +655,60 @@ describe(`DateTimeUtil Unit tests`, () => {
         expect(DateTimeUtil.isValidDate(false)).toBeFalse();
         expect(DateTimeUtil.isValidDate(true)).toBeFalse();
     });
+
+    it('should correctly identify formats that would resolve to only numeric parts (and period) for the date/time parts', () => {
+        // test with locale covering non-ASCII characters as well
+        registerLocaleData(localeBg, 'bg');
```
damyanpetev, line 661: shouldn't this be with the next test that actually uses the locale data?

projects/igniteui-angular/src/lib/date-common/util/date-time.util.ts
```diff
@@ -506,6 +529,58 @@ export abstract class DateTimeUtil {
...
+
+    /** Gets the locale-based format from an array of date parts */
+    public static getLocaleInputFormatFromParts(locale: string, dateParts: DateParts[]): string {
+        const options = {};
+        dateParts.forEach(p => {
+            if (p !== DateParts.Year && p !== DateParts.AmPm) {
+                options[p] = FormatDesc.TwoDigits;
+            } else {
+                options[p] = FormatDesc.Numeric;
+            }
```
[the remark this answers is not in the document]
damyanpetev: @ddaribo The second comment is still active - at least for `getLocaleInputFormatFromParts`, I guess the numeric check could be useful as is

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -162,34 +163,28 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
...
      */
     @Input()
-    public override inputFormat: string;
+    public override set inputFormat(value: string) {
+        if (value) {
+            this._inputFormat = value;
+            if (this.hasProjectedInputs) {
+                this.updateInputFormat();
+            }
+        }
```
damyanpetev, line 182: Same logic with the picker base - the old functionality around `updateInputFormat` didn't care if there was a value or not.
Also you could use `super.inputFormat` to avoid rewriting the same get/set logic, IIRC it should work with our compile options. It's already in the state directive.

damyanpetev: Same could go for the `displayFormat`, assuming the override is for documentation snippet purposes?

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
```diff
@@ -135,7 +136,17 @@ export class IgxDateTimeEditorDirective extends IgxMaskDirective implements OnCh
      * ```
      */
     @Input()
-    public displayFormat: string;
+    public set displayFormat(value: string) {
+        if (value) {
+            this._displayFormat = value;
+        }
+        this.updateDefaultFormat();
+        this.setMask(this.inputFormat);
```
damyanpetev, line 144: Same here for the value check. Also if this set mask popped here, wouldn't there be an equivalent place it moved from?

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
```diff
@@ -500,8 +512,9 @@ export class IgxDateTimeEditorDirective extends IgxMaskDirective implements OnCh
     /** @hidden */
     protected override setPlaceholder(_value: string): void { }
 
-    private updateDefaultFormat(): void {
-        this._defaultInputFormat = DateTimeUtil.getDefaultInputFormat(this.locale);
+    private updateDefaultFormat(dataType: DataType = DataType.DateTime): void {
```
damyanpetev, line 515: Um wouldn't this make the editor default to date-time format if not specified, where it used to default to date-only? Don't think that'd be expected and the param isn't used, so just leave it to the old default in `getDefaultInputFormat`

CHANGELOG.md
damyanpetev: @ddaribo I'd still describe the auto localized input format and infer from the display(pipe) format as a separate (first) point. That'll could even be more impactful for more people that the new option which is an override if the new functionality doesn't produce quite what you'd like :)

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1592,6 +1592,33 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
...
+     * <igx-column dataType="date" [editorOptions]="editorOptions"></igx-column>
+     * ```
+     * @memberof IgxColumnComponent
+     */
+    @notifyChanges()
+    @WatchColumnChanges()
+    @Input()
+    public set editorOptions(value: IColumnEditorOptions) {
+        this._editorOptions = value;
+        this.grid.pipeTrigger++;
```
damyanpetev, line 1616: Assuming this is based on the `pipeArgs` setter, however I don't believe the `editorOptions` are used anywhere in the pipes, so triggering them is unnecessary extra work

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
ddaribo: Right, I changed this, but since this is related to the default format for the `dateTime` column in the grid, thought before trying to come up with anything, it would be better to discuss how it should be addressed, after this change and according to the new implementation. Currently some tests will be failing.
So far, the input format was set to a default format property ([ref](https://github.com/IgniteUI/igniteui-angular/blob/a34e44c06d8abe05d57f3d1fa790516840ecb90e/projects/igniteui-angular/src/lib/grids/cell.component.html#L143)) defined on the FieldType interface ([ref](https://github.com/IgniteUI/igniteui-angular/blob/a34e44c06d8abe05d57f3d1fa790516840ecb90e/projects/igniteui-angular/src/lib/grids/common/grid.interface.ts#L308)). So it is hardcoded and also includes time parts.

So, if we'd like that to remain the same, I guess it should be additionally handled for the `dateTime` column?

To give some more context, the input format will not be inferred from the displayFormat in this case, as the default display format for `dateTime` col (`medium`) contains non-numeric parts (ref [here](https://github.com/IgniteUI/igniteui-angular/blob/a34e44c06d8abe05d57f3d1fa790516840ecb90e/projects/igniteui-angular/src/lib/grids/columns/column.component.ts#L62) and [here](https://github.com/IgniteUI/igniteui-angular/blob/a34e44c06d8abe05d57f3d1fa790516840ecb90e/projects/igniteui-angular/src/lib/grids/columns/column.component.ts#L1841)). So, it will resolve to the default one for the dateTimeEditorDirective, which will be date-only.

projects/igniteui-angular/src/lib/grids/grid/column.spec.ts
```diff
@@ -1101,12 +1086,215 @@ describe('IgxGrid - Column properties #grid', () => {
...
+            fix.detectChanges();
+
+            expect(firstCell.nativeElement.innerText).toEqual('01-10-2015 11:37 AM');
+
+            firstCell.setEditMode(true);
+            fix.detectChanges();
+            tick();
+
+            let input = firstCell.nativeElement.querySelector('.igx-input-group__input');
+            expect((input as any).value).toEqual('01-10-2015 11:37 AM');
```
damyanpetev, line 1150: Since you started cleaning up the hardcoded selector elsewhere, might as well continue to query `By.directive(IgxDateTimeEditorDirective)` in all tests. Or make the selector a constant.
Also judging by the other tests at this point the editor isn't focused yet, right? So checking the the value will give you the display instead of checking the input format. IIRC, I think you need to wait for a rAF for the focus directive to kick in (i.e. `tick(16)`), though you could as well just check the `intputFormat` prop on the editor to confirm the column passed the correct options and the rest in theory is already covered by the editor tests.

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
damyanpetev: Okay, was going to ask how the columns even manage the difference between date/date-time because either default on the editor won't work for both. That `medium` format is unfortunate, but I guess it's the only reasonable option w/ time included :/
Regardless, changing the behavior of another component to suit the grid usage probably not the way to go. Would've also been nice if the spec actually specified the default inputFormat behavior rather than a remark on the API doc. Which reminds me multiple specs are also likely due an update once we're done moving around changes.

projects/igniteui-angular/src/lib/directives/date-time-editor/date-time-editor.directive.ts
```diff
@@ -177,6 +185,18 @@ export class IgxDateTimeEditorDirective extends IgxMaskDirective implements OnCh
...
+     * Specify the default input format type. Defaults to `date`, which includes
+     * only date parts for editing. Other valid options are `time` and `dateTime`.
+     *
+     * @example
+     * ```html
+     * <input igxDateTimeEditor [defaultInputFormatType]="'dateTime'">
+     * ```
+     */
+    @Input()
+    public defaultInputFormatType: DataType = DataType.Date;
```
damyanpetev, line 198: Besides this eliminating the need for various picker components to recalculate their own formats as discussed offline, I'm thinking this can easily be shortened to just `defaultFormatType` - unless the `displayFormat` is explicitly overridden, this will also affect it.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -334,6 +334,50 @@ export class IgxDatePickerComponent extends PickerBaseDirective implements Contr
...
+     * @example
+     * ```html
+     * <igx-date-picker displayFormat="EE/M/yy"></igx-date-picker>
+     * ```
+     *
+     */
+    @Input()
+    public override set displayFormat(value: string) {
+        if (this.dateTimeEditor) {
+            this.dateTimeEditor.displayFormat = value;
```
damyanpetev, line 352: Why did this switch away from direct bindings? Was this so the tests can check the inferred values? Because tests can also just query for the editor directive for checks

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
```diff
@@ -330,6 +339,39 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
...
+    /**
+     * @example
+     * ```html
+     * <igx-date-range-picker locale="jp"></igx-date-range-picker>
+     * ```
+     */
+    /**
+     * Gets the `locale` of the date-range-picker.
+     * If not set, defaults to applciation's locale.
+     */
```
damyanpetev, line 351:
```suggestion
     * If not set, defaults to application's locale.
```

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.html
```diff
@@ -155,7 +156,10 @@
             igxInput
             tabindex="0"
             [placeholder]="placeholder"
-            [igxDateTimeEditor]="column.defaultDateTimeFormat"
+            [locale]="filteringService.grid.locale"
+            [displayFormat]="column.pipeArgs.format"
+            [igxDateTimeEditor]="column.editorOptions?.dateTimeFormat"
+            [defaultFormatType]="'dateTime'"
```
damyanpetev, line 162:
```suggestion
            defaultFormatType="dateTime"
```

damyanpetev: That's for pretty much everywhere

projects/igniteui-angular/src/lib/date-range-picker/date-range-picker.component.ts
damyanpetev: Wow suggestion landed on the wrong line, it's just a typo :)

projects/igniteui-angular/src/lib/time-picker/time-picker.component.spec.ts
```diff
@@ -1229,6 +1233,71 @@ describe('IgxTimePicker', () => {
...
+                fixture.detectChanges();
+
+                expect(dateTimeEditor.inputFormat.normalize('NFKC')).toEqual('hh:mm tt');
+
+                timePicker.locale = 'bg-BG';
+                timePicker.displayFormat = 'fullTime';
+                fixture.detectChanges();
+
+                expect(dateTimeEditor.inputFormat).toEqual('HH:mm');
+            }));
```
damyanpetev, line 1299: This test seems to be a duplicate of the above

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1592,6 +1592,32 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
         return this._columnPipeArgs;
     }
 
+    /**
+     * @remarks
+     * Pass optional parameters to control properties of the default column editors.
+     * Accepts an `IColumnEditorOptions` object with the `dateTimeFormat` property.
+     * For more details see https://angular.io/api/common/DatePipe
```
damyanpetev, line 1599: The DatePipe reference is no longer quite acurate now, right?

```suggestion
     * Pass optional properties for the default column editors.
     * @remarks
     * Options may be applicable only to specific column type editors. 
```
Honestly if there should be a link it'd be https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-types#datetime-date-and-time (which I suppose can use an update now too lol), but I'm not sure 
Also despite being the only option, the more specific description belongs to the `dateTimeFormat` option itself.

projects/igniteui-angular/src/lib/grids/columns/interfaces.ts
```diff
@@ -51,6 +51,13 @@ export interface IFieldPipeArgs {
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface IColumnPipeArgs extends IFieldPipeArgs {}
 
+export interface IFieldEditorOptions {
+    /** A custom or a predefined format string used for the built-in editors of date/time columns */
```
damyanpetev, line 55:
```suggestion
    /** A custom input format string used for the built-in editors of date/time columns */
```
And here is where we could drop a link. With API I'd love to be able to link to either the directive or time picker `inputFormat` option, but none of those explain their supported formats. So the best option it to explain this sets the `inputFormat` of the editor + see the Editing section under https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-types#datetime-date-and-time

damyanpetev (approved):

ChronosSF:
@ddaribo , please check the migration tests failures

PR #14477, opened 2024-07-04
Allow negative values in filter input

### LINKED ISSUE #14086: Grid: minus sign disappears when filtering with a negative number
## Description  
When trying to filter for a negative number then the minus sign disappears.

 * igniteui-angular version: 17.1.7

## Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/filtering#angular-grid-filtering-example
2. Filter Unit Price with a negative nuzmber, e.g. -1

## Result  
The minus sign disappears.

## Expected result  
The minus sign should not disappear.

## Attachments  
2024-04-09_14h34_25

MonikaKirkova:
Closes #14086 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.html
```diff
@@ -27,7 +27,7 @@
             autocomplete="off"
             [value]="value"
             (input)="onInput($event)"
-            [type]="type"
+            type="text"
```
gedinakova, line 30: This is a really major change in the filtering UI which may cause other problems. The number input actually allows typing - & + signs, so the problem seems to be in the filtering logic itself. I believe the problem is in the IgxGridFilteringRowComponent's "onInput event handler. When the target's value is just "-" (or + actually) it is resolved as NaN, hence the behavior you are observing. I sugges we account for this specific scenario by adding something similar to:
public onInput(eventArgs) {
.....
        const target = eventArgs.target;
        if (!target.value && this.column.dataType === GridColumnDataType.Number) {
            return;
        }
        if (this.column.dataType === GridColumnDataType.DateTime) {
          ....

gedinakova (changes requested):

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.ts
```diff
@@ -98,6 +98,9 @@ export class IgxGridFilteringRowComponent implements AfterViewInit, OnDestroy {
                 return;
             }
         } else {
+            if (val === '') {
```
gedinakova, line 101: Doesn't this if eliminate the need of the newly added check in onInput?

ChronosSF:
Closing this as it will duplicate with the mass merges.

PR #14482, opened 2024-07-05
Allow negative values in filter input - 17.2.x

### LINKED ISSUE #14086: Grid: minus sign disappears when filtering with a negative number

MonikaKirkova:
Closes #14086 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


gedinakova:
@MonikaKirkova We still have some issues when the -, + and . characters are typed in the editor.

1. Run Dev Demos & go to Grid Filtering
2. Enable filter row
3. Type in "1.2" in Employees
4. Hit backspace to delete the 2
5. Note where the caret is placed 
Result: The caret moves to the start although the 1 is still there

1. Run Dev Demos & go to Grid Filtering
2. Enable filter row
3. Attempt to type "+12" in Employees
Result: After typing the '1' the caret moves to the start and the value becomes 21 (without the +)

1. Run Dev Demos & go to Grid Filtering
2. Enable filter row
3. Type "-5" in Employees
4. Hit Backspace
Result: The whole value is cleared, not the 5 only.

We should try stop the caret from jumping to the start. We should also preserve the +-. until the value is actually committed.

gedinakova (approved):
Approving despite the scenarios from my previous comment.
Removing the minus is indeed not such a big problem and it really is not an actual value, so let's leave it as is.
As for the behavior when a floating value is entered, the issue with deleting the decimal part seems to be present for non-negative values as well, so we should probably log and review it separately.

PR #14498, opened 2024-07-09
feat(IgxIconService): POC - Add broadcast channel to sync icons betwe…

MayaKirova:
…en angular and wc.

Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -90,6 +91,7 @@ export class IgxIconService {
         }
     }
 
+
```
damyanpetev, line 94: Would appreciate it greatly if those are cleaned out so no changes are made to this file without need :)

projects/igniteui-angular/src/lib/icon/icon.broadcast.service.ts
```diff
@@ -0,0 +1,79 @@
+import { Injectable, Optional } from '@angular/core';
```
damyanpetev, line 1: Should this be in the Elements project source?
Or did we consider the option for this to be available for Angular apps, though with the current functionality I'm not sure that'd be of much use since it's not two-way yet and the registries parity is also not 1:1 yet.

damyanpetev (commented):
Can consider re-targeting to 18.1.x so this can be used with the next xplat releases, especially if we move the service to elements and it doesn't affect the main Angular project at all.

projects/igniteui-angular/src/lib/icon/icon.broadcast.service.ts
MayaKirova: I was thinking of making it optional, so that if someone wants to inject it for some reason into an angular project they could. But we could just have it in elements if that makes more sense.

projects/igniteui-angular-elements/src/app/app.module.ts
```diff
@@ -31,20 +31,22 @@ import { registerConfig } from "../analyzer/elements.config";
...
-    { provide: ELEMENTS_TOKEN, useValue: true }
+    { provide: ELEMENTS_TOKEN, useValue: true },
+    IgxIconBroadcastService
   ],
 //   bootstrap: []
 })
 export class AppModule {
 
-  constructor(private injector: Injector) {}
+  constructor(private injector: Injector, private _iconBroadcast: IgxIconBroadcastService) {}
```
damyanpetev, line 49: Okay, kind of better we're keeping this in Elements for now.
A bit awkward in terms of usability as it stands having to inject just to trigger the functionality. So it either doesn't need to go through the DI system or perhaps we should consider some API on it like `start` & `stop` basically - both to make sense as usage and to have a cleanup mechanism.
Alternatively, since this isn't root-provided it should be able to take advantage of `ngOnDestroy` for cleanup.
Both of those things technically don't matter much for Elements, since this module will never destroy until the entre app does, but still should be noted.

projects/igniteui-angular-elements/src/lib/icon.broadcast.service.ts
```diff
@@ -0,0 +1,82 @@
...
+            this.iconBroadcastChannel.onmessage = (event) => {
+                const message = event.data as BroadcastIconsChangeMessage;
+                if (message.actionType === ActionType.SyncState ||
+                     message.actionType === ActionType.RegisterIcon) {
+                    this.updateIconsFromCollection(message.collections);
+                }
+
+                if (message.actionType === ActionType.SyncState ||
+                     message.actionType === ActionType.UpdateIconReference) {
+                    this.updateRefsFromCollection(message.references);
```
damyanpetev, line 45: Assuming it's the responsibility of the other side to only emit the icons that do need changing for `UpdateIconReference`?

damyanpetev (approved):
Besides the usage/lifecycle comment that's not important right now, the rest LGTM.

projects/igniteui-angular-elements/src/lib/icon.broadcast.service.ts
MayaKirova: Yes. The other side sends the state of the collections and this side updates its own state.

damyanpetev: Hehe, then it's not a "yes" answer :) The question was if "the other side to **only** emits the icons that do **need changing**".
If the other side sends the entire state of the collections, **every** API for change on that side will also cause the Angular side to re-register all previous icons for example, which is at least redundant if it doesn't affect performance much.

damyanpetev: Nvm, checked the implementation, the other side creates a fresh collection for each change, so it is sending just the modified refs. No need to break-check me like that 😄

PR #817, opened 2018-03-22
igxGrid - deferred column resizing feature

SAndreeva:
## Spec

Initial implementation of deferred column resizing feature, based on Wiki spec:
https://github.com/IgniteUI/igniteui-angular/wiki/Column-Resizing

## Tasks
- [x] Implementation
- [x] Testing
- [x] Documentation
- [x] Samples

Closes #797

PR #14513, opened 2024-07-10
docs(column): mark mistakenly left exposed getters as hidden

damyanpetev:
@hanastasov @dkamburov Might need some help from your teams whether some of the exposed API makes sense or has seen usage in custom samples/cases.
So far I've made the following internal:
- `filterCell`, `headerCell` and `headerGroup` - I can find no usage anywhere outside internals and see little reason - those expose hidden internal structure components (the `IgxGridFilteringCellComponent`  is not even in the public API, though IMO the rest shouldn't be either).
- `defaultMinWidth` - it's hooked into the public `minWidth` input already and ideally will be removed and moved entirely to CSS at some point. Also not even in the `ColumnType` interface
- `gridRowSpan`/`gridColumnSpan` - those are shortcuts to very simple calculations with the `rowStart/End` & `colStart/End` public inputs, don't think they were meant to be public at all.

Candidates for the chopping block that I still can't find any public use, but at least sound possibly useful:
- `cells` - Only used in tests, added likely for internal resize handling in #817 and that logic no longer exists; Also not on the `ColumnType` interface and it's a convenience path alternative instead of going through row.cells and picking one. There's also an internal `_cells` that does similar thing, so someone really wanted this public, just not sure of any use case for it.
- `columnLayoutChild` - The only reason I can think of for this is to make sense of the values or lack there of behind `rowStart/End` & `colStart/End` which is mostly how we use it as well. Suppose it could be a useful shortcut for that check
- `topLevelParent` - since we're maintaining children and parent API, again only as a convenience I suppose.
- `filteringExpressionsTree` - A shortcut (again) to the grid's property scoped to the column with `find()`. We haven't found need for either in the samples yet and the specs remain silent as with the rest, so not sure on it.

Would appreciate input.

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
+++ b/projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
@@ -311,17 +311,9 @@ export interface ColumnType extends FieldType {
     get childColumns(): ColumnType[];
     /** @hidden @internal */
     allChildren: ColumnType[];
-    /**
-     * The header group component associated with this column.
-     * Could be of any type
-    */
-    // TYPE
+    /** @hidden @internal */
     headerGroup: any;
-    /**
-     * The header cell component associated with this column.
-     * Could be of any type
-     */
-    // TYPE
+    /** @hidden @internal */
     headerCell: any;
     validators: any[];
 
@@ -418,6 +410,7 @@ export interface ColumnType extends FieldType {
      * If the value is true, the result will not depend on the case (example: `E` will match `e`)
      */
     sortingIgnoreCase: boolean;
+    /** @hidden @internal */
     filterCell: any;
     filteringIgnoreCase: boolean;
     /**
@@ -447,7 +440,9 @@ export interface ColumnType extends FieldType {
     rowEnd: number;
     colStart: number;
     colEnd: number;
+    /** @hidden @internal */
     gridRowSpan: number;
+    /** @hidden @internal */
     gridColumnSpan: number;
     columnLayoutChild: boolean;
     width: string;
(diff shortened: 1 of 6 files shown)


damyanpetev:
Update: Decided to go ahead and also hide the `cells` collection - it's way too odd, considering the grid is usually navigated by rows and cells primarily and a good portion of the `CellType` API reflects back column/row state or the grid itself has methods to do the same thing, so I struggle to see a reason to have that collection exposed on the column.

hanastasov:
> Update: Decided to go ahead and also hide the `cells` collection - it's way too odd, considering the grid is usually navigated by rows and cells primarily and a good portion of the `CellType` API reflects back column/row state or the grid itself has methods to do the same thing, so I struggle to see a reason to have that collection exposed on the column.
> 
> Unless there's any feedback, I'll leave it at that.

Agreed with you on DM. Let's remove `cells` as well.

For all the above, I would keep `topLevelParent`

Havent looked for `columnLayoutChild`, but maybe someone for Grinders could tell for it.

hanastasov (approved):

projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -447,7 +440,9 @@ export interface ColumnType extends FieldType {
     rowEnd: number;
     colStart: number;
     colEnd: number;
+    /** @hidden @internal */
     gridRowSpan: number;
+    /** @hidden @internal */
     gridColumnSpan: number;
```
ddincheva, line 446: these are also present in the cell component... can you please check if there is any necessity for these getters there

damyanpetev: Only see them on `IgxGridCellComponent` which should not be exposed in any public API, the `CellType` interface doesn't contain those thankfully.

dkamburov (approved):

PR #14525, opened 2024-07-16
fix(advance-filtering): handle error when columns are dynamically changed - master

### LINKED ISSUE #14314: Grid Advanced Filtering throws if grid has recreated its columns colection
## Steps to reproduce  

1. Add grid with Advanced filtering (you can use the `GridFilteringComponent` in dev demos
2. Through Advanced filtering dialog, filter some column, lets say `CompanyName` 
3. Change the collection, which the grid uses to create columns based on;

this.columns = this.columns2; // `columns2` has different columns than `columns`

OR change the grid data and have tje grid auto generate its columns:

```
        this.grid1.shouldGenerate = true;
        this.data = this.data3;
```

4. Open Advanced Filtering dialog

## Result  
Error is thrown and Advanced filtering dialog is not opened
image

georgianastasov:
Closes #14314   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
```diff
@@ -50,6 +50,30 @@ export class IgxGridToolbarAdvancedFilteringComponent implements AfterViewInit {
         this.grid?.advancedFilteringExpressionsTreeChange.subscribe(filteringTree => {
             this.numberOfColumns = this.extractUniqueFieldNamesFromFilterTree(filteringTree).length;
         });
+        this.grid?.filteringExpressionsTreeChange.subscribe(() => {
+            setTimeout(() => {
+                if (this.grid.advancedFilteringExpressionsTree) {
+                    const dataKeys = Object.keys(this.grid.data[0]);
```
ddincheva, line 56: why is this code necessary to be wrapped in setTimeout and even have another one inside this logic on row 69?

georgianastasov: I used a custom sample to check different scenarios, and the setTimeout functions were part of testing one specific scenario. They were inadvertently left in the code from the large amount of testing done. Thank you for pointing this out;.I’ll remove them immediately.

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
```diff
@@ -50,6 +50,26 @@ export class IgxGridToolbarAdvancedFilteringComponent implements AfterViewInit {
         this.grid?.advancedFilteringExpressionsTreeChange.subscribe(filteringTree => {
             this.numberOfColumns = this.extractUniqueFieldNamesFromFilterTree(filteringTree).length;
         });
+        this.grid?.filteringExpressionsTreeChange.subscribe(() => {
```
ddincheva, line 53: Advanced filtering is using advancedFilteringExpressionsTree collection why is necessary to listen on filteringExpressionsTreeChange here?!

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
```diff
@@ -50,6 +50,26 @@ export class IgxGridToolbarAdvancedFilteringComponent implements AfterViewInit {
...
+                        return this.checkFilteringOperands(operand, dataKeys);
+                    } else {
+                        return dataKeys.includes(operand.fieldName);
+                    }
+                });
+
+                if (!allOperandsPresent) {
+                    this.grid.advancedFilteringExpressionsTree = this.grid.filteringExpressionsTree;
+                    this.numberOfColumns = 0;
+                }
```
ddincheva, line 70: advanced filtering and excel style filtering are working independently in the grid -- so setting esf collection to advanced filtering collection causes some strange behavior .. please check the following recordings:
advFilteringAndESF

ddincheva (changes requested):

ddincheva:
When we have advanced filtering applied and change the grid data the grid is left in some incorrect state: the advanced filter counter is not reset but when you open the Advance filtering dialog you can see no filtering expression applied and also the grid's data is not displayed /cause the data is still filtered out/;
advFilteringAndRegenerating

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
georgianastasov: I listen for a change in the filtering expression tree `filteringExpressionsTreeChange` because when we change the data by default the filters are cleared and `filteringExpressionsTreeChange` is executed while `advancedFilteringExpressionsTreeChange` is not executed, and it is executed only if we clear the filters from the **'Clear Filters'** button in the advanced filtering dialog.

That's why in this scenario when the data changes, apart from this PR fixing the issue about the error that appears in the console when opening the advance filter dialog, I also handled that to clear the advance filtering expression tree and to reset the number that shows how many filters are attached. 

That's why when changing the data and clearing the filters, I listen on `filteringExpressionsTreeChange` event to see if it is necessary to clear the advanced filter and also to clear the number, because `advancedFilteringExpressionsTreeChange` is not executed in this case.

**Not sure if this should be handled in this PR or if a new issue should be logged for this case?!**

projects/igniteui-angular/src/lib/grids/toolbar/grid-toolbar-advanced-filtering.component.ts
georgianastasov: On my end, everything works as expected as you can see from the next gif.

About setting esf collection (`filteringExpressionsTree`) to advanced filtering collection (`advancedFilteringExpressionsTree`).. setting this collection happens only when we change the grid data and therefore the filter expression tree is cleared and contains nothing.

{
 fieldName: undefined
 "filteringOperands": [],
 "operator": 0,
 "type": 0
}

Therefore I also clear the advance filtering tree by setting it to the empty default filtering expression tree which is the same thing when it is cleared. That's why here I set `numberOfColumns` to be equal to 0 because we cleared the advanced filters by setting it to an empty filtering tree.

I'm not sure if this is correct though and if something else should be thought of like null undefined or something else.

georgianastasov:
> When we have advanced filtering applied and change the grid data the grid is left in some incorrect state: the advanced filter counter is not reset but when you open the Advance filtering dialog you can see no filtering expression applied and also the grid's data is not displayed /cause the data is still filtered out/; advFilteringAndRegenerating 

Again on my end, everything works as expected as you can see from the next gif!

georgianastasov:
@ddincheva after a discussion with @gedinakova, we have decided that these PRs will focus solely on fixing the specific issue where error is thrown and advanced filtering dialog is not opened after the grid dataset is changed. No other changes or resets to the filters will be made in these PRs.

After these PRs are merged, we will create a new task to address the user experience (UI/UX) aspect. This task will focus on how to effectively notify users about remaining applied filters in the advance filtering dialog after a dataset change, considering the best practices in UI/UX design.

ChronosSF:
Closing this as it will duplicate with the mass merges.

PR #14598, opened 2024-07-29
refactor(icon-service): update icon reference names and registration

simeonoff:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/icon/icon.references.ts
```diff
@@ -0,0 +1,586 @@
...
+        family: 'material',
+    },
+});
+addIcon('search', {
+    default: {
+        name: 'search',
+        family: 'material',
+    },
+});
+addIcon('sum', {
```
damyanpetev, line 371: AFAICT, this is only used for the Pivot and in both places I found it it was used to open dropdown to switch functions, so the alias name `functions` was better IMO.

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-pinning-actions.component.ts
```diff
@@ -141,23 +141,6 @@ export class IgxGridPinningActionsComponent extends IgxGridActionsBaseDirective
             this.iconService.addSvgIconFromText(unpinLeft.name, unpinLeft.value, 'imx-icons', true);
             this.iconService.addSvgIconFromText(jumpDown.name, jumpDown.value, 'imx-icons', true);
             this.iconService.addSvgIconFromText(jumpUp.name, jumpDown.value, 'imx-icons', true);
```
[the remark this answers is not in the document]
damyanpetev: @ChronosSF I'll just be committing this one as a fix

projects/igniteui-angular/src/lib/date-picker/date-picker.component.html
```diff
@@ -20,7 +20,7 @@
         [attr.aria-labelledby]="label?.id" aria-haspopup="dialog" aria-autocomplete="none" role="combobox">
 
     <igx-suffix *ngIf="!clearComponents.length && value" (click)="clear()">
-        <igx-icon family="default" name="clear"></igx-icon>
+        <igx-icon family="default" name="input_clear"></igx-icon>
```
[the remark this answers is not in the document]
damyanpetev: Managed to dig up this from the [Simple Combo spec](https://github.com/IgniteUI/igniteui-angular/wiki/Simple-Combo#3-functionality): https://xd.adobe.com/view/f0a2ac12-9c3d-48bd-89c8-b9f1b0ee6a25-a5f1/ so um maybe it's the Combos that are not consistent with the rest?

projects/igniteui-angular/src/lib/icon/icon.references.ts
```diff
@@ -0,0 +1,586 @@
...
+ * Likewise, modifying existing references should be handled with caution as many component in the framework already
+ * share icons that have equivellent semantic meaning. For example, the `Paginator`, `Grid Filtering Row`,
+ * and `Tabs` components in Ignite UI for Angular all use the `prev` and `next` icons as for navigating between pages
+ * or lists of items. Changing the underlying target for those icons should be done in a way that suits all components.
+ *
+ * Keep in mind that icon aliases and their underlying names are shared between Ignite UI component frameworks
+ * and changing an alias name here should be reflected in the other frameworks as well.
+ *
+ * To get aquainted with which component uses what icon, please make sure to read the
+ * [docs](https://infragistics.com/products/ignite-ui-angular/Angular/components/icon-service#internal-usage).
```
damyanpetev, line 25: Just typos, otherwise 👌. Thanks for adding this bit.
```suggestion
 * Before you add/modify an icon reference, please think about the semantics of the icon you are adding/modifying.
 *
 * Icon aliases have sematic meaning depending on the context in which they are used.
 * For instance, if your component handles toggling between expanded and collapsed states,
 * you may want to use the already existing `expand` and `collapse` aliases that point to
 * the `expand_more` and `expand_less` icons in the material font set.
 *
 * It may so happen, however, that the design of your component requires you to use the `chevron_right` for the
 * expand icon and the `expand_more` for the collapse icon. In this case the `tree_expand` and `tree_collapse` aliases
 * would be appropriate.
 * This distinction is important when choosing which icon to use for your component as it will have an impact
 * when a user decides to rewire the `expand`/`collapse` icons to some other icons.
 *
 * Likewise, modifying existing references should be handled with caution as many component in the framework already
 * share icons that have equivalent semantic meaning. For example, the `Paginator`, `Grid Filtering Row`,
 * and `Tabs` components in Ignite UI for Angular all use the `prev` and `next` icons as for navigating between pages
 * or lists of items. Changing the underlying target for those icons should be done in a way that suits all components.
 *
 * Keep in mind that icon aliases and their underlying names are shared between Ignite UI component frameworks
 * and changing an alias name here should be reflected in the other frameworks as well.
 *
 * To get acquainted with which component uses what icon, please make sure to read the
```

projects/igniteui-angular/src/lib/icon/icon.references.ts
```diff
@@ -0,0 +1,586 @@
+/** READ BEFORE YOU MODIFY THIS FILE
+ *
+ * Before you add/modify an icon reference, please think about the sematics of the icon you are adding/modifying.
```
damyanpetev, line 3:
```suggestion
 * Before you add/modify an icon reference, please think about the semantics of the icon you are adding/modifying.
```

projects/igniteui-angular/src/lib/icon/icon.references.ts
```diff
@@ -0,0 +1,586 @@
...
+ * the `expand_more` and `expand_less` icons in the material font set.
+ *
+ * It may so happen, however, that the design of your component requires you to use the `chevron_right` for the
+ * expand icon and the `expand_more` for the collapse icon. In this case the `tree_expand` and `tree_collapse` aliases
+ * would be appropriate.
+ * This distinction is important when choosing which icon to use for your component as it will have an impact
+ * when a user decides to rewire the `expand`/`collapse` icons to some other icons.
+ *
+ * Likewise, modifying existing references should be handled with caution as many component in the framework already
+ * share icons that have equivellent semantic meaning. For example, the `Paginator`, `Grid Filtering Row`,
```
damyanpetev, line 17:
```suggestion
 * share icons that have equivalent semantic meaning. For example, the `Paginator`, `Grid Filtering Row`,
```

projects/igniteui-angular/src/lib/icon/icon.references.ts
```diff
@@ -0,0 +1,586 @@
...
+ * and `Tabs` components in Ignite UI for Angular all use the `prev` and `next` icons as for navigating between pages
+ * or lists of items. Changing the underlying target for those icons should be done in a way that suits all components.
+ *
+ * Keep in mind that icon aliases and their underlying names are shared between Ignite UI component frameworks
+ * and changing an alias name here should be reflected in the other frameworks as well.
+ *
+ * To get aquainted with which component uses what icon, please make sure to read the
```
damyanpetev, line 24:
```suggestion
 * To get acquainted with which component uses what icon, please make sure to read the
```

damyanpetev (commented):
Just typos, otherwise 👌. Thanks for adding this bit.

projects/igniteui-angular/src/lib/icon/types.ts
```diff
@@ -0,0 +1,37 @@
...
+    prefix?: string;
+}
+
+export interface IconFamily {
+    name: string;
+    meta: FamilyMeta;
+}
+
+// Exported public types
+export interface IconMeta {
```
damyanpetev, line 33: Might want to throw `export { IconMeta } from './types';` in the `public_api.ts` for that to be true ;)

projects/igniteui-angular/src/lib/icon/types.ts
```diff
@@ -0,0 +1,37 @@
...
+export type IconType = "svg" | "font" | "liga";
+
+export type IconReference = IconMeta & FamilyMeta;
+
+export type MetaReference = {
+  alias: IconMeta;
+  target: Map<IconThemeKey, IconMeta>;
+};
+
+export interface FamilyMeta {
```
damyanpetev, line 21: And I think this one should also be moved in the public API part too since it's used in the service `setFamily` method.
And with it `IconType` apparently.

projects/igniteui-angular/src/lib/icon/types.ts
```diff
@@ -0,0 +1,37 @@
...
+    className: string;
+    type: IconType;
+    prefix?: string;
+}
+
+export interface IconFamily {
```
damyanpetev, line 27: And `defaultFamily` on the service makes this public API too. Unless we hide that property and consider it internal?

projects/igniteui-angular/src/lib/icon/types.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable @typescript-eslint/consistent-type-definitions */
+import { Theme } from 'igniteui-webcomponents/theming/types.js';
```
damyanpetev, line 2: Um, need to check if this gets compiled away entirely in the declarations, otherwise it'll make `igniteui-webcomponents` a direct dependency of the library which it isn't.
Also also, there's an identical `IgxTheme` in the theme service expored already which can be used instead, no?

projects/igniteui-angular/src/lib/icon/types.ts
damyanpetev: And `IconReference` because of `getIconRef`, basically all the old exports from `icon.service.ts` should be exported from the public API

projects/igniteui-angular/src/lib/date-picker/date-picker.component.html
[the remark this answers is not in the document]
simeonoff: I will consult with Svilen on this one, but I'd also like to keep the clear icon consistent across input, combo, select, and the various pickers we have.

projects/igniteui-angular/src/lib/grids/cell.component.html
```diff
@@ -46,7 +46,7 @@
         }}</div>
     <igx-icon
         family="default"
-        [name]="value ? 'check' : 'close'"
+        [name]="value ? 'confirm' : 'close'"
```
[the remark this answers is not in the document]
simeonoff: I couldn't come up with a scenario in which changing the confirm in the other places will affect the cell in any way, I believe semantically they are all the same action in all components, including the cell.

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.html
```diff
@@ -215,7 +215,7 @@
                         (click)="toggleOperatorsDropDown($event, i)"
                         [igxDropDownItemNavigation]="operators"
                         [style.--ig-size]="filteringElementsSize">
-                        <igx-icon family="default" name="expand"></igx-icon>
+                        <igx-icon family="default" name="expand_more"></igx-icon>
```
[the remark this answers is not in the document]
simeonoff: > Would this also make sense w/ just `expand`? Technically even `input_expand` if this https://getbootstrap.com/docs/4.0/components/dropdowns/#single-button-dropdowns is anything to go by and we're okay with a minor icon change other themes. From what I can see the `expand_more` is pretty much dedicated to this button and we can drop it if we assign it one of the other two aliases.

I don't agree with input_expand as it changes from theme to theme, and is semantically different. The expand is closer in meaning, but I believe choosing the `arrow_drop_down` icon was intentional, thus I kept it as a separate reference used only in a few awkward places.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-data-selector.component.html
```diff
@@ -128,15 +128,15 @@ <h6 class="igx-pivot-data-selector__header-title">
...
+                                name="sum"
                                 class="igx-pivot-data-selector__action-summary"
                                 *ngIf="panel.type === null"
                                 (click)="onSummaryClick($event, item, dropdown)"
                                 [igxDropDownItemNavigation]="dropdown">
                             </igx-icon>
                             <igx-icon
                                 family="default"
-                                name="drag_handle"
+                                name="drag_indicator"
```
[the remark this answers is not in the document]
simeonoff: I will consult with Svilen about this one. I really want to reduce the amount of ambiguous icons sprinkled all over. Technically both icons are used as drag handles.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-data-selector.component.html
```diff
@@ -173,9 +173,9 @@ <h6 class="igx-pivot-data-selector__header-title">
         [class.igx-pivot-data-selector__item-ghost--no-drop]="!dropAllowed"
     >
         <div class="igx-pivot-data-selector__item-ghost-text">
-            <igx-icon family="default" name="unfold_more"></igx-icon>
+            <igx-icon family="default" name="import_export"></igx-icon>
```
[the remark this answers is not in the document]
simeonoff: Yes, it does. To be honest this icon makes a ton more sense in this scenario than the `unfold_more` icon as you are exporting an item from one area and importing it into another vertical area either above or below. I will consult with Svilen about this change as well. To me, the icon that should be different is the one used for the export button.

projects/igniteui-angular/src/lib/grids/filtering/base/grid-filtering-row.component.html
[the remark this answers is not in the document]
damyanpetev: They do indeed - IIRC Fluent Design consistently uses a chevron for both select and split/menu buttons, Material seems to be using this type of arrow for selects & buttons pretty consistently too, though we've strayed from that. And yes, I know that's opening another can of worms and can be dealt with separately.

projects/igniteui-angular/src/lib/date-picker/date-picker.component.html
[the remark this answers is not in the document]
simeonoff: So we agreed to make changes to the date/time picker in the UI kit. From now on, they will use the same icon as the combo component.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-data-selector.component.html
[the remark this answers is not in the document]
simeonoff: Svilen and I agreed that the drag_indicator will be used in both places.

projects/igniteui-angular/src/lib/icon/icon.references.ts
```diff
@@ -0,0 +1,598 @@
...
+
+    iconReferences.push(icon as MetaReference);
+}
+
+addIcon('more_vert', {
+    default: {
+        name: 'more_vert',
+        family: 'material',
+    }
+});
```
damyanpetev, line 59: Unfortunately this doesn't tree-shake very well 😶
I guess understandable - top level function calls don't are assumed to be a possible side effect and the tree-shake bails.
And those can't really be marked at pure calls either (will permanently be dropped) since they really have a side effect we need when the references are in use + it's mildly infuriating to have to mark each call.

Two alternatives I can think of
- either leave this as a straight up object and give up and give up on the Map convenience (the "default" family assign can live in the service)
- Move the loading of icons in a separate function so it's call can be marked as pure, something like:
image
  That's the one setup that I've landed on that does the job. And no, spreading the calls in the var array directly doesn't imply the calls are pure and doesn't allow shaking.
  

PS: Working off of https://github.com/javascript-compiler-hints/compiler-notations-spec/blob/main/pure-notation-spec.md that seems to have more support than the side-effect flag.

PR #14634, opened 2024-08-15
fix(toggle): Change the way how the dropdown passes size to toggle

### LINKED ISSUE #14618: ExpressionChangedAfterItHasBeenCheckedError when setting the size of the action strip
## Description  
When the size of the grid with action strip is set through --ig-size, ExpressionChangedAfterItHasBeenCheckedError is thrown on hover.

 * igniteui-angular version: 18.0
 * browser: 

## Steps to reproduce  

1. Open [this sample](https://stackblitz.com/edit/ccfbef?file=src%2Fapp%2Fgrid%2Fgrid-action-strip%2Fgrid-action-strip-sample.html)
2. Hover on some row

## Result  
ExpressionChangedAfterItHasBeenCheckedError is thrown.

## Expected result  
There should be no error.

mddragnev:
Closes #14618   
Removing the size input because when action strip is creating/destroying on mouse move between the rows it creates and destroys a dropdown within the actionstrip which causes the error.
 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -680,6 +681,11 @@ export class IgxOverlayService implements OnDestroy {
         if (info.initialSize.width !== 0) {
             info.elementRef.nativeElement.parentElement.style.width = info.initialSize.width + 'px';
         }
+
+        // set content div size
```
wnvko, line 685: too late to set the `size` here. If we attach a component ref, e.g. `igxCalendar` in `igxDatePicker` on line 677 we will get the initial size without taking into account the its potential `--ig-size`. Then here we are setting the `--ig-size` and initial size may end up with wrong value. Move the setting of `--ig-size` in the very beginning of `updateSize` method.

wnvko (commented):
Please name your PRs with some descriptive names. Having something like this is really far away from helpfull:
image

projects/igniteui-angular/src/lib/drop-down/drop-down.component.html
```diff
@@ -1,6 +1,5 @@
 <div class="igx-drop-down__list" [style.width]="width"
     igxToggle
-    [style.--ig-size]="dropDownSize"
```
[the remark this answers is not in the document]
wnvko: the idea was to delete `dropDownSize` getter from `IgxDropDownBaseDirective`.

wnvko (commented):
please add also test for this change

projects/igniteui-angular/src/lib/drop-down/drop-down.component.html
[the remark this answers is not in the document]
mddragnev: I have already deleted it from the IgxDropDownBaseDirective.

mddragnev:
@wnvko  There are couple of tests that use that inline style to see whether the items in the dropdown are correctly sized so I dont think it is neccessary to write another one which is just checking whether the list has size when I did not only delete it but just changed that the sizing is set to its parent container. I would be testing css inheritance this way.

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -977,4 +983,14 @@ export class IgxOverlayService implements OnDestroy {
             info.closeAnimationPlayer.finish();
         }
     }
+
+    private addComponentSize(info: OverlayInfo) {
+        if (info?.elementRef?.nativeElement && info.elementRef.nativeElement instanceof Element) {
```
damyanpetev, line 988: The info really always exists looking from where this is called, so no need to check and can be shortened to:
```suggestion
        if (info.elementRef?.nativeElement instanceof Element) {
```

projects/igniteui-angular/src/lib/services/overlay/overlay.ts
```diff
@@ -977,4 +983,14 @@ export class IgxOverlayService implements OnDestroy {
...
+            const styles = window.getComputedStyle(info.elementRef.nativeElement);
```
damyanpetev, line 989: Perhaps unlikely scenario for attach to run on some component's init in SSR, but not entirely impossible. Even for consistency, we should make it a habit to avoid directly accessing `window` properties.
`globalThis.window?.getComputedStyle` is one way used widely (there's even `getComponentSize` in utils that does the next line like that, tho only used in.. tests for some reason. Alt you can prob also use the injected doc token (`this._document.defaultView`) though need to see if a check is required for the method existence in domino

damyanpetev (commented):
The only test I think we might want to add for this is an explicit cover of the overlay functionality that moves the computed size var (similar to the width/height, guessing there is one). There might be some that indirectly go though the logic, but I' not sure they guarantee this works

damyanpetev (approved):

wnvko (approved):

PR #14647, opened 2024-08-19
feat(query-builder): support for nested queries and other improvements

### LINKED ISSUE #14642: Query Builder: Update theme to match latest design

### LINKED ISSUE #14979: Query Builder component update

### LINKED ISSUE #15389: Chip: long suffix and prefix wrap on 2 lines
If the chip suffix or prefix contains long text, it wraps onto two lines. Since the chip has a fixed height, this causes a broken layout. A better approach is to apply an ellipsis to maintain visual consistency.

Image

_Originally posted by @simeonoff in https://github.com/IgniteUI/igniteui-angular/pull/14647#discussion_r1965060459_

igdmdimitrov:
Closes #14642   
Closes #14979
Closes #15389 

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


gedinakova:
1. Run QB dev demo
	2. Click on the group line in the inner query

Result: The dialog to change the group operator is not shown. Only the group line style changes to selected. 
Note: The dialog is shown when the query is in edit mode.

This may be by design but I still think we should probably discuss this UX.

gedinakova:
1. Run QB Dev Demo
	2. Hover over the first condition of the inner query
	3. Click the "+" button to add a new condition
	4. Click the "+condition", "and group" or "or group" button 

The row to add a condition is shown but the following error is logged in the console (3 times):
TypeError: Cannot read properties of null (reading 'name')
    at IgxQueryBuilderTreeComponent2_ng_template_9_div_0_Template (template.html:99:21)
    at executeTemplate (core.mjs:11621:9)
    at refreshView (core.mjs:13239:13)
    at detectChangesInView (core.mjs:13455:9)
    at detectChangesInViewIfAttached (core.mjs:13415:5)
    at detectChangesInEmbeddedViews (core.mjs:13374:13)
    at refreshView (core.mjs:13267:9)
    at detectChangesInView (core.mjs:13455:9)
    at detectChangesInViewIfAttached (core.mjs:13415:5)
    at detectChangesInEmbeddedViews (core.mjs:13374:13)

gedinakova:
1. Run the dev demo
	2. Reset all filters & start a new group
	3. Select entity & add the first condition
	4. The add a new group, e.g an and group
	5. Add the first condition to it (note how all the buttons to add new filters are inside the group)
	6. Submit the condition - only the condition remains in the group, the buttons seem to be shifted to the outer group
	7. Attempt to add another group

Result: All inputs and buttons in the new group are disabled.
Expected result: The buttons should stay part of the current group and it should be possible to add another condition to it (not having all inputs disabled on step 7)

gedinakova:
1. Run the dev demo
    2. Click to edit the second condition
    3. In the filter dropdown change from "Equals" to "In"
    4. Click the commit icon button

The button is not disabled on switching to In/Not In conditions (as the value the input is). The following error is logged in the console on clicking it:
TypeError: Cannot read properties of undefined (reading 'operator')
    at IgxQueryBuilderTreeComponent2.commitOperandEdit (query-builder-tree.component.ts:618:123)
    at IgxQueryBuilderTreeComponent2_ng_template_9_div_1_Template_button_click_18_listener (template.html:351:21)
    at executeListenerWithErrorHandling (core.mjs:26660:16)

Otixa:
```
- Run the dev demo
- Select both inner conditions
- Create OR group
```
WonkyRegrouping
This seems to create a deadlock where you're unable to modify the inner condition anymore properly.
Additionally, there doesn't seem to be a way to change the logic(AND/OR) of the inner group. Unless I'm mistaken?

teodosiah:
1. Add field of type date to the entity
2. Add 'after' condition to the date type column

An error is thrown:
TypeError: Cannot read properties of undefined (reading 'formatter')
    at IgxQueryBuilderTreeComponent2.getFormatter (query-builder-tree.component.ts:1014:57)
    at IgxQueryBuilderTreeComponent2_ng_template_9_div_0_span_6_ng_container_2_Template (template.html:107:21)
    at executeTemplate (core.mjs:11621:9)
    at refreshView (core.mjs:13239:13)
    at detectChangesInView (core.mjs:13455:9)
    at detectChangesInViewIfAttached (core.mjs:13415:5)
    at detectChangesInEmbeddedViews (core.mjs:13374:13)
    at refreshView (core.mjs:13267:9)
    at detectChangesInView (core.mjs:13455:9)
    at detectChangesInViewIfAttached (core.mjs:13415:5)

teodosiah:
1. Add query builder without initially set expression tree:
` <igx-query-builder #queryBuilder
            [entities]="this.entities"></igx-query-builder>
`

The '+' icons in the add "'and'/'or' group" buttons are not displayed.

_Suggestion: move the registerSVGIcons method to the query-builder.component.ts and call it on initial loading because when there is no expression tree a query-builder-tree is not created and the method is not called_

teodosiah:
1. Add query builder without initially set expression tree
2. Add new group: 
image
3. Commit it:
image
4. Delete the group
5. Click on some of the add new group buttons
The selected entity is the same as the previously deleted:
image

Should we keep the previous selection?

teodosiah:
1. Add query builder without initially set expression tree
2. Click on some of the add new group buttons (all inputs are empty)
3. Click on the entity select

**Result:**
The Confirmation dialog is shown.

Should we open the dialog when initially setting (not changing) the entity?

teodosiah:
1. Add query builder without initially set expression tree
2. Hover the condition action button area
queryBuilderIssue

'close' button should be disabled

teodosiah:
1. Start editing some query
The conditions drop-downs are placed in the igx-query-builder__outlet, however the entity and fields drop-downs are outside:
qb_overlay
This could be the reason for the following scenario:
image

gedinakova:
1. Add an entity without fields
```typescript
{
       name: 'Empty', fields: undefined
}
```
3. Run the dev demo
4. Select the empty entity from the dropdown & confirm

Result: An error is thrown.  (ERROR TypeError: Cannot read properties of undefined (reading 'length'))
Expected result: We should guard the fields as it is possible to get an undefined for instance.

gedinakova:
An empty tooltip is shown on hovering over an unary condition.
image

gedinakova:
@desig9stein 
1. Run the dev demo with indigo theme
2. Expand "Select return fields" combo

Result: The search & select all are not aligned with the actual items.
image

Expected result: All elements should be aligned.

gedinakova:
1. Run the dev demo
2. Clear all conditions
3. Start a new group
4. Add a new condition without a value
5. Click the line group indicator

Result: The changes are discarded.

desig9stein:
> @desig9stein
> 
> 1. Run the dev demo with bootstrap or fluent theme
> 2. Put one of the conditions in edit mode
> 3. Note the commit & discard buttons alignment
> 
> Result: image image
> 
> Expected result: The buttons should be aligned with the inputs.

@AnjiManova Since this is by design, I am unsure how to proceed. the problem is that those are huge icon buttons, maybe we can use just icons instead?

AnjiManova:
> > @desig9stein
> > 
> > 1. Run the dev demo with bootstrap or fluent theme
> > 2. Put one of the conditions in edit mode
> > 3. Note the commit & discard buttons alignment
> > 
> > Result: image image
> > Expected result: The buttons should be aligned with the inputs.
> 
> @AnjiManova Since this is by design, I am unsure how to proceed. The problem is that those are icon buttons, and they are big, even in small variant.
> 
> Maybe we can use just icons instead? We already have just icons here image

Can we just try to apply align-item: flex-end and align the buttons to the bottom instead of changing the used components? 
tmp_aeda9c61-73d7-4571-92e8-c1977b10acbb

desig9stein:
Won't work l, since the buttons are very big and there are labels on top of the inputs

desig9stein:
> > > @desig9stein
> > > 
> > > 1. Run the dev demo with bootstrap or fluent theme
> > > 2. Put one of the conditions in edit mode
> > > 3. Note the commit & discard buttons alignment
> > > 
> > > Result: image image
> > > Expected result: The buttons should be aligned with the inputs.
> > 
> > 
> > @AnjiManova Since this is by design, I am unsure how to proceed. The problem is that those are icon buttons, and they are big, even in small variant.
> > Maybe we can use just icons instead? We already have just icons here image
> 
> Can we just try to apply align-item: flex-end and align the buttons to the bottom instead of changing the used components? tmp_aeda9c61-73d7-4571-92e8-c1977b10acbb

Won't work l, since the buttons are very big and there are labels on top of the inputs
image

AnjiManova:
> > > > @desig9stein
> > > > 
> > > > 1. Run the dev demo with bootstrap or fluent theme
> > > > 2. Put one of the conditions in edit mode
> > > > 3. Note the commit & discard buttons alignment
> > > > 
> > > > Result: image image
> > > > Expected result: The buttons should be aligned with the inputs.
> > > 
> > > 
> > > @AnjiManova Since this is by design, I am unsure how to proceed. The problem is that those are icon buttons, and they are big, even in small variant.
> > > Maybe we can use just icons instead? We already have just icons here image
> > 
> > 
> > Can we just try to apply align-item: flex-end and align the buttons to the bottom instead of changing the used components? tmp_aeda9c61-73d7-4571-92e8-c1977b10acbb
> 
> Won't work, since the buttons are very big and there are labels on top of the inputs. the only way to align the buttons with the inputs is to use icons, I can try to make the buttons absolute and play with the position, but i don't like this idea image
> 
> absolute position image
> 
> Or maybe we can override the size of the buttons to match the size of the inputs in the context of the query builder

If it's up to me I will not override the size of the buttons. The problem is only in the Large size in Bootstrap and in the Small size in Fluent and it is because of the fixed size of the Input Group components (Small size) in the Advanced Filtering and the Query Builder. We could consider future improvment here.

About the buttons misalignment we will discuss your concerns with the team and we will come back to you, but for now it can stay as it is.

In addition we added visual examples in all sizes for the Query Builder for all themes in the Figma file.

ivanvpetrov:
1. Open Query Builder demo page
2. Select both available conditions
3. Group the as new And/Or group
4. Repeat 2 and 3 again
5. The Select expression disappears - issue
6. Now start creating new select expression
7. Cancel the creation.
8. The two conditions disappear - issue

https://github.com/user-attachments/assets/6e5377fb-3a19-4432-95d4-b8e0287c2854

gedinakova:
1. Run the dev demo
2. Put the inner query in edit mode
3. Click the Or Group button to add a group at the bottom
4. Enter a valid condition and submit it
5. Attempt to add a new condition to the newly added group

Result: The + button does not show up on the group.

Expected: One should be allowed to add new conditions to groups.

gedinakova:
1. Run the dev demo
2. Clear all conditions and start a new group > all add buttons are disabled
3. Select an entity > all add buttons get enabled
4. Click "+ condition" button > Two empty conditions are in view
5. Select a field in the first one

Result: Both conditions get updated with the newly selected field.
Expected result: The add buttons should not be enabled before the first condition is submitted.

teodosiah:
> 1. Run the dev demo
> 
>     2. Put the 'Id equals 123' condition in edit mode (or create a new inner query)
> 
>     3. Change 'Equals' to 'In' > the value input gets disabled
> 
>     4. Add a valid condition in the inner query & submit it
> 
>     5. Commit the parent
> 
> 
> Result: The condition remains in edit mode although the parent condition's commit button gets enabled. On attempting to commit the parent the following error is thrown: ERROR TypeError: Cannot read properties of null (reading 'name') at _a150.createExpressionGroupItem (query-builder-tree.component.ts:1305:64) at _a150.init (query-builder-tree.component.ts:1526:31) at set expressionTree (query-builder-tree.component.ts:221:14) at applyValueToInputField (core.mjs:4045:29) at writeToDirectiveInput (core.mjs:11443:13) at setInputsForProperty (core.mjs:12749:9) at elementPropertyInternal (core.mjs:12050:9)
> 
> Expected result: Adding an In/Not condition should be working fine.

The issue could be reproduced when there is no initially set expressionTree and a new group is added as well:
QB_issue3

teodosiah:
1. Add new group

Result:
image

Еxpected behavior:
image

igdmdimitrov:
@desig9stein , please add background to 'Select All' item (when selected) and align the item: 
image
image

teodosiah:
1. Open Grid Filtering dev demo
2. Open Advanced Filtering dialog
3. Set some expression
4. Click on the Clear Filter button

Result:
The filtering expression tree is not cleared
QB_issue5

Expected result:
To be cleared

desig9stein:
> @desig9stein , please add background to 'Select All' item (when selected) and align the item: ![image](https://private-

I fixed only the color, as far as I understand there is an issue that needs to be addressed in order to fix the sizing, there is an inline height on the combo items that needs to be removed first.

teodosiah:
Committing query should commit all of its committed expressions and discard all uncommitted. Currently, the uncommitted condition remains in edit mode.
QB_issue6

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
```diff
@@ -424,7 +424,7 @@ export class IgxQueryBuilderTreeComponent implements AfterViewInit, OnDestroy {
     /**
      * @hidden @internal
      */
-    public searchValue: any;
+    public searchValue: { value: any } = { value: null };
```
gedinakova, line 427: Why do we need this? If we need null, why not just:
```
public searchValue: any = null;
``` 
Or is it in case we eventually need to add more props to the search value object?

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -2211,8 +2211,8 @@ export class IgxQueryBuilderSampleTestComponent implements OnInit {
     template: `
      <igx-query-builder #queryBuilder [entities]="this.entities" [expressionTree]="this.expressionTree">
          <igx-query-builder-header [title]="'Custom Title'" [showLegend]="showLegend"></igx-query-builder-header>
-         <ng-template #searchValueTemplate igxQueryBuilderSearchValue let-expression="expression">
-            <input type="text" class="custom-class" required [(ngModel)]="expression.searchVal"/>
+         <ng-template #searchValueTemplate igxQueryBuilderSearchValue let-searchValue>
+            <input type="text" class="custom-class" required [(ngModel)]="searchValue.value"/>
```
gedinakova, line 2215: Maybe we should add tests for the selectedField and selectedCondition vars?

gedinakova (changes requested):

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -19,7 +21,14 @@ export declare interface IFilteringExpressionsTree extends IBaseEventArgs, IExpr
     /* alternateName: treeType */
     type?: FilteringExpressionsTreeType;
 
+    /**
+     * @deprecated
```
gedinakova, line 25: Revert this. The  change has been done separate in https://github.com/IgniteUI/igniteui-angular/pull/14815 .

projects/igniteui-angular/src/lib/query-builder/query-builder-functions.ts
```diff
@@ -0,0 +1,903 @@
+import { DebugElement } from '@angular/core';
```
gedinakova, line 1: As this whole file is related to the tests, I suggest we add spec to its name. i.e. query-builder-functions.spec.ts.

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
ivanvpetrov: We need it to be an object so we can pass it between components as a reference and keep it's data bound both ways

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
ivanvpetrov: Done

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -19,7 +21,14 @@ export declare interface IFilteringExpressionsTree extends IBaseEventArgs, IExpr
...
+     */
     find(fieldName: string): IFilteringExpressionsTree | IFilteringExpression;
+
+    /**
+     * @deprecated
```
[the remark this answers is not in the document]
ivanvpetrov: Done

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
ivanvpetrov: Added those two fields in the test

igdmdimitrov:
@imincheva , the PR is ready for design validation.

imincheva:
### UX 
@igdmdimitrov 
- By design when all fields are selected from the "Select return fields" combo, the input value should be "All fields" instead of listing the column names
- By design there's supposed to be an "end" button, when 2 or more conditions in a group have been created (For reference the "end" button is supposed to be positioned after "or group" button
image

### Design (all themes):
@desig9stein 
- when you make a selection in the Combo, the width of the field jumps, which shouldn't happen. By design all inputs/selects/combos should have max-width: 300px and should be set to fill, so this might solve the issue, but in general this shouldn't happen in any scenario
https://github.com/user-attachments/assets/0f89fb96-8d11-43d7-8059-9b00bce3e95f

- "and" and "or" labels in the header should have font-weight: 400 and line-height: 16px
- in medium size the paddings on the children/nested queries should be 16px 12px
- in small size the paddings on the children/nested queries should be 16px 8px
- gap between query builder and output area should be 24px
- input actions (check and cancel buttons) should be aligned to the bottom line of the row (“align-items: end” I suppose)
- background color on the output area should be surface
- empty filter title should be grays.500 in both light and dark mode (except for Indigo, where it’s grays.600, which is correct)

### Fluent
- Border radius on the nested queries should be 4px. (This was our mistake, which I'll be fixing shortly, so you might not see it in the handoff)

### Bootstrap
- Border color of the output area should be grays.100 in light and dark mode
- Bottom padding on the header should be removed


### Indigo
- Border radius on the legend squares and lines (“and” and “or” group elements) should be 4px. I know it’s because these elements use the border radius of the query builder, but it doesn’t look okay, so I suggest we override them only for indigo
- There’s some jumping on some chips. I only noticed it in Indigo

https://github.com/user-attachments/assets/0ee8f02e-1ddd-41b4-99bf-c18598206365

- Border color on the query builder in dark mode should be grays.100. This should be applied on the nested queries frame as well
- Border color on the output area in dark mode should be grays.100

yradoeva:
**UX Design**

**All themes**

1. **Unfold_more / unfold_less icons** in the Details buttons are exchanged
When the nested queries are expanded, the icon should be unfold_less and vice versa.
Design - expanded:
Pasted Graphic 86
Design - collapsed:
Pasted Graphic 89


2. **The chips in S size Query builder should be S as well** (they are M currently)
+ CONDITION

yradoeva:
**Indigo Light & Dark**

1. Title of the Query builder should be H5 (not H6)
Pasted Graphic 93

2. Gap between Edit and Add icons next to the chip should be 8px (not 16px)
ACCESSIBILITY
.
Pasted Graphic 95



**Indigo - Light mode**
3. Background of the nested query area should be grays.50 (not surface light)
Pasted Graphic 91
.
--tree-background

gedinakova:
@desig9stein 

1. Run the demo
2. Edit the first condition's return fields to be all except for Id
3. Note the alignment in the chip

**Result**:
Fluent theme
image

Bootstrap theme
image

Indigo theme
image

**Expected result:**
The fields should be on one line and end with ellipsis if the string is too long.

gedinakova:
@igdmdimitrov 
"Details" button is not accessible with the keyboard as it is a div without a tabstop.

imincheva:
**Issue** 
image

**Steps to reproduce**
1. Start with an empty query.
2. Create a new group by clicking and/or group button
3. Select ID column
4. Select filter, which supports value, e.g. contains
5. Type in value e.g. 123
6. Change the selected filter to filter, which doesn't support value, e.g. Empty

**Result**
The "Value" input text still says 123.

**Expected result**
The input text from "Value" input should be removed and the input should be empty.

igdmdimitrov:
> **Issue** image
> 
> **Steps to reproduce**
> 
> 1. Start with an empty query.
> 2. Create a new group by clicking and/or group button
> 3. Select ID column
> 4. Select filter, which supports value, e.g. contains
> 5. Type in value e.g. 123
> 6. Change the selected filter to filter, which doesn't support value, e.g. Empty
> 
> **Result** The "Value" input text still says 123.
> 
> **Expected result** The input text from "Value" input should be removed and the input should be empty.

@imincheva This can be reproduced in the previous iteration of the query builder so please log an issue and we will fix it in all versions.

imincheva:
@desig9stein As we've previously discussed, the opacity applied on the Label of the Chip should be updated:
- Material, Fluent, Bootstrap: 0.6 in both light and dark mode
- Indigo: 0.8 in both light and dark mode

imincheva:
@desig9stein 
Some chips have gap on the right after the clear icon, which is unnecessary. This is also present, when there's no clear icon on the right.

**Issue**
image
image

**Expected**
image

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -89,8 +92,14 @@ export class FilteringExpressionsTree implements IFilteringExpressionsTree {
      */
     public type?: FilteringExpressionsTreeType;
 
-    constructor(operator: FilteringLogic, fieldName?: string) {
+    public entity?: string;
```
gedinakova, line 95: We need descriptions for the two new properties - entity & return fields.

CHANGELOG.md
```diff
@@ -8,10 +8,29 @@ All notable changes for each version of this project will be documented in this
...
 - `IgxGrid`, `IgxTreeGrid`, `IgxHierarchicalGrid`
     - To streamline the sorting of columns with custom formats, a new `FormattedValuesSortingStrategy` has been introduced. This strategy simplifies the sorting process by allowing direct sorting based on formatted values, eliminating the need to extend the `DefaultSortingStrategy` or implement a custom `ISortingStrategy`. This enhancement improves the ease of handling sorting with custom column formatters.
-
 - `IgxCarousel`
     - Added support for vertical alignment. Can be configured via the `vertical` property. Defaults to `false`.
     - Added support for showing/hiding the indicator controls (dots). Can be configured via the `indicators` property. Defaults to `true`.
+- `IgxQueryBuilder`
+    - Introduced ability to create nested queries by specifying IN/NOT IN operators.
+    - Added the `entities` property that accepts an array describing the entity name and an array of its fields. The `fields` input property has been deprecated and will be removed in a future version. Automatic migrations are available and will be applied on `ng update`.
```
gedinakova, line 16:
```suggestion
    - Added the `entities` property that accepts an array of `EntityType` objects describing an entity with its name and an array of fields. The `fields` input property has been deprecated and will be removed in a future version. Automatic migrations are available and will be applied on `ng update`.
```

gedinakova (changes requested):

CHANGELOG.md
```diff
@@ -8,11 +8,26 @@ All notable changes for each version of this project will be documented in this
...
     - To streamline the sorting of columns with custom formats, a new `FormattedValuesSortingStrategy` has been introduced. This strategy simplifies the sorting process by allowing direct sorting based on formatted values, eliminating the need to extend the `DefaultSortingStrategy` or implement a custom `ISortingStrategy`. This enhancement improves the ease of handling sorting with custom column formatters.
-
 - `IgxCarousel`
     - Added support for vertical alignment. Can be configured via the `vertical` property. Defaults to `false`.
     - Added support for showing/hiding the indicator controls (dots). Can be configured via the `indicators` property. Defaults to `true`.
-
+- `IgxQueryBuilder`
+    - Introduced ability to create nested queries by specifying IN/NOT IN operators.
+    - Added the `entities` property that accepts an array describing the entity name and an array of its fields. The `fields` input property has been deprecated and will be removed in a future version. Automatic migrations are available and will be applied on `ng update`.
+    - Add option to template the search value input:
```
gedinakova, line 17:
```suggestion
    - Added option to template the search value input:
```

desig9stein:
@igdmdimitrov When is this dialog supposed to open? I'm asking because I see inconsistent behavior—sometimes when I click on the chip, it opens, and other times it doesn't.
image


https://github.com/user-attachments/assets/83795cfd-7c0b-4526-8e18-fe37fc21be2b

igdmdimitrov:
> @igdmdimitrov When is this dialog supposed to open? I'm asking because I see inconsistent behavior—sometimes when I click on the chip, it opens, and other times it doesn't. 

@desig9stein , when you select at least two chips.

gedinakova:
@Otixa The conditionName is not updated when changing the filter of an existing condition.
[screen-capture (1).webm](https://github.com/user-attachments/assets/39cf9bdf-2caf-4325-bee2-c3e323180566)

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.html
```diff
@@ -0,0 +1,645 @@
...
+                            [style.display]="isInEditMode() ? 'block' : 'none'"
+                        >
+                            <ng-template igxComboHeader>
+                                <div
+                                    class="igx-drop-down__item igx-drop-down__item--query-builder"
+                                    (click)="onSelectAllClicked($event)"
+                                    [ngClass]="{'igx-drop-down__item--selected': this.selectedEntity && this.selectedReturnFields && this.selectedEntity.fields.length === this.selectedReturnFields.length}"
+                                >
+                                    <igx-checkbox
+                                            [checked]="this.selectedEntity && this.selectedReturnFields && this.selectedEntity.fields.length === this.selectedReturnFields.length"
```
ivanvpetrov, line 490: Forwarding Galia's suggestion:
_To be safe, we should probably check the fields before getting the length (as we do for indeterminate):_
```suggestion
                                            [checked]="this.selectedEntity && this.selectedReturnFields && this.selectedEntity.fields?.length === this.selectedReturnFields.length"
```

gedinakova (approved):

gedinakova:
@igdmdimitrov In the dev demo, attempt to bind the "Commit" button's disabled prop to "!queryBuilder.canCommit()". An exception is thrown in the console - apparently the queryTree is not yet existent.
image

gedinakova:
_On behalf of Javier._
In the dev demo attempt the following initial setup:
        const jsonTree = {
            "filteringOperands": [
              {
                "fieldName": "categoryId",
                "condition": {
                  "name": "lessThan",
                  "isUnary": false,
                  "iconName": "filter_less_than"
                },
                "conditionName": "lessThan",
                "searchVal": 4,
                "searchTree": null,
                "ignoreCase": true
              },
              {
                "fieldName": "categoryId",
                "condition": {
                  "name": "greaterThan",
                  "isUnary": false,
                  "iconName": "filter_greater_than"
                },
                "conditionName": "greaterThan",
                "searchVal": 6,
                "searchTree": null,
                "ignoreCase": true
              },
              {
                "fieldName": "name",
                "condition": {
                  "name": "in",
                  "isUnary": false,
                  "isNestedQuery": true,
                  "iconName": "in"
                },
                "conditionName": "in",
                "ignoreCase": true,
                "searchVal": null,
                "searchTree": {
                  "filteringOperands": [
                    {
                      "fieldName": "name",
                      "condition": {
                        "name": "contains",
                        "isUnary": false,
                        "iconName": "filter_contains"
                      },
                      "conditionName": "contains",
                      "ignoreCase": true,
                      "searchVal": "a",
                      "searchTree": null
                    },
                    {
                      "fieldName": "name",
                      "condition": {
                        "name": "contains",
                        "isUnary": false,
                        "iconName": "filter_contains"
                      },
                      "conditionName": "contains",
                      "ignoreCase": true,
                      "searchVal": "b",
                      "searchTree": null
                    }
                  ],
                  "operator": 0,
                  "entity": "categories",
                  "returnFields": [
                    "name",
                    "categoryId",
                    "description"
                  ]
                }
              }
            ],
            "operator": 1,
            "entity": "categories",
            "returnFields": [
              "categoryId",
              "description",
              "name"
            ]
          };
        this.expressionTree = jsonTree as IExpressionTree;

An error is thrown in getFormatter function.
image

gedinakova:
@igdmdimitrov After deleting all inner conditions and submitting the parent, the two buttons for adding a group remain.
1. In the dev demo, expand the inner query via "Details" button (without entering edit mode on the parent)
2. Put the first inner condition in edit mode & discard
3. Delete all inner conditons
4. Commit the parent 

The buttons stay:
[screen-capture (2).webm](https://github.com/user-attachments/assets/1f510df1-ab54-418b-8a7d-6cf43f2c0387)

gedinakova:
@igdmdimitrov @ivanvpetrov 
1. In the dev demo, change the entity (do not change the return fields)
2. Click one of the add group buttons
3. Discard the empty condition

Result: In addition to removing the empty condition, the return fields are also cleared.

Expected result: Return fields not to be changed.

teodosiah:
1. start editing an expression
2. click the edit icon of another expression

Result:
- if the edited expression is valid i.e. can be committed, it exits edit mode properly, but the next expression does not enter edit mode
- if the edited expression is not valid, it exits edit mode and discards the invalid changes, and the next one enters edit mode as expected
QB-edit-issue

Expected:
To exit edit mode, apply or discard the changes based on the validity of the expression and the next expression should enter edit mode properly

andiesm813:
@desig9stein I made a comment in Figma, but i'm adding it here too. These gaps need to be reduced for the following themes (lightt and dark):

- Material
- Fluent
- Indigo

image

ig-georgeA:
When query builder is used in the dialog mode, the ghost chips are not shown. 
The functionality is working, where "drop here to insert" target is shown. 

image

desig9stein:
> When query builder is used in the dialog mode, the ghost chips are not shown. The functionality is working, where "drop here to insert" target is shown.
> 
> image

I’m unsure of the best approach for handling this. While I could add z-index: 0 and position: relative to the overlay, it doesn’t feel like an appropriate solution. @igdmdimitrov, would it be possible to render the chip ghost inside the `igx-query-builder`?

gedinakova:
> > When query builder is used in the dialog mode, the ghost chips are not shown. The functionality is working, where "drop here to insert" target is shown.
> > image
> 
> I’m unsure of the best approach for handling this. While I could add z-index: 0 and position: relative to the overlay, it doesn’t feel like an appropriate solution. @igdmdimitrov, would it be possible to render the chip ghost inside the `igx-query-builder`?

@ivanvpetrov

CHANGELOG.md
```diff
@@ -2,6 +2,29 @@
 
 All notable changes for each version of this project will be documented in this file.
 ## 19.1.0
+### New Features
+- `IgxQueryBuilder`
+    - Introduced ability to create nested queries by specifying IN/NOT IN operators.
```
gedinakova, line 7:
```suggestion
    - Introduced the ability to create nested queries by specifying IN/NOT IN operators.
```

CHANGELOG.md
```diff
@@ -2,6 +2,29 @@
...
+                let-searchValue
+                let-selectedField = "selectedField" 
+                let-selectedCondition = "selectedCondition"
+                let-defaultSearchValueTemplate = "defaultSearchValueTemplate">
+        @if (selectedField?.field === 'Id' && selectedCondition === 'equals'){
+            <input type="text" required [(ngModel)]="searchValue.value"/>
+        } @else {  
+            <ng-container #defaultTemplate *ngTemplateOutlet="defaultSearchValueTemplate"></ng-container>
+        }
+    </ng-template>
```
gedinakova, line 22: Add the drag & drop functionality

gedinakova (changes requested):

teodosiah:
1. Click on the add condition/group button of some expression
2. Hover the same expression
Result: The button is not shown
QB-add-cond-btn-issue

teodosiah:
The classes and selectors stings in the drag service should be replaced as suggested in the https://github.com/IgniteUI/igniteui-angular/pull/15122 - https://github.com/IgniteUI/igniteui-angular/pull/15122#discussion_r1942475292

ivanvpetrov:
_Moved from previous PR:_

[Bug] "Select All" is not accessible via the keyboard.

Open "Return fields" drop down
Using the arrow keys attempt to navigate to "Select All" option
Result: It's not possible.

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
```diff
@@ -0,0 +1,534 @@
...
+
+            this._ghostChipMousemoveSubscription$ = mouseMoves.pipe(sampleTime(100)).subscribe(() => {
+                this.onChipLeave();
+            });
+        }
+
+        this.setDragCursor('grab');
+
+        //TODO z-index is set, but ghost still not visible in Dialog 
+        if(this.dragGhostElement.style) this.dragGhostElement.style.zIndex = "9999";
```
ivanvpetrov, line 302: ### DRAG GHOST NOT VISIBLE IN DIALOG

Event after setting the **z-index** of the drag ghost to **9999**, it's still not  visible over the dialog.
As you can see, the ghost is rendered as last child of the body, outside app-root and overlay and we don't have control over this:
image

ivanvpetrov (changes requested):

ivanvpetrov:
Please see my review below titled DRAG GHOST NOT VISIBLE IN DIALOG

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
ivanvpetrov: Seems that the overlay is with z-index 10005. 
That's why 9999was not working. 10006 is...

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -1162,6 +1162,35 @@ describe('IgxQueryBuilder', () => {
       expect(queryBuilder.queryTree.hasEditedExpression).toBeFalse();
     }));
 
+    it('Should show add expression button when there is expression in add mode.', fakeAsync(() => {
```
gedinakova, line 1165:
```suggestion
    it('Should show add expression button when there is an expression in add mode.', fakeAsync(() => {
```

gedinakova: This test doesn't fail before the fix!

ivanvpetrov: Currently, all tests are passing locally

gedinakova: Yes, they do. The issue is that the without the fix that caused the addition of this test, the test would still pass. We do not want tests that don't test anything. @teodosiah

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
desig9stein: @ivanvpetrov Why can't move the ghost element inside the query builder?
Adding a z-index to the ghost element is not a viable solution, as any future changes to the z-index of the overlay could result in unintended regressions.

ivanvpetrov: We have almost no control over the drag ghost. It's part of the igx-chip's drag directive and we use it out of the box.
I've looked at other examples along the documentation, the drag ghost is always a sibling of the body, outside any enclosure.

sbayreva:
**Indigo**
There is some inconsistency with the colors for the footer:

Screenshot 2025-02-10 at 17 38 11
The background  + the background of the footer should be grays 50

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
```diff
@@ -0,0 +1,1673 @@
...
+        expressionTree.filteringOperands.forEach(o => isTree(o) ? exprTreeCopy.filteringOperands.push(this.getExpressionTreeCopy(o)) : exprTreeCopy.filteringOperands.push(o));
+
+        if (!this.innerQueryNewExpressionTree && shouldAssignInnerQueryExprTree) {
+            this.innerQueryNewExpressionTree = exprTreeCopy;
+        }
+
+        return exprTreeCopy;
+    }
+
+    public onSelectAllClicked(_event) {
```
ChronosSF, line 1430: Always type params

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
```diff
@@ -0,0 +1,1673 @@
...
+        ) {
+            this.returnFieldsCombo.deselectAllItems();
+        } else {
+            this.returnFieldsCombo.selectAllItems();
+        }
+    }
+
+    public onReturnFieldSelectChanging(event: IComboSelectionChangingEventArgs | ISelectionEventArgs) {
+        let newSelection = [];
+        if (event.newSelection instanceof Array) {
```
ChronosSF, line 1443: Not the biggest fan of instanceof when it can be avoided. It probably wouldn't matter in this case but have in mind that it can be erroneous . As there is Array.isArray method, you don't have to use it here.

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
```diff
@@ -0,0 +1,1673 @@
...
+
+    private createExpressionGroupItem(expressionTree: IExpressionTree, parent?: ExpressionGroupItem, entityName?: string): ExpressionGroupItem {
+        let groupItem: ExpressionGroupItem;
+        if (expressionTree) {
+            groupItem = new ExpressionGroupItem(expressionTree.operator, parent);
+            if (!expressionTree.filteringOperands) {
+                return groupItem;
+            }
+
+            for (const expr of expressionTree.filteringOperands) {
```
ChronosSF, line 1548: for of loops look cool but they are a lot ot slower than traditional for i; i< array.length . Sometimes traditional is better :)

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
```diff
@@ -0,0 +1,568 @@
...
+    private _isKeyboardDrag: boolean;
+    private _dropZonesList: HTMLElement[];   //stores a flat ordered list of all chips, including +Condition button, while performing the keyboard drag&drop
+    private _expressionsList: ExpressionItem[]; //stores a flat ordered list of all expressions, including +Condition button, while performing the keyboard drag&drop
+    private _timeoutId: any;
+
+
+
+    //Get the dragged ghost as a HTMLElement
+    private get dragGhostElement(): HTMLElement {
+        return (document.querySelector('.igx-chip__ghost[ghostclass="igx-chip__ghost"]') as HTMLElement);
```
ChronosSF, line 41: Have you checked what happens if you have two QBs on the page?

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
```diff
@@ -0,0 +1,568 @@
...
+        //if (event == null) return true;
+        const ghostBounds = this.dragGhostElement.getBoundingClientRect();
+        const targetBounds = ofElement.getBoundingClientRect();
+
+        return ((ghostBounds.top + ghostBounds.bottom) / 2) >= ((targetBounds.top + targetBounds.bottom) / 2);
+    }
+
+    //Create the drop ghost node based on the base chip that's been dragged
+    private createDropGhost(keyboardMode?: boolean) {
+        const dragCopy = this.sourceElement.cloneNode(true);
```
ChronosSF, line 250: This is DOM magic galore. Isn't there a way to template the ghost instead?

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
```diff
@@ -0,0 +1,568 @@
...
+            (dragCopy.firstChild.firstChild as HTMLElement).classList.add(QueryBuilderSelectors.FILTER_TREE_EXPRESSION_ITEM_GHOST);
+        }
+        return dragCopy;
+    }
+
+    //Make a copy of the drag chip and place it in the DOM north or south of the drop chip
+    private renderDropGhostChip(appendToElement: HTMLElement, appendUnder: boolean, keyboardMode?: boolean): void {
+        const dragCopy = this.createDropGhost(keyboardMode);
+
+        //Append the ghost
```
ChronosSF, line 272: See above. We want to minimize DOM manipulation in Angulara as much as possible. I am not going to stop the pr from being merged because of this but double check if it's possible to tempalte this please.

ig-georgeA:
Keyboard interactions via tab and shift+tab stop working.
Steps:
1. use the tab key to focus the drag indicator
2. use up/down arrow to rearrange conditions
3. hit tab or shift+tab to change focus to a different element
4. Unable to focus
image

ivanvpetrov:
> Keyboard interactions via tab and shift+tab stop working. Steps:
> 
> 1. use the tab key to focus the drag indicator
> 2. use up/down arrow to rearrange conditions
> 3. hit tab or shift+tab to change focus to a different element
> 4. Unable to focus
> 
> image

This behavior is made intentionally. Once you start the drag&drop procedure (hance hitting up/down) the only way to exit this state is confirming(Enter) or canceling(Esc). That's why Tabs is inhibited. 
If not inhibited, you would loose focus and leave the drag&drop in unfinished state.

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -0,0 +1,3227 @@
...
+      group = QueryBuilderFunctions.getQueryBuilderTreeRootGroup(fix) as HTMLElement;
+      QueryBuilderFunctions.verifyRootAndSubGroupExpressionsCount(fix, 4, 7);
+      expect(queryBuilder.expressionTreeChange.emit).toHaveBeenCalled();
+    }));
+
+    it(`Should add a new 'Or' group to existing group by using add buttons.`, fakeAsync(() => {
+      queryBuilder.expressionTree = QueryBuilderFunctions.generateExpressionTree();
+      fix.detectChanges();
+      tick(100);
+      fix.detectChanges();
```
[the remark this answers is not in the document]
teodosiah: After the first detectChanges the chips are rendered but the values of the entities select and fields combo are still not. Before the re-design some of these tests were failing because of this. Now it seems everything is ok without it

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
ivanvpetrov: This is the drag ghost element, which:
- is rendered upon mouse drag (pointer down + mouse move)
- is destroyed upon mouse release (pointer up)
So, at any given time, only one such ghost can exist for the whole page. Having multiple QBs is not a problem. The dev demo sample it self has three, actually.

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
ivanvpetrov: Syntax changed to `Array.isArray(event.newSelection)`

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
```diff
@@ -0,0 +1,1673 @@
...
+        return !this.parentExpression || (this.parentExpression && this.parentExpression.inEditMode);
+    }
+
+    public onInEditModeChanged(expressionItem: ExpressionOperandItem) {
+        if (!expressionItem.inEditMode) {
+            this.enterExpressionEdit(expressionItem);
+        }
+    }
+
+    public getExpressionTreeCopy(expressionTree: IExpressionTree, shouldAssignInnerQueryExprTree?: boolean): IExpressionTree {
```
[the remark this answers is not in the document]
igdmdimitrov: @ChronosSF, I just tried this but it seems like this type of complex object is not supported: 
image

ig-georgeA:
> 
> This behavior is made intentionally. Once you start the drag&drop procedure (hance hitting up/down) the only way to exit this state is confirming(Enter) or canceling(Esc). That's why Tabs is inhibited. If not inhibited, you would loose focus and leave the drag&drop in unfinished state.

When using the keyboard to move the chip, hitting the down-arrow or up-arrow, works. And to me, the operation is complete.  Your explanation is that the user should still hit enter or escape to exit drag mode first. Is it necessary for the user to confirm again?

ivanvpetrov:
> > This behavior is made intentionally. Once you start the drag&drop procedure (hance hitting up/down) the only way to exit this state is confirming(Enter) or canceling(Esc). That's why Tabs is inhibited. If not inhibited, you would loose focus and leave the drag&drop in unfinished state.
> 
> When using the keyboard to move the chip, hitting the down-arrow or up-arrow works as expected. And to me, the operation is complete. Your explanation is that the user should still hit enter or escape to exit drag mode first. Is it necessary for the user to confirm again? In any case, not a blocker. Mostly asking to understand.

With the arrow keys, you are actually moving a drop ghost, it only looks like the real chip (like moving mouse pointer in mouse drag)
Enter confirms the move (like mouse pointer up does)

sbayreva:
**Material Light**

1. The footer should be color surface white, now it is just white:

Screenshot 2025-02-11 at 17 24 30

** Bootstrap Light**

1. The footer should be color surface white, now it is just white:

Screenshot 2025-02-11 at 17 29 43


**Fluent Light**

1. The footer should be color surface white, now it is just white:
2. 
Screenshot 2025-02-11 at 17 31 52

@desig9stein

ivanvpetrov:
> > > > This behavior is made intentionally. Once you start the drag&drop procedure (hance hitting up/down) the only way to exit this state is confirming(Enter) or canceling(Esc). That's why Tabs is inhibited. If not inhibited, you would loose focus and leave the drag&drop in unfinished state.
> > > 
> > > 
> > > When using the keyboard to move the chip, hitting the down-arrow or up-arrow works as expected. And to me, the operation is complete. Your explanation is that the user should still hit enter or escape to exit drag mode first. Is it necessary for the user to confirm again? In any case, not a blocker. Mostly asking to understand.
> > 
> > 
> > With the arrow keys, you are actually moving a drop ghost, it only looks like the real chip (like moving mouse pointer in mouse drag) Enter confirms the move (like mouse pointer up does)
> 
> not ideal, nor a blocker. But users will feel think it's not working. Two options:
> 
> 1. if it's still a ghost till hitting enter, then give it the same visual style with the darker gray and drop shadow
> 2. OR when arrow key is pressed, move + commit the "drop"

Personally, I'm all for styling the drop ghost. It was actually styled this way while ago

gedinakova (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,33 @@
...
+                let-selectedCondition = "selectedCondition"
+                let-defaultSearchValueTemplate = "defaultSearchValueTemplate">
+        @if (selectedField?.field === 'Id' && selectedCondition === 'equals'){
+            <input type="text" required [(ngModel)]="searchValue.value"/>
+        } @else {  
+            <ng-container #defaultTemplate *ngTemplateOutlet="defaultSearchValueTemplate"></ng-container>
+        }
+    </ng-template> 
+    ```
+- `IFilteringExpression`
```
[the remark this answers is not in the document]
igdmdimitrov: These are siblings to the QB section

pmoleri:
Query builder has a huge shadow compared to other components:
image

ig-georgeA:
> Query builder has a huge shadow compared to other components:
From what I can tell, QB does not need any shadow. Especially if one is using it inline. And if it's used inside a dialog, the shadow will come from the dialog. May be it's accidental?

ivanvpetrov:
@ig-georgeA Hi, George
After a demonstration of the drag&drop, some people expressed mixed feeling about a limitation that was set intentionally.
When a inner query is expanded (it's in edit mode), the drag and drop for the parent query chips is been disabled. There is no problem for it to work, it's only disabled by an if in the code.
But this leaves people wandering why it doesn't work. After all the inner query is only a chip that's in edit mode. If you open a simple condition chip in edit mode, it doesn't disable the functionality, why should this disable it? And I can agree with them.
What's you opinion on removing this limitation. Again, there won't be any problems for it to work properly.

P.S. It seems I forgot to disable the keyboard drag&drop in that scenario too. So I'll await your comment before disabling the keyboard one.

https://github.com/user-attachments/assets/0f79d921-8391-440f-839f-0863e2e7490d

ig-georgeA:
> @ig-georgeA Hi, George After a demonstration of the drag&drop, some people expressed mixed feelings about a limitation that was set intentionally. When a inner query is expanded (it's in edit mode), the drag and drop for the parent query chips is been disabled. There is no problem for it to work, it's only disabled by an if in the code. But this leaves people wandering why it doesn't work. After all the inner query is only a chip that's in edit mode. If you open a simple condition chip in edit mode, it doesn't disable the functionality, why should this disable it? And I can agree with them. What's you opinion on removing this limitation. Again, there won't be any problems for it to work properly.
> 
> P.S. It seems I forgot to disable the keyboard drag&drop in that scenario too. So I'll await your comment before 

If I understood the suggestion, this kind of free-form drag and drop from outside subquery into a sub-query in edit mode only makes sense if it's the same entity. And usually one is creating a subquery to pull something from a different entity. Am I mistaken?  

In the Angular samples we have, Entity A and Entity B seem to be identical.  If the entities don't match, will this not create scenarios for preventing drop in such a situation? I mean yes, at a primitive level, it's a chip, and we have drag and drop. I want to make sure we are supporting a realistic query by relaxing the restriction.

pmoleri:
Query builder adds this text by default.

image

And there's no way to remove the text or the header. I can only override it by setting a different non-empty text.

projects/igniteui-angular/src/lib/query-builder/query-builder.component.ts
```diff
@@ -220,225 +160,68 @@ export class IgxQueryBuilderComponent implements AfterViewInit, OnDestroy {
     @Output()
     public expressionTreeChange = new EventEmitter();
```
pmoleri, line 161: This gets typed as `EventEmitter<any>` which is really confusing to the consumer of the component.
Perhaps it can be typed as `EventEmitter<void>`

src/app/query-builder/query-builder.sample.html
```diff
@@ -1,17 +1,52 @@
-<div class="sample-content">
-    <div class="sample-column">
-        <size-selector #sizeSelector></size-selector>
+<div class="qb-sample-content">
+    <div class="qb-sample-column">
+        <button igxButton="contained" (click)="qbDialog.open()">Open QB inside a dialog</button>
+
+        <igx-dialog #qbDialog closeOnOutsideSelect>
+            <igx-query-builder [entities]="this.entities">
+                <igx-query-builder-tree ></igx-query-builder-tree>
```
pmoleri, line 7: Is this `<igx-query-builder-tree>` really meant to be here?

projects/igniteui-angular/src/lib/query-builder/query-builder.component.ts
pmoleri: Which also makes this documentation misleading:
image

It's not really two-way bindable if it doesn't emit a value.
It only has a change event that you can subscribe to.

What's more, attempting to two-way bind it, like in the docs, breaks it completely after the first update.

teodosiah:
> Query builder adds this text by default.
> 
> image
> 
> And there's no way to remove the text or the header. I can only override it by setting a different non-empty text.

In order to remove the title, you can use:
`<igx-query-builder-header [title]="' '"></igx-query-builder-header>`

src/app/query-builder/query-builder.sample.html
teodosiah: It shouldn't be here

ChronosSF:
> > Query builder adds this text by default.
> > image
> > And there's no way to remove the text or the header. I can only override it by setting a different non-empty text.
> 
> In order to remove the title, you can use: `<igx-query-builder-header [title]="' '"></igx-query-builder-header>`

It's better for the title to be opt-in right (i.e. only have the header and title if the input is set)? Let's do it this way but with a separate pr as this one is quite overwhelming already.

ChronosSF (approved):

pmoleri (dismissed):
We need a flag to disabled the _Select fields_ too:
image
(only at the root level, nested conditions can be changed).

ChronosSF:
> We need a flag to disabled the _Select fields_ too: image (only at the root level, nested conditions can be changed).

please thumbs it up, the PR is already unreviewable even without further changes. @igdmdimitrov is logging a separate task for this which will be marked blocking for the release. We'll have it as a separate pr by EOW

ig-georgeA:
> > Query builder has a huge shadow compared to other components: !
> 
> @ig-georgeA , @desig9stein , is this implemented according to the design handoff and if so should we expect the design to change prior to merging the pr?

the design hand-off was showing off the query builder in a dialog.  But the dialog itself is not part of the component.  In either case, the shadow does not seem necessary coz we have better control this way.

desig9stein:
> > > Query builder has a huge shadow compared to other components: !
> > 
> > 
> > @ig-georgeA , @desig9stein , is this implemented according to the design handoff and if so should we expect the design to change prior to merging the pr?
> 
> the design hand-off was showing off the query builder in a dialog. But the dialog itself is not part of the component. In either case, the shadow does not seem necessary coz we have better control this way.

@ig-georgeA Should I completely remove the shadow, if yes, can we have a screen in the handoff that shows the query builder outside of a dialog?

package.json
```diff
@@ -58,46 +58,46 @@
...
+    "@angular/platform-server": "^19.1.6",
+    "@angular/router": "^19.1.6",
+    "@angular/ssr": "^19.1.7",
     "@igniteui/material-icons-extended": "^3.1.0",
     "@lit-labs/ssr-dom-shim": "^1.2.1",
     "@types/source-map": "0.5.2",
     "express": "^4.21.1",
     "fflate": "^0.8.1",
-    "igniteui-theming": "^14.3.1",
+    "igniteui-theming": "^14.5.0-beta.4",
```
ChronosSF, line 77: @simeonoff , @desig9stein , can we have a final 14.5 version to use here before merging?

CHANGELOG.md
```diff
@@ -6,11 +6,44 @@ All notable changes for each version of this project will be documented in this
 - `IgxCarousel`
     - **Behavioral Changes** - the `maximumIndicatorsCount` input property now defaults to `10`.
     - **Deprecation** - `CarouselIndicatorsOrientation` enum members `top` and `bottom` have been deprecated and will be removed in a future version. Use `start` and `end` instead.
+- `IgxQueryBuilderHeader
```
teodosiah, line 9:
```suggestion
- `IgxQueryBuilderHeader`
```

projects/igniteui-angular/src/lib/core/styles/components/query-builder/_query-builder-theme.scss
```diff
@@ -103,48 +117,154 @@
...
         background-color: var-get($theme, 'background');
-        border-radius: border-radius(rem(4px));
-        box-shadow: elevation(12);
+        border-radius: $border-radius;
+        box-shadow: var-get($theme, 'elevation');
+        overflow: hidden;
+	    
+	    .igx-chip__ghost {
+		    position: relative;
+	    }
```
simeonoff, line 126: The chip from the context of the query builder should be targeted in different way.

projects/igniteui-angular/src/lib/core/styles/components/query-builder/_query-builder-theme.scss
```diff
@@ -103,48 +117,154 @@
...
+            @include type-style('h6');
+        }
+
+        @if $variant == 'indigo' {
+            @include type-style('h5');
         }
+
+        @if $variant == 'bootstrap' {
+            @include type-style('h5');
+        }
```
simeonoff, line 244: Shouldn't these be a part of the typography mixin?

projects/igniteui-angular/src/lib/core/styles/components/chip/_chip-theme.scss
```diff
@@ -698,7 +698,9 @@
 
     %igx-chip__prefix,
     %igx-chip__suffix {
-        display: inline-flex;
+        @include ellipsis();
+
+        display: inline-block;
```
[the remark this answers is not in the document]
desig9stein: Yes, `inline-block` is necessary because the `ellipsis` does not work with inline-flex. The `ellipsis` is essential to prevent issues like this one.
image

I modified it directly in the component since it's not related just to QB, if you want I can remove it from this PR, log another issue, and fix it there.

projects/igniteui-angular/src/lib/core/styles/components/query-builder/_query-builder-theme.scss
desig9stein: Even if I add a class to the `igx-chip`, I still need to target the inner element containing `igx-chip__ghost` to override its position.
image

projects/igniteui-angular/src/lib/core/styles/components/query-builder/_query-builder-theme.scss
desig9stein: if I add them to the typography mixin, we have to remove them after the refactoring for the scoped themes.

projects/igniteui-angular/src/lib/query-builder/query-builder-drag.service.ts
```diff
@@ -0,0 +1,563 @@
...
+    private sourceElement: HTMLElement;
+    private targetExpressionItem: ExpressionItem;
+    private targetElement: HTMLElement;
+    private dropUnder: boolean;
+    private _ghostChipMousemoveSubscription$: Subscription;
+    private _keyboardSubscription$: Subscription;
+    private _keyDragOffsetIndex: number = 0;
+    private _keyDragFirstMove: boolean = true;
+    private _isKeyboardDrag: boolean;
+    private _dropZonesList: HTMLElement[];   //stores a flat ordered list of all chips, including +Condition button, while performing the keyboard drag&drop
```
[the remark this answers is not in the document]
damyanpetev: Also check if this pops up in the API docs and mark as `@hidden` accordingly

projects/igniteui-angular/src/lib/query-builder/query-builder.component.ts
```diff
@@ -218,227 +158,70 @@ export class IgxQueryBuilderComponent implements AfterViewInit, OnDestroy {
...
     /**
      * @hidden @internal
      */
-    public fieldSelectOverlaySettings: OverlaySettings = {
-        scrollStrategy: new AbsoluteScrollStrategy(),
-        modal: false,
-        closeOnOutsideClick: false
-    };
+    @ContentChild(IgxQueryBuilderSearchValueTemplateDirective, { read: TemplateRef })
+    public searchValueTemplate: TemplateRef<any>;
```
damyanpetev, line 173: Again, would prefer if these props were directly private/protected

projects/igniteui-angular/src/lib/query-builder/query-builder.component.ts
```diff
@@ -874,344 +257,68 @@ export class IgxQueryBuilderComponent implements AfterViewInit, OnDestroy {
...
-        if (parent) {
-            if (parent.selected) {
-                parent.selected = false;
-                const index = this._selectedGroups.indexOf(parent);
-                this._selectedGroups.splice(index, 1);
-            }
-            this.deselectParentRecursive(parent);
-        }
-    }
+        const inIcon = '<svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 -960 960 960" width="24px" fill="#5f6368"><path d="M560-280H120v-400h720v120h-80v-40H200v240h360v80Zm-360-80v-240 240Zm560 200v-120H640v-80h120v-120h80v120h120v80H840v120h-80Z"/></svg>';
```
damyanpetev, line 285: Are those supposed to go into the actual package at some point?

CHANGELOG.md
```diff
@@ -6,11 +6,44 @@ All notable changes for each version of this project will be documented in this
 - `IgxCarousel`
     - **Behavioral Changes** - the `maximumIndicatorsCount` input property now defaults to `10`.
     - **Deprecation** - `CarouselIndicatorsOrientation` enum members `top` and `bottom` have been deprecated and will be removed in a future version. Use `start` and `end` instead.
+- `IgxQueryBuilderHeader`
+    - **Behavioral Changes**
+    - Legend is no longer shown.
+    - **Deprecation**
+    - The `showLegend` input property has been deprecated and will be removed in a future version. Automatic migrations are available and will be applied on `ng update`.
```
damyanpetev, line 13: This is still mostly relevant under the QB section rather than general, no?

damyanpetev: Also this can be tied with a point that I'm missing - there is a redesign of the component we might want to note.

projects/igniteui-angular/src/lib/core/i18n/query-builder-resources.ts
```diff
@@ -35,26 +38,46 @@ export interface IQueryBuilderResourceStrings {
...
+    igx_query_builder_where_label?: string;
     igx_query_builder_and_group?: string;
     igx_query_builder_or_group?: string;
     igx_query_builder_end_group?: string;
     igx_query_builder_and_label?: string;
     igx_query_builder_or_label?: string;
+    igx_query_builder_switch_group?: string;
     igx_query_builder_add_condition?: string;
-    igx_query_builder_create_and_group?: string;
-    igx_query_builder_create_or_group?: string;
```
damyanpetev, line 45: Removing those is technically a breaking change, however small. The resource interfaces are public API, just sayin :)

damyanpetev (commented):
Leaving some comments, most of those can be addressed later (thus non-blocking for now)

simeonoff (approved):

projects/igniteui-angular/src/lib/core/i18n/query-builder-resources.ts
teodosiah: added migration in this PR - https://github.com/IgniteUI/igniteui-angular/pull/15381

ChronosSF (approved):

damyanpetev (approved):

PR #14686, opened 2024-08-26
Ignore time portions when filtering 'date' type columns - master

### LINKED ISSUE #14643: Excel Style Filter: incorrect filter result when values have time portions for 'date' type columns
## Description  
If the date values in the data are stored as ISO strings which have different times, or as Dates havin time portions different than 0, and when performing a filtering action, the result is not as expected and all duplicated dates are discarded.

ESF-issue

While the unique date values are correctly extracted for the filter list of the ESF, the underlying values of the items are the original strings that contain the time portions as well. Because of this, only the records matching the first encountered unique date value are included: for example 'Dec 29: 2023-12-29T17:**25**:51.615+00:00' is included, but not '2023-12-29T19:**30**:51.615+00:00'.

 * browser: any

## Steps to reproduce  

1. Open [this sample](https://stackblitz.com/edit/igx-grid-date-filtering?file=src%2Fapp%2Fgrid%2Fgrid-filtering-sample%2Fgrid-filtering-sample.component.html).
2. Add a filter through the ESF UI, for example exclude the "Dec 13, 2024" date.

## Result  
The result set has excluded any duplicated date values.

## Expected result  
The result set to include all dates, whose date parts match the ones of the filter values set.
When comparing values of columns of type `date` or `time`, only the relevant date-time portions to be accounted for.

## Attachments  

mtsvyatkova:
Closes #14643 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


gedinakova (changes requested):

projects/igniteui-angular/src/lib/grids/filtering/excel-style/excel-style-filtering.component.ts
```diff
@@ -487,37 +487,34 @@ export class IgxGridExcelStyleFilteringComponent extends BaseFilteringComponent
...
+                    return new Date(value).toISOString();
+                case GridColumnDataType.Time:
+                    return typeof value === 'string' ? value : new Date(value).toLocaleTimeString();
+                default:
+                    return value;
+            }
+        };
+
+        const processExpression = (arr: any[], e: any): any[] => {
+            if (e.expression.condition.name === 'in') {
```
gedinakova, line 511: Not sure, but maybe we should add the 'not in' condition here as well?

mtsvyatkova: When filter is applied and there more than two selected items in the `IgxExcelStyleSearchComponent`, a single filtering operand with 'in' condition and a set of search values is created. So here we are only checking for that 'in' condition to handle the `searchVal` properly.

gedinakova (approved):

ddaribo:
The duplicate values disappearig issue is addressed, however, I noticed that after deselecting a value from the ESF dropdown list, so that it gets filtered out, applying the filter condition and then reopening the filter dialog, all items become deselected: 

after

PR #14689, opened 2024-08-27
fix(pivot-grid-navigation): rxjs operators import to support v6

### LINKED ISSUE #14654: igniteui-angular 18.x throws for apps having 'rxjs' version 6.x
## Description  
As per the Angular Version compatibility page, among the [actively supported versions](https://angular.dev/reference/versions#actively-supported-versions) for 18.0.x, 18.1.x and 18.2.x for the `rxjs` package are `^6.5.3 \^7.4.0`.

However, for apps having `rxjs` version lower than 7.x, the following error is thrown from the igniteui-angular library:

```
Error: export 'timeout' (imported as 'timeout') was not found in 'rxjs' (possible exports: ArgumentOutOfRangeError, AsyncSubject, BehaviorSubject, ConnectableObservable, EMPTY, EmptyError, GroupedObservable, NEVER, Notification, NotificationKind, ....
```

Most likely coming from line 9 in `lib\grids\pivot-grid\pivot-grid-navigation.service.ts` ([ref](https://github.com/IgniteUI/igniteui-angular/blob/427d6eee33d96766136147405befc0109e6a6a53/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-navigation.service.ts#L9)):
`import { take, timeout } from 'rxjs';`

It appears that in versions 6.x or below the operators have to be imported from `rxjs/operators`.

The report is coming from a customer and according to them the supported by Angular 18.1.x `rxjs` versions should also be supported by `igniteui-angular`.

 * igniteui-angular version: 18.x
 * browser: any

## Steps to reproduce  

1. Check out [this sample](https://stackblitz.com/edit/18-x-sample-rxjs-6-x?file=package.json).

## Result  
Build fails.

## Expected result  
Successful build.

## Attachments  

ddaribo:
Closes #14654 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


damyanpetev:
This fix is good, considering `./operators` is an active export path still, even though it is considered deprecated after 7.2 https://rxjs.dev/guide/importing#new-in-rxjs-v720

Unfortunately, if we're to adhere to 6 support alongside Angular there's currently nothing preventing new code from going with the latest as documented. Haven't gone though the entire changelog since 6 is really quite old at this point, so somewhat hesitant to suggest downgrading the entire monorepo to 6. If possible only for the "bundle-test" project or perhaps an special CI step to downgrade and re-build to guarantee no other imports like that creep in.

Also noticed we don't really list `rxjs` explicitly in the lib `peerDependencies` and wondering if that's even right. If we do, we could list 7+ at some point too.

damyanpetev (approved):
After discussion, we've decided it's probably not worth it for now to dive into trying to enforce the compatibility. The only similar fix I've landed on the Angular repo was https://github.com/angular/angular/pull/54193 and I can see no enforcement mechanic anywhere there too. So perhaps just everyone should be advised to check carefully the rsjx usage when reviewing.
Otherwise, LGTM

PR #14717, opened 2024-09-03
Add missing blazor attribute for selectedRows prop

ig-robstoffers:
The `selectedRows` property in `GridBaseDirective` was missing a Blazor attribute that would allow it to function correctly when setting it programmatically.


damyanpetev:
@ig-robstoffers The `blazor-angular-elements` branch is very much archive at this point (need to see if it can be locked), so this should go into 18.1.x and master. I can rebase this one but you should also pick the fix for master as well.

damyanpetev (approved):

PR #14734, opened 2024-09-05
fix(grid): nav service cdr & event emit while row edit

damyanpetev:
We noticed a some extra edit template refresh in Elements which turned out to be a `cdr.detectChanges()` call from the navigation service when it shouldn't be handling in such cases. That stemmed from this check:

https://github.com/IgniteUI/igniteui-angular/blob/618b0b439788b8261f5c3d2c637f9a6a553df8e0/projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts#L69-L74

Where while row editing `this.grid.crudService.rowInEditMode` is always available and that early return can never happen.
So given a grid like this:
```html
<igx-grid #gridRowEdit [data]="data" [primaryKey]="'ProductID'" width="700px" height="400px" [rowEditable]="true" (gridKeydown)="keyDown($event)">
```
you get both `gridKeydown` event and an extra change detection on every key:
grid-row-edit-nav-service

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


rkaraivanov (dismissed):

ddincheva:
The same issue is reproduced also with cell editing
editingEvent

 .. and also the issue is still reproducible for the hierarchical grid during rowEditing;
The event gridKeydown should not be triggered if the key combination is not supported by the grid /https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/igxgridcomponent.html#gridKeydown/

projects/igniteui-angular/src/lib/grids/grid-navigation.service.ts
```diff
@@ -69,7 +69,7 @@ export class IgxGridNavigationService {
     public dispatchEvent(event: KeyboardEvent) {
         const key = event.key.toLowerCase();
         if (!this.activeNode || !(SUPPORTED_KEYS.has(key) || (key === 'tab' && this.grid.crudService.cell)) &&
-            !this.grid.crudService.rowEditingBlocked && !this.grid.crudService.rowInEditMode) {
+            !this.grid.crudService.rowEditingBlocked && !!this.grid.crudService.rowInEditMode) {
```
ddincheva, line 72: we have the same check and inside hierarchical-grid-navigation.service ... please address the change correspondingly

damyanpetev: Sigh, itch to completely rework all of this intensifies. But will try to resists for the fix PRs :D

damyanpetev:
Okay, after more digging - yeah initial fix would turn the cell edits permanently the other way around.
The `rowEditingBlocked` check might be entirely redundant/obsolete IMO - that flag is set on cancel of `rowEdit` but remains until next attempt for/successful exit, so after a cancel the original issue still repros too.

ddincheva (approved):

ChronosSF:
As discussed with @damyanpetev , we'll only include this for 18.2

PR #14772, opened 2024-09-16
feat(combo, simple-combo): filter out non-existent items from selection - master

### LINKED ISSUE #14732: Mismatch in combo/simple-combo: Non-matching values should not appear in the selection
## Description  
When binding an array of values to the igx-combo and programmatically setting selected values that include non-existent items, the combo displays all values, including the invalid ones, in a comma-separated string. The expected behavior is that only the matched values in the bound array should be shown in the combo.

 * igniteui-angular version: 17.2 18.1 master
 * browser: all

## Steps to reproduce  
1. Bind an array of values to the igx-combo.
2. Programmatically set a list of selected values that includes some which are not part of the bound data.
3. Observe the combo box after the selection is applied.

## Result  
The igx-combo shows all the selected values, including ones that are not present in the bound array, in a comma-separated format.

## Expected result  
Only the values that match the data bound to the combo should appear in the selection, and the invalid values should be ignored or handled.

## Enhancement Note
This is more of an enhancement request

This issue isn't a bug per se but more of an enhancement request. The current behavior is expected since the API selects values as provided, even if they aren't part of the bound data. However, it would be ideal if the combo box could automatically filter out invalid values to improve user experience and data accuracy.

## Attachments  
```
<div class="sample-wrapper">
  <div class="combo-section">
    <igx-combo class="combo" #withValueKey [data]="cities" valueKey="id" displayKey="name" groupKey="country"
        [(ngModel)]="selectedValueKey">
    </igx-combo>
    <button igxButton="contained" (click)="selectValues()">Set Values</button>
  </div>
  <h6>Selected Cities:</h6>
  <div>{{ selectedValueKey }}</div>
</div>
```
```
public cities = [{
    name: 'Sofia',
    id: 'BG01',
    country: 'Bulgaria'
  }, {
    name: 'Plovdiv',
    id: 'BG02',
    country: 'Bulgaria'
  }];

public selectedValueKey = [];

// This method sets values, some of which do not exist in the bound data
public selectValues() {
   this.selectedValueKey = ['UK01', 'JP01', 'DE01', 'BG01'];  // Non-existent values
}
```

georgianastasov:
Closes #14732  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -93,6 +93,47 @@ export const mergeObjects = (obj1: any, obj2: any): any => mergeWith(obj1, obj2,
...
+ * Recursively checks if two objects are deeply equal.
+ * Handles circular references by keeping track of seen objects.
+ *
+ * @param obj1 First object to compare
+ * @param obj2 Second object to compare
+ * @param seen Set of already visited objects to handle circular references
+ * @returns true if objects are equal, false otherwise
+ * @hidden
+ */
+export const areObjectsEqual = (obj1: any, obj2: any, seen = new Set()): boolean => {
```
jackofdiamond5, line 106: We don't need this function - we have a dependency to `lodash-es` which exposes an [isEqual](https://lodash.com/docs#isEqual) function which is pretty good at deep object comparisons.

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -267,6 +267,14 @@ export class IgxComboComponent extends IgxComboBaseDirective implements AfterVie
         }
     }
 
+    /** @hidden @internal */
+    public ngOnChanges(changes: SimpleChanges): void {
+        if (changes['data'] && this.data && this.data.length > 0 && this._value && this._value.length) {
+            const oldSelection = this.selection;
+            this.applySelection(oldSelection);
```
jackofdiamond5, line 274:
```suggestion
            this.applySelection(this.selection);
```

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -267,6 +267,14 @@ export class IgxComboComponent extends IgxComboBaseDirective implements AfterVie
         }
     }
 
+    /** @hidden @internal */
+    public ngOnChanges(changes: SimpleChanges): void {
+        if (changes['data'] && this.data && this.data.length > 0 && this._value && this._value.length) {
+            const oldSelection = this.selection;
```
jackofdiamond5, line 273:
```suggestion
```

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -469,4 +485,30 @@ export class IgxComboComponent extends IgxComboBaseDirective implements AfterVie
             selection.join(', ');
         return value;
     }
+
+    private applySelection(oldSelection: any[]): void {
```
jackofdiamond5, line 489: This method is not really needed since we already have the `setSelection` method. It does more or less the same thing and the only difference is that in the `setSelection` we also emit an event to notify that the selection has been applied and this can be toggled with an additional argument that says whether it should emit or not.

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -419,15 +434,18 @@ export class IgxComboComponent extends IgxComboBaseDirective implements AfterVie
...
+    protected setSelection(selection: Set<any>, event?: Event, emitEvent: boolean = true): void {
+        const filteredSelection = this.isRemote ? Array.from(selection) : Array.from(selection).filter(item => this.isItemInData(item));
+
         const currentSelection = this.selectionService.get(this.id);
-        const removed = this.convertKeysToItems(diffInSets(currentSelection, selection));
-        const added = this.convertKeysToItems(diffInSets(selection, currentSelection));
-        const newValue = Array.from(selection);
+        const removed = this.convertKeysToItems(diffInSets(currentSelection, new Set(filteredSelection)));
+        const added = this.convertKeysToItems(diffInSets(new Set(filteredSelection), currentSelection));
+        const newValue = filteredSelection;
```
jackofdiamond5, line 443: `filteredSelection` is the new value now so we can move this to be at the top of the method.

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -309,8 +316,16 @@ export class IgxComboComponent extends IgxComboBaseDirective implements AfterVie
...
      */
     public select(newItems: Array<any>, clearCurrentSelection?: boolean, event?: Event) {
-        if (newItems) {
-            const newSelection = this.selectionService.add_items(this.id, newItems, clearCurrentSelection);
+        if (!newItems) {
+            return;
+        }
+
+        if (this.isRemote || (this.data && this.data.length)) {
+            const validItems = !this.isRemote
```
jackofdiamond5, line 324: We can simplify this so that we don't check if we are remote twice, something like this maybe:
```ts
let validItems = newItems;
if (!this.isRemote && this.data && this.data.length) {
    validItems = newItems.filter(item => this.isItemInData(item));
}

const newSelection = this.selectionService.add_items(this.id, validItems, clearCurrentSelection);
this.setSelection(newSelection, event);
```

CHANGELOG.md
```diff
@@ -43,10 +43,11 @@ For Firefox users, we provide limited scrollbar styling options through the foll
 
     - **Behavioral Changes** - the `keyboardSupport` input property now defaults to `false`.
     - **Deprecation** - the `keyboardSupport` input property has been deprecated and will be removed in a future version. Keyboard navigation with `ArrowLeft`, `ArrowRight`, `Home`, and `End` keys will be supported when focusing the indicators' container via ` Tab`/`Shift+Tab`. 
-
 - `IgxBadge`
     - **Breaking Change** The `$border-width` property has been removed from the badge theme.
     - New outlined variant of the badge component has been added. Users can switch to `outlined` by adding the newly created `outlined` property to a badge.
+- `IgxCombo`
```
jackofdiamond5, line 49: We may want to mention that this applies to when the combo is bound to local data specifically since we can't really query the whole data source to check if the value is there or not.

projects/igniteui-angular/src/lib/combo/combo.component.ts
```diff
@@ -449,7 +471,10 @@ export class IgxComboComponent extends IgxComboBaseDirective implements AfterVie
             } else {
                 this._displayValue = this.createDisplayText(this.selection, args.oldSelection);
             }
-            this._onChangeCallback(args.newValue);
+            this.cdr.markForCheck();
+            if (emitEvent) {
```
jackofdiamond5, line 475: Not sure about this one - it shouldn't be tied to setting the value imo. Since we can set the value through the API and not emit an event, yet we still have to update the model.

RivaIvanova:
Setting the selection via the `select` method emits the `selectionChanging` event. It does not matter if it is a valid selection or not. The event should be emitted only on user interaction.

georgianastasov:
> Setting the selection via the `select` method emits the `selectionChanging` event. It does not matter if it is a valid selection or not. The event should be emitted only on user interaction.

@RivaIvanova @jackofdiamond5 
Currently, the `select()` method emits the `selectionChanging` event even when the selection is set programmatically, regardless of whether the selection is valid.

Are you suggesting that the `selectionChanging` event should only be emitted on user interactions and not when the selection is changed programmatically via the `select()` method? If so, would you like me to modify the `select()` method to prevent it from emitting this event during programmatic changes?

Please let me know if I’ve understood your suggestion correctly or if there’s another approach you’d prefer me to consider.

RivaIvanova:
You are correct and the `select` method along with the `selectAllItems`, `setSelectedItem`, `deselect`, `deselectAllItems` methods emit the `selectionChanging` event prior to this PR so you can disregard the previous comment. However, there are some other behaviors that I observed when the combo has the `required` attribute set:

```html
<igx-combo
  #comboModel
  #myComboModel="ngModel"
  required
  [data]="items"
  [(ngModel)]="ngModelValue"
>
  <label igxLabel>Fruits</label>
</igx-combo>

combo selection {{ comboModel.selection.length }}
combo valid {{ myComboModel.valid }}
```
```ts
public items: any[] = ['Mango', 'Banana', 'Apple'];
public ngModelValue = ['invalid value'];
```

In this case, the combo selection is 0, and nothing is displayed in the read-only input field, which is correct according to the new changes. However, the controls [`valid`](https://angular.dev/api/forms/AbstractControlDirective#valid) property is `true` but should be `false`.

If the custom `ngModelValue` is set to an empty array, i.e., `public ngModelValue = [];`, the `valid` property is properly set to `false`.

igxCombo - required

The same is true if we have a combo inside a reactive form.

```html
<form [formGroup]="user">
  <igx-combo
    [data]="genres"
    formControlName="genres"
  >
    <label igxLabel>Movies</label>
  </igx-combo>
</form>
reactive form valid {{ user.valid }}
```
```ts
public user: UntypedFormGroup;
public genres: any[] = ['Action', 'Comedy', 'Adventure'];

constructor(fb: UntypedFormBuilder) {
  this.user = fb.group({
    genres: [['test'], Validators.required],
  });
}
```

If we set `genres` to an empty array `genres: [[], Validators.required],` the reactive form's `valid` property is `false`.



[Here](https://stackblitz.com/edit/lpkkxe-ow3dh5?file=src%2Fapp%2Fcombo-main%2Fcombo-main.component.html,src%2Fapp%2Fcombo-main%2Fcombo-main.component.ts) is a sample that demonstrates the configuration.

Also, this enhancement aligns the Angular combo with the WC one, which is great, but this should be addressed for the simple combo as well. 🙂

georgianastasov:
We have separated this behavior into a separate issue [here](https://github.com/IgniteUI/igniteui-angular/issues/14949) to ensure it is considered independently and does not block the new enhancement. The enhancement is already implemented and ready for review.

projects/igniteui-angular/src/lib/combo/combo.component.spec.ts
```diff
@@ -2606,6 +2610,32 @@ describe('igxCombo', () => {
                 expect(selectionSpy).toHaveBeenCalledWith(expectedResults);
                 expect(input.nativeElement.value).toEqual(expectedDisplayText);
             });
+            it('should only select and display existing values in the data when programmatically selecting non existing items with ngModel', fakeAsync(() => {
```
jackofdiamond5, line 2613: This one directly modifies the value of the field bound to the model and the one below changes the value through an API call. So, we should change the tests' descriptions accordingly.

projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts
```diff
@@ -192,10 +192,17 @@ export class IgxSimpleComboComponent extends IgxComboBaseDirective implements Co
      * ```
      */
     public select(item: any): void {
-        if (item !== undefined) {
-            const newSelection = this.selectionService.add_items(this.id, item instanceof Array ? item : [item], true);
-            this.setSelection(newSelection);
+        if (item === undefined) {
```
jackofdiamond5, line 195: At this point the `select` methods for both combos are very similar. So, it should be safe to extract the logic to a common method.

projects/igniteui-angular/src/lib/simple-combo/simple-combo.component.ts
```diff
@@ -212,12 +219,8 @@ export class IgxSimpleComboComponent extends IgxComboBaseDirective implements Co
...
     /** @hidden @internal */
     public writeValue(value: any): void {
-        const oldSelection = super.selection;
-        this.selectionService.select_items(this.id, this.isValid(value) ? [value] : [], true);
-        this.cdr.markForCheck();
-        this._displayValue = this.createDisplayText(super.selection, oldSelection);
-        this._value = this.valueKey ? super.selection.map(item => item[this.valueKey]) : super.selection;
-        this.filterValue = this._displayValue?.toString() || '';
+        this._value = value !== undefined ? [value] : [];
+        this.setSelection(new Set(this._value), undefined, false, false);
```
jackofdiamond5, line 223: Should we check if we also have the data loaded here, like we do with the other combo 🤔 In which case, it could be possible to have the `writeValue` method cease being abstract in `combo.common` as we could merge the logic of both methods into a common one.

PR #14815, opened 2024-09-24
Depricated find and findIndex from IFilteringExpressionsTree, added FilteringUtil

Otixa:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


ivanvpetrov (changes requested):
The circular reference in filtering-utils.ts must be addressed

projects/igniteui-angular/src/lib/data-operations/filtering-util.ts
```diff
@@ -0,0 +1,58 @@
+import { FilteringExpressionsTree, IFilteringExpression, IFilteringExpressionsTree } from 'igniteui-angular';
```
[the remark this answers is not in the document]
Otixa: Thanks! 🙂

ivanvpetrov (dismissed):

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -19,7 +20,14 @@ export declare interface IFilteringExpressionsTree extends IBaseEventArgs, IExpr
     /* alternateName: treeType */
     type?: FilteringExpressionsTreeType;
 
+    /**
+     * @deprecated in version 19.1. Use `FilteringUtil.find` instead.
```
gedinakova, line 24: Use the full version - 19.1.0.

gedinakova (changes requested):
Please add a section about this breaking change in the CHANGELOG.md. For more info, refer to CONTRIBUTING.md > "Breaking changes and migrations" section.

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -19,7 +20,14 @@ export declare interface IFilteringExpressionsTree extends IBaseEventArgs, IExpr
     /* alternateName: treeType */
     type?: FilteringExpressionsTreeType;
 
+    /**
+     * @deprecated in version 19.1.0. Use `FilteringUtil.find` instead.
```
gedinakova, line 24: 18.2.0 instead of 19.1.0

projects/igniteui-angular/src/lib/data-operations/filtering-util.ts
```diff
@@ -0,0 +1,59 @@
...
+     * Returns the index of the filtering expression for a column with the provided tree and fieldName.
+     * ```typescript
+     * let filteringExpressionIndex = FilteringUtil.findIndex(gridExpressionTree, 'Column Field');
+     * ```
+     */
+    public static findIndex(tree: IFilteringExpressionsTree, fieldName: string): number {
+        let expr;
+        for (let i = 0; i < tree.filteringOperands.length; i++) {
+            expr = tree.filteringOperands[i];
+            if (expr instanceof FilteringExpressionsTree) {
```
damyanpetev, line 31: Since we touched on the subject - this is still technically a circular reference as `FilteringExpressionsTree` depends on `FilteringUtil` (runtime) and vice versa, so the imports between the two modules remain. I suppose since usage is in methods, the order is not important and that might be why the tooling doesn't complain.
Regardless, what's more is that the actual class `FilteringExpressionsTree` is not supposed to be manadatory so these really shouldn't be checking for instance like that. Perhaps a simple check for `filteringOperands` and the rest of these can remain as pure types (can even be explicitly imporeted as such)?

damyanpetev: And noticed `expr` isn't really typed, but if you want to do so or this could also just be for-of instead - the check can be with a type guard function to narrow the type for the inner call and skip the assert on the else :)

Otixa: I'm a bit wary of modifying the original code, but I've changed the way the type checks are done, and removed the `FilteringExpressionsTree` requirement.

damyanpetev (approved):

damyanpetev:
@gedinakova Good catch! Also (facepalm), got stuck with this util being internal for some reason, but on the off chance it is useful to someone... yeah.

Have to wonder if the `FilterUtil` was exposed by mistake - tracing it back to an refactor part of another feature and both don't seem to be related to 'expose util to run filter standalone' in some manner. ¯\\_(ツ)_/¯
Regardless, it's out there now, so we either integrate into it (two filter utils sounds off to me) or we renamed the new util and export (`ExpressionsTreeUtil`? takers?).

damyanpetev (approved):

PR #14859, opened 2024-10-01
refactor(icons): remove unnecessary icons and reuse the icon registration logic in Grid Filtering and QueryBuilder

### LINKED ISSUE #14604: Grid Filtering & Query builder redundant extended icons registration
Noticed while going over changes in #14598 that both the filtering service and Query Builder (possibly copied over) pull **all** icons from the Material extended Editors category and register them all:
https://github.com/IgniteUI/igniteui-angular/blob/0451f1194a6d6ec396f7cc47a8cab6c2c92e3aaf/projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts#L320-L323

https://github.com/IgniteUI/igniteui-angular/blob/0451f1194a6d6ec396f7cc47a8cab6c2c92e3aaf/projects/igniteui-angular/src/lib/query-builder/query-builder.component.ts#L1210-L1213

That's regardless if all icons are used and I'm fairly certain we're not using all - e.g. previous quarter (even though we could in the future), match, YTD, etc. Now that the icons in use are very clearly listed and documented, would be very nice if we can shake the excess of SVG's we register for no reason.

Also, since the filtering the UI and Query Builder use the same set of icons for filtering operands, it'd be lovely if this operation is bundled somewhere shared around those and reused for the Grid and Query builder components.

teodosiah:
Closes #14604  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/icon/filtering-icons.ts
```diff
@@ -0,0 +1,44 @@
...
+    notEqual,
+    selectAll,
+    startsWith,
+    thisMonth,
+    thisYear,
+    today,
+    ungroup,
+    yesterday
+];
+export const FILTERING_ICONS = new InjectionToken<void>('FILTERING_ICONS', {
```
damyanpetev, line 35: Okay, not exactly what I had in mind, but could consider.
First issue with the token approach is it doesn't tree-shake without explicit `/*@__PURE__*/` before it (see others in the repo) and thus drags the factory function and the icons collection with the exports from the package, thus making those permanently in the bundle. See the bundle app build warning here:
https://github.com/IgniteUI/igniteui-angular/actions/runs/11162862017/job/31028455568#step:16:63
(just FYI same thing happens w/ just a chip component active).

The other issue with this is that due to this being injected (thus on component init) - these icons will be registered even when not in use, so again might be better as a simpler reusable function.

Also no reason for `filteringIcons` and `registerFilteringSVGIconsFactory` to be exported as far as I can tell.

damyanpetev: There's also something going on with the extended icons - I get all of them bundled, which makes no sense and there might be an additional issue there as well.

damyanpetev: Okay, seems to be to do with the `all` export. If split like this:
```ts
import {
    contains, doesNotContain, endsWith, equals, greaterThan, greaterThanOrEqual, isAfter, isBefore,
    isEmpty, isFalse, isNull, isNotNull, isTrue, lastMonth, lastYear, lessThan, lessThanOrEqual, nextMonth,
    nextYear, notEmpty, notEqual, selectAll, startsWith, thisMonth, thisYear, today, ungroup, yesterday
} from '@igniteui/material-icons-extended/editor';

import { type IMXIcon } from '@igniteui/material-icons-extended';
```
Tree-shaking seems to work fine

damyanpetev (dismissed):

damyanpetev (approved):

PR #14879, opened 2024-10-07
isNested prop added to IFilteringOperation

### LINKED ISSUE #14878: Query Builder. Add filter prop "isNestedQuery" rather than using 'in'/'notIn' strings conditions
Currently, for some logic around nested queries, we use conditions with 'in'/'notIn' strings which is not ideal. 

Add a new prop "isNestedQuery" to IFilteringOperation and use that in those conditions. This will also leave the door open for creating custom filters triggering nested query creation

ivanvpetrov:
Closes #14878

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ All notable changes for each version of this project will be documented in this 
 
 - `IFilteringExpression`
     - A new optional property called `conditionName` has been introduced. This would generally be equal to the existing `condition.name`.
-    - A new optional property called `isNestedQuery` has been introduced. It's used to indicate whether the condition leads to a sub-query creation.
+
+- `IFilteringOperation`
+    - A new optional property called `isNestedQuery` has been introduced. It's used to indicate whether the condition leads to a nested query creation.
 
 #### Scrollbar: New CSS variables
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ All notable changes for each version of this project will be documented in this 
 
 - `IFilteringExpression`
     - **Breaking Change** There is a new `conditionName` property which is required. This would generally be equal to the old `condition.name`.
-
+    - A new optional property called `isNestedQuery` has been introduced. It's used to indicate whether the condition leads to a sub-query creation.
 
 #### Scrollbar: New CSS variables
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes for each version of this project will be documented in this 
     - Added support for showing/hiding the indicator controls (dots). Can be configured via the `indicators` property. Defaults to `true`.
 
 - `IFilteringExpression`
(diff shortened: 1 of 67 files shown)


projects/igniteui-angular/src/lib/data-operations/filtering-condition.ts
```diff
@@ -890,6 +947,7 @@ export class IgxStringFilteringOperand extends IgxFilteringOperand {
 export interface IFilteringOperation {
     name: string;
     isUnary: boolean;
+    isNestedQuery: boolean;
```
gedinakova, line 950: We should mark this optional, otherwise we will introduce a breaking change.

gedinakova (changes requested):

gedinakova:
@ivanvpetrov We need to update the CHANGELOG.md as well.

projects/igniteui-angular/src/lib/data-operations/filtering-condition.ts
teodosiah, line 1: Since isNestedQuery is optional, I believe that there is no need to set it everywhere to false, only when it should be true. Similar to the hidden property. Also, then it will not be visible in the grid filtering state

projects/igniteui-angular/src/lib/data-operations/filtering-condition.ts
ivanvpetrov: Marked as optional

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ All notable changes for each version of this project will be documented in this
 
 - `IFilteringExpression`
     - A new optional property called `conditionName` has been introduced. This would generally be equal to the existing `condition.name`.
-
+    - A new optional property called `isNestedQuery` has been introduced. It's used to indicate whether the condition leads to a sub-query creation.
```
gedinakova, line 18: This is not on IFilteringExpression, but on IFilteringOperation.

projects/igniteui-angular/src/lib/grids/state.pivotgrid.spec.ts
```diff
@@ -168,7 +168,7 @@ describe('IgxPivotGridState #pivotGrid :', () => {
                             {
                                 "filteringOperands":[
                                     {
-                                        "condition": {"name":"in","isUnary":false,"iconName":"is-in","hidden":true},
+                                        "condition": {"name":"in","isUnary":false,"isNestedQuery": true,"iconName":"is-in","hidden":true},
```
gedinakova, line 171: I am a little concerned that this is-in is not the same in condition we are refering to. We should check this in the base PR as well.

gedinakova (changes requested):

gedinakova (approved):

igdmdimitrov (approved):

PR #14887, opened 2024-10-07
feat(query-builder): added rehydration to expression trees

Otixa:
Added serialization and deserialization logic for `IExpressionTree` and `IFilteringExpression` to allow for the rehydration of `IFilteringExpression` upon deserialization.
i.e. Correctly recreating the `condition` field with the correct `logic` function, icon, etc.

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -96,14 +101,28 @@ export class FilteringExpressionsTree implements IFilteringExpressionsTree {
 
     public returnFields?: string[];
 
+    /**
+     * Array of filtering operand providers to be used when generating type hints
+     * during serialization.
+     * @memberof FilteringExpressionsTree
+     */
+    public expressionTypes? = [
```
gedinakova, line 109: Expression types are all serialized with null. Maybe we should serialize the names?
image

gedinakova (changes requested):

projects/igniteui-angular/src/lib/data-operations/filtering-util.ts
```diff
@@ -0,0 +1,127 @@
...
+    public static recreateTreeFromJson(jsonString: string): FilteringExpressionsTree {
+        return this.recreateTree(JSON.parse(jsonString));
+    }
+
+    /**
+     * Recreates a `FilteringExpressionsTree` from its serialized model.
+     * @param model The serialized model to recreate the tree from.
+     * @returns A fully reconstructed `FilteringExpressionsTree`.
+     */
+    public static recreateTree(model: ISerializedFilteringExpressionTree): FilteringExpressionsTree {
```
gedinakova, line 20: This method does not return properly a condition for a Boolean operand in an inner query.

In the dev demo add the following code and compare the two trees:
```
var treeS = FilteringUtil.recreateTree((this.expressionTree as FilteringExpressionsTree).toJSON());
var treeSJ = FilteringUtil.recreateTreeFromJson(JSON.stringify((this.expressionTree as FilteringExpressionsTree).toJSON()));
```
image

gedinakova (changes requested):

projects/igniteui-angular/src/lib/data-operations/filtering-util.ts
Otixa: I've resolved this in the latest commit.

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
Otixa: I've removed the `expressionTypes` field from the serialized model.

gedinakova (changes requested):

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -193,4 +212,58 @@ export class FilteringExpressionsTree implements IFilteringExpressionsTree {
         }
         return false;
     }
+
+    /**
+     * Converts the expression tree to a an intermediate object which
```
gedinakova, line 217:
```suggestion
     * Converts the expression tree to an intermediate object which
```

projects/igniteui-angular/src/lib/data-operations/filtering-util.ts
```diff
@@ -0,0 +1,127 @@
...
+import { IFilteringExpression, ISerializedFilteringExpression } from './filtering-expression.interface';
+import { FilteringExpressionsTree, IFilteringExpressionsTree, ISerializedFilteringExpressionTree } from './filtering-expressions-tree';
+
+export class FilteringUtil {
+    /**
+     * Recreates a `FilteringExpressionsTree` from a given JSON serialized string.
+     * @param jsonString The serialized string to recreate the tree from.
+     * @returns A fully reconstructed `FilteringExpressionsTree`.
+     */
+    public static recreateTreeFromJson(jsonString: string): FilteringExpressionsTree {
```
gedinakova, line 11: Attempt to do:
var tree1 = FilteringUtil.recreateTreeFromJson(JSON.stringify(this.expressionTree));

This method throws the following error:
ERROR TypeError: Cannot read properties of undefined (reading 'toString')
    at filtering-expressions-tree.ts:230:55
    at Array.find (<anonymous>)
    at containsCondition (filtering-expressions-tree.ts:229:40)
    at filtering-expressions-tree.ts:244:43
    at Array.forEach (<anonymous>)
    at filtering-expressions-tree.ts:243:42
    at Array.forEach (<anonymous>)
    at packTree (filtering-expressions-tree.ts:237:37)
    at filtering-expressions-tree.ts:252:49
    at Array.forEach (<anonymous>)

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
Otixa: I've changed `expressionTypes` to be static so the are not present on the model, and do not get lost during serialization.

projects/igniteui-angular/src/lib/data-operations/filtering-util.ts
Otixa: I've added a check to prevent this.

damyanpetev (commented):
Okay, without seeing usage I can only guess where this is going, but not sure we're aligned on the idea. Part of the main reason for requesting the declarative API was direct object assign to/from the component and out of the box serialization.

Any rehydrated state should remain internal hopefully, but it's only really needed for convenience to satisfy the existing usage or if resolving the logic on the fly right before it's really needed to run could cost us in performance (though I'm almost sure we can probably do with a mapped registry).

Regardless, the expression tree is tied to the fields/columns/entries by design pretty much, so I'm not sure it's worth replicating the type info again if it can't be used without the other anyway. In which case both the  `ISerializedFilteringExpressionTree` and the `toJson` override are free to go as they shouldn't be needed anymore ;)

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.spec.ts
```diff
@@ -0,0 +1,319 @@
...
+
+        const serializedTree = serialize(tree, true);
+        const deserializedTree = ExpressionsTreeUtil.recreateTreeFromEntities(JSON.parse(serializedTree), entities);
+        const firstOperand = deserializedTree.filteringOperands[0] as IFilteringExpression;
+
+        expect(firstOperand.condition.logic(new Date(2020, 9, 2, 18, 30, 0, 0), firstOperand.searchVal)).toBe(true);
+    });
+
+    it('Nested tree should deserialize correctly', () => {
+        const tree = new FilteringExpressionsTree(FilteringLogic.Or, 'myField', 'myEntity', ['Id']);
```
gedinakova, line 278: On row 285 myEntity's fields collection consists of 'date' field only.

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.spec.ts
```diff
@@ -0,0 +1,319 @@
...
+            const reconstructedOp = deserializedTree.filteringOperands[index] as IFilteringExpression;
+
+            expect(reconstructedOp.condition.logic).not.toBeNull();
+            expect(reconstructedOp.condition.name).toBe(op.conditionName);
+            expect(reconstructedOp.conditionName).toBe(op.conditionName);
+        }
+    });
+
+    it('Sub-queries should deserialize correctly', () => {
+        const tree = new FilteringExpressionsTree(FilteringLogic.Or, undefined, 'myEntity', ['Id']);
```
gedinakova, line 110: myEntity has no 'Id' field.

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
```diff
@@ -1,7 +1,116 @@
...
+
+        for (let i = 0; i < tree.filteringOperands.length; i++) {
+            const operand = tree.filteringOperands[i];
+            if (isTree(operand)) {
+                tree.filteringOperands[i] = this.recreateTreeFromEntities(operand, entities);
+            } else {
+                if (operand.searchTree) {
+                    operand.searchTree = this.recreateTreeFromEntities(operand.searchTree, entities);
+                }
+                tree.filteringOperands[i] = this.recreateExpression(operand, entity?.fields);
```
gedinakova, line 24: What if `entity` does not exist, respectively the fields collection is empty? Do we just silently recreate whatever we can?

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
```diff
@@ -1,7 +1,116 @@
+import { DateTimeUtil } from '../date-common/util/date-time.util';
+import { EntityType, FieldType } from '../grids/common/grid.interface';
+import { GridColumnDataType } from './data-util';
+import { IFilteringOperation, IgxBooleanFilteringOperand, IgxDateFilteringOperand, IgxDateTimeFilteringOperand, IgxFilteringOperand, IgxNumberFilteringOperand, IgxStringFilteringOperand, IgxTimeFilteringOperand } from './filtering-condition';
 import { IFilteringExpression } from './filtering-expression.interface';
-import { IFilteringExpressionsTree } from './filtering-expressions-tree';
+import { IExpressionTree, IFilteringExpressionsTree } from './filtering-expressions-tree';
 
 export class ExpressionsTreeUtil {
+    public static recreateTreeFromEntities(tree: IExpressionTree, entities: EntityType[]): IExpressionTree {
```
gedinakova, line 9: Why don't we just name this 'recreateTree'?

gedinakova (changes requested):

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
Otixa: It would essentially return the expression as is, with a possible addition of `conditionName`. It's not ideal, but we decided against rehydrating based on inferred type.

damyanpetev (commented):
Some outstanding things that I think we can file under separate issues if you want this to be integrated sooned into the branch:
- An odd behavior we encountered while digging through initial errors was the global filtering where we currently instruct users to apply string `contains` condition on all colmns (date & number included). The surprising part is it does work, but it does throw a wrench in our restore logic since we might need to conside string operands somewhat of a fallback for all types. Still not settled in my head, just saying it's something we should probably addresss.
- I was sort of expecting the state persistance logic that does the same type of rehydration to be removed as it should be replaced by this
- Seeing the QB tests reminded me we might want ensure `conditionName` is now also added by all the Grid Filtering UI-s.

PS: I see one more `instanceof FilteringExpressionsTree` check in the filtering service, should probably nix that as well.

gedinakova (approved):

projects/igniteui-angular/src/lib/grids/filtering/grid-filtering.service.ts
```diff
@@ -146,7 +146,7 @@ export class IgxFilteringService implements OnDestroy {
         this.isFiltering = true;
 
         let expressionsTree;
-        if (expressions instanceof FilteringExpressionsTree) {
+        if ('operator' in expressions) {
```
damyanpetev, line 149: expressions can be null, so:
```suggestion
        if (expressions && 'operator' in expressions) {
```

projects/igniteui-angular/src/lib/query-builder/query-builder.component.ts
```diff
@@ -53,10 +54,26 @@ export class IgxQueryBuilderComponent implements OnDestroy {
...
+
+    /**
+     * Sets the entities.
+     * @hidden
      */
     @Input()
-    public entities: EntityType[];
+    public set entities(entities: EntityType[]) {
+        this._entities = entities;
+        if (JSON.stringify(entities) !== JSON.stringify(this._entities)) {
```
damyanpetev, line 71: `JSON.stringify(entities)` can fail with circular ref error if `fields` are actual column instances provided from the grid (advanced filter)

damyanpetev: PS: Also above `this._entities = entities;` 👀

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
```diff
@@ -54,3 +58,159 @@ export class ExpressionsTreeUtil {
...
+function recreateExpression(expression: IFilteringExpression, fields: FieldType[]): IFilteringExpression {
+    const field = fields?.find(f => f.field === expression.fieldName);
+
+    if (field && !expression.condition?.logic) {
+        if (!field.filters) {
+            expression.condition = recreateOperatorFromDataType(expression, field.dataType);
+        } else {
+            expression.condition = field.filters.condition(expression.conditionName || expression.condition?.name);
+        }
+    }
```
damyanpetev, line 150: in case the wrong conditionName is set on the column
```suggestion
        if (!expression.condition) {
            throw Error('Wrong `conditionName`, TBD message');
        }
```

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -1846,18 +1847,17 @@ export abstract class IgxGridBaseDirective implements GridType,
...
+                if (!(isTree(value.filteringOperands[index]))) {
+                    const newExpressionsTree = new FilteringExpressionsTree(FilteringLogic.And, value.filteringOperands[index].fieldName);
+                    newExpressionsTree.filteringOperands.push(value.filteringOperands[index]);
+                    value.filteringOperands[index] = newExpressionsTree;
                 }
             }
 
             value.type = FilteringExpressionsTreeType.Regular;
-            this._filteringExpressionsTree = value;
+            this._filteringExpressionsTree = recreateTreeFromFields(value, this.columns) as IFilteringExpressionsTree;
```
damyanpetev, line 1860: If this is set before columns populate the restore might not work

damyanpetev (approved):
I think there might be more we can remove from the state directive (perhaps later as this PR has been sitting long enough already), otherwise LGTM :)

gedinakova (approved):

Issue #14020, opened 2024-03-25
InlineEditorTemplate of the grid is not working correct if not two way bound to the editValue of the cell


mddragnev:
## Description  
If you try to use inlineEditorTemplate like this and not bound to ngModel: 
```
       <ng-template igxCellEditor let-cell="cell">
            <input name="units"  [value]="cell.editValue" (change)="test($event,cell)"/>
        </ng-template>
```
then the edited value is not updated.
 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Use the above inlineEditorTemplate
2. Try to edit the cell and update the value
3 .Press enter

## Result  
The value of the cell is not updated

## Expected result  
The value of the cell should be updated

This is happening because the change event is called after the editing is done (after the cell template is destroyed) because pressing enter goes to the `keydown` event of the grid body which is before change in the sequence

hanastasov:
Due to the same reason (the grid keyboard navigation handling keydown, which might come before a value is updated), the cell.editValue is not updated in another scenario too:

```
       <ng-template igxCellEditor let-cell="cell">
            <input name="units"  [(ngModel)]="cell.editValue" [ngModelOptions]="{ updateOn: 'blur' }" />
        </ng-template>
```

Let's verify it works for all possible values of upddateOn option (see updateOn: Defines the event upon which the form control value and validity update. Defaults to 'change'. Possible values: 'change' | 'blur' | 'submit'.)

mddragnev:
@hanastasov I dont think that `submit` as `ngModelOptions` makes much sense to me for an editing template. Do you have anything in mind? The PR fixes the `blur` part

hanastasov:
> @hanastasov I dont think that `submit` as `ngModelOptions` makes much sense to me for an editing template. Do you have anything in mind? The PR fixes the `blur` part

I agree, I think a form control in the editing template is an edge case rather than the common case.

PR #14894, opened 2024-10-08
fix(editing): re-trigger activation clear on endedit #14842

### LINKED ISSUE #14842: Active cell is not cleared if editing is ended on focusout of the grid
## Description  
If users want to close editing when the focus moves away from the grid (using the `endEdit` method) the active state of the cell that was in edit mode is not cleared.

 * igniteui-angular version: 17.2, 18.1
 * browser: n/a

## Steps to reproduce  

1. Open the attached sample
2. Start editing in the first grid's title column
3. While editing click away to trigger the custom focusout implementation that calls `endEdit`

## Result  
While editing closes correctly, the active style of the cell remains.

## Expected result  
The cell should clear it's active state just like it would if you select a cell and click away from the grid

## Attachments  
[end-edit.zip](https://github.com/user-attachments/files/17293271/end-edit.zip)

Animation

ChronosSF:
Closes #14842   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
+++ b/projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
@@ -926,7 +926,9 @@ export class IgxDatePickerComponent extends PickerBaseDirective implements Contr
             }
             // do not focus the input if clicking outside in dropdown mode
             if (this.getEditElement() && !(args.event && this.isDropdown)) {
-                this.inputDirective.focus();
+                // this avoids the grid receiving a focus out because the overlay is removed in the same execution
+                // chain causing the browser to focus the body element first
+                setTimeout(() => this.inputDirective.focus());
             } else {
                 this._onTouchedCallback();
                 this.updateValidity();
--- a/projects/igniteui-angular/src/lib/grids/common/crud.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/common/crud.service.ts
@@ -765,6 +765,8 @@ export class IgxGridCRUDService extends IgxRowAddCrudState {
             this.grid.navigateTo(rowIndex, visibleColIndex);
         }
 
+        this.grid.notifyChanges(false);
+
         return false;
     }
 }
--- a/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
@@ -3597,19 +3597,9 @@ export abstract class IgxGridBaseDirective implements GridType,
      */
     public _setupListeners() {
         const destructor = takeUntil<any>(this.destroy$);
-        fromEvent(this.nativeElement, 'focusout').pipe(filter(() => !!this.navigation.activeNode), destructor).subscribe((event) => {
-            if (!this.crudService.cell &&
-                !!this.navigation.activeNode &&
-                ((event.target === this.tbody.nativeElement && this.navigation.activeNode.row >= 0 &&
-                    this.navigation.activeNode.row < this.dataView.length)
-                    || (event.target === this.theadRow.nativeElement && this.navigation.activeNode.row === -1)
-                    || (event.target === this.tfoot.nativeElement && this.navigation.activeNode.row === this.dataView.length)) &&
-                !(this.rowEditable && this.crudService.rowEditingBlocked && this.crudService.rowInEditMode)) {
-                this.navigation.lastActiveNode = this.navigation.activeNode;
-                this.navigation.activeNode = {} as IActiveNode;
-                this.notifyChanges();
-            }
-        });
+        fromEvent(this.nativeElement, 'focusout')
+            .pipe(filter(() => !!this.navigation.activeNode), destructor)
+            .subscribe(event => this.onFocusOut(event));
         this.rowAddedNotifier.pipe(destructor).subscribe(args => this.refreshGridState(args));
         this.rowDeletedNotifier.pipe(destructor).subscribe(args => {
             this.summaryService.deleteOperation = true;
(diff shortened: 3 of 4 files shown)


projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6245,7 +6235,12 @@ export abstract class IgxGridBaseDirective implements GridType,
     // TODO: do not remove this, as it is used in rowEditTemplate, but mark is as internal and hidden
     /* blazorCSSuppress */
     public endEdit(commit = true, event?: Event): boolean {
-        return this.crudService.endEdit(commit, event);
+        const success = this.crudService.endEdit(commit, event);
+        // simulate the process of focusing out of the cell so that the active element is cleared 
+        // while this should naturally happen when clicking a button (as a most likely endEdit called)
+        // the intentional focusout ignore during edit mode prevents the expected behavior
+        this.onFocusOut({ target: this.tbody.nativeElement } as unknown as FocusEvent);
```
damyanpetev, line 6242: Eh, I think it might be a bit too risky to assume all calls to this API are coming when the Grid is focused out and clear selection.
Actually, there's a case for the opposite that currently has an issue. See Row Edit sample in dev demos, check Custom Template (it should have been using `endEdit()` as exposed by the template context, but I digress). Point being the buttons on the row edit overlay are intended to call into `endEdit` but loose focus breaking KB nav. Compare w/ what the default template use - the internal `endRowEditTabStop` that checks for active node and returns focus to the grid tbody. Note that check is also silly and fails if the last active node is header, etc.

My thinking is for this method to handle both cases correctly, it should probably check the focused state of the grid before restoring focus or clearing active cell. Something in the lines of:
```ts
        const activeCell = this.navigation.activeNode;
        const document = this.nativeElement?.getRootNode() as Document | ShadowRoot;

        if (!activeCell) {
            return success;
        }

        if (this.nativeElement?.contains(document.activeElement) && activeCell.row !== -1) {
            // restore focus for navigation, TODO: navigation method to handle all cases...
            this.tbody.nativeElement.focus();
        } else {
            // grid already lost focus, clear active
            this.navigation.lastActiveNode = this.navigation.activeNode;
            this.navigation.activeNode = {} as IActiveNode;
            this.notifyChanges();
        }
```
This else can be refactored to be shared with the onFocusOut handler, but I see no reason for the additional checks - if the grid lost focus prior to end edit, might as well cleanup the active node that editing was blocking.

damyanpetev: Oh, okay, so the logic to restore the correct focus is actually baked into the focus out checks :)
```ts
if (activeNode.row >= 0 && activeNode.row < this.dataView.length) this.tbody.nativeElement.focus()
if (activeNode.row === -1) this.theadRow.nativeElement.focus();
if (activeNode.row === this.dataView.length) this.tfoot.nativeElement.focus();
```
eh 😶

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -926,7 +926,9 @@ export class IgxDatePickerComponent extends PickerBaseDirective implements Contr
             }
             // do not focus the input if clicking outside in dropdown mode
             if (this.getEditElement() && !(args.event && this.isDropdown)) {
-                this.inputDirective.focus();
+                // this avoids the grid receiving a focus out because the overlay is removed in the same execution
+                // chain causing the browser to focus the body element first
+                setTimeout(() => this.inputDirective.focus());
```
damyanpetev, line 931: Kinda sus. The grid receives multiple focus out events, _most_ of which are from in-between focus shifting to the document body. Pretty much every cell click does that and that's why the checks on the handler are so strict about what we want to handle.

damyanpetev: Okay, the cell click traced down to `‎IgxCellCrudState.updateCell`:
https://github.com/IgniteUI/igniteui-angular/blob/d6e7b0e74fd51fa7bb5fb65db309c6197d721121/projects/igniteui-angular/src/lib/grids/common/crud.service.ts#L259-L263

That blur will force the focus to jump temporarily outside the grid and return, tricking the app scenario usage. This IIRC can be considered a minor bug on the Grid side and it might be better if we avoid blanket blur in general. If replaced with `this.grid.tbody.nativeElement.focus()` instead it should work fine, though the fix for #14020 should be re-validated.

damyanpetev: As for the Date Picker - that's another component issue. The entire KB nav of the date picker is broken:
https://www.infragistics.com/products/ignite-ui-angular/angular/components/date-picker#angular-datepicker-example Pick a value, focus _does not_ return to the input.

projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -765,6 +767,8 @@ export class IgxGridCRUDService extends IgxRowAddCrudState {
             this.grid.navigateTo(rowIndex, visibleColIndex);
         }
 
+        this.grid.notifyChanges(false);
+
```
damyanpetev, line 771: From what I can see, this is also not required anymore (if it was indeed). Unless you want to refactor all other calls into this method that end up calling `notifyChanges` as well one way or another, I'd rather not have this extra here.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6760,6 +6762,23 @@ export abstract class IgxGridBaseDirective implements GridType,
         });
     }
 
+    /**
+     * @hidden @internal
+     */
+    protected onFocusOut(event: FocusEvent) {
```
damyanpetev, line 6768: At this point this can also be reverted since the checks are not needed. But we could refactor the repeated save of `lastActiveNode` and clear of active into a method

damyanpetev (commented):
And we should still add a test for the additional functionality of `endEdit` - part of it is covered by the usage in the row edit templates that restores focus, but specifically calling after blur still isn't covered

dkamburov (dismissed):

damyanpetev (approved):

PR #14910, opened 2024-10-15
feat(query-builder): add public api for commit and discard current state

igdmdimitrov:
Closes #  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


gedinakova:
@igdmdimitrov Currently, if I edit a condition in a subquery, it would commit the inner condition only. Nothing happens for the parent condition:
image
On commit, the result is:
image
Even if I commit at this point, the parent condition is not committed.

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -1984,6 +1984,120 @@ describe('IgxQueryBuilder', () => {
 
             QueryBuilderFunctions.verifyRootAndSubGroupExpressionsCount(fix, 3, 6);
         }));
+
+        it('canCommit should return the correct validity state of curently edited condition.', fakeAsync(() => {
```
gedinakova, line 1988:
```suggestion
        it('canCommit should return the correct validity state of currently edited condition.', fakeAsync(() => {
```

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -1984,6 +1984,120 @@ describe('IgxQueryBuilder', () => {
...
+            queryBuilder.expressionTree = QueryBuilderFunctions.generateExpressionTree();
+            fix.detectChanges();
+            tick(100);
+            fix.detectChanges();
```
gedinakova, line 1992: Do we need this second change detection?

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -1984,6 +1984,120 @@ describe('IgxQueryBuilder', () => {
...
+
+            // Double-click the existing chip to enter edit mode.
+            QueryBuilderFunctions.clickQueryBuilderTreeExpressionChip(fix, [1], true);
+            tick(50);
+            fix.detectChanges();
+            expect(queryBuilder.canCommit()).toBeTrue();
+
+            // Verify the Query Builder validity state while editin a condition.
```
gedinakova, line 2000:
```suggestion
            // Verify the Query Builder validity state while editing a condition.
```

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -1984,6 +1984,120 @@ describe('IgxQueryBuilder', () => {
...
+            queryBuilder.discard();
+            tick(100);
+            fix.detectChanges();
+
+            // Verify expression is not commited
+            QueryBuilderFunctions.verifyExpressionChipContent(fix, [1], 'OrderId', 'Greater Than', '3');
+
+            // Verify event is not fired
+            expect(queryBuilder.expressionTreeChange.emit).toHaveBeenCalledTimes(0);
+        }));
```
gedinakova, line 2100: Please add a couple tests covering the scenario where we actually commit an invalid condition and check it doesn't throw.

gedinakova (changes requested):

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -2098,6 +2096,28 @@ describe('IgxQueryBuilder', () => {
             // Verify event is not fired
             expect(queryBuilder.expressionTreeChange.emit).toHaveBeenCalledTimes(0);
         }));
+
+        it(`Should NOT throw errors when an invalid condition is commited through API.`, fakeAsync(() => {
```
gedinakova, line 2100:
```suggestion
        it(`Should NOT throw errors when an invalid condition is committed through API.`, fakeAsync(() => {
```

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -2098,6 +2096,28 @@ describe('IgxQueryBuilder', () => {
...
+            tick(100);
+            fix.detectChanges();
+            expect(queryBuilder.canCommit()).toBeFalse();
+
+            // Apply the changes
+            queryBuilder.commit();
+            tick(100);
+            fix.detectChanges();
+
+            expect(console.error).not.toHaveBeenCalled();
```
gedinakova, line 2119: Add a check what actually was committed.

gedinakova (changes requested):

gedinakova (changes requested):
We need a test with subQuery - verify that both the inner & the parent condition are committed (we actually had a bug initially).

teodosiah:
1. Change the entity in the nested query
2. Commit through API
QB_issue12
The empty expression is still in edit mode

teodosiah:
If we clear the return fields and commit through API as follows:
QB_issue13
We would have 
image
@imincheva  Should we change the appearance in this scenario in order to notify the user that this is not a valid expression?

teodosiah:
- If we commit through API an expression with a column of type number without selecting operator the chip looks like the following:

image

- For other data types the chip in this scenario will be the following:
image

However, there is a valid Null operator that could be chosen and the chip would be the same:
image

@imincheva Do you have any suggestions how we should mark such invalid expressions once they are committed (through API), so they look similar regardless of their column data type and not the same as some valid expression?

gedinakova:
An error is thrown on attempting to commit an inner query without conditions:
image
image

gedinakova (approved):

gedinakova:
> * If we commit through API an expression with a column of type number without selecting operator the chip looks like the following:
> 
> image
> 
> * For other data types the chip in this scenario will be the following:
>   image
> 
> However, there is a valid Null operator that could be chosen and the chip would be the same: image
> 
> @imincheva Do you have any suggestions how we should mark such invalid expressions once they are committed (through API), so they look similar regardless of their column data type and not the same as some valid expression?

A decision was taken to throw an error on attempting to commit invalid state instead of going into too many weird scenarios after committing an invalid condition.

gedinakova:
@igdmdimitrov 
1. In the dev demo, modify any existing condition by deleting the search value
2. Exit edit mode on the editor

Result: The "commit" button stays enabled. canCommit() returns true.
image


Note: The issue is not observed when creating a new condition.

gedinakova:
@igdmdimitrov 
1. In the dev demo, clear the root group's return fields
2. Check the canCommit result > it returns false
3. Enter edit mode on the subquery without any other changes
4. Check the canCommit result

Result: It returns true.
Expected result: It should return false because of the root group's state.

gedinakova (approved):

PR #14935, opened 2024-10-22
feat(banner): add expanded input property to the banner

### LINKED ISSUE #14890: IgxBannerComponent - Support collapsed input
I have a global notification banner list component. Whenever banners need to be shown, a service manages a list of banner components to put on the page. I loop through this list in the html, and render an igx-banner component for each message that needs to be shown. Looking through the source code, I don't see any way to set the initial state of the banner component to be expanded when rendered, so that I don't have to call `open()` on each individual banner after it's added to the DOM.

Support an input property `collapsed` getter/setter, or some other property name.

Currently right now I have an effect that opens the banner whenever the viewChildren signal changes. It would be nice to not have to do this though:

```html
@for (banner of banners(); track banner.id) {
  <igx-banner
    #igxBanner
    class="tt-banner tt-banner--{{ banner.type }}"
    (closed)="bannerStore.remove(banner.id)"
  >
    @if (banner.icon) {
      <igx-icon>{{ banner.icon }}</igx-icon>
    }
    {{ banner.message }}
  </igx-banner>
}

```

```ts
import { explicitEffect } from 'ngxtension/explicit-effect';

export class BannerListComponent {
  protected readonly bannerStore = inject(BannerStore);
  readonly banners = this.bannerStore.prioritized;
  protected readonly bannerComponents = viewChildren(IgxBannerComponent);

  constructor() {
    explicitEffect([this.bannerComponents], ([bannerComponents]) => {
      bannerComponents.forEach(bannerComponent => {
        bannerComponent.open();
      });
    });
  }
}
```

georgianastasov:
Closes #14890   

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


wnvko:
Setting the `collapsed` input will not fire `opening`, `opened`, `closing` and `closed` events. Also banner changes its state without playing any animation. In this way if you:
- call `banner.open` a nice animation is played, `opening` and `opened` events are fired and banners opens.
- set `collapsed` to false no animation is played, no events are fired and banner gets opened.
Is this behavior desired?

georgianastasov:
> Setting the `collapsed` input will not fire `opening`, `opened`, `closing` and `closed` events. Also banner changes its state without playing any animation. In this way if you:
> 
> * call `banner.open` a nice animation is played, `opening` and `opened` events are fired and banners opens.
> * set `collapsed` to false no animation is played, no events are fired and banner gets opened.
>   Is this behavior desired?

@wnvko I initially expected this to be the behavior of the `collapsed ` property since it defines the initial state of the banner (expanded or collapsed). I didn’t think that events like `opening` `opened` `closing` or `closed` should fire when the property is setting the initial state. As for the animations, I’m unsure whether they should be included when toggling collapsed, considering its purpose of defining the starting state.

What’s your suggestion on how this should be handled? Should animations and events be tied to the collapsed property as well?

wnvko:
@georgianastasov 
The new property could be set at any moment. It is not limited to the initial state of the component. You can even change it during the opening/closing animation of the banner. I am not sure how component should behave in these cases. This should be described in the component specification.

georgianastasov:
@wnvko You're absolutely right, the new `collapsed` property can be set at any point, even during the banner's open/close animations. I believe the key question here is whether the banner events should be triggered from the `collapsed` property or not, and if animation should be incorporated during these state transitions. This should definitely be detailed in the component specification to clarify the intended behavior.

@damyanpetev @simeonoff — Would you mind sharing your thoughts on how we should best implement this? Specifically, whether banner events should be triggered from the `collapsed` property and how to handle animations.

simeonoff:
Two things I noticed:

1. No events should be fired when setting this property programatically. Also, no animations should be triggered initially. Animations trigger only when the property change programmatically on runtime.
2. To me a boolean property named `expanded` that is set initially to false sounds more logical.

damyanpetev:
I'll echo parts of what @wnvko and @simeonoff already said:
- I personally agree events shouldn't fire for API calls, though we've had that discussion internally multiple times over the years and never quite settled; it's not entirely consistent across the product even and I don't think we want to open that can of worms for this. But we might not need to, more below.
- The name `collapsed` sort of stuck from toggle directive and propagated to multiple components. Can't say I like it much either, more like used to it. There's also the curios case of two other components - the Nav Drawer and Dialog have **`isOpen`** which I think is a decent alternative. And don't mind that co-existing with a `collapsed` getter.
- Animations honestly a toss depending on the component. For the banner I'm leaning towards without.

Funnily enough, `collapsed` is already an `@Input` on the Expansion Panel that does the same thing and it should be possible for the banner to use it and since _that_ doesn't do animations it sets a pattern. The panel itself has a small issue with this being set while the closing animation runs since it sets the value on end, but that's a bit edge case and can be logged separately.

projects/igniteui-angular/src/lib/banner/banner.component.ts
```diff
@@ -195,6 +233,8 @@ export class IgxBannerComponent implements IToggleView {
     @ContentChild(IgxBannerActionsDirective)
     private _bannerActionTemplate: IgxBannerActionsDirective;
 
+    private _expanded: boolean = false;
+    private _isProgrammaticExpanded = false;
```
wnvko, line 237: can we rename this to something like `shouldFireEvent` or `fireEvent` or anything like this.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Ignite UI for Angular Change Log
 
 All notable changes for each version of this project will be documented in this file.
+## 19.1.0
+### New Features
+- `IgxBanner`
+    - Introduced a new `expanded` input property, enabling dynamic control over the banner's state. This property allows the banner to be programmatically set as expanded (visible) or collapsed (hidden) both initially and during runtime.
```
wnvko, line 7: add more information about when animation will be played, which events will be fired and when.

wnvko (changes requested):
If you change `expanded` to `true` while close animation is playing nothing happens!

Here is why. If you have banner which is not collapsed, `expanded` is `true` and you set `expanded` to `false` this will call `close` method to the internal `expansionPanel`. This will then call `collapse` method and it will start the close animation. At this point `collapsed` is still `false`. It will change to `true` once animation is finish.
While animation is playing if you change `expanded` to `true` it will call `open` method of the internal expansion panel. This will then call `expand` method. In this method the first check is this:
```ts
        if (!this.collapsed) { // If the panel is already opened, do nothing
            return;
        }
```
As expansion panel's `collapsed` is still `false` nothing will happen here.

CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
...
     - **Behavioral Changes** - the `maximumIndicatorsCount` input property now defaults to `10`.
     - **Deprecation** - `CarouselIndicatorsOrientation` enum members `top` and `bottom` have been deprecated and will be removed in a future version. Use `start` and `end` instead.
+### New Features
+- `IgxBanner`
+    - Introduced a new `expanded` input property, enabling dynamic control over the banner's state. The banner can now be programmatically set to expanded (visible) or collapsed (hidden) both initially and at runtime. Animations will trigger during runtime updates — the **open animation** plays when `expanded` is set to `true`, and the **close animation** plays when set to `false`. However, no animations will trigger when the property is set initially.
+    - The banner's event lifecycle (`opening`, `opened`, `closing`, `closed`) only triggers through **user interactions** (e.g., clicking to open/close). Programmatic updates using the `expanded` property will not fire any events.
+    - If the `expanded` property changes during an ongoing animation, the current animation will **cancel immediately**, applying the new state without additional animations or events.
```
wnvko, line 13: This is not how component behaves now.

With current implementation if `expanded` is changed during an animation banner will either do nothing or will start the opposite animation at the point where previous animation has reached.

For example, if the open animation duration is 10 seconds and the close animation duration is 5 seconds, and if `expanded` is set to `true` the open animation will start. Let say animation is at 6 seconds and at this point `expanded` is changed to `false`. Open animation will be stopped, and close animation will start from 3rd second.

If is quite interesting now that `expanded` property and `toggle` method have different behavior.

wnvko:
One more thing. Please when creating a PR add descriptive names, like `add-expanded-to-banner`. Having this at my side does not help me a lot
pr-names

wnvko (approved):

PR #14976, opened 2024-10-29
fix(navigation-drawer): fix ARIA role issue warning.

### LINKED ISSUE #14847: igx-nav-drawer uses aside semantic element with incompatible role
## Description  
The `aside` element is semantic for `role="complementary"` but is being applied `role="navigation"` in the `igx-nav-drawer`.

 * igniteui-angular version: 18.1.x
 * browser: all

## Steps to reproduce  

1. Define a nav drawer with template and mini template
2. Define some nav items
3. Run lighthouse analysis on the application

## Result  
There's an accessibility issue reported.

## Expected result  
There should be no accessibility issues reported.

## Attachments  
image

desig9stein:
Change` <aside>` to a simple `<div>,` and remove `role="navigation"` since we're using `<nav>`, which is a semantic tag that doesn’t need to be wrapped in an element with a `role="navigation"`.

Closes #14847 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


simeonoff (approved):

projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.html
```diff
@@ -14,7 +14,7 @@
     [class.igx-nav-drawer--disable-animation]="disableAnimation"
     (click)="close()" #overlay>
 </div>
-<aside role="navigation"
+<div
```
kdinev, line 17: Shouldn't the correct semantic element be `<nav>`?

kdinev (changes requested):

projects/igniteui-angular/src/lib/navigation-drawer/navigation-drawer.component.html
desig9stein: We already have the nav element inside, that's why I changed it to div, i removed the role since we are using the semantic `<nav>` element, and no role is needed for it.

image

desig9stein: I think we should leave that for the user to decide, we don't know what will be inside, if we change it to nav this means that the drawer can only be used for navigation purposes, if that's what we want. then OK, I will change it.

desig9stein: @simeonoff, @damyanpetev, what is your opinion?

kdinev: @desig9stein I'm saying this because so far, we'd decided by putting `role="navigation"`.

desig9stein: @kdinev you are correct, otherwise I will introduce a breaking change.

kdinev (approved):

damyanpetev (approved):
Yup, from what I can see from the few sites I checked out (https://www.youtube.com, https://developer.mozilla.org/en-US/docs/Web, https://m3.material.io and https://w3c.github.io/aria/) just navigation seems to be quite enough.
Not sure where the `aside` came from, I probably took it from the simpler MDN description that mentions sidebar, but looking at https://w3c.github.io/aria/#complementary it seems it's intended for other purposes.

Anyway, will this involve an update on the docs too? Because those instruct users to add their own `<nav>` manually as well which shouldn't be needed even if it's not technically wrong. www.infragistics.com/products/ignite-ui-angular/angular/components/navdrawer

desig9stein:
> www.infragistics.com/products/ignite-ui-angular/angular/components/navdrawer

Yes @damyanpetev, thanks for pointing this out, I will update the doc samples as well

kdinev (approved):

PR #14980, opened 2024-10-29
fix(ExcelExport): Remove the not supported null char. #14944 

### LINKED ISSUE #14944: igx-grid - Exporting data with null character from grid generates invalid Excel file
## Description  
Exporting null character (\u0000) from grid generates invalid Excel file

Sample of data
````
export const INVOICE_DATA = [{
    ShipName: "Alfred's Futterkiste",
    ShipCity: "\u0000 Berlin(1)",
    ExtendedPrice: 60.0000,
    Freight: 23.9400
}];
````

 * igniteui-angular version:  18.0.0
 * browser:  Chrome 130.0.6723.59 

## Steps to reproduce  

1. Open sample
2. Click export
3. Open Excel

## Result  
An invalid Excel file is generated.

## Expected result  
An valid Excel file is generated.

## Attachments  
Null character https://en.wikipedia.org/wiki/Null_character

Screenshot 2024-10-23 132055 - Copy
Screenshot 2024-10-23 132025 - Copy

[igniteui-angular-export-excel.zip](https://github.com/user-attachments/files/17497606/igniteui-angular-export-excel.zip)

### PR DESCRIPTION
**Pull request by gedinakova**
Closes #14944  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### CODE REVIEW

File: projects/igniteui-angular/src/lib/services/exporter-common/export-utilities.ts [hunk 1]
```diff
@@ -68,7 +68,9 @@ export class ExportUtilities {
                               .replace(/</g, '&lt;')
                               .replace(/>/g, '&gt;')
                               .replace(/"/g, '&quot;')
-                              .replace(/'/g, '&apos;');
+                              .replace(/'/g, '&apos;')
+                              // Bug #14944 - Replace the not supported null character (\u0000, \x00) with a space
+                              .replace(/\0/g, '&#32;');
```
damyanpetev, line 73: Just cuz `\0` looked odd to me - I checked. It's technically a [legacy escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_escape#0), so perhaps we should avoid using it and just have `\x00` instead.
PS: Also still thinking the null char should be empty space (see [wiki](https://en.wikipedia.org/wiki/Null_character)), but really don't feel too strongly about it.

Otixa: IMO, we shouldn't be displaying it at all. I don't particularly mind the current implementation though.
It's almost tempting to say that we should cut off the string at the null byte as it is intended (why would you even be pushing null bytes into there unless you're trying to do something wonky?), but that's a whole separate can of worms...

gedinakova: > Just cuz `\0` looked odd to me - I checked. It's technically a [legacy escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_escape#0), so perhaps we should avoid using it and just have `\x00` instead. 
Not sure what you mean? We don't use it, we just check if it somehow ended up as part of the cell value and make sure to replace it with something that won't break the XML.  
PS: Also still thinking the null char should be empty space (see [wiki](https://en.wikipedia.org/wiki/Null_character)), but really don't feel too strongly about it.
If I understand correctly, you prefer the non-breaking space (&#160;), right?

gedinakova: > Just cuz `\0` looked odd to me - I checked. It's technically a [legacy escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Character_escape#0), so perhaps we should avoid using it and just have `\x00` instead.

Not sure what you mean? We don't use it, we just check if it somehow ended up as part of the cell value and make sure to replace it with something that won't break the XML.

gedinakova: > PS: Also still thinking the null char should be empty space (see [wiki](https://en.wikipedia.org/wiki/Null_character)), but really don't feel too strongly about it.

If I understand correctly, you prefer the non-breaking space (&#160;), right?

gedinakova: > It's almost tempting to say that we should cut off the string at the null byte as it is intended (why would you even be pushing null bytes into there unless you're trying to do something wonky?), but that's a whole separate can of worms...

Agreed, it's very tempting to cut it off but I sincerely doubt they would expect that. I'm assuming this is data coming from a db and those characters have accidentally ended up as cell values.

Otixa (approved):

projects/igniteui-angular/src/lib/services/exporter-common/export-utilities.ts
damyanpetev: > Not sure what you mean? We don't use it, we just check if it somehow ended up as part of the cell value and make sure to replace it with something that won't break the XML.

@gedinakova The comment in on the line with the regex using octal escape (`\0`) which is technically in the deprecated and obsolete page (see link). It might never be removed, however just saying we could use the hex escape instead :)

Also
> If I understand correctly, you prefer the non-breaking space (`&#160;`), right?

Nope, quite literally empty, as in replace with `''`.

PR #14987, opened 2024-10-31
QueryBuilder Validation

teodosiah:
Related to #14647  
Specification - https://github.com/IgniteUI/igniteui-angular/wiki/Query-Builder-Validation-Specification

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/query-builder/query-builder.directives.ts
```diff
@@ -21,3 +23,44 @@ import { Directive, TemplateRef } from '@angular/core';
...
+            } else {
+                this.validators.forEach(validator => {
+                    if (validator.type === 'required') {
+                        this.element.nativeElement.required = true;
+                    } else if (!validator.type.includes('date')) {
+                        this.element.nativeElement.setAttribute(validator.type, validator.value);
+                    }
+                });
+            }
+        }
```
gedinakova, line 64: With this if-else structure the pickers (including dateTimeEditor) would never get the required attribute. 
I would suggest using a switch for better readability:
```suggestion
        if (this.validators) {
            this.validators.forEach(validator => {
                switch (validator.type) {
                    case 'required':
                        this.element.nativeElement.required = true;
                        break;
                    case 'mindate':
                        if (this.picker)
                            this.picker.minValue = validator.value;
                        break;
                    case 'maxdate':
                        if (this.picker)
                            this.picker.maxValue = validator.value;
                        break;
                    default:
                        this.element.nativeElement.setAttribute(validator.type, validator.value);
                        break;      
                }
            });
        }
```

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -2043,6 +2042,204 @@ describe('IgxQueryBuilder', () => {
             QueryBuilderFunctions.verifyEditModeExpressionInputValues(fix, 'OrderId', 'In', '', 0);
             QueryBuilderFunctions.verifyEditModeExpressionInputValues(fix, 'Released', 'True', '', 1);
         }));
+
+        it(`Should apply 'string' column validators properly and calculate 'canCommit' base on it.`, fakeAsync(() => {
```
gedinakova, line 2046: I beleive that simply checking the props/attributes are applied on the inputs/pickers should be enough of a test. The actual effect of these props should have been tested in the inputs themselves.

gedinakova (changes requested):

gedinakova (changes requested):

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ All notable changes for each version of this project will be documented in this
         }
     </ng-template> 
     ```
+    - Added `IgxFieldValidators` that could be used to define basic validation of the search value input based on the field data type.
```
gedinakova, line 25: Can we make this a little more verbose and probably add some snippet?

projects/igniteui-angular/src/lib/query-builder/query-builder.component.spec.ts
```diff
@@ -1904,13 +1905,13 @@ describe('IgxQueryBuilder', () => {
             QueryBuilderFunctions.verifyRootAndSubGroupExpressionsCount(fix, 3, 6);
         }));
 
-        it('Should not make bug where existing inner query is leaking to a newly created one', fakeAsync(() => {
+        it('Should not make bug where existing inner query is leaking to a newly created one,', fakeAsync(() => {
```
gedinakova, line 1908:
```suggestion
        it('Existing inner query's conditions should not leak to a newly created one,', fakeAsync(() => {
```

gedinakova (changes requested):

projects/igniteui-angular/src/lib/grids/common/grid.interface.ts
```diff
@@ -307,6 +307,7 @@ export interface FieldType {
     pipeArgs?: IFieldPipeArgs;
     defaultTimeFormat?: string;
     defaultDateTimeFormat?: string;
+    validators?: any[];
```
damyanpetev, line 310: This is still the base for `ColumnType` right? So this change would affect the the grid and we might have to consider that in the way we expose it for the QB.
Also not sure `any[]` is a good type for this, technically it's `IFieldValidator[]` from what I see. Guessing that's because the column already has an _internal_ `validators: Validator[]` which doesn't match and would error out (that should be a hint).
While the declarative `IFieldValidator` would actually be better for Xplat purposes,  it could be a bit limiting to the validators we support without a mechanism to extend them. This might force us to consider the options for the column validation on xplat as well to keep those in sync.

PR #15150, opened 2024-12-06
fix(*): icon service doesn't work with scoped themes

### LINKED ISSUE #15094: In multi-theme applications, icons automatically include classes based on the global theme.
## Description  
In applications with multiple themes, if an icon is placed inside a wrapper that uses a different theme from the global theme, the icon's class will still reflect the global theme. For example, it may appear as `internal_indigo`, additionally the loaded icon references are loaded for the global one.

 * igniteui-angular version: 18.2.x
 * browser: 

## Steps to reproduce  

1. Create an application using multiple themes.
2. Set 'Indigo' theme to the main component/wrapper.
3. Add a component utilizing different theme .e.g Material inside of the main wrapper.
4. Observe icon styles and classes

## Result  
The icon styles are to based on the current theme.

## Expected result  
The correct icon styles and classes to be included based on the theme.

## Attachments  
internal-indigo-icons-class
material-grid-icons-class

simeonoff:
Closes #15094  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [x] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -133,13 +138,36 @@ export class IgxIconService {
...
+            this._setIconRef(
+                alias.name,
+                alias.family,
+                icon,
+                overwrite
+            );
+        }
+    }
+
+    public _setIconRef(name: string, family: string, icon: IconMeta, overwrite = false) {
```
damyanpetev, line 159: Ehm, public?

damyanpetev: Also by the looks of it, this is more/less the same as the public `setIconRef` and perhaps it can simply call in this method with the modified as external meta + overwrite by default?

projects/igniteui-angular/src/lib/icon/icon.service.ts
```diff
@@ -59,17 +59,22 @@ export class IgxIconService {
...
         @Optional() private _httpClient: HttpClient,
         @Optional() private _platformUtil: PlatformUtil,
+        @Optional() @Inject(THEME_TOKEN) private _themeToken: ThemeToken,
         @Optional() @Inject(DOCUMENT) protected document: Document,
     ) {
+
         this.iconLoaded = this._iconLoaded.asObservable();
         this.setFamily(this._defaultFamily.name, this._defaultFamily.meta);
 
+        this._themeToken?.onChange((theme) => {
```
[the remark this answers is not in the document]
damyanpetev: IMO, not sure there's much use for the callback as public and we might as well keep it internal to skip the need for hoops to hide the observable and hook thinks with `takeUntilDestroyed` :)

projects/igniteui-angular/src/lib/icon/types.ts
```diff
@@ -23,6 +23,7 @@ export interface IconMeta {
     name: string;
     family: string;
     type?: IconType;
+    external?: boolean;
```
damyanpetev, line 26: Any chance we can keep this on the internal types only?

projects/igniteui-angular/src/lib/checkbox/checkbox.component.ts
```diff
@@ -492,29 +493,43 @@ export class IgxCheckboxComponent implements EditorProvider, AfterViewInit, Cont
...
+        protected themeToken: ThemeToken,
         @Optional() @Self() public ngControl: NgControl,
     ) {
         if (this.ngControl !== null) {
             this.ngControl.valueAccessor = this;
         }
 
-        this.theme = this.themeService.globalTheme;
+        this.theme = this.themeToken.theme;
+        this._prefersTokenizedTheme = this.themeToken.preferToken;
```
damyanpetev, line 512: Very minor: You already have the whole token locally and can possibly skip another local prop for the prefer flag. That said `theme` can also be a simple getter for `this.themeToken.theme` too.
Was going to say the same applied for the input group, but there it's a public prop (0.0)

projects/igniteui-angular/src/lib/icon/types.ts
simeonoff: This has also been exposed like that in the Web Components registry.

damyanpetev (approved):
LGTM

damyanpetev (dismissed):

projects/igniteui-angular/src/lib/checkbox/checkbox.component.ts
```diff
@@ -493,28 +494,40 @@ export class IgxCheckboxComponent implements EditorProvider, AfterViewInit, Cont
...
-        this.destroyRef.onDestroy(() => this._subscription.unsubscribe());
+        this.destroyRef.onDestroy(() => unsubscribe);
+    }
+
+    private setComponentTheme() {
+        if(!this.themeToken.preferToken) {
+            const theme = getComponentTheme(this.elRef.nativeElement);
+
+            if (theme && theme !== this.theme) {
+                this.themeToken.set(theme);
```
damyanpetev, line 527: Sorry, missed this one - shouldn't this be `this.theme = theme` instead of setting it on the token which the component doesn't own and is in this case ignoring too?

projects/igniteui-angular/src/lib/services/theme/theme.token.ts
```diff
@@ -0,0 +1,52 @@
...
+import { mkenum } from "../../core/utils";
+import { BehaviorSubject } from "rxjs";
+import { DOCUMENT } from "@angular/common";
+
+export class ThemeToken {
+    private document = inject(DOCUMENT);
+    public subject: BehaviorSubject<IgxTheme>;
+
+    constructor(private t?: IgxTheme) {
+        const globalTheme = globalThis.window
```
damyanpetev, line 11: And I think I preferred this more on the root factory level or was this intentional for a sneaky auto-init on a scoped bit of the DOM? I guess that could be useful, just figured the main use would be to always set something anyway which would make the computed styles call redundant.

simeonoff: The sneaky thing this allows me to do is to initialize a ThemeToken without passing a theme to the constructor in the root factory, making the preferToken getter return false when no theme is passed to the factory.

projects/igniteui-angular/src/lib/checkbox/checkbox.component.ts
simeonoff: You are correct. This shouldn't modify the global token theme.

damyanpetev (approved):

PR #15201, opened 2025-01-06
feat(igxHierarchicalGrid): Add api to navigate to a specific child grid.

### LINKED ISSUE #14912: Add functionality to navigate to a specific child grid within the hierarchical grid
Currently, an API is available for navigating to a specific cell in the hierarchical grid. However, there is no efficient way to navigate to a specific child grid to ensure it is visible.

I am looking for an API that enables navigation to a specific component by scrolling it into view.

This feature is essential for working with the Hierarchical Grid in App Builder. Within the context of the App Builder, we use the Outline Tree and Data Hierarchy Editor to select specific Row Islands. When a Row Island is selected, we want it to automatically scroll into view, making it visible to users. This allows users to see the layout of the specific Child Grid and configure it, along with its columns and nested child grids, as needed.

In addition, it would be beneficial if this new functionality included the ability to expand and collapse rows on demand. Currently, when a Hierarchical Grid in App Builder isn’t selected, its rows remain collapsed.

To achieve a similar outcome, we are currently using a workaround with the functionality of the [positionInParent ](https://github.com/IgniteUI/igniteui-angular/blob/593a033d6f2b8f565e84e617f7365f57ea4fd86c/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts#L154C1-L155C1)method, which partially works for navigation. However, this approach fails in cases where we need to navigate to a grid with multiple scrollable parent elements or when the target grid is outside the DOM. Adding these features would greatly improve the usability and configurability of nested grids within App Builder.

## Additional context
image

MayaKirova:
Closes #14912  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -179,6 +179,47 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
...
+
+            const childGrid =  this.grid.gridAPI.getChildGrid([pathElem]);
+            const positionInfo = this.getElementPosition(childGrid.nativeElement, false);
+            if (positionInfo.offset > 0) {
+                this.grid.verticalScrollContainer.addScrollTop(positionInfo.offset);
+                this.grid.verticalScrollContainer.chunkLoad.pipe(first()).subscribe(() => {
+                    childGrid.navigation.navigateToChildGrid(pathToChildGrid, cb);
+                });
+            } else {
+                childGrid.navigation.navigateToChildGrid(pathToChildGrid, cb);
```
mddragnev, line 218: Is there a case where you can hit that because you are comparing the position of the childGrid nativeElement with the parent positioning? IMO, that `positionInfo.offset` will be > 0 always.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
```diff
@@ -179,6 +179,47 @@ export class IgxHierarchicalGridNavigationService extends IgxGridNavigationServi
...
+    public navigateToChildGrid(pathToChildGrid: IPathSegment[], cb?: () => void) {
+        if (pathToChildGrid.length == 0) {
+            if (cb) {
+                cb();
+            }
+            return;
+        }
+        const pathElem = pathToChildGrid.shift();
+        const rowKey = pathElem.rowKey;
+        const rowIndex = this.grid.gridAPI.get_rec_index_by_id(rowKey, this.grid.dataView);
```
mddragnev, line 197: Probably a good idea to check whether `rowIndex` is -1 (<0) because it is causing a pending navigation.
A case where it is reproducible is:
In a hgrid, try to use the API with a key that does not exist in the parent, i.e (there is not a rowKey = 16) 
`this.grid.navigation.navigateToChildGrid([{rowID:"16",rowKey: "16", rowIslandKey: "childData"}]);`
Nothing happens.
But if you try to expand a row then an error is thrown:
Screenshot 2025-01-06 at 18 09 51

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
MayaKirova: I guess not, since it always scrolls to the parent row first.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid-navigation.service.ts
MayaKirova: Checks are added.

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.navigation.spec.ts
```diff
@@ -950,6 +951,73 @@ describe('IgxHierarchicalGrid Navigation', () => {
...
+        it('should navigate to exact child grid with navigateToChildGrid.', (done) => {
+            hierarchicalGrid.primaryKey = 'ID';
+            hierarchicalGrid.expandChildren = false;
+            fixture.detectChanges();
+            const path: IPathSegment = {
+                rowKey: 10,
+                rowIslandKey: 'childData2',
+                rowID: 10
+            };
+            (hierarchicalGrid.navigation as IgxHierarchicalGridNavigationService).navigateToChildGrid([path], () => {
```
mddragnev, line 978: Probably those castings can now be omitted

mddragnev (changes requested):

mddragnev (dismissed):

viktorkombov (changes requested):

viktorkombov (commented):
In AppBuilder, the implementation works correctly in the design surface, but only when the zoom level is set to 100%. If the zoom level is changed (e.g., below 100%), the scrolling behavior breaks, and navigation to the child grids no longer works as expected.

You can reproduce this issue in the dev demos by adding a `transform: scale(0.5, 0.5)` style to the `.content` class in the `_app-layout.scss` file. This might indicate that the scrolling logic does not properly handle transformations or scaled layouts.

image

viktorkombov (approved):

damyanpetev:
@MayaKirova @dkamburov 
Okay, just jumping in to note that services like `grid.navition` are sitting public only because nobody bothered with access modifiers in the beginning and are still such because we've also had the bright idea to use some bit in docs/samples. Regardless, those are not documented or exported so using them is a workaround more than anything and exposing them further I don't think is the way to go.

Hopefully this isn't going in docs of any kind, I'd much prefer any other alternative. The Grids already have navigation API exposed (see https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/keyboard-navigation#custom-keyboard-navigation) - perhaps that can be extended to handle HGrid's more complex cases?

dkamburov:
@damyanpetev Sure, it won't be documeneted, I'm logging one to expose as part of the hgrid API

PR #15246, opened 2025-01-16
fix(excel-export): Get correct grid column collection from row island…

### LINKED ISSUE #15169: HGrid: Excel Export is not working properly when there are collapsible column groups in the child grid
## Description  
When we use collapsible column groups in a child grid, downloading the hgrid data in Excel does not work properly. The header of the column that is marked as ` [visibleWhenCollapsed]="true"` is skipped in the excel data, but the data itself is there. Which leads to shift in the data for each of the next columns.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Create hgrid that has rowisland with at least one column group like this

```html
<igx-column-group header="Delivery" [collapsible]="true">
     <igx-column field="delivery.dateOrdered" header="Date Ordered" [dataType]="'string'" [width]="'180px'" [sortable]="true" [resizable]="true" [visibleWhenCollapsed]="true"></igx-column>
     <igx-column field="delivery.dateOrdered" header="Date Ordered" [dataType]="'string'" [width]="'180px'" [sortable]="true" [resizable]="true" [visibleWhenCollapsed]="false"></igx-column>
     <igx-column field="delivery.dateShipped" header="Date Shipped" [dataType]="'string'" [width]="'180px'" [sortable]="true" [resizable]="true" [visibleWhenCollapsed]="false"></igx-column>
</igx-column-group>
```
2. Try to export in excel format

## Result  
Result includes only the non-collapsed columns headers.

## Expected result  
Result should include the collapsed columns content (if collapsed) or the non-collapsed (otherwise).

## Attachments  
1. We have this child grid so we expect to see all 3 columns in the excel downloaded
image

2. Instead we see this 
Excel download issue

3. We see the above result regardless whether we have column collapsed or not

mddragnev:
… when exporting collapsible MCH

Closes #15169   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/services/exporter-common/base-export-service.ts
```diff
@@ -1232,7 +1232,12 @@ export abstract class IgxBaseExporter {
 
             columnList = this.getAutoGeneratedColumns(islandData);
         } else {
-            const islandColumnList = island.columns;
+            /*
+            * If there is a registered child grid then get its columns because the collection of columns within
+            * the rowIsland might be outdated. If no then get the row islands column because they will be up to date.
+            */
+            const rowIslandGrid = island.rowIslandAPI.getChildGrids()[0];
```
MayaKirova, line 1239: Not sure about getting an individual child grid since each one can have different visibility per column as a result of user interactions, that are different than the initial state of the row island.

MayaKirova: When I try to export the hierarchical grid from dev demos: http://localhost:4200/hierarchicalGrid
With collapsible groups:
image

It shows a warning for content format, which should be checked:
image

Also, as I suspected, it current gets the state of the first child grid that was expanded and applies it to all other. So if I collapse the column in just the first grid:
image

And export it, all other grids will get that state, even though they are not collapsed:

image

It should either reflect the state of each grid correctly, or if that was not implemented as a feature yet,  then show the initial state from the row island for all grids.

mddragnev: I have no idea why that warning from excel is there. I dont get a warning.

MayaKirova (approved):

mddragnev:
Just to be clear - If there is a row island with collapsible MCH, then the exporter would generate the MCH state as it is initially. Meaning that if any of the column groups is either collapsed or expanded it wont affect the exported excel. This will be working like this because currently we do not process the state of each row island grid. I will document this in the topic as a note.

mddragnev:
@IMinchev64 Could you re-verify this please?

projects/igniteui-angular/src/lib/services/exporter-common/base-export-service.ts
MayaKirova: Warning was happening only in older version of excel. Works fine with newer versions.

PR #15304, opened 2025-01-27
refactor(grids): Move all elements specific code to angular-element's source

### LINKED ISSUE #15235: [Elements] Move all elements specific members from the angular grids into angular-element's source.
Currently there are a lot of props/events/collections, as well as other code changes, in the angular grids base code that are added for correct cross-platform code generation but are not relevant to the angular component itself. Such as all members that are marked as hidden/internal but also have /* blazorInclude */ tag. There are probably other instances of fixes and overrides specifically added for elements.

We want all of those to exist only for the elements grid's and be removed from angular. Hence the grids can be extended in the elements projects as separate components and have all such members moved there.

MayaKirova:
Closes #15235   
Goes with: https://github.com/IgniteUI/igniteui-api-analyzer/pull/129

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


skrustev (dismissed):

skrustev (dismissed):

dkamburov:
@MayaKirova please resolve the conflicts here

skrustev (dismissed):

dkamburov:
@MayaKirova there is a conflict here

skrustev (dismissed):

projects/igniteui-angular-elements/src/analyzer/elements.config.ts
```diff
@@ -259,6 +271,7 @@ export var registerConfig = [
       },
     ],
     additionalProperties: [
+      { name: "actionStripComponents", writable: true },
```
damyanpetev: Not really a fan of those showing up in the API if there weren't there before, since that means they've lost their `@hidden` jsdoc tag probably in the move and will now show up in API docs and won't have the tags in TS definitions as well.

projects/igniteui-angular-elements/src/lib/grids/grid.component.ts
```diff
@@ -0,0 +1,166 @@
...
+    @Output()
+    public childrenResolved = new EventEmitter<void>();
+
+    /* contentChildren */
+    /* blazorTreatAsCollection */
+    /* blazorCollectionName: ActionStripCollection */
+    /* blazorCollectionItemName: ActionStrip */
+    /* ngQueryListName: actionStripComponents */
+    @ContentChildren(IgxActionStripToken)
+    public actionStripComponents: QueryList<IgxActionStripToken>;
```
damyanpetev: Yeah this bit - I'd absolutely keep this with `@hidden @internal` and the `blazorInclude` tag. These query props/collections are/were not meant to be public API and we only keep them as long as Blazor needs them (and it really shouldn't one day)

projects/igniteui-angular-elements/src/lib/grids/grid.component.ts
```diff
@@ -0,0 +1,166 @@
...
+ * The Ignite UI Grid is used for presenting and manipulating tabular data in the simplest way possible.  Once data
+ * has been bound, it can be manipulated through filtering, sorting & editing operations.
+ * @example
+ * ```html
+ * <igx-grid [data]="employeeData" [autoGenerate]="false">
+ *   <igx-column field="first" header="First Name"></igx-column>
+ *   <igx-column field="last" header="Last Name"></igx-column>
+ *   <igx-column field="role" header="Role"></igx-column>
+ * </igx-grid>
+ * ```
```
damyanpetev:
```suggestion
```
same for the others

projects/igniteui-angular-elements/src/lib/state.component.ts
```diff
@@ -35,7 +35,7 @@ export interface IGridStateInfo {
 /* jsonAPIManageCollectionInMarkup */
 /**
  * State component allows saving and restoring the state of the grid features.
- * @igxParent IgxGridComponent, IgxTreeGridComponent, IgxHierarchicalGridComponent, IgxPivotGridComponent, *
+ * @igxParent IgxGridElementsComponent, IgxTreeGridElementsComponent, IgxHierarchicalGridElementsComponent, IgxPivotGridComponent, *
```
damyanpetev: Oh, well that's awkward to say the least. This goes directly into the public API docs for Ignite UI for Angular, so not really great to list the elements class names.

Two options:
1. Make the parent resolve more flexible so old `IgxGridComponent` also recognizes the extended `IgxGridElementsComponent`
2. OR, bit a odd idea, but might work: Re-define `IgxGridComponent` as the extended class in Elements and simply alias the main component as something else:
   ```ts
   import { IgxGridComponent as IgxGrid } from 'igniteui-angular/grids/grid/src/grid.component';
   @Component({
       selector: 'igx-grid',
       templateUrl: '../../../../igniteui-angular/grids/grid/src/grid.component.html'
   })
   export class IgxGridComponent extends IgxGrid {
   ```

Honestly, the second option would mean we touch almost nothing in the main source and the config for elements too, merely change the import path for `IgxGridComponent` to come from elements instead.

projects/igniteui-angular-elements/src/lib/grids/hierarchical-grid.component.ts
```diff
@@ -0,0 +1,162 @@
...
+        IgxGridHierarchicalPipe,
+        IgxGridHierarchicalPagingPipe,
+        IgxStringReplacePipe,
+        IgxGridCellMergePipe,
+        IgxScrollInertiaDirective,
+        IgxGridUnmergeActivePipe
+    ],
+    schemas: [CUSTOM_ELEMENTS_SCHEMA]
+})
+export class IgxHierarchicalGridElementsComponent extends IgxHierarchicalGridComponent {
```
damyanpetev: One last concern I have is with the HGrid - I know we had functionality so toolbars and such as templates can discover their actual `igx-hierarchical-grid` parent that's not an Element. The selector is the same but the instance prototype won't be.
Not sure if the original Angular class missing from the elements config will cause issues with that functionality (must check carefully there)

projects/igniteui-angular-elements/src/lib/state.component.ts
MayaKirova: Why would this go in the public api docs for Angular? It's under the `projects/igniteui-angular-elements` and the components should only ship with wc.

projects/igniteui-angular-elements/src/lib/grids/hierarchical-grid.component.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    constructor() {
+        super();
+        this.selectionService.selectedRowsChange.pipe(takeUntil(this.destroy$)).subscribe((args: any[]) => {
+            this.selectedRowsChange.emit(args);
+        });
+    }
```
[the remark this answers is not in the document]
MayaKirova: Fixed it so that it exists only in elements.

projects/igniteui-angular-elements/src/lib/state.component.ts
damyanpetev: Okay, left the comment on the wrong file for this, but applies to action strip and grid columns, those tags end up here:
https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/igxactionstripcomponent.html
https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/igxcolumncomponent.html

and so on.

projects/igniteui-angular-elements/src/analyzer/elements.config.ts
MayaKirova: Removed.

projects/igniteui-angular-elements/src/lib/grids/hierarchical-grid.component.ts
MayaKirova: Seems to work for toolbar, at least in the dev sample:

image

image

Also for paginator:

image


image

projects/igniteui-angular-elements/src/lib/state.component.ts
MayaKirova: Seems to work with the aliases. Added the change.

projects/igniteui-angular-elements/src/analyzer/elements.config.ts
```diff
@@ -870,36 +876,168 @@ export var registerConfig = [
   {
     component: IgxRowIslandComponent,
     selector: "igc-row-island",
-    parents: [IgxHierarchicalGridComponent, IgxRowIslandComponent],
+    parents: [
+      IgxGridElementsComponent,
+      IgxHierarchicalGridElementsComponent,
+      IgxRowIslandElementsComponent,
+    ],
+    contentQueries: [],
```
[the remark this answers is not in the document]
MayaKirova: No longer applicable.

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -1172,19 +1158,13 @@ export abstract class IgxGridBaseDirective implements GridType,
...
-    /* blazorInclude */
-    /* blazorTreatAsCollection */
-    /* blazorCollectionName: ActionStripCollection */
-    /* blazorCollectionItemName: ActionStrip */
-    /* ngQueryListName: actionStripComponents */
     /** @hidden @internal */
-    @ContentChildren(IgxActionStripToken)
-    protected actionStripComponents: QueryList<IgxActionStripToken>;
+    @ContentChild(IgxActionStripToken)
+    public actionStripComponent: IgxActionStripToken;
```
damyanpetev: That was protected before, can we keep it as such? Would like to avoid adding more "internal public" props where possible.

projects/igniteui-angular-elements/src/analyzer/elements.config.ts
```diff
@@ -555,6 +557,7 @@ export var registerConfig = [
       },
     ],
     additionalProperties: [
+      { name: "childLayoutList", writable: true },
```
damyanpetev: Unless these are needed (they weren't exposed before), I'd really prefer if they stay hidden

projects/igniteui-angular-elements/src/analyzer/elements.config.ts
```diff
@@ -870,26 +874,30 @@ export var registerConfig = [
   {
     component: IgxRowIslandComponent,
     selector: "igc-row-island",
-    parents: [IgxHierarchicalGridComponent, IgxRowIslandComponent],
+    parents: [
+      IgxGridComponent,
```
damyanpetev: Yeah, that's off a bit, IgxGridComponent is not a valid parent or am I missing something?

damyanpetev (dismissed):

skrustev:
@MayaKirova It seems like the `columnsAutogenerated` event doesn't fire for the RowIslands in Elements. I tested it with a previous version of the WC Grid and it seems to be working. For some reason also after building in the analyzer the Hierarchical Grid event doesn't fire, even though it fires here in the samples. Not sure if I'm doing something wrong

MayaKirova:
@skrustev `columnsAutogenerated` should re-emit for row islands now.

damyanpetev (approved):

PR #15327, opened 2025-02-03
fix(igxTreeGrid): Re-eval correct row after collapsing. Add test.

### LINKED ISSUE #15322: IgxTreeGrid: Wrong parent record when adding new child row
## Description  
When a new child record is added, the parent record is wrong. The following precondtions should be met to reproduce this: 
1) all records must have been previously collapsed - either manually or by invoking the tree grid's `collapseAll()` method. 
2) the grid should be scrollable after expanding the records
3) the grid needs to be scrolled a bit after some records are expanded

 * igniteui-angular version: all supported
 * browser: any

## Steps to reproduce  

1. Open [this](https://www.infragistics.com/products/ignite-ui-angular/angular/components/treegrid/row-adding#angular-tree-grid-row-adding-example) demo
2. Collapse all records
3. Now expand "Victoria Lincoln" and "Yang Wang" so that the grid gets scrollable again.
4. Scroll the grid a bit.
5. Attempt to add a child row to "Yang Wang" through the action strip. As a result the target record is collapsed.
6. Observe where the add row UI is rendered. When editing is finished, the new record is added as a child to the previous record, and not the target one.

## Result  
Add row UI is not in the proper place and the new record is added as child to the record above.

Image

## Expected result  
Have the add row UI in the correct place and for the new record to be added as child to the target record.

## Attachments  

MayaKirova:
Closes #15322   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/icon-button/_icon-button-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/icon-button/_icon-button-theme.scss
@@ -369,7 +369,6 @@
                 content: '';
                 inset: 0;
                 box-shadow: inset 0 0 0 rem(1px) var-get($outlined-theme, 'border-color');
-                z-index: -1;
                 border-radius: inherit;
             }
 


projects/igniteui-angular/src/lib/grids/common/crud.service.ts
```diff
@@ -686,7 +686,12 @@ export class IgxGridCRUDService extends IgxRowAddCrudState {
         this.endEdit(true, event);
 
         if (parentRow != null && this.grid.expansionStates.get(parentRow.key)) {
-            this.grid.collapseRow(parentRow.key);
+            const key = parentRow.key;
+            this.grid.collapseRow(key);
+            // detect changes and get correct row by key
+            // in case collapsing changes row context.
+            this.grid.cdr.detectChanges();
```
mddragnev, line 693: Why do we really need to detect any changes here? The `dataView` used in `getRowByKey` is updated after the `collapseRow` method. Is there a specific case that this is fixing. Also, what do you think of handling this with a copy of the parentRow since it comes from the action strip context and might be changed. Thus, something like this 
` const row = Object.assign(copyDescriptors(parentRow),parentRow);`
And then use this `row` instance further so that it doesnt change?

MayaKirova: I added the change detection it as precaution since nothing in the collapse logic does change detection nor pipe triggers and I wanted to be sure everything is up-to-date.

Also I don’t mind using a copy. I’m somewhat concerned with copying the whole IgxTreeGridRowComponent, since it’s a whole component object and the cloning of some of the angular specific props might fail or be incomplete. I’d much rather work with what the IgxTreeGridRow object that the API returns instead. But that’s just my preference.

mddragnev: About the cloning the row object, we are cloning the `rowType` interface so I dont think there will be an issue but I undestand what are your concerns. I`m just not a big fan of invoking change detection in the middle of the method. 
Also, collapse logic does trigger change detection here: https://github.com/IgniteUI/igniteui-angular/blob/959435d7ab9b17720db1fd511349d5389fb92ef2/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts#L4174.

PR #15378, opened 2025-02-18
chore(*): bumping dependencies

mddragnev:
This is needed because of https://github.com/angular/angular/issues/59632 . We are using such self-referencing components in column groups and those are not working in  `19.1.2`

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [x] CI/CD

### Checklist:


kdinev (approved):

mddragnev:
> Funny thing... The new query builder had the same issue and we've already updated the deps in #14647 We've also tested on a separate application that even with 19.1.2 it was not affected, i.e. it's in the dev demos thankfully (prob quite outdated setup) so we don't need to update the _actual_ library dependencies to require greater than that version, otherwise those packages don't really affect end-users.
> 
> Can you also confirm this doesn't reproduce in a separate application?

Seems like it works just fine in a separate application using latest master of igniteui-angular and 19.1.2 version of angular

PR #15381, opened 2025-02-19
IgxQueryBuilderComponent: expose disableReturnFieldsChange & empty header title

teodosiah:
Closes #  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


teodosiah:
@ig-georgeA What should be the distance from above when there is no header?
image

ig-georgeA:
@desig9stein thoughts?

Any ideas on how we can achieve the following? Alternate suggestions are also welcome.

When header is present, the entity and field selectors are  separated from header by 1rem.  From what I could think of, the header should have no bottom padding, and the padding top comes from the `filter-tree_inputs`. 

image

When header is not present, it might look like this:

image

When used in the nested subquery, we don't need the top padding. coz it it comes from `igx-query-builder-tree igx-query-builder-tree--level-1`

image

When used in a dialog,  I take it dialog does not come with any padding .  It comes from the content, which in this case is the Query Builder component? Here we don't need the top padding on `igx-filter-tree__inputs` as well.

image

p.s.  I suspect we must have run into similar situations with alert vs. confirmation dialog? Or may be card header? Is there a pattern we can resuse?

desig9stein:
@teodosiah can we remove the header if it's empty or add a class to it when it is? 

I am unable to remove it only with styles since the combination of `:has` and `:not` pseudo selectors is not working in Safari
```SCSS
// This selector is not working in safari
// in chrome this selector sets display none to the header if it has empty title.
%advanced-filter__header {
  &:has(.igx-query-builder__title:empty) {
      display: none;
  }
}
```

desig9stein:
> @desig9stein thoughts?
> 
> Any ideas on how we can achieve the following? Alternate suggestions are also welcome.
> 
> When header is present, the entity and field selectors are separated from header by 1rem. From what I could think of, the header should have no bottom padding, and the padding top comes from the `filter-tree_inputs`.
> 
> image
> When header is not present, it might look like this:
> 
> image
> When used in the nested subquery, we don't need the top padding. coz it it comes from `igx-query-builder-tree igx-query-builder-tree--level-1`
> 
> image
> When used in a dialog, I take it dialog does not come with any padding . It comes from the content, which in this case is the Query Builder component? Here we don't need the top padding on `igx-filter-tree__inputs` as well.
> 
> image
> p.s. I suspect we must have run into similar situations with alert vs. confirmation dialog? Or may be card header? Is there a pattern we can resuse?

@ig-georgeA  If there is no header, the padding remains unchanged. I am simply adjusting the container on which the padding is applied. Please test and confirm if this aligns with the expected behavior. When used within a dialog, the padding remains consistent with its behavior outside the dialog. We have chosen not to override any default padding of the Query Builder within the standard dialog, leaving that customization to the user.

ig-georgeA:
Did something get missed @desig9stein 

This does not look right.

image

desig9stein:
> Did something get missed @desig9stein
> 
> This does not look right.
> 
> image

run `npm install`

gedinakova (approved):

PR #15428, opened 2025-02-28
fix(pivot-grid): Handle currency pivot values with count aggregator

### LINKED ISSUE #15148: Pivot Grid data type `currency` is applied on `COUNT` aggregator as well
## Description  
When the `dataType` property for a pivot value is set to `currency`, it applies currency formatting to the `Count` aggregator as well, while this doesn't make sense in that case.

 * igniteui-angular version: 19.0.x
 * browser: any

## Steps to reproduce  

1. Run the attached sample
2. Change the `Value` aggregator to `Count`
3. Observe the data in the grid

## Result  
The aggregated data for `Count` has currency formatting.

## Expected result  
The aggregated data for `Count` shouldn't have any currency formatting.

## Attachments 
[currency_count.zip](https://github.com/user-attachments/files/18027467/currency_count.zip)

currency_count

IMinchev64:
Closes #15148 

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-state.service.ts
+++ b/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-state.service.ts
@@ -1,0 +1,22 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class IgxPivotGridStateService {
+    public currencyColumnSet: Set<string> = new Set();
+
+    constructor() {}
+
+    public addCurrencyColumn(columnName: string): void {
+        this.currencyColumnSet.add(columnName);
+    }
+
+    public removeCurrencyColumn(columnName: string): void {
+        this.currencyColumnSet.delete(columnName);
+    }
+
+    public isCurrencyColumn(columnName: string): boolean {
+        return this.currencyColumnSet.has(columnName);
+    }
+}
--- a/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-data-selector.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-data-selector.component.ts
@@ -47,6 +47,8 @@ import { IgxIconComponent } from "../../icon/icon.component";
 import { IgxInputGroupComponent } from "../../input-group/input-group.component";
 import { fadeIn, fadeOut } from 'igniteui-angular/animations';
 import { Size } from '../common/enums';
+import { GridColumnDataType } from '../../data-operations/data-util';
+import { IgxPivotGridStateService } from './pivot-grid-state.service';
 
 interface IDataSelectorPanel {
     name: string;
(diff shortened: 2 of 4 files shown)


projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-grid-state.service.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class IgxPivotGridStateService {
```
MayaKirova, line 6: I don't think we need a whole new service on the root level. The pivot selector already has a reference to the related pivot and you can get the information you need from the config or modify it as needed directly. There's no need to go through a global service.

The service, as it is now, might even cause more issues, since it would be a single shared instance between grids on the page but it doesn't hold a reference to its owner. So if there are multiple grids with the same memberName it might return incorrect results (result added from another pivot on the page).

MayaKirova (changes requested):

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-util.ts
```diff
@@ -508,5 +508,28 @@ export class PivotUtil {
         }
     }
 
+    public static handleCountAggregator(columns: ColumnType[], value: IPivotValue, isSingleValue: boolean): void {
+        const isCountAggregator = value.aggregate.aggregator?.name?.toLowerCase() === 'count' || value.aggregate.aggregatorName?.toLowerCase() === 'count';
 
+        if ((value.dataType === GridColumnDataType.Currency || value.dataType === GridColumnDataType.Percent) && isCountAggregator) {
+            columns.forEach(column => {
+                console.log(column.field?.includes(value.member))
```
MayaKirova: Remove console log.

projects/igniteui-angular/src/lib/grids/pivot-grid/pivot-util.ts
```diff
@@ -508,5 +508,28 @@ export class PivotUtil {
         }
     }
 
+    public static handleCountAggregator(columns: ColumnType[], value: IPivotValue, isSingleValue: boolean): void {
+        const isCountAggregator = value.aggregate.aggregator?.name?.toLowerCase() === 'count' || value.aggregate.aggregatorName?.toLowerCase() === 'count';
 
+        if ((value.dataType === GridColumnDataType.Currency || value.dataType === GridColumnDataType.Percent) && isCountAggregator) {
```
MayaKirova, line 523: These checks are a bit repetitive as is the code in the ```if``'s. The same ones are in `updateColumnDataTypeByAggregator` and in `generateColumnHierarchy`. Maybe just extract it in a common util method that gets the column type for a given PivotValue and reuse it. It would also make the code a bit more readable:

```
public static updateColumnTypeByAggregator(columns: ColumnType[], value: IPivotValue, isSingleValue: boolean): void {
  const targetColumnType = getColumnDataTypeForValue(value);   
   columns.forEach(column => {
                  if (column.field?.includes(value.member) || isSingleValue) {
                      column.dataType = targetColumnType;
                  }  
              });
}
```

MayaKirova (changes requested):

MayaKirova (approved):

MayaKirova (approved):

PR #15449, opened 2025-03-04
feat(igxGrid): Apply min/max width constraints on user-set and auto-w…

### LINKED ISSUE #15383: The minWidth property of the columns is not working on inital render of the grid
## Description  
The minWidth property of the columns is not working on inital render of the grid, it's only being applied if the resizing of the columns is used

 * igniteui-angular version: 18.2.*
 * browser: N\A

## Steps to reproduce  

1. Open the attachment
2. `ng serve`
3. Open the page on mobile view

## Result  
The column `ContactName` has `minWidth` applied of 80px, even though it's getting smaller than that.
Image

## Expected result  
The `minWidth` property should applied when the widths of the columns are calculated initially.
The same applies to `maxWidth` property.

MayaKirova:
…idths.

Closes #15383  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@ChronosSF @dkamburov @damyanpetev @kdinev

A question arose around one of the existing tests:
https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-resizing

In that test the scenario is the following:

```
<igx-column [field]="'ID'" [width]="'50px'" [header]="'ID'" [filterable]="true"></igx-column>
```

Small width, no user set min-width. Previously this allowed the user to set such a width initially. Resulting in a column with 50px width.

However for user interaction, there's a defaultMinWidth, based on the current density:
https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/src/lib/grids/columns/column.component.ts#L1185
Which gets applied by default if there's no user set min-width. Hence you cannot resize a column below that threshold.

For instance, here: https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/column-resizing#angular-grid-column-resizing-example
First column has no min-width, but resizing is still constrained to 80px (the defaultMinWidth for that density).

I believe these defaults had something to do with the different paddings in the cells, which are different for the different densities. Since below a certain threshold with the paddings you can no longer see the content of the cell and it's no longer a meaningful render. @simeonoff might have more insight. 

**So the question is: Should this defaultMinWidth be applied as constraint to the initially set width of the column or not?**

If it does, this might be a breaking change, since user that have set a small width initially, with no min-width, will render the column with default min instead. For example, user has set 50px width, but since defaultMinWidth  is 80, the column will be 80px when rendered.

Let me know what you think.

kdinev:
I don't think this `defaultMinWidth` should be applied in this scenario, but I'll let others comment. Currently if width is set in % and the width is smaller than this `defaultMinWidth`, then it doesn't get applied.

kdinev (approved):

dkamburov:
@MayaKirova please add a note to the changelog

dkamburov (approved):

Issue #15500, opened 2025-03-11
Assigning a filteringExpressionsTree created by parsing a JSON string crashes


gedinakova:
* igniteui-angular version:  19.1
 * browser: N/A

## Steps to reproduce  

1. Attempt to bind the grid's filteringExpressionsTree to an object parsed from a JSON

## Result  
An error is thrown in the console.
TypeError: cond.logic is not a function    at FilteringStrategy.findMatchByExpression (infragistics-igniteu…ngular.mjs:19275:17)     at FilteringStrategy.matchRecord (infragistics-igniteu…ngular.mjs:19304:21)     at FilteringStrategy.matchRecord (infragistics-igniteu…ngular.mjs:19286:33)     at FilteringStrategy.filter (infragistics-igniteu…ngular.mjs:19420:64)     at FilterUtil.filter (infragistics-igniteu…ngular.mjs:19266:27)     at _IgxGridFilteringPipe.transform (infragistics-igniteu…gular.mjs:100832:31)     at IgxGridComponent_Template (infragistics-igniteu…ular.mjs:103824:332)

## Expected result  
The filtering operation should be successful.

PR #15506, opened 2025-03-12
fix(grid): fixed advancedFilteringExpressionsTree rehydration #15500

Otixa:
Closes #  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
@@ -6671,6 +6671,9 @@ export abstract class IgxGridBaseDirective implements GridType,
             if (this._columns && this._filteringExpressionsTree) {
                 this._filteringExpressionsTree = recreateTreeFromFields(this._filteringExpressionsTree, this._columns) as IFilteringExpressionsTree;
             }
+            if (this._columns && this._advancedFilteringExpressionsTree) {
+                this._advancedFilteringExpressionsTree = recreateTreeFromFields(this._advancedFilteringExpressionsTree, this._columns) as IFilteringExpressionsTree;
+            }
         }
 
         this.initColumns(this._columns, (col: IgxColumnComponent) => this.columnInit.emit(col));


gedinakova (changes requested):
We need a test for this.

projects/igniteui-angular/src/lib/grids/grid/grid-filtering-advanced.spec.ts
```diff
@@ -1389,6 +1389,22 @@ describe('IgxGrid - Advanced Filtering #grid - ', () => {
             // Verify no filtered data
             expect(grid.filteredData).toBe(null);
         }));
+
+    });
+
+    describe('Expression tree serialization - ', () => {
+        it('should correctly filter with a deserialized expression tree.', fakeAsync(() => {
+            const errorSpy = spyOn(console, 'error');
+            let fix = TestBed.createComponent(IgxGridAdvancedFilteringSerializedTreeComponent);
```
hanastasov, line 1398: I dont see anything by searching for `IgxGridAdvancedFilteringSerializedTreeComponent`, and tests also fail because of that.

hanastasov (changes requested):

hanastasov:
Is this PR still in progress? Because it has the verified badge, but commits are made which are not related to some comment requesting changes.

For the last changes, maybe they are OK but I dont see why such change is required:

image

Obviously it does not bring along any change in how the code works with those interfaces, so it made me question if this is neccesary.

gedinakova:
> Is this PR still in progress? Because it has the verified badge, but commits are made which are not related to some comment requesting changes.
> 
> For the last changes, maybe they are OK but I dont see why such change is required:
> 
> image
> 
> Obviously it does not bring along any change in how the code works with those interfaces, so it made me question if this is neccesary.

Yes, they are necessary and should cover the scenario where the expression tree is set as follows (this is coming from a sample provided by PwnJS):
advancedFilteringExpressionsTree2: IExpressionTree =	{
    "filteringOperands": [
      {
        "fieldName": "productID",
        "condition": {
          "name": "equals",
          "isUnary": false,
          "iconName": "filter_equal"
        },
        "conditionName": "equals",
        "ignoreCase": true,
        "searchVal": 33,
        "searchTree": null
      }
    ],
    "operator": 0,
    "returnFields": []
  };

hanastasov:
Thanks Gale for the explanation.

gedinakova (approved):

projects/igniteui-angular/src/lib/data-operations/filtering-expression.interface.ts
```diff
@@ -14,9 +14,9 @@ export enum FilteringLogic {
  */
 export declare interface IFilteringExpression {
     fieldName: string;
-    condition?: IFilteringOperation;
-    conditionName?: string;
-    searchVal?: any;
-    searchTree?: IExpressionTree;
+    condition?: IFilteringOperation | null;
+    conditionName?: string | null;
+    searchVal?: any | null;
```
damyanpetev, line 19: any is already null? :) Kinda redundant on this one.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -1856,8 +1856,8 @@ export abstract class IgxGridBaseDirective implements GridType,
             }
 
             value.type = FilteringExpressionsTreeType.Regular;
-            if (value && this.columns) {
-                this._filteringExpressionsTree = recreateTreeFromFields(value, this.columns) as IFilteringExpressionsTree;
+            if (value && this._columns && this._columns.length > 0) {
```
damyanpetev, line 1859: Also could be just `this._columns?.length` but by the looks of it it's always initialized and not checked elsewhere, so:
```suggestion
            if (value && this._columns.length > 0) {
```

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6668,9 +6675,12 @@ export abstract class IgxGridBaseDirective implements GridType,
...
             this._columns = this.getColumnList();
-            if (this._columns && this._filteringExpressionsTree) {
-                this._filteringExpressionsTree = recreateTreeFromFields(this._filteringExpressionsTree, this._columns) as IFilteringExpressionsTree;
-            }
+        }
+        if (this._columns && this._filteringExpressionsTree) {
+            this._filteringExpressionsTree = recreateTreeFromFields(this._filteringExpressionsTree, this._columns) as IFilteringExpressionsTree;
+        }
+        if (this._columns && this._advancedFilteringExpressionsTree) {
+            this._advancedFilteringExpressionsTree = recreateTreeFromFields(this._advancedFilteringExpressionsTree, this._columns) as IFilteringExpressionsTree;
```
damyanpetev, line 6683: At some point we might want to dedupe this into a private function we can call in all places :)

damyanpetev (changes requested):
Eh, sure - I suppose the props on the filtering expression/tree technically accept nullish and some odd scenario in an app _could_ evaluate to a `null` type, but generally shouldn't happen. Don't mind it typed as such regardless.

The change request mostly for the type on return fields, the rest of the comments are merely small improvements.

projects/igniteui-angular/src/lib/data-operations/filtering-expressions-tree.ts
```diff
@@ -11,9 +11,9 @@ export enum FilteringExpressionsTreeType {
 export declare interface IExpressionTree {
     filteringOperands: (IExpressionTree | IFilteringExpression)[];
     operator: FilteringLogic;
-    fieldName?: string;
-    entity?: string;
-    returnFields?: string[];
+    fieldName?: string | null;
+    entity?: string | null;
+    returnFields?: (string | null)[];
```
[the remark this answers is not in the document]
Otixa: Actually one of the test cases provided was `null`s in the array, but I suppose having the entire array as null should also be considered so... `returnFields?: (string | null)[] | null`? 😐

projects/igniteui-angular/src/lib/data-operations/filtering-expression.interface.ts
Otixa: Fair

gedinakova (approved):

gedinakova (approved):

damyanpetev (approved):

hanastasov (approved):

PR #15523, opened 2025-03-14
Fix filtering expression throwing error with no columns. Related fixes for elements Blazor grid.

### LINKED ISSUE #15527: Setting Filtering Expression Tree without condition function throws error if columns are not yet defined
## Description  
When specifying `filteringExpression` or `advancedFilteringExpressionsTree`, the user should be able to specify only `conditionName` per spec without the need for `condition` now. But if there are not `columns` yet defined for the grid it throws error, event if it works after the columns are added.

 * igniteui-angular version: 19.1.x
 * browser: any

## Steps to reproduce  

1. Open the `grid-external-filtering` demo
2. Remove all cases where `condition` property is used and leave `conditionName` from filtering expression trees.
3. Set `columns` to empty array 
4. Run the demo
5. (optional) Add the columns using `setTimeout` to delay them. This works if the error is not thrown/removed from code.

## Result  
```
hook.js:608 ERROR Error: Wrong `conditionName`, `condition` or `field` provided!
    at recreateExpression (expressions-tree-util.ts:153:15)
    at recreateTreeFromFields (expressions-tree-util.ts:215:41)
    at set advancedFilteringExpressionsTree (grid-base.directive.ts:1909:54)
    at propDesc.set (watch-changes.ts:24:32)
    at _GridExternalFilteringComponent.ngAfterViewInit (grid-external-filtering.sample.ts:41:20)
    at callHookInternal (core.mjs:4195:10)
    at callHook (core.mjs:4219:7)
    at callHooks (core.mjs:4179:9)
    at executeInitAndCheckHooks (core.mjs:4134:5)
    at refreshView (core.mjs:13970:11)
```

## Expected result  
There should be no errors and should be able to use `conditionName` only initially before `columns` are set.

skrustev:
Closes #15527

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
+++ b/projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
@@ -149,7 +149,7 @@ function recreateExpression(expression: IFilteringExpression, fields: FieldType[
         }
     }
 
-    if (!expression.condition && !expression.conditionName) {
+    if (!expression.condition && expression.conditionName) {
         throw Error('Wrong `conditionName`, `condition` or `field` provided! It is possible that there is a type mismatch between the condition type and field type.');
     }
 
--- a/projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
+++ b/projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
@@ -149,7 +149,7 @@ function recreateExpression(expression: IFilteringExpression, fields: FieldType[
         }
     }
 
-    if (!expression.condition && expression.conditionName) {
+    if (!expression.condition && !expression.conditionName) {
         throw Error('Wrong `conditionName`, `condition` or `field` provided!');
     }
 
--- a/projects/igniteui-angular/src/lib/data-operations/filtering-expression.interface.ts
+++ b/projects/igniteui-angular/src/lib/data-operations/filtering-expression.interface.ts
@@ -7,7 +7,6 @@ export enum FilteringLogic {
     Or
 }
 
-/* tsPlainInterface */
 /* marshalByValue */
 /**
  * Represents filtering expressions.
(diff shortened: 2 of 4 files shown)


projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
```diff
@@ -149,7 +149,7 @@ function recreateExpression(expression: IFilteringExpression, fields: FieldType[
         }
     }
 
-    if (!expression.condition && expression.conditionName) {
+    if (!expression.condition && !expression.conditionName) {
```
Otixa, line 152: This error appears when rehydration has failed. If a `conditionName`  is present and the `condition` has not been populated - it should throw.
This change would likely make the code silently fail when rehydration is unsuccessful and be challenging to debug.

Otixa (dismissed):

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
gedinakova: Most cases where rehydration would fail should be coverred in https://github.com/IgniteUI/igniteui-angular/pull/15506.

skrustev: @Otixa Why it would throw an error when it is supposed to work without `condition` set?

skrustev: Hm I will check again to see the expression I tested again, might be related to that why `condition` was invalid in this case.

Otixa (approved):

projects/igniteui-angular/src/lib/data-operations/expressions-tree-util.ts
skrustev: Lastly it seems like the linked PR resolves most of the issues. I've added only a few safety check if the columns array is an empty string in this case.

gedinakova (approved):

MayaKirova (approved):

PR #15556, opened 2025-03-20
feat(GridBase): Apply spec for server-side rendering for relative gri…

### LINKED ISSUE #15202: SSR Grid Improvements
When in SSR mode (or prerender) the correct behavior of the grid should be:

- render all elements declared in the template
- do not change sizes (% width should render as % width)
- render all defined columns without virtualization (auto-sizing should work as it does if there is no data)
- render empty grid template or loading indicator based on options

MayaKirova:
…d sizes.

Closes #15202  

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova:
@dkamburov @ChronosSF 

FYI, the grid will now retain 100% size (width/height) from the SSR render, however note that things inside the grid may still flicker during hydration on the client.

For example:
- Scrollbars may show, since we're going from an empty grid to a grid with data.
- Columns that have relative size (in %), may change their size after hydration (and as a result of that also spawn a scrollbar). This makes sense because they were originally measured to the size the sever has used for pre-render.

    So it looks similar to this, when hydration occurs:
grid_hydration

kdinev (approved):

mddragnev:
@MayaKirova If you have a grid width a fixed sizes ( let say 600px width and height) then the data should be rendered on the server but it doesnt. It renders on the client

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3363,12 +3363,26 @@ export abstract class IgxGridBaseDirective implements GridType,
...
+    protected get totalCalcWidth() {
+        return this.platform.isBrowser ? this.calcWidth : undefined;
+    }
+
+    protected get renderData() {
+        // omit data if not in the browser and size is %
+        return !this.platform.isBrowser && this.isPercentHeight ? undefined : this.data;
+    }
+
+    @HostBinding('style.display')
```
dkamburov, line 3375: Shouldn't those host bindings be in the scss in the themes

MayaKirova: They already are (you can find the same styles in the theme), but they don't get applied until the theme is fully loaded and they are necessary for the body content to stretch correctly in SSR. Hence why they are set inline as well.

dkamburov (approved):

ChronosSF (changes requested):
This PR needs to update the changelog. It's actually one of the features we want to promote for 19.2.

dkamburov (approved):

ChronosSF (approved):

PR #15564, opened 2025-03-21
Grid validate filtering expressions tree

### LINKED ISSUE #15563: Grid throws if passed an expressionTree with a field name that does not exist in Grid.
## Description  
Grid supports accepting a FilteringExpressionsTree with missing logic function. The grid will recreate the logic based on the field it refers to.

However, if such field does not exist in the grid data, condition logic will not be recreated (Expected). Still, the grid will try to call condigion.logic without verifying it does exist

 * igniteui-angular version: 19.1.4

## Steps to reproduce  

1. Add grid
2. Pass the following filteringExpressionsTree:
grid.advancedFilteringExpressionsTree = JSON.parse('{"filteringOperands":[{"fieldName":"**Missing**","condition":{"name":"notEmpty","isUnary":true,"iconName":"filter_not_empty"},"conditionName":"notEmpty","ignoreCase":true,"searchVal":null,"searchTree":null}],"operator":0,"returnFields":[],"type":1}')

3. Grid throws:
cond.logic is not a function

## Result  
the filtering strategy should return false when trying to find a value in a field, which does not exist. Should try to call the condigion logic function to check a value.

Image

## Expected result  
Grid should appear empty, because for every record, false would have been returned by the filtering strategy
AdvancedFilteringDialog should still display the filtering expression:

Image

hanastasov:
Closes #15563 

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/data-operations/filtering-strategy.spec.ts
+++ b/projects/igniteui-angular/src/lib/data-operations/filtering-strategy.spec.ts
@@ -49,6 +49,20 @@ describe('Unit testing FilteringStrategy', () => {
         const res = fs.matchRecord(rec, expressionTree);
         expect(res).toBeTruthy();
     });
+    it ('tests `matchRecordByExpressions` for working with filtering operands with missing condition', () => {
+        const rec = data[0];
+        const expressionTree = new FilteringExpressionsTree(FilteringLogic.Or);
+        expressionTree.filteringOperands = [
+            {
+                conditionName: 'contains',
+                fieldName: 'string',
+                ignoreCase: false,
+                searchVal: 'ROW'
+            }
+        ];
+        const res = fs.matchRecord(rec, expressionTree);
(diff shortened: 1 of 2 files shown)


gedinakova:
@Otixa Hristo's scenario is actually a little different. The idea is that there's an option that a field is let's say no longer available. This would lead to breaking the user's app because the expression tree has not been checked & updated accordingly. Maybe it's not that bad to simply show an empty grid which would hint them into the direction there's something wrong with the filters.

Otixa (approved):

hanastasov:
Oh i;m such a rookie - fix should be "

```
        if (expr.condition.logic) {
            return expr.condition?.logic(val, expr.searchVal, expr.ignoreCase);
        }
```

will make a new pr

PR #15568, opened 2025-03-24
fix(grid): apply conditon on condition.logic, not on condition

### LINKED ISSUE #15563: Grid throws if passed an expressionTree with a field name that does not exist in Grid.

hanastasov:
Closes #15563

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/data-operations/filtering-strategy.ts
+++ b/projects/igniteui-angular/src/lib/data-operations/filtering-strategy.ts
@@ -40,7 +40,9 @@ export abstract class BaseFilteringStrategy implements IFilteringStrategy  {
     // protected
     public findMatchByExpression(rec: any, expr: IFilteringExpression, isDate?: boolean, isTime?: boolean, grid?: GridType): boolean {
         const val = this.getFieldValue(rec, expr.fieldName, isDate, isTime, grid);
-        return expr.condition?.logic(val, expr.searchVal, expr.ignoreCase);
+        if (expr.condition.logic) {
+            return expr.condition?.logic(val, expr.searchVal, expr.ignoreCase);
+        }
     }
 
     // protected
--- a/projects/igniteui-angular/src/lib/data-operations/filtering-strategy.spec.ts
+++ b/projects/igniteui-angular/src/lib/data-operations/filtering-strategy.spec.ts
@@ -51,15 +51,9 @@ describe('Unit testing FilteringStrategy', () => {
     });
     it ('tests `findMatchByExpression` for working with filtering operands with missing condition', () => {
         const rec = data[0];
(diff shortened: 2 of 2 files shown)


projects/igniteui-angular/src/lib/data-operations/filtering-strategy.ts
```diff
@@ -40,7 +40,9 @@ export abstract class BaseFilteringStrategy implements IFilteringStrategy  {
     // protected
     public findMatchByExpression(rec: any, expr: IFilteringExpression, isDate?: boolean, isTime?: boolean, grid?: GridType): boolean {
         const val = this.getFieldValue(rec, expr.fieldName, isDate, isTime, grid);
-        return expr.condition?.logic(val, expr.searchVal, expr.ignoreCase);
+        if (expr.condition.logic) {
```
gedinakova, line 43:
```suggestion
        if (expr.condition.logic) {
```
```suggestion
        if (expr.condition?.logic) {
```

gedinakova, line 43: Well, if there's no condition, this would crash. Maybe you meant something like:
```suggestion
        if (expr.condition?.logic) {
```

gedinakova (changes requested):

projects/igniteui-angular/src/lib/data-operations/filtering-strategy.ts
hanastasov: done, thanks

Issue #15461, opened 2025-03-06
crypto.randomUUID() does not work unless https is enabled


xdinterface:
## Description  
We recently updated our ignite version and i saw that [this refactor](https://github.com/IgniteUI/igniteui-angular/pull/15049) changed the way rowIds in a grid are generated.

While i think it's great to move away from external packages if they are not necessary this change is breaking applications that are not using secure context.

While most application use secure context/https anyways and it still works on the localhost, i would like my application to function even when the https has to fallback to http for whatever reason.

 * igniteui-angular version:     "@infragistics/igniteui-angular": "19.0.6",
 * browser: Firefox & Chrome

## Steps to reproduce  

1. Disable or do not set up https
2. try adding a row in a grid that uses crypto.randomUUID (tree-grid for example)

## Result  
```
chunk-IUX3FXRM.js:7 ERROR TypeError: crypto.randomUUID is not a function
    at n.generateRowID (chunk-7MNBVYAO.js:46:217020)
    at n.getEmptyRecordObjectFor (chunk-7MNBVYAO.js:47:272979)
    at n.createAddRow (chunk-7MNBVYAO.js:45:108177)
    at n.enterAddRowMode (chunk-7MNBVYAO.js:45:110660)
    at n.beginAddRowForIndex (chunk-7MNBVYAO.js:47:6235)
    at n._addRowForIndex (chunk-7MNBVYAO.js:47:5774)
    at n.beginAddRowByIndex (chunk-7MNBVYAO.js:47:272024)
    at r.addRow (chunk-LCS62TL7.js:23:2675)
    at chunk-LCS62TL7.js:10:217
    at $c (chunk-IUX3FXRM.js:7:103118)
```

## Expected result  
Adding a row should function at any time, even when secure context is disabled.

## Attachments  
[https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)
[https://github.com/IgniteUI/igniteui-angular/pull/15049](https://github.com/IgniteUI/igniteui-angular/pull/15049)

xdinterface:
If i can make a suggestion you could use getRandomValues instead and generate a uuid from that.

```
function generateUUID(): string {
    const randomBytes = new Uint8Array(16);

    crypto.getRandomValues(randomBytes);

    randomBytes[6] = (randomBytes[6] & 0x0f) | 0x40;
[...]

PR #15597, opened 2025-03-25
fix(*): adding random function for unsecure context #15461

ChronosSF:
Closes #  

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/common/random.ts
+++ b/projects/igniteui-angular/src/lib/grids/common/random.ts
@@ -1,0 +1,22 @@
+/**
+ * Use the function to get a random UUID string when secure context is not guaranteed making crypto.randomUUID unavailable.
+ * @returns A random UUID string.
+ */
+export function getUUID() {
+    if (typeof crypto.randomUUID === "function") {
+        return crypto.randomUUID();
+    }
+    // Secure fallback using crypto.getRandomValues()
+    const bytes = new Uint8Array(16);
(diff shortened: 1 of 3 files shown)


projects/igniteui-angular/src/lib/grids/common/random.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Use the function to get a random UUID string when secure context is not guaranteed making crypto.randomUUID unavailable.
+ * @returns A random UUID string.
+ */
+export function getUUID() {
+    if (typeof crypto.randomUUID === "function") {
```
damyanpetev, line 6: Minor nit, but per the repo settings we use single quotes (goes for the rest in the file obv). Also do set the return type instead of the jsdoc tag since we do use TS. You can be even more specific and type it exactly like `randomUUID` is with the pattern if you'd like.
```suggestion
 * Get a random UUID string. Uses `crypto.randomUUID` if available or a fallback implementation when not in secure context.
 */
export function getUUID(): string {
    if (typeof crypto.randomUUID === 'function') {
```

damyanpetev: PS: Don't mind this getting release and switched later too :)

damyanpetev (approved):

PR #15649, opened 2025-04-02
fix(elements): Handle template setting of components passed as event args through proxy. 

### LINKED ISSUE #15615: Applying headerTemplate in columnInit event breaks the pivot grid in Elements
## Description  
Applying  `headerTemplate` to a columns in the `columnInit` event arguments breaks the Pivot Grid. Possibly other grids as well.

 * igniteui-angular version: 19.1.x
 * browser: any

## Steps to reproduce  

1. Open the elements demos index.thml
2. Add to the first grid:
```
        grid1.addEventListener("columnInit", (args) => {
            const col = args.detail;
            col.headerTemplate = (ctx) => html`<span>${col.field}AAAA</span>`;
        });
```
3. Run the demos

## Result  
Observe the console and the first rendered grid. The following error is thrown:

```
hook.js:608 ERROR TypeError: r.createEmbeddedViewImpl is not a function
    at XR.createEmbeddedView (combined.js:6365:50)
    at _n39.ngOnChanges (combined.js:9811:27)
    at _n39.tB (combined.js:3604:30)
    at dE (combined.js:3902:7)
    at TB (combined.js:3909:97)
    at O2 (combined.js:3896:65)
    at Cf (combined.js:3883:3)
    at bR (combined.js:5206:19)
    at u (combined.js:5201:3)
    at Toe (combined.js:37483:618)
```

## Expected result  
There should be no errors and the headers template be applied.

## Attachments  

Image

skrustev:
Closes #15615

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


MayaKirova (approved):

projects/igniteui-angular-elements/src/app/custom-strategy.ts
```diff
@@ -388,6 +388,65 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
...
+                const emitter: EventEmitter<any> = componentRef.instance[propName];
+                return emitter.pipe(map((value: any) => ({ name: templateName, value: this.patchOutputComponents(value) })));
+            },
+        );
+
+        (this as any).eventEmitters.next(eventEmitters);
+    }
+
+    protected patchOutputComponents(eventArgs: any) {
+        let componentConfig: ComponentConfig = this.findConfig(eventArgs);
```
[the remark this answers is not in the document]
skrustev: I've added a handling that should cover different variations of object passed as argument.
Was thinking about what you are saying a proxy that potentially spawns other proxies for nested components, but maybe we shouldn't encourage user to set template on such nested items. Most likely if something can be assigned template is available in the event args.

projects/igniteui-angular-elements/src/app/custom-strategy.ts
```diff
@@ -403,26 +405,33 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
...
+     * Nested search of event args that contain angular components and replace them with proxies.
+     * If event args are array of angular component instances should return array of proxies of each of those instances.
+     * If event args are object that has a single property being angular component should return same object except the angular component being a proxy of itself.
+     */
+    protected createProxyForComponentValue(value: any, depth: number) {
+        if (depth > this.maxEventProxyDepth) {
+            return value;
+        }
+
+        if (value?.__ngContext__) {
```
damyanpetev, line 421: Okay, I agree we probably need some way to avoid checking _everything_ against the config, but if this stays in (due to lack of alternative) it should stay with a note and/or TODO about `__ngContext__` which is not public API and thus not stable, so not something we'd like to keep as permanent solution.

Also, somewhat expected to have `get` handled as well to create the nested proxies on-demand as options are being accessed, otherwise we're going through the entire structure (which can be heavy I suppose) for no reason. Which is why the 1/2 depth was fine for starting point and the fix :)

PS: Also thinking, if we can't efficiently proxy props, we might want to have some part of the analyzer scan the event args and really the entire public API surface. That could give us explicit paths if you will down specific events' args we need to patch

skrustev: We've decided to be as is for now. For further more generic implementation that will include methods that access components like IgxColumn could be a further improvement, that will be nice to look into and probably sooner or later will need to be looked into.

projects/igniteui-angular-elements/src/app/custom-strategy.ts
```diff
@@ -405,6 +407,73 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
...
+            if (componentConfig?.templateProps) {
+                return this.createElementsComponentProxy(value, componentConfig);
+            }
+        } else if (Array.isArray(value)) {
+            return value.map(item => this.createProxyForComponentValue(item, depth + 1))
+        } else if (typeof value === "object" && Object.entries(value).length) {
+            for (const [key, item] of Object.entries(value)) {
+                value[key] = this.createProxyForComponentValue(item, depth + 1);
+            }
+        }
```
damyanpetev, line 450: Not a huge fan of the re-map of arrays (could possibly hit some big ones that are pretty irrelevant like data or selection stuff). I have some concern those reference recreations could cause trouble, besides the performance implications (on the basic samples - from what I see `createProxyForComponentValue` took around .8ms which isn't a lot but also not nothing; still probably acceptable).

What's curious is the amount, despite targeting pretty much 1 event only:
image
So... yeah, that's not great. The bulk of the patching is going over internal events and args that don't need which is honestly something we should address on the library level if we can avoid those semi-public outputs.

I'm seriously considering if we should limit the proxies to just specific event(s) like columnInit for now.

skrustev: Hm ok, I will all a check to the array handling to be of only depth 1 and if the first item returned is a Proxy generated from us. If not then no need to map the whole array.

I looked into the grid events and there are a lot of arrays indeed that don't need parsing, so that should cover it.

skrustev: Yeah maybe we should have a list in the config with the public events and in general handle only those?

damyanpetev (approved):

PR #15688, opened 2025-04-07
fix(igxGrid): Use default scheduler when throttling.

### LINKED ISSUE #15644: [SSR]: Setting column width to auto in a grid within SSR project throws an error and shut down the process
## Description  
If you set column width to auto in angular SSR grid then the whole application stops

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1.  Open the attached sample
2. npm install
3. npm run start

## Result  
An error is thrown and the application stops

## Expected result  
The app should work correct and the column should be auto sized

## Attachments  

[ig-cli-ssr.zip](https://github.com/user-attachments/files/19551281/ig-cli-ssr.zip)

MayaKirova:
Closes #15644   

Use default scheduler instead of explicitly using the animationFrameScheduler since it throws errors in SSR env.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -3751,7 +3751,7 @@ export abstract class IgxGridBaseDirective implements GridType,
 
         // notifier for column autosize requests
         this._autoSizeColumnsNotify.pipe(
-            throttleTime(0, animationFrameScheduler, { leading: false, trailing: true }),
+            throttleTime(0),
```
skrustev: Since this is mainly related to SSR, in this case I think would be better to use the `isBrowser` util check and otherwise provide undefined scheduler. Might not make much difference in this case, but to be sure, since `calculateGridSizes` can be called multiple times initially.

skrustev: From testing it a bit further it seems like the default scheduler actually has a bit more delay between calls when throttled with 0, compared to the `animationFrameScheduler`, which I am not sure is either beneficial or not, but there are some differences in how quickly the column resizing reacts.

PR #15706, opened 2025-04-09
fix(grid): strict string typing the grid paging mode

ChronosSF:


projects/igniteui-angular/migrations/update-11_1_0/index.spec.ts
```diff
@@ -708,47 +708,4 @@ export class CsvExportComponent {
             )
         ).toEqual(expectedContent);
     });
-
-    it('should update GridPagingMode enum from lowerCase to TitleCase', async () => {
```
damyanpetev, line 712: Assuming these don't work because the resolution is based on the TypeScript LS checking `GridPagingMode` and verifying it's an igniteui-angular export, the member is based pure string match. Perhaps we have a class declaration or enum filter that fails the check for the new const. Might want to check, even though not really critical for this PR

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1422,7 +1422,8 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
         return this.grid.dataView
             .map((rec, index) => {
                 if (!this.grid.isGroupByRecord(rec) && !this.grid.isSummaryRow(rec)) {
-                    this.grid.pagingMode === 1 && this.grid.page !== 0 ? index = index + this.grid.perPage * this.grid.page : index = this.grid.dataRowList.first.index + index;
+                    this.grid.pagingMode === GridPagingMode.Remote && this.grid.page !== 0 ?
```
damyanpetev, line 1425: Just use the actual value, the constant is there mostly for backwards compat
```suggestion
                    this.grid.pagingMode === 'remote' && this.grid.page !== 0 ?
```
Same for the other places ofc

ChronosSF:
TODO: check `StateDirective` restoring of the new property type.

damyanpetev (approved):
@ChronosSF Also changelog, since it should be noted the input on the grid can now be set in markup directly as `'remote'` instead of importing in the enum (requires extra prop/code).
Otherwise, LGTM.

PR #15813, opened 2025-05-15
fix(tooltip): make touch events passive - 19.2.x

### LINKED ISSUE #14518: [igxTooltipTarget] [performance] Permanently attached touch document listeners can cause the UI to become unresponsive with large number of instances
## Description 
Followup of #7638 which only addressed key events.
As suspected _touch events_ present the same problem.

 * igniteui-angular version: 18.0.4
 * browser: Chrome Mobile or Touchscreen (or emulated)

## Steps to reproduce  

[Manual]
1. Create a 1000 items with tooltip target directive
2. Run in mobile browser, touchscreen or emulate with dev tools
3. Touch any item

[Stackblitz]
1. Open in Chrome: https://stackblitz.com/edit/b9ujjo?file=src%2Fapp%2Ftooltip-simple%2Ftooltip-simple.component.html
2. Pop out the preview to a different tab
3. Open the devtools and emulate mobile (see screenshot)
4. Touch any item, or the Test checkbox - slow
5. Uncheck `use tooltips` checkbox
6. Try the Test checkbox (it's now fast).

## Result  
There's a big delay to show the tooltip, the UI is unresponsive the whole time.

## Expected result  
The tooltip to open immediately

## Attachments  
image

RivaIvanova:
Closes #14518  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts
```diff
@@ -110,6 +120,19 @@ export class IgxTooltipDirective extends IgxToggleDirective {
         @Optional() navigationService: IgxNavigationService) {
         // D.P. constructor duplication due to es6 compilation, might be obsolete in the future
         super(elementRef, cdr, overlayService, navigationService);
+
+        this.opening.pipe(first(), takeUntil(this._destroy$)).subscribe(() => {
+            document.addEventListener('touchstart', this.onDocumentTouchStart, { passive: true });
```
pmoleri, line 125: I don't like the idea of passing around the onDocumentTouchStart handler.

And I think that the current code is wrong.
1st target opens the tooltip -> adds event listener  (no one removes it)
...
Nth target opens the tooltip -> add event listener

When the tooltip gets destroyed only the last listener is detached.
And also, there's the missed opportunity of cleaning up after each target was left.

RivaIvanova: Actually, adding the event listener in the `opening` event of `this.target` (tooltip) will add it 1000 times, while the current approach adds it once. Even if the event listeners are removed in the `closing` event, the `onDocumentTouchStart` method will be called 1000 times before that. This also seems to throw the `[Violation] 'touchstart' handler took 1102ms` warning.

documentTouchStart


I'm not sure that adding/removing the document:touchStart event should be handled in the tooltipTarget since there are multiple tooltipTargets (purple squares) with a single tooltip.

projects/igniteui-angular/src/lib/directives/tooltip/tooltip-target.directive.ts
```diff
@@ -308,13 +310,17 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
                 event.cancel = true;
```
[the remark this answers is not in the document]
RivaIvanova: Yes, you are correct. I've added a check that prevents emitting the `tooltipHide` event for all the other targets.

projects/igniteui-angular/src/lib/directives/tooltip/tooltip.directive.ts
pmoleri: It can be checked that only the specific target opens it.

I don't think that you have addressed the issue of leaking listeners.
You attach on each open event, but you only detach on destroy.

projects/igniteui-angular/src/lib/directives/tooltip/tooltip-target.directive.ts
```diff
@@ -300,21 +303,40 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
         this._overlayDefaults.closeOnOutsideClick = false;
         this._overlayDefaults.closeOnEscape = true;
 
+        this.target.opening.pipe(takeUntil(this.destroy$)).subscribe(() => {
+            if (this.target.tooltipTarget === this) {
+                document.addEventListener('touchstart', this.onDocumentTouchStart = this.onDocumentTouchStart.bind(this), { passive: true });
```
pmoleri, line 308: It's wrong to do `this.onDocumentTouchStart = this.onDocumentTouchStart.bind(this)`
inside the event handler. You will be creating a brand new instance each time it opens. You can bind just once in the constructor or on init.

projects/igniteui-angular/src/lib/directives/tooltip/tooltip-target.directive.ts
```diff
@@ -300,21 +303,40 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
...
             const hidingArgs = { target: this, tooltip: this.target, cancel: false };
             this.tooltipHide.emit(hidingArgs);
 
             if (hidingArgs.cancel) {
                 event.cancel = true;
             }
         });
+
+        this.target.forceClosed.pipe(takeUntil(this.destroy$)).subscribe(() => {
+            document.removeEventListener('touchstart', this.onDocumentTouchStart);
```
pmoleri, line 327: I'm not sure if it can be expensive to call removeEventListener from all targets, when only one is needed.

pmoleri (approved):

PR #15877, opened 2025-06-03
Elements child component projection & DOM manipulation

damyanpetev:
Current projection in Elements moves some traditionally non-visible children of the Grid in a specialized container:
image
Which moves them after the Grid template initializes and as such causes API use like `parent.removeChild` and `parent.insertBefore` to fail.

Moving those to the grid root requires some preventive styling to avoid those elements from messing with the grid layout and also a **behavior change** on the Action Strip to prevent it from being initially visible in the grid with the standard setup (haven't added migration for that, might end up being a migration guide entry, if anything).

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
@@ -6739,7 +6739,7 @@ export abstract class IgxGridBaseDirective implements GridType,
     }
 
     protected getColumnList() {
-        return this.columnList.toArray();
+        return this.columnList.toArray().filter((col) => col.grid === this);
     }
 
     /**
@@ -6797,6 +6797,9 @@ export abstract class IgxGridBaseDirective implements GridType,
             let removed = false;
             let pinning = false;
             diff.forEachAddedItem((record: IterableChangeRecord<IgxColumnComponent>) => {
+                if (record.item.grid !== this) {
+                    return;
+                }
                 added = true;
                 if (record.item.pinned) {
                     this._pinnedColumns.push(record.item);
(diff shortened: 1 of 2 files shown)


ChronosSF (dismissed):

projects/igniteui-angular/src/lib/grids/grid/grid.component.html
```diff
@@ -300,8 +300,7 @@
 }
 <div class="igx-grid__outlet" #igxFilteringOverlayOutlet igxOverlayOutlet></div>
 @if (platform.isElements) {
-    <div #sink style="display: none;">
-        <ng-content select="igx-grid-state,igc-grid-state"></ng-content>
-        <ng-content select="igx-column,igc-column,igx-column-group,igc-column-group,igx-column-layout,igc-column-layout,igx-action-strip,igc-action-strip"></ng-content>
-    </div>
+    <div #sink style="display: none;"></div>
```
MayaKirova, line 303: Should check if this breaks logic in Blazor when getting ref to projected children, since previously all projected content was inside this div. Test things like whether you can call api methods on any of the child components like state, column etc.

damyanpetev: I truly hope that doesn't matter (it shouldn't). I don't think there are issues with the paginator that projects in a different container and the toolbar that doesn't move.
Regardless, note taken. The good part is Blazor is at 19.2 currently, so the 20 update there will be a separate step and we can address if needed then.

PR #15927, opened 2025-06-11
feat(h-grid): update adv filtering schema on row expand

### LINKED ISSUE #15827: Hide and/or deprecate the schema prop on HGrid
Course-correcting change added by #15562

The `schema` prop is of `EntityType[]` which represents a structure, except it's the QB config that's not native to the grids (that's why Field type in that config are directly compatible with columns for example, because the Grids uses those instead) and directly duplicates the Row Islands and Columns structure.

Point being, I don't see the need for that prop - the schema defined can be expressed with the expected HGrid setup and it likely should be.
I see no explanation in the PR or spec that reflects what the intended use is, we've mentioned perhaps load on demand with auto generate maybe, but even in that case I don't see it. Remember, we can create components dynamically, so nothing stopping you from doing:
```html
@for (col of wouldBeLoadedColumnsConfig | asnyc; track col.key) {
    <igx-column [field]="col.key" [header]="col.key"></igx-column>
}
```
If you can create the `schema` object in-advance for it, you can create the columns too - they are configuration declarative components after all.

Also one more point: If it is indeed auto-generate AND you configure the columns (as you're meant to) on `columnInit` with formatters and other i18n options, which ones do you think counts - the ones configure on the columns or the ones in the schema prop which is separate? That seems to me like it creates an opportunity for conflict besides the duplication.

Would like to know more, but if the above stands, I don't think we need the prop and should hide it before it gets used if possible.
Already hidden from other platforms with [#15824](https://github.com/IgniteUI/igniteui-angular/pull/15824/commits/36d199284eec6f745b5bd7de79a50855213ada72)

### LINKED ISSUE #15908: [Elements] Hierarchical grid Advanced filtering does not open with WC 6.0.0-beta.0
## Description  
Hierarchical grid with Advanced filtering in the toolbar does not open when you click the Advanced Filtering button.

You can refer to this sample: https://staging.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/hierarchical-grid/advanced-filtering#web-components-hierarchical-grid-advanced-filtering-example

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Click the advanced filtering button on the root level of the grid.

## Result  
Error is thrown: 

ERROR TypeError: Cannot read properties of undefined (reading '0')
    at 1623.a2d35d56.chunk.js:8:585747
    at Array.reduce (<anonymous>)
    at wl.reduce (1623.a2d35d56.chunk.js:2:554487)
    at e.generateChildEntity (1623.a2d35d56.chunk.js:8:585686)
    at 1623.a2d35d56.chunk.js:8:585321
    at Array.reduce (<anonymous>)
    at wl.reduce (1623.a2d35d56.chunk.js:2:554487)
    at e.generateSchema (1623.a2d35d56.chunk.js:8:585292)
    at get schema (1623.a2d35d56.chunk.js:8:576158)
    at e.generateEntity (1623.a2d35d56.chunk.js:2:1578003)

And dialog does not open.

## Expected result  
Dialog opens with no error.

## Attachments  

igdmdimitrov:
Closes #15827 
Closes #15908 

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


gedinakova:
@igdmdimitrov There's an issue when expanding inner levels:

1. Load HGrid Advanced Filtering dev demo
2. Expand the first record, then its first child

Result: The following error is thrown in the console:
grid-base.directive.ts:7198 ERROR TypeError: Cannot set properties of undefined (setting 'fields')
    at _IgxHierarchicalGridComponent.updateRootGridSchema (hierarchical-grid.component.ts:1292:25)
    at Object.next (hierarchical-grid.component.ts:686:18)
    at _IgxHierarchicalGridComponent.autogenerateColumns (grid-base.directive.ts:7198:35)

Expected result: No errors

gedinakova:
As currently setup, the AF dialog cannot be open in HGrid Advanced Filtering dev demo. The following console error is thrown:
ERROR TypeError: Cannot read properties of undefined (reading 'pipeArgs')
    at _IgxQueryBuilderTreeComponent.getFormat (query-builder-tree.component.ts:1413:58)
    at IgxQueryBuilderTreeComponent_ng_template_11_Conditional_0_Conditional_1_Conditional_10_Conditional_2_Conditional_1_Conditional_1_Template (query-builder-tree.component.html:211:45)
    at executeTemplate (debug_node-Dn-GvQJo.mjs:7142:9)
    at refreshView (debug_node-Dn-GvQJo.mjs:8887:13)
    at detectChangesInView (debug_node-Dn-GvQJo.mjs:9107:9)
    at detectChangesInViewIfAttached (debug_node-Dn-GvQJo.mjs:9067:5)
    at detectChangesInEmbeddedViews (debug_node-Dn-GvQJo.mjs:9024:13)
    at refreshView (debug_node-Dn-GvQJo.mjs:8916:9)
    at detectChangesInView (debug_node-Dn-GvQJo.mjs:9107:9)
    at detectChangesInViewIfAttached (debug_node-Dn-GvQJo.mjs:9067:5)

gedinakova:
@igdmdimitrov  In the `generateSchema` method, please uncomment the label & header (I thought this was fixed at some point, but apparently not).
image

gedinakova (approved):

gedinakova (changes requested):
We need a Changelog update for this change.

gedinakova (dismissed):

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes for each version of this project will be documented in this file.
 
+## 21.0.0
```
gedinakova: I believe 20.2.0 would be before 21 :)
```suggestion
## 20.2.0
```

projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.component.ts
```diff
@@ -590,11 +590,10 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
      * const schema = this.grid.schema;
      * this.grid.schema = [{ name: 'Products', fields: [...], childEntities: [...] }];
      * ```
+     * @deprecated in version 21.0.0.
```
gedinakova:
```suggestion
     * @deprecated in version 20.2.0.
```

gedinakova (changes requested):

gedinakova (dismissed):

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.component.ts
```diff
@@ -562,11 +562,10 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
      * const schema = this.grid.schema;
      * this.grid.schema = [{ name: 'Products', fields: [...], childEntities: [...] }];
      * ```
+     * @deprecated in version 20.2.0.
      */
     @Input()
-    public set schema(entities: EntityType[]) {
-        this._hGridSchema = entities;
-    }
+    public set schema(entities: EntityType[]) {}
```
damyanpetev, line 555: A bit iffy on the setter being noop - while it's doubtful that its seen major usage, ideally we'd have it fully functional until its removed.
Otherwise, at least make sure whatever scenario this prop was meant to support, that even with the inactive setter doesn't change with the new functionality handling schema internally.

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.component.ts
```diff
@@ -679,6 +678,10 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
                 this.batchEditing = val;
             });
         }
+        this.columnsAutogenerated.subscribe((e) => {
```
damyanpetev, line 662: With `pipe(takeUntilDestroyed())` like other subs?

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.component.ts
```diff
@@ -1253,44 +1257,62 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
...
+                return acc.concat(this.generateChildEntity(rowIsland));
             }
             , []);
         }
 
         return entities;
     }
 
-    private generateChildEntity(rowIsland: IgxRowIslandComponent, firstRowData: any[]): EntityType {
+    private updateRootGridSchema(event: any) {
```
damyanpetev, line 1291: I'm sure the even has a type, other than `any`

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.component.ts
```diff
@@ -1253,44 +1257,62 @@ export class IgxHierarchicalGridComponent extends IgxHierarchicalGridBaseDirecti
...
+                                dataType: f.dataType,
+                                label: f.label,
+                                header: f.header,
+                                editorOptions: f.editorOptions,
+                                filters: f.filters,
+                                pipeArgs: f.pipeArgs,
+                                defaultTimeFormat: f.defaultTimeFormat,
+                                defaultDateTimeFormat: f.defaultDateTimeFormat
+                            })) as FieldType[];
+            }
```
damyanpetev, line 1324: While the optimization for a single path update sounds nice, grid setups rarely changes w/ high frequency and the structures being processed are rarely, but sure - I can recall a few massive column lists, but even those are likely in the 100s, which is nothing much for list.

Regardless, two questions about this:
- This logic is more/less a repeat of the one in `generateSchema()`? If so, would be nice if those are either combined, reused or otherwise refactored into a single thing that does the islands/columns -> entity/fields mapping
- And isn't the column directly compatible with `FieldType`? Does the QB modify something in the schema fields? Otherwise, I wonder why not just take the filtered columns directly?

damyanpetev (commented):
Not entirely sure all dynamic column scenarios are handled, though that might've been the case even before - `generateSchema()` is called only on init and for some reason when setting the expression tree. `updateColumns/onColumnsChanged`?
`schema` will likely benefit greatly being a `computed()` signal eventually (will make it lazy too), sadly don't think we have enough of the rest of the pieces there yet (columnList/childLayoutList).

And we'll need to update the deprecated versions :)

PR #15982, opened 2025-06-24
Address grids column headers accessibility issues - active descendant, what is announced by SRs - 20.0.x

### LINKED ISSUE #15962: Grids: column headers accessibility issues
## Description  
Child of https://github.com/IgniteUI/igniteui-angular/issues/15836.
Discussing the accessibility state of the grid headers, the following items were outlined:

- [x] 1. FIX aria-activedescendant on headers + TEST ❗(likely a regression)
- [x] 2. Add aria-sort on column headers + update SPEC
- [x] 3. Icons should be aria-hidden (i.e. not read by SR/UA) - [separate PR](https://github.com/IgniteUI/igniteui-angular/pull/16010)
- [x] 4. For a virtualized grid, cell row/col indexes are not announced properly

 * igniteui-angular version: 20.0.x, 19.2.x
 * browser: any

## Steps to reproduce  

1. Use a screen reader and go through the grid headers.

TL;DR: What the SR announces would depend on its mode.
NVDA has two modes - Browse (default) and Focus. Browse targets static content, while focus - interactive, such as forms and grids. The latter should be used when navigating the grid to leverage its own keyboard navigation. The browse mode behavior varies per SR. The equivalent in JAWS, I believe, would be whether the "Virtual PC cursor" is on (browse) or off (focus/interaction mode), and in Narrator - Scan on and off.
Additionally, the expectation would be that the SR automatically enters the target mode (focus in the case of the grid), when landing elements of ceratin roles (columnheader, gridcell, etc.). NVDA/Narrator currently don't do this, probably because of the messed `activedescendant`.

## Result  
Nothing is announced when traversing the headers.

## Expected result  
Announce the column title, its sorting state and other related features, such as selection, expansion state (if the header is a group parent), etc.

## Attachments  

ddaribo:
Closes #15962

- [x] 1. FIX aria-activedescendant on headers + TEST ❗(likely a regression)
- [x] 2. Add aria-sort on column headers + update SPEC
- [x] 3. Icons should be aria-hidden (i.e. not read by SR/UA) - [separate PR](https://github.com/IgniteUI/igniteui-angular/pull/16010)
- [x] 4. For a virtualized grid, cell row/col indexes are not announced properly

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -517,7 +517,8 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
     /** @hidden @internal */
     @HostBinding('attr.aria-describedby')
     public get ariaDescribeBy() {
-        let describeBy = (this.gridID + '_' + this.column.field).replace('.', '_');
+        let describeBy = this.grid.headerGroupsList
+                        .find(hg => hg.column.field === this.column.field)?.headerID || '';
```
damyanpetev: Hm, is this because the column didn't have a field or because the description moved? Either way, while the header group list might be large in most cases, would generally avoid having getters for bindings that run on each change detection do any meaningful amount of work (like search) if it can be avoided.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -1839,6 +1839,16 @@ export abstract class IgxGridBaseDirective implements GridType,
...
 
+    @HostBinding('attr.aria-colcount')
+    protected get ariaColCount(): number {
+        return this.visibleColumns.length;
+    }
+
+    @HostBinding('attr.aria-rowcount')
+    protected get ariaRowCount(): number {
+        const totalRows = (this as any).totalItemCount ?? this.data?.length ?? 0;
+        return (this.paginator ? this._totalRecords : totalRows) + 1;
```
damyanpetev: Eh, `(this as any)`? Surely that's not needed 👀
Also, rows are manipulated not just by the paginator, so that might not be the only case. Was going to ask if `dataView` length might do the trick, but by the look of it we do want to announce the full data set?
If so, I'm not entirely sure the virtualization index (which I believe drive row index) will match that in all cases.
Also would filtering and grouping need to be involved to determine the final count? Because those will affect the row index regardless IIRC. Basically, anything in the pipeline that manipulates the data collection is before the for-of even gets the data.

It might be the simplest way to poke the for-of state for the length of items and only detect remote/paginated scenarios to look for _totalRecords / totalItemCount

rkaraivanov (approved):

PR #15996, opened 2025-06-25
fix(grid): Cancel endEdit if there is no active editing

dkamburov:
Closes #  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/drop-down/_drop-down-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/drop-down/_drop-down-theme.scss
@@ -243,7 +243,7 @@
         -webkit-overflow-scrolling: touch;
         position: relative;
 
-        igx-display-container {
+        .igx-display-container--scrollbar {
             padding-inline-end: var(--vhelper-scrollbar-size);
         }
     }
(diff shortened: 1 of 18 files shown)


projects/igniteui-angular/src/lib/grids/grid/grid-cell-selection.spec.ts
```diff
@@ -84,6 +84,10 @@ describe('IgxGrid - Cell selection #grid', () => {
 
             UIInteractions.simulatePointerOverElementEvent('pointerup', endCell.nativeElement);
             detect();
+            // Invoke endEdit() to make sure if no editing is going on,
+            // the cell activation shouldn't be lost (https://infragistics.visualstudio.com/Indigo_Platform/_workitems/edit/37933)
+            grid.endEdit(true, null);
```
MayaKirova, line 89: IMO it would be better if this was a separate test rather than modifying an existing basic one. If we automate everything related to a certain feature in the same test it will soon become too convoluted to read and maintain.

dkamburov: Yeah, that's true, but it's a single line and we are at a stage were we have too many test and we should be careful when adding new ones, also Damyan suggested adding this to an already existing one.

MayaKirova: Could you elaborate on what you mean by too many tests? I'm not aware of there being an issue with adding new ones.

dkamburov: Mainly because of the disconnects, I know there is another reason for them, but the bigger amount of the tests is not helping.

MayaKirova (approved):

PR #16034, opened 2025-07-08
refactor(*): New localization implementation.

### LINKED ISSUE #15181: Grid Grouping ISO Dates displays the wrong value
## Description  
Describe the issue.

 * igniteui-angular version: 18.2.8
 * browser: Any

## Steps to reproduce  

1. Add a grid
2. Set to data with dates as date-only ISO strings
3. Set the date column as group-able `[groupable]="true"`
4. Use a custom template to format the date as 'shortDate':
    ```
    <igx-column field="OrderDate" header="Order Date" width="200px" [groupable]="true" dataType="date">
        <ng-template igxCell let-value>
            {{ value | date:'shortDate' }}
        </ng-template>
    </igx-column>
    ```
5. Set your machine timezone to any negative GMT
6. Try grouping by that column

## Result  
The group Date is wrong, expanding the group shows the correct date in the rows.
e.g. 2016-11-23 gets grouped as 2016-11-22

## Expected result  
The group Date to be the same as the rows that it groups.

## Attachments  
image

Repro stackblitz: https://stackblitz.com/edit/mt2zmdk2?file=src%2Fapp%2Fdata%2FinvoiceData.ts,src%2Fapp%2Fgrid%2Fgrid-groupby-custom-sample%2Fgrid-groupby-custom-sample.component.html

skrustev:
Closes #15181

Requires PR: https://github.com/IgniteUI/igniteui-i18n/pull/1

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/calendar/src/calendar/days-view/days-view.component.spec.ts
+++ b/projects/igniteui-angular/calendar/src/calendar/days-view/days-view.component.spec.ts
@@ -127,10 +127,13 @@ describe("Days View Component", () => {
 
         // Mock the formatter behavior
         // Simulate a locale (like zh-CN) that adds a suffix to the day number.
-        // Cast to 'any' to overwrite the protected 'formatterDay' property used by formattedDate()
-        (daysView as any).formatterDay = {
-            format: () => '25日',
-        } as Intl.DateTimeFormat;
+        Object.defineProperty(daysView, "formatterDay", {
+            get() {
+                return {
+                    format: () => '25日',
+                } as Intl.DateTimeFormat
+            }
+        });
 
         // 1. Verify Programmatic Access (formattedDate method)
         // Should return the raw formatted string from the formatter (with suffix)


projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -1,10 +1,11 @@
-import { CurrencyPipe, formatDate as _formatDate, isPlatformBrowser } from '@angular/common';
+import { formatDate as _formatDate, getLocaleCurrencyCode, getLocaleFirstDayOfWeek as ngGetLocaleFirstDayOfWeek, isPlatformBrowser } from '@angular/common';
```
MayaKirova: Should we be using the `getLocaleCurrencyCode` from angular?

It's deprecated and also does not seem to work correctly when you pass a region specific lang tag. For example if you pass "en-GB" it should return "£", but it returns "$" instead.

Hence you get incorrect formatting:

image


Same for `getLocaleFirstDayOfWeek`.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -2884,9 +2884,7 @@ export abstract class IgxGridBaseDirective implements GridType,
         if (this._currencyPositionLeft !== undefined) {
             return this._currencyPositionLeft;
         }
-        const format = getLocaleNumberFormat(this.locale, NumberFormatStyle.Currency);
-        const formatParts = format.split(',');
-        const i = formatParts.indexOf(formatParts.find(c => c.includes('¤')));
+        const i = getI18nManager().getCurrencyPosition(this.locale);
```
MayaKirova: `currencyPositionLeft` gets cached, so if you change the locale runtime it will not reflect the new position in the editing template. It should be cleared when the locale changes.

projects/igniteui-angular/src/lib/calendar/calendar-base.ts
```diff
@@ -656,9 +654,15 @@ export class IgxCalendarBaseDirective implements ControlValueAccessor {
         protected keyboardNavigation?: KeyboardNavigationService,
         protected cdr?: ChangeDetectorRef,
     ) {
-        this.locale = _localeId;
+        initi18n(_localeId);
```
MayaKirova: Previously the locale setter was also setting a default `weekStart`:

```
if (!this.weekStart) {
            this.weekStart = getLocaleFirstDayOfWeek(this._locale);
}
```

But now, nothing seems to set it and if the user does not explicitly set one, it throws an error:
this.monthWeeks.at is not a function or its return value is not iterable

You can reproduce it on a plain calendar with no props:

```
<igx-calendar></igx-calendar>
```

projects/igniteui-angular/src/lib/date-picker/date-picker.component.ts
```diff
@@ -516,7 +517,11 @@ export class IgxDatePickerComponent extends PickerBaseDirective implements Contr
         private cdr: ChangeDetectorRef,
         @Optional() @Inject(IGX_INPUT_GROUP_TYPE) _inputGroupType?: IgxInputGroupType) {
         super(element, _localeId, _inputGroupType);
-        this.locale = this.locale || this._localeId;
+        initi18n(_localeId);
+        this.locale = this.locale || getCurrentI18n();
+        getI18nManager().onResourceChange(() => {
+            this._resourceStrings = getCurrentResourceStrings(DatePickerResourceStringsEN, false);
```
MayaKirova: Shouldn't the locale be updated on change, like in the grid?:
```
        getI18nManager().onResourceChange((args: ResourceChangeEventArgs) => {
            this._defaultLocale = args.newLocale;
```

Because currently the picker does not react to a change in the locale of the page and stays in the original language/formattings.

It works only if you set it before the picker initializes.

projects/igniteui-angular/src/lib/calendar/calendar-base.ts
```diff
@@ -656,9 +654,15 @@ export class IgxCalendarBaseDirective implements ControlValueAccessor {
...
+        this._defaultLocale = getCurrentI18n();
         this.viewDate = this.viewDate ? this.viewDate : new Date();
         this.initFormatters();
+
+        getI18nManager().onResourceChange((args: ResourceChangeEventArgs) => {
+            this._defaultLocale = args.newLocale;
```
MayaKirova: This works for most parts of the calendar, however there's a `formattedMonth` func that currently does not take the locale in consideration, so the month in the calendar remains in the default locale:

image

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -579,14 +580,113 @@ export const isConstructor = (ref: any) => typeof ref === 'function' && Boolean(
...
  */
-export const formatDate = (value: string | number | Date, format: string, locale: string, timezone?: string): string => {
+export function formatDate(value: Date | string | number | null | undefined, format: string, locale: string, timezone?: string): string {
     if (value === null || value === undefined || value === '') {
         return '';
     }
-    return _formatDate(value, format, locale, timezone);
-};
+    if (typeof value === "string" || typeof value === "number") {
+        value = new Date(value);
```
pmoleri: Using `new Date(string)` is most probably wrong for ISO dates.
See issue #15181 for details.

skrustev: I've replied in the issue, this is how it should be imo.
The date time string format the Date object universally supports is based off of the ISO format, so there should be even less issues for it.

projects/igniteui-angular/src/lib/core/utils.ts
skrustev: Yes, this is for backwards compatibility so we don't break any old use of Angular locale, until we fully deprecate it.

MayaKirova: Do we have plans to replace it with some of our own custom logic or have the user pass the currency type in the pipes himself?

If it's going to break anyway, we might as well introduce the breaking change now with the major refactor and document it. 

It also bugs me that it works, but only for some locales. It's a bit inconsistent. For example, it's correct for DE, but not for IT.

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -579,14 +580,113 @@ export const isConstructor = (ref: any) => typeof ref === 'function' && Boolean(
...
+    }
+    let dateStyle = undefined, timeStyle = undefined;
+    if (format === 'short' || format === 'medium' || format === 'long' || format === 'full') {
+        dateStyle = format;
+        timeStyle = format;
+    } else if (format.includes('Date')) {
+        dateStyle = format.replace('Date', '');
+    } else if (format.includes('Time')) {
```
MayaKirova: Maybe you need to set the timeStyle too here? Because currently it does not construct correct options for time formats, for example: 'mediumTime':
image

So it returns a medium date instead of time. You can see it in the grid if you define a `time` column. It will not format correctly.

projects/igniteui-angular/src/lib/core/utils.ts
pmoleri: > Date object universally supports is based off of the ISO format

Not at all. It's based on bad decisions made almost 30 years ago, not honoring ISO standard.

See Angular's explanation:
https://github.com/angular/angular/blob/4a8d0b79001ec09bcd6f2d6b15117aa6aac1932c/packages/common/src/i18n/format_date.ts#L915C1-L921C65

With this change you're introducing a regression. Angular pipe was formatting as expected by ISO, the new code is not.

projects/igniteui-angular/src/lib/core/utils.ts
skrustev: Hmm the Angular `getLocaleFirstDayOfWeek` throws error if you haven't registered the specified locale with their `registerLocaleData`, but I see that `getLocaleCurrencyCode` actually works without any locale date registered by the user, so for that it might need a smarter way of determining if user uses the old locale setters.

If both method relied on `registerLocaleData` it would be as easy as removing any use of it, to use the new locales, but I guess it's not the same for all API

projects/igniteui-angular/src/lib/core/utils.ts
pmoleri: With "regression" I mean this:
image
both from the same data: '2025-08-01'

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -1852,7 +1852,7 @@ export abstract class IgxGridBaseDirective implements GridType,
     }
 
     public get resourceStrings(): IGridResourceStrings {
-        return this._resourceStrings;
+        return this._resourceStrings || this._defaultResourceStrings;
```
MayaKirova, line 1818: Are we going to continue supporting overwriting just a subset of the strings via the property? Like in the docs: https://www.infragistics.com/products/ignite-ui-angular/angular/components/general/localization#localize-specific-strings-for-a-specific-instance-of-a-component

Because currently the way its setup, if you set a single string as custom, all others will resolve as undefined aka you have to overwrite all of the strings yourself.

projects/igniteui-angular/src/lib/core/utils.ts
skrustev: Yes, that is because Angular handles the date formatting themselves and the Date object on its own is not enough. That is the whole point of the Intl we are using now.

I personally am ok with this breaking change, because I do not think previously the default behavior of Angular is the better one. There are a lot of discussion what is the "correct" behavior, but to me the date displayed should by default be in local timezone first and if the dev wants to specify target timezone should set it anyway.

@damyanpetev @rkaraivanov What do you think?

pmoleri: > That is the whole point of the Intl we are using now.

You're confusing parsing and formatting.
You're using Intl for formatting, that's ok.
You're choosing to parse with `new Date()`, this is the mistake.

skrustev: We can change it manually if no timezone is specified, that the date displayed to be the same set in the data, but that should be deliberate and differs from the practice of the backend dates to be in UTC. This is purely cosmetic hack that needs to be done and doesn't really serve purpose in practice.

pmoleri: > There are a lot of discussion what is the "correct" behavior.

I don't think there's anything to discuss '2025-01-01' should NEVER be formatted as 31 Dec 2024.

skrustev: Like I said the Angular pipes and method are cosmetic and purely frontend oriented while the grid data object provided is supposed to be a backend one. These are quite different apples and oranges and we cannot apply the same logic.

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
skrustev: Yes, at least for now that is the goal. I've yet to mark them as `deprecated`, but I'll mark them and in a future release we can get rid of them.
Hm yeah the second thing sounds like an issue, I'll look into it. The rest resource strings should default to `en`

projects/igniteui-angular/src/lib/core/utils.ts
skrustev: It turned out to be the locales registered through angular in the main.ts of the demos messing with how it works. Should be ok for now how it is to not introduce breaking changes.

projects/igniteui-angular/src/lib/core/utils.ts
skrustev: I've updated the handling. I agree on supporting the ISO standard, just the way Angular handles it is not the best imo

skrustev: If we are to support the whole range of ISO standard though with the newest things, that will require quite a lot of other things though. Not sure if worth it really.

projects/igniteui-angular/input-group/src/input-group/input-group.component.ts
```diff
@@ -50,7 +51,7 @@ export class IgxInputGroupComponent implements IgxInputGroupBase, AfterContentCh
      * Returns the resource strings.
      */
     public get resourceStrings(): IInputResourceStrings {
-        return this._resourceStrings;
+        return this._resourceStrings || this._defaultResourceStrings;
```
MayaKirova, line 60: Some getters, like the one for `fileNames` get directly the `_resourceStrings`, so they throw an error on init:

Cannot read properties of null (reading 'igx_input_file_placeholder')
    at get fileNames (input-group.component.ts:381:62)

They can use the public one.

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -579,14 +580,113 @@ export const isConstructor = (ref: any) => typeof ref === 'function' && Boolean(
...
+        value = parseFloat(value);
+    }
+    const parsedDigitsInfo = parseDigitsInfo(digitsInfo);
+    let currencyDisplay: keyof Intl.NumberFormatOptionsCurrencyDisplayRegistry;
+    if (display !== 'code' && display !== 'symbol' && display !== 'symbol-narrow' && display !== 'narrowSymbol' && display !== "name") {
+        currencyDisplay = 'symbol';
+    } else if (display === 'symbol-narrow') {
+        currencyDisplay = 'narrowSymbol';
+    } else {
+        currencyCode = display || undefined;
```
MayaKirova: Shouldn't this be `currencyDisplay`?  It's setting the code, which is  a different thing.
So if you set a pipeArg, for example:

```
   public pipeArgs: IColumnPipeArgs = {
        display: 'code',
        currencyCode: "EUR"
    };
```

It would generate incorrect options:

image

So it would throw error: RangeError: Invalid currency code : code

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
```diff
@@ -5,7 +5,7 @@ import {
     Output,
     TemplateRef
 } from '@angular/core';
-import { getLocaleFirstDayOfWeek, NgTemplateOutlet, NgClass, DatePipe } from '@angular/common';
+import { NgTemplateOutlet, NgClass, DatePipe } from '@angular/common';
```
MayaKirova: Do we plan to change the DatePipe used in the templates? Currently it's using angular's date pipe, so it's not using our `IgxDateFormatterPipe` with the Intl implementation.

Same for the `base-export-service.ts`, where it uses it to format dates.
And also for the `TimeFormatPipe`, which uses it to format time.

projects/igniteui-angular/src/lib/core/utils.ts
pmoleri: We can scope it to support the ones defined in [swagger](https://swagger.io/docs/specification/v3_0/data-models/data-types/#strings) (date and date-time)
plus everything else that `new Date()` can parse.

pmoleri: I didn't find the new handling. Is it pushed already?

projects/igniteui-angular/src/lib/core/utils.ts
```diff
@@ -576,6 +576,26 @@ export const yieldingLoop = (count: number, chunkSize: number, callback: (index:
 
 export const isConstructor = (ref: any) => typeof ref === 'function' && Boolean(ref.prototype) && Boolean(ref.prototype.constructor);
 
+/**
+ * Bind to the i18n manager's onResourceChange event
+ * @param destroyObj Object responsible for signaling destruction of the handling object
+ * @param context Reference to the object's this context
+ */
+export function onResourceChangeHandle(destroyObj: Subject<any> | DestroyRef, callback: (event?: CustomEvent<IResourceChangeEventArgs>) => void, context: any) {
```
pmoleri: If this is always meant to be called form the constructor (directly or indirectly) you could `inject(DestroyRef)` here. And avoid passing it from every component.
You can still keep it as an optional parameter.

```ts
export function onResourceChangeHandle(callback: (event?: CustomEvent<IResourceChangeEventArgs>) => void, context: any, destroyRef? DestroyRef) {
    destroyRef ??= inject(DestroyRef);
    ...
}

File: projects/igniteui-angular/src/lib/core/utils.ts [hunk 13]
```diff
@@ -576,6 +576,26 @@ export const yieldingLoop = (count: number, chunkSize: number, callback: (index:
...
+    getI18nManager().addEventListener("onResourceChange", onResourceChangeHandler);
+
+    // Handle removal of listener on context destroy
+    const removeHandler = () => {
+        getI18nManager().removeEventListener("onResourceChange", onResourceChangeHandler);
+    }
+    if (destroyObj instanceof DestroyRef) {
+        destroyObj.onDestroy(() => removeHandler());
+    } else {
+        destroyObj.pipe(finalize(() => removeHandler()));
```
pmoleri: I may be wrong, but I think this finalize() handler will never be called.
When you .pipe() you're creating a new observable than no one will ever subscribe to. It will be GC as soon as this method finishes.

projects/igniteui-angular/src/lib/core/utils.ts
skrustev: This should be working now

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
skrustev: This as well should be working as previously.
In the old behavior it added the partial new resource strings to the current ones. For example if you have switched to BG resource strings and changed to the HI in the demo, it applied them on top of the BG resources instead of EN.
It's a bit weird but that should remain unchanged.

MayaKirova: If you set it initially via the property binding it breaks, for example:

```
<igx-grid [resourceStrings]="partialCustomHindi" ...
```

All other strings show as undefined:

image

It does work if you first load a locale and then change it. But it does not, if you set it initially via the prop.

skrustev: Hm I didn't change much but maybe I fixed it somehow because it seem to work on my end by setting it initially the `resourceStrings`.

projects/igniteui-angular/src/lib/core/utils.ts
pmoleri: I see it's resolved now.

projects/igniteui-angular/src/lib/query-builder/query-builder-tree.component.ts
skrustev: I've logged separate issues for the pickers and the exporter by #16138 and #16139, since they required a bit more look into regarding their internal logic, aside from using Angular pipes or methods.

projects/igniteui-angular/src/lib/core/utils.ts
skrustev: For now I don't want to force it to constructor only and just less side effects in general.

pmoleri: It's not force it if you keep the optional parameter.
It's true that is less "side effects" (or magic), but it's just like takeUntilDestroyed(), which is a recommended practice.
Also, currently you're adding destroyRef or destroy$ observable to many components that didn't need it. I think that with this change you'll get predictable behavior and cleaner code all over.

projects/igniteui-angular/src/lib/core/utils.ts
MayaKirova: Should be this one in the core package, I believe: 
https://github.com/IgniteUI/igniteui-i18n/pull/1/commits/b6decebde0fadfa94668c74f5ab760125947f2db#diff-822597eac4832fa682cb5ce0ad7869ba5df2b69df35ad1b66a3be7af651acb03R170

pmoleri: Thanks. I didn't realize that Github was taking me to the old state of the PR, so I kept seeing the old code.

pmoleri: While this improves the PR. I noticed that there's still other code using the old parsing:
```
export const parseDate = (value: any): Date | null => {
    // if value is Invalid Date return null
    if (isDate(value)) {
        return !isNaN(value.getTime()) ? value : null;
    }
    return value ? new Date(value) : null;
};
```

So, there's probably still things to be addressed in other places. Unrelated to this PR though.

dkamburov:
@skrustev there are conflicts please resolve them

dkamburov:
@skrustev hiding in toolbar is not changing title in hierarchicalgrid and treegrid
image

dkamburov:
@skrustev 

1. Pivot is not changing it's aggregators for the values
2. Search in the pivot data selector

image

dkamburov:
@skrustev Calendar is not changing the `select date` text 
image

skrustev:
> @skrustev
> 
> 1. Pivot is not changing it's aggregators for the values
> 2. Search in the pivot data selector
> 
> image

Pivot value aggregations doesn't appear to be localized in general due to how they are retrieved at the moment. I think this can be logged separately.

PS: I already have it logged actually from earlier during development #16159

skrustev:
> @skrustev hiding in toolbar is not changing title in hierarchicalgrid and treegrid image

A little more details please on how to reproduce. The default lables for Toolbar's column hiding is `Visible columns` and for me it is localized:
image

PS: It appears you have set title for the hiding action:
```
<igx-grid-toolbar-hiding title="Column Hiding"></igx-grid-toolbar-hiding>
```

skrustev:
> @skrustev Calendar is not changing the `select date` text image

This exact resource is the same for almost all languages except I think Bulgarian. No idea why, seems to be a separate issue.

mddragnev:
Setting the locale directly on the pivot grid does not change the date dimension months and display localizations

projects/igniteui-angular/grids/pivot-grid/src/pivot-grid.component.ts
```diff
@@ -991,6 +991,11 @@ export class IgxPivotGridComponent extends IgxGridBaseDirective implements OnIni
                 this.generateConfig();
             }
             this.setupColumns();
+            // Bind to onResourceChange after the columns have initialized the first time to avoid premature initialization.
+            onResourceChangeHandle(this.destroy$, () => {
+                // Since the columns are kinda static, due to assigning DisplayName on init, they need to be regenerated.
```
MayaKirova: Is this because of the predefined dimensions like the `IgxPivotDateDimension` that have localization? If so it might be a bit of an overkill to regenerate all columns, especially since there might not be a date dimension to localize for.

MayaKirova, line 996: Is this because of the predefined dimensions like the `IgxPivotDateDimension` that have localization? If so it might be a bit of an overkill to regenerate all columns, especially since there might not be a date dimension to localize for.

projects/igniteui-angular/core/src/core/i18n/formatters/formatter-intl.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    describe('other', () => {
+        it('getCurrencyCode should return default USD as currency code for locale', () => {
+            expect(i18nFormatter.getCurrencyCode('en-US')).toEqual('USD');
+            expect(i18nFormatter.getCurrencyCode('it')).toEqual('USD');
+        });
+
+        it('getCurrencySymbol should return correct currency symbol', () => {
+            expect(i18nFormatter.getCurrencySymbol('USD', 'en-US')).toEqual('$');
+            expect(i18nFormatter.getCurrencySymbol('BGN', 'bg')).toEqual('лв.');
```
MayaKirova: Not sure if this will be valid as of January. Same for all other bg currency tests. But I guess we could change it afterward.

projects/igniteui-angular/core/src/core/i18n/formatters/formatter-intl.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+        it('should format string to dateTime using Intl', () => {
+            expect(i18nFormatter.formatDate('2025-01-25T14:15:00+02:00', 'short', 'en-US', "Europe/Sofia")).toEqual('1/25/25, 2:15 PM');
+            expect(i18nFormatter.formatDate('2025-01-25T14:15:00+02:00', 'medium', 'en-US', "Europe/Sofia")).toEqual('Jan 25, 2025, 2:15:00 PM');
+            expect(i18nFormatter.formatDate('2025-01-25T14:15:00+02:00', 'long', 'en-US', "Europe/Sofia")).toEqual('January 25, 2025 at 2:15:00 PM GMT+2');
+            expect(i18nFormatter.formatDate('2025-01-25T14:15:00+02:00', 'full', 'en-US', "Europe/Sofia")).toEqual('Saturday, January 25, 2025 at 2:15:00 PM Eastern European Standard Time');
+        });
+
+
+        it('should format string to time using Intl', () => {
+            expect(i18nFormatter.formatDate('2025-01-25T14:15:00+02:00', 'shortTime', 'en-US', "Europe/Sofia")).toEqual('2:15 PM');
```
MayaKirova, line 50: It might be better not to have the timezone offset hardcoded but maybe resolve it runtime. Since  the tests will probably fail on a machine in another timezone.

skrustev: The tests on the CI already run on another machine in different timezone. This test checks if you set it to explicitly different timezone, it should return the time in that timezone.

projects/igniteui-angular/grids/pivot-grid/src/pivot-grid.component.ts
skrustev: Well localization shouldn't be changed often. The other solution would require some updates to how dimension columns are created and way to sync them with their corresponding dimensions and values.
Sounds like a good improvement, don't think this is the place to do.

PR #16036, opened 2025-07-08
test(hgrid): add a test

### LINKED ISSUE #16035: The display container of the header row spans further than the parent container when having row selection/expanding
## Description  
Having row selection and/or expanding causes the display container of the headers to exceed the width of the header row itself. 

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open any hierachical grid sample and add row selection
2. Add with 100%

## Result  

Image

See that the width of the display container is the same as the width of the header row resulting in exceeding the grid width (because of row selectors and expanders)
Image

## Expected result  
The display container of the headers should shrink so that the selectors + expanders + display container equals the width of the whole header-row

## Attachments  

mddragnev:
Closes #16035   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.spec.ts
+++ b/projects/igniteui-angular/src/lib/grids/hierarchical-grid/hierarchical-grid.spec.ts
@@ -668,6 +668,18 @@ describe('Basic IgxHierarchicalGrid #hGrid', () => {
             );
             jasmine.getEnv().allowRespy(false);
         });
+
+        it('should calculate correct column headers width when rowSelection + expand indicators', () => {
+            hierarchicalGrid.rowSelection = 'multiple';
+            fixture.detectChanges();
+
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/core/styles/components/grid/_grid-theme.scss
```diff
@@ -1274,6 +1274,7 @@
     %grid-display-container-tr {
         width: 100%;
         overflow: visible;
+        flex: 1;
```
MayaKirova, line 721: The change itself is fine, however since it changes the size of the cell's virtual container you'll also have to change how right pinning positions the cells. It currently sets an incorrect offset so if you pin any column to the right, you'll see that it's not correctly positioned to the right edge of the grid:

image

MayaKirova (changes requested):

MayaKirova (approved):

PR #16053, opened 2025-07-16
Handle validation for fields with '.' - 19.2.x

### LINKED ISSUE #15932: [Grid] Validations now working when there is a dot (".") in field property
## Description  
Validations now working when there is a dot (".") in field property

 * igniteui-angular version: 19.2
 * browser: Chrome

## Steps to reproduce  

1. Create a grid
2. Add a column
3. Add validation, for example minlength
4. Add field property with a "." in the value, doesn't matter if it's a real one

## Result  
The validation is not working

## Expected result  
Validation should be working

## Attachments  

<pre> ```html <igx-grid [data]="northwindSwaggerEmployeeDto" primaryKey="employeeId" [allowFiltering]="true" filterMode="excelStyleFilter" class="grid" rowEditable="true"> <igx-column field="title" dataType="string" header="title" [filterable]="true" [minlength]="5" [sortable]="true" [maxlength]="10" [selectable]="false"></igx-column> <igx-column field="address.street" dataType="string" header="addressstreet" [filterable]="true" [sortable]="true" [minlength]="5" [maxlength]="10" [selectable]="false"></igx-column> <igx-column field="test.1" dataType="string" header="test1" [filterable]="true" [sortable]="true" [minlength]="5" [maxlength]="10" [selectable]="false"></igx-column> <igx-column field="test2" dataType="string" header="test2" [filterable]="true" [sortable]="true" [minlength]="5" [maxlength]="10" [selectable]="false"></igx-column> </igx-grid> ``` </pre>

MonikaKirkova:
Closes #15932 

Changes:
1. Do not replace . in field names with _ when passing a name for the ngControl, Use the column field name as a name for the native ngControl added
image

2. When trying to get a control using ngControl.get, wrap the name in an array. This way ngControl.get will treat it as a literal key. Otherwise, default behavior of ngControl,get("address.street") is to look for a control named "address" and then look for a control nested named "street" , which is nested under address.

image

3. Assign formControl.errors in a let variable in order to be resolved once, not to be resolved once for each usage in the template

image

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -542,6 +542,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
             const isInvalid = this.grid.validation?.isFieldInvalid(this.formGroup, this.column?.field);
             return !this.intRow.deleted && isInvalid;
         }
+        return false;
     }
 
     /**
@@ -554,6 +555,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
             const isValidAfterEdit = this.grid.validation?.isFieldValidAfterEdit(this.formGroup, this.column?.field);
             return this.editMode && isValidAfterEdit;
         }
+        return false;
     }
 
     /**
--- a/projects/igniteui-angular/src/lib/grids/cell.component.ts
+++ b/projects/igniteui-angular/src/lib/grids/cell.component.ts
@@ -1182,7 +1182,7 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
      * @internal
      */
     public getTransformedFieldName(field: string): string {
-        return field.replace(/\./g, '_');
+        return field?.replace(/\./g, '_');
     }
 
     /**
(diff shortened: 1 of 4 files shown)


projects/igniteui-angular/src/lib/grids/cell.component.ts
```diff
@@ -1175,6 +1177,14 @@ export class IgxGridCellComponent implements OnInit, OnChanges, OnDestroy, CellT
         return meta;
     }
 
+    /**
+     * @hidden
+     * @internal
+     */
+    public getTransformedFieldName(field: string): string {
+        return field?.replace(/\./g, '_');
```
dkamburov: I know it's an edge case, but I'm not sure about it. If we have column test.1 and test_1, we'll mess up the validation, I guess we are good with this, but should be at least very well described in the docs.

Or even better we should use a bit more complex identifier to replace with.

dkamburov: Also I tested it and seems like it's not working, I can't get "green" validated field.
image

I used grid-updates sample from the repo I only added editing and required field:
image

dkamburov (changes requested):

deyvidnenchev:
getFieldKey can be deleted because it is not used anymore.

deyvidnenchev (approved):

PR #16134, opened 2025-08-12
fix(column): use minWidth when sizing columns - master

### LINKED ISSUE #16057: `minWidth` doesn't override the default 136px min width when autosizing columns without width
## Description  
In a grid with columns that don't have width set, the width defaults to 136px with the property `minWidth` that should affect this (by spec) not having any effect.

 * igniteui-angular version: latest
 * browser: n/a

## Steps to reproduce  

1. Open https://stackblitz.com/edit/d37qt97a-gqnfqw6y 
2. Observe the column widths of columns after the first one

## Result  
The columns disregard the min width and render at the default width of 136px

## Expected result  
The columns to render at the specified minWidth of 44px (as they are overflowing the container anyway, even at 44px)

RivaIvanova:
Closes #16057  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/core/styles/components/calendar/_calendar-theme.scss
+++ b/projects/igniteui-angular/src/lib/core/styles/components/calendar/_calendar-theme.scss
@@ -814,11 +814,9 @@
         %date-inner {
             color: var-get($theme, 'weekend-color');
 
-            @if $variant == 'indigo' {
-                &:hover {
-                    color: var-get($theme, 'date-hover-foreground');
-                }
-            }
+           &:hover {
+               color: var-get($theme, 'date-hover-foreground');
+           }
         }
     }
 


projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -1012,6 +1012,7 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
             return;
         }
         this._defaultMinWidth = value;
+        this.minWidthSetByUser = true;
```
MayaKirova: I don't think you need an additional prop, since we already have the getter for `userSetMinWidthPx` that returns the user set min width (if there is one). You could directly use that.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -8008,4 +8010,16 @@ export abstract class IgxGridBaseDirective implements GridType,
...
         this._rowCount += 1; // include header row
     }
+
+    private setPossibleMinColumnWidth(column: IgxColumnComponent): void {
+        if (this.columnWidthSetByUser || this.width == null) {
+            return;
+        }
+
+        if (column.minWidthSetByUser || column.hidden) {
+            this._columnWidth = this._possibleColumnWidth;
```
MayaKirova: I don't see why the default width of the columns on the grid level (associated with the `columnWidth` prop) would change based on the individual props on each column. That should be constant and apply for all columns, regardless of whether any of them have individual min-widths.

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -5534,8 +5535,8 @@ export abstract class IgxGridBaseDirective implements GridType,
         computedWidth -= this.featureColumnsWidth();
 
         const columnWidth = !Number.isFinite(sumExistingWidths) ?
-            Math.max(computedWidth / columnsToSize, this.minColumnWidth) :
-            Math.max((computedWidth - sumExistingWidths) / columnsToSize, this.minColumnWidth);
+            Math.max(computedWidth / columnsToSize, this.defaultHeaderGroupMinWidth) :
```
MayaKirova: `defaultHeaderGroupMinWidth` is a property we intent to remove, since it's a hardcoded left-over from the display density property deprecation. I'd prefer if we don't use it anymore.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -2699,4 +2706,26 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
...
+     * @hidden
+     */
+    public set minWidthSetByUser(value: boolean) {
+        this._minWidthSetByUser = value;
+    }
+
+    /**
+     * @hidden
+     */
+    public get minWidthSetByUser(): boolean {
```
MayaKirova: Probably don't need this getter and setter.

projects/igniteui-angular/src/lib/grids/columns/column.component.ts
```diff
@@ -2699,4 +2706,26 @@ export class IgxColumnComponent implements AfterContentInit, OnDestroy, ColumnTy
...
+        return this._minWidthSetByUser;
+    }
+
+    private getPossibeWidth(): string {
+        const possibleWidth = this.grid.getPossibleColumnWidth();
+        if (!this.minWidthSetByUser && parseFloat(possibleWidth) < this.grid.minColumnWidth) {
```
MayaKirova: Can check `userSetMinWidthPx` instead.

MayaKirova (approved):

ChronosSF (approved):

PR #16251, opened 2025-09-26
Perf. improvements with activation and cell merge.

### LINKED ISSUE #16146: Improve performance of navigation when there are merged cells.
Related to: https://github.com/IgniteUI/igniteui-angular/pull/16024

## Description  
When there are 1mil records and some columns are merged, navigation lags when using the arrow keys.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Run the cell merging sample with 1mil records.
2. Try to navigate up/down in the grid.

## Result  
It feels slow.

## Expected result  
It should be about as fast as when there is no merging (which is also a bit slow when scrolling, but that's its own issue).

MayaKirova:
Closes #16146  

Split pipes in two, where one does the merging and the other tracks active rows and unmerges within related merge sequences.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
@@ -126,24 +126,25 @@ export class IgxGridUnmergeActivePipe implements PipeTransform {
             const index = result.indexOf(x);
             const colKeys = [...x.cellMergeMeta.keys()];
             const cols = colsToMerge.filter(x => colKeys.indexOf(x.field) !== -1);
+            let res = [];
             for (const col of cols) {
-                const rs = x.cellMergeMeta.get(col.field).rowSpan;
+
                 let childData = x.cellMergeMeta.get(col.field).childRecords;
                 const childRecs = childData.map(x => x.recordRef);
                 const isDate = col?.dataType === 'date' || col?.dataType === 'dateTime';
                 const isTime = col?.dataType === 'time' || col?.dataType === 'dateTime';
-                const res = this.grid.mergeStrategy.merge(
+                res = this.grid.mergeStrategy.merge(
                     [x.recordRef, ...childRecs],
                     col.field,
                     col.mergingComparer,
-                    [],
+                    res,
                     activeRowIndexes,
                     isDate,
                     isTime,
                     this.grid);
-                result.splice(index, index + rs, ...res);
-            }
 
+            }
+            result.splice(index, (index + res.length - 1), ...res);
         });
 
 
--- a/projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
@@ -121,8 +121,8 @@ export class IgxGridUnmergeActivePipe implements PipeTransform {
                 rootsToUpdate.push(root);
             });
         });
-
-        rootsToUpdate.forEach(x => {
+        const uniqueRoots = rootsToUpdate.filter((val, idx, arr) => arr.indexOf(val) === idx);
+        uniqueRoots.forEach(x => {
             const index = result.indexOf(x);
             const colKeys = [...x.cellMergeMeta.keys()];
             const cols = colsToMerge.filter(x => colKeys.indexOf(x.field) !== -1);
@@ -138,13 +138,13 @@ export class IgxGridUnmergeActivePipe implements PipeTransform {
                     col.field,
                     col.mergingComparer,
                     res,
-                    activeRowIndexes,
+                    activeRowIndexes.map(x => x - index),
                     isDate,
                     isTime,
                     this.grid);
 
             }
-            result.splice(index, (index + res.length - 1), ...res);
+            result.splice(index, res.length, ...res);
         });
 
 
--- a/projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
+++ b/projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
@@ -86,14 +86,67 @@ export class IgxGridCellMergePipe implements PipeTransform {
 
     constructor(@Inject(IGX_GRID_BASE) private grid: GridType) { }
 
-    public transform(collection: any, colsToMerge: ColumnType[], mergeMode: GridCellMergeMode, mergeStrategy: IGridMergeStrategy, activeRowIndexes: number[], pinned: boolean, _pipeTrigger: number) {
+    public transform(collection: any, colsToMerge: ColumnType[], mergeMode: GridCellMergeMode, mergeStrategy: IGridMergeStrategy, _pipeTrigger: number) {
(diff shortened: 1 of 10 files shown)


dkamburov:
Openned rowPinning sample and pinned a row: 
image
Also when I pin a row from cell merge sample I'm getting the following violation:
image

MayaKirova:
@dkamburov First one, I was not able to reproduce:
rowpin

 Are you sure you have the latest changes in the branch or are there any additional steps needed to reproduce?

Second one is a sample issue. Because of the way the data is added, the same object refs are used in the grid sample:

```
for (let i = 1; i <= 600_000; i++) {
            const rnd = Math.floor(Math.random() * length);
            allData.push(INVOICE_DATA[rnd]);
        }
```

Hence, duplicate keys. 

I fixed it in the sample.

projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
```diff
@@ -86,14 +86,74 @@ export class IgxGridCellMergePipe implements PipeTransform {
...
+
+    public transform(collection: any, colsToMerge: ColumnType[], activeRowIndexes: number[], pinned: boolean, _pipeTrigger: number) {
         if (colsToMerge.length === 0) {
             return collection;
         }
         if (this.grid.hasPinnedRecords && !pinned && this.grid.pinning.rows !== RowPinningPosition.Bottom) {
             activeRowIndexes = activeRowIndexes.map(x => x - this.grid.pinnedRecordsCount);
         }
-        const result = DataUtil.merge(cloneArray(collection), colsToMerge, mergeStrategy, activeRowIndexes, this.grid);
+        activeRowIndexes = activeRowIndexes.filter((val, idx, arr) => arr.indexOf(val) === idx).filter(x => !isNaN(x));
```
mddragnev: If this is only removing the duplicates can we use a set? Probably, it would be too much of a work to refactor everywhere `activeRowIndexes` is used but you can use it here like :
```
Array.from(new Set(activeRowIndexes)).filter(x => !isNan(x));
```

dkamburov:
@MayaKirova appologize on the non reproducible sample: I do had some changes on that sample(didn't noticed them at first as they were from my previous testing on the cell merging):
image

`[cellMergeMode]="'always'"` to the grid and  `[merge]="true"` to the columns

projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
```diff
@@ -86,14 +86,74 @@ export class IgxGridCellMergePipe implements PipeTransform {
...
+            const target = collection[index];
+            colsToMerge.forEach(col => {
+                const colMeta = target.cellMergeMeta.get(col.field);
+                const root = colMeta.root ||  (colMeta.rowSpan > 1 ? target : null);
+                if (root) {
+                    rootsToUpdate.push(root);
+                }
+            });
+        });
+        const uniqueRoots = rootsToUpdate.filter((val, idx, arr) => arr.indexOf(val) === idx);
```
mddragnev: `rootsToUpdate` can be a Set here as well

mddragnev (changes requested):
Additional, issues I found:
1. If you click on any of the headers in the dev sample an error is thrown.
Screenshot 2025-09-29 at 4 01 33 pm

2. Go to the performance project and enable the merging on all columns. Run the performance samples using `npm run start:performance` and go to the grid with 1 million records. Sort the `Position` column. Click on the first cell of the `Registered` column. An error is thrown with maximum callstack exceeded.
Screenshot 2025-09-29 at 3 53 58 pm
3. For 1 million records on the performance sample, it takes 4 to 5 seconds to merge the cells. Sometimes the browser crashes out of memory.  This makes all other features that are pipe related to feel very slow.

projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
MayaKirova: Changes are applied and issues should be solved. The performance comments is up for discussion I guess.

projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
```diff
@@ -152,7 +152,7 @@ export class IgxGridUnmergeActivePipe implements PipeTransform {
                     this.grid);
 
             }
-            result.splice(index, res.length, ...res);
+            result = result.slice(0, index).concat(res, result.slice(index + res.length));
```
mddragnev, line 155: Could we use a regular for-of loop here and instead of creating 3 new array coppies we can mutate the result (which is cloned beforehand) in place?

MayaKirova: Do you mean going through all items in `res` and replacing each one in the original array with the unmerged version?

mddragnev: Yes.

MayaKirova: I mean, we could. But it would look a bit ugly and also not sure how much more performant it would be considering that in the particular scenario:
> 2. Go to the performance project and enable the merging on all columns. Run the performance samples using `npm run start:performance` and go to the grid with 1 million records. Sort the `Position` column. Click on the first cell of the `Registered` column. An error is thrown with maximum callstack exceeded.
> 
> Screenshot 2025-09-29 at 3 53 58 pm

It would need to loop through 300k+ records and replace each one.

mddragnev: I thought that 
```
for(let i =0;i<res.length;++i) {
    result[index+i] = res[i];
}
```
would be faster because there is no array copying but changes the result in place. However, seems like the `slice` and `concat` version is perhaps more optimized by the js engine. I have tested this and, indeed, your current version is faster.

mddragnev (approved):

dkamburov:
mouse wheel when there is cell merging is becoming choppy, it was faster before the changes and there is a difference when compared to a grid without cell merging
https://github.com/user-attachments/assets/da5b5e05-a0aa-48da-a9ae-6bd464a9584b

MayaKirova:
> mouse wheel when there is cell merging is becoming choppy, it was faster before the changes and there is a difference when compared to a grid without cell merging https://github.com/user-attachments/assets/da5b5e05-a0aa-48da-a9ae-6bd464a9584b

@dkamburov I can confirm that none of the merge related pipes are actually triggered on mouse wheel, so it doesn't make sense for the change to effect it.

Are you sure it's not simply because of the large amount of data in the grid from the sample? It's choppy for me even without merging.

dkamburov:
Seems like that's due to the huge amount of data in that sample(same behavior without merging)

PR #16270, opened 2025-10-02
fix(action-strip): Assign parent to not destroy on DOM move. 

### LINKED ISSUE #15768: [Elements] Row Island Action strip is destroyed after interacting outside the child grid
## Description  
Per the title, in a Row Island with an Action Strip, after expanding and hovering over a few rows to confirm it's active, moving the pointer outside the child grid causes the strip to be destroyed (permanently removed from DOM) and it's never re-created again.

 * igniteui-angular version: Reproducible in 19.2.x before changes from #15766. Assuming it worked fine before, likely a regression
 * browser: N/A

## Steps to reproduce  

1. Open the Element demos
2. On the HGrid expand a row
3. Hover over some rows
4. Move the pointer our
5. Move back into the child grid and hover rows OR expand additional grids

## Result  
Once the pointer moves out the action strip is destroyed for good:
Image

## Expected result  
Actions strip should still work on other expanded child grids and after interacting outside one.

## Attachments  

### LINKED ISSUE #15911: [Elements] Action strip is destroyed after deleting last row with WC 6.0.x
## Description  
Action strip gets destroyed when deleting last row (or several last rows in succession) in grid, hierarchical grid, and tree grid

 * igniteui-angular version: 
 * browser: N/A

## Steps to reproduce  

1. Navigate to any sample that uses action strip
2. Delete the last row (or several in succession)

## Result  
Acrion strip disappears

## Expected result  
Action strip should not disappear

## Attachments  
  https://staging.infragistics.com/webcomponents-demos/grids/grid/action-strip
  https://staging.infragistics.com/webcomponents-demos/grids/hierarchical-grid/action-strip
  https://staging.infragistics.com/webcomponents-demos/grids/tree-grid/action-strip

Possibly related to #15768

MayaKirova:
Closes #15768   
Closes #15911
Closes https://github.com/IgniteUI/igniteui-webcomponents/issues/1861

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular-elements/src/app/custom-strategy.ts
+++ b/projects/igniteui-angular-elements/src/app/custom-strategy.ts
@@ -120,12 +120,6 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
             if (parent?.ngElementStrategy) {
                 this.angularParent = parent.ngElementStrategy.angularParent;
 
-                // action strip is reused in row island child grid
-                // assign parent so it's not destroyed on detach/attach.
-                if (element.tagName.toLocaleLowerCase() === 'igc-action-strip') {
-                    this.angularParent = (parent.ngElementStrategy as any).componentRef;
-                }
-
                 this.parentElement = new WeakRef(parent);
                 let parentComponentRef = await parent?.ngElementStrategy[ComponentRefKey];
                 parentInjector = parentComponentRef?.injector;
@@ -137,6 +131,12 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
                     // this.componentRef = parentAnchor.createComponent(this.componentFactory.componentType, { projectableNodes, injector: childInjector });
                     parentComponentRef = await parent?.ngElementStrategy[ComponentRefKey];
                     parentAnchor = parentComponentRef?.instance.anchor;
+                }
+
+                // action strip is reused for all rows
+                // assign parent so it's not destroyed on detach/attach.
+                if (element.tagName.toLocaleLowerCase() === 'igc-action-strip' || configParent.selector === 'igc-action-strip') {
+                    this.angularParent = parentComponentRef;
                 }
             } else if ((parent as any)?.__componentRef) {
                 this.angularParent = (parent as any).__componentRef;
--- a/projects/igniteui-angular-elements/src/app/custom-strategy.ts
+++ b/projects/igniteui-angular-elements/src/app/custom-strategy.ts
@@ -119,7 +119,6 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
             // ngElementStrategy getter is protected and also has initialization logic, though that should be safe at this point
             if (parent?.ngElementStrategy) {
                 this.angularParent = parent.ngElementStrategy.angularParent;
-
                 this.parentElement = new WeakRef(parent);
                 let parentComponentRef = await parent?.ngElementStrategy[ComponentRefKey];
                 parentInjector = parentComponentRef?.injector;
@@ -133,11 +132,6 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
                     parentAnchor = parentComponentRef?.instance.anchor;
                 }
 
-                // action strip is reused for all rows
-                // assign parent so it's not destroyed on detach/attach.
-                if (element.tagName.toLocaleLowerCase() === 'igc-action-strip' || configParent.selector === 'igc-action-strip') {
-                    this.angularParent = parentComponentRef;
-                }
             } else if ((parent as any)?.__componentRef) {
                 this.angularParent = (parent as any).__componentRef;
                 parentInjector = this.angularParent.injector;
--- a/projects/igniteui-angular-elements/src/app/custom-strategy.ts
+++ b/projects/igniteui-angular-elements/src/app/custom-strategy.ts
@@ -119,6 +119,13 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
             // ngElementStrategy getter is protected and also has initialization logic, though that should be safe at this point
             if (parent?.ngElementStrategy) {
                 this.angularParent = parent.ngElementStrategy.angularParent;
+
+                // action strip is reused in row island child grid
+                // assign parent so it's not destroyed on detach/attach.
+                if (element.tagName.toLocaleLowerCase() === 'igc-action-strip') {
+                    this.angularParent = (parent.ngElementStrategy as any).componentRef;
+                }
+
                 this.parentElement = new WeakRef(parent);
                 let parentComponentRef = await parent?.ngElementStrategy[ComponentRefKey];
                 parentInjector = parentComponentRef?.injector;
--- a/projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
+++ b/projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
@@ -279,6 +279,9 @@ export class IgxActionStripComponent implements IgxActionStripToken, AfterConten
      * ```
      */
     public show(context?: any): void {
+        if(!this._originalParent) {
+            this._originalParent = this._viewContainer.element.nativeElement?.parentElement;
+        }
         this.hidden = false;
         if (!context) {
             return;
--- a/projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
+++ b/projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
@@ -147,7 +147,7 @@ export class IgxGridEditingActionsComponent extends IgxGridActionsBaseDirective 
         const context = this.strip.context;
         const grid = context.grid;
         grid.deleteRow(context.key);
-
+        this.grid.cdr.detectChanges();
         this.strip.hide();
     }
 


projects/igniteui-angular/src/lib/action-strip/action-strip.component.ts
```diff
@@ -279,6 +279,9 @@ export class IgxActionStripComponent implements IgxActionStripToken, AfterConten
      * ```
      */
     public show(context?: any): void {
+        if(!this._originalParent) {
```
damyanpetev: Hm, if every workflow goes through `show` we might even move the caching of the original parent here `ngAfterViewInit` where it was originally, so both cache and return are on the show/hide pair.

Anyway, I'm assuming this fixes the disappearing strip if it inits in some case as detached, but still is attached before showing.

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
```diff
@@ -147,7 +147,7 @@ export class IgxGridEditingActionsComponent extends IgxGridActionsBaseDirective
         const context = this.strip.context;
         const grid = context.grid;
         grid.deleteRow(context.key);
-
+        this.grid.cdr.detectChanges();
```
damyanpetev: Okay, the original parent bit I get, but do we really need this? Any context why a full grid change detection is needed/doing?

damyanpetev: Surely the deletion and CRUD service will also do that if needed, so this one would be extra even.

MayaKirova: `deleteRow` removes the record from the data, but the actual removal of the DOM element happens on the next change detection. This creates a bit of a timing issue. 

I no longer remember the details, since it has been a while, however I can confirm that the code does not work without the change detection (the issue reproduces without the change detection and is fixed with the change detection).

MayaKirova: @damyanpetev Here are a few observation.

The issue is specifically when you delete the last row and there's no next row in the grid, for example you have a grid with only 2 rows and you delete the second one. So this means that the RowComponent associated with the row will be destroyed. The issue happens when the moment the row is destroyed the action strip is still inside it:

image

Hence the destroy and removal of the row, also removes and destroys the action strip.

To me it seems that the logic that moves the action strip back to the root of the grid, the one inside the `hide` API:

```
  this.renderer.appendChild(this._originalParent, this._viewContainer.element.nativeElement);
```

Happens asynchronously - so it does not immediately move it to the root in time, before angular goes and destroys the parent. So regardless of the order:

```
        this.strip.hide();
        grid.deleteRow(context.key);
```

or

```
 grid.deleteRow(context.key);
 this.strip.hide();
```

It never manages to move the strip out of the row in time.

Also another fun observation - if you debug the `deleteRowHandler` code or the one inside the `hide` method, the issue stops reproducing as then the code executes synchronously and moves the strip to the root just fine.

Adding a change detection somewhere in the `deleteRowHandler` (doesn't matter where - before, between or after the other actions) seems to force it to resolve the operation and causes it work again.

damyanpetev: @MayaKirova you're right to suspect something async happening, sure, except it's not the `appendChild` being async (that one is very much synchronous for the dom version). The The actual delete calls `markForCheck` and the actual re-render happens on next cycle so order of calls in that location doesn't change the outcome.

What I'm actually observing is quite interesting:
```
this.strip.hide();
```
works exactly as it should and moves the strip back to the grid element (can confirm with a simple log to avoid messing with timing during debug). Except something else moves the strip _back_ to the row and gets it deleted permanently from the DOM by destroying the row. That's why the `detectChanges()` helps, cuz it forces the deletion to happen first, so the return to DOM on `hide()` sticks.

Logpoint at the end of `hide` like so
image
and one on `renderer.removeChild` yields the following when I hit the delete on a single row:
image

So the actual issue which I haven't dug out yet - why on earth is the action strip returning to its row just before it gets destroyed so it can deleted along with its DOM host.

MayaKirova: @damyanpetev It seems to me that the `show` gets called after the `hide`, so it moves it back:
image
 Probably because the row is not yet deleted and the mouse is still over it.

damyanpetev: image
Yeah, was just going to note that there's a secondary show()

projects/igniteui-angular-elements/src/app/custom-strategy.ts
```diff
@@ -131,6 +131,7 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
                     parentComponentRef = await parent?.ngElementStrategy[ComponentRefKey];
                     parentAnchor = parentComponentRef?.instance.anchor;
                 }
+
```
damyanpetev:
```suggestion
```

projects/igniteui-angular/src/lib/action-strip/grid-actions/grid-editing-actions.component.ts
```diff
@@ -147,7 +147,6 @@ export class IgxGridEditingActionsComponent extends IgxGridActionsBaseDirective
         const context = this.strip.context;
         const grid = context.grid;
         grid.deleteRow(context.key);
-
         this.strip.hide();
```
damyanpetev, line 151:
```suggestion

        this.strip.hide();
```

damyanpetev:
PS This should also address https://github.com/IgniteUI/igniteui-react/issues/105, right?

projects/igniteui-angular/src/lib/grids/grid-base.directive.ts
```diff
@@ -6438,6 +6438,12 @@ export abstract class IgxGridBaseDirective implements GridType,
         }
     }
 
+    protected viewDetachHandler(args) {
+        if (this.actionStrip && args.view.rootNodes.find(x => x === this.actionStrip.context?.element.nativeElement)) {
```
damyanpetev, line 6442: I'd flip that to:
```suggestion
        if (this.actionStrip?.context?.element.nativeElement && args.view.rootNodes.includes(this.actionStrip.context.element.nativeElement)) {
```
just so the check doesn't run if there's no context on the strip set at all.

---

In any case, this is getting to be a bit of a rabbit hole and I got GroupBy and possibly paging as extra cases. So far I've been able to find the grouping scenario with deleting the last row of a group that does the trick _and_ once again there's a rogue `show()` in there (two actually) before the remove:
image
The paging one haven't caught yet.

To that end:
- Do add _some_ explanation as to what's not working and what/why that code is doing there, `beforeViewDetach` is a _very_ chatty event to just take in general and connect to the specific case.
- This is becoming more and more involved of a fix and at this stage I don't think we can do without test coverage - this should be automatable both in Angular and Elements-specific tests and we should have the scenarios and checks. If we do allow the fix in without tests, those should be the next thing to add immediately after.
- And lastly, as the screenshot above shows _and_ our previous debug logs from the initial case - there's always a show _after_ the action strip has already once been hidden correctly. Both of those fixes are merely covering up the fact that the show is likely not managed correctly and the strip can't properly close itself as it deletes rows and we might need to address the actual cause instead.

MayaKirova: @damyanpetev 
1. The specific scenario shown to me by @MarielaTihova was a grid with paging and groupby (the one in the elements grid sample) where if you keep deleting the last data row until it's replaced by group by row, the action strip gets destroyed. 

   In that specific scenario the row is not destroyed but instead cached (see the `igxTemplateOutlet`), which means it gets detached and it detaches together with the action strip inside. And since it's removed from the DOM, elements destroys it. We already have `beforeViewDetach` event, which fires before the row detaches, so I just added the same logic there as for the row's destroy handler.

   Hope it makes more sense now.

2. Sure I can work on automation either as part of this PR or a separate one, depending on whether @dkamburov wants the fix to be included quicker with the next patch or if it can wait.

3.  I'm open to your suggestions on a better fix if you have one.

damyanpetev: Not sure I got the grouping case right, but okay.
As for the alternative suggestion, it's been the same as the last two times - try to avoid the extra `show()` when the strip closed itself. That seems to revolve around the mouseenter handeler (should it be pointer instead?), so I tried a check for the related target, excluding children of the row (possibly redundant) and children of the action strip itself:
image
```ts
@HostListener('mouseenter', ['$event'])
public showActionStrip(e: MouseEvent) {
    const from = e.relatedTarget as HTMLElement | null;
    if (this.grid.actionStrip && !(!from || this.nativeElement.contains(from) || this.grid.actionStrip.el.nativeElement.contains(from))) {
        this.grid.actionStrip.show(this);
    }
    this.grid.hoverIndex = this.index;
}
```
That handles the initial bug just fine from what I saw (delete the single root of the Tree Grid in Elements demo) and the Grouping scenario (delete the last row of a group). Still something iffy on open from when another row takes the place under the cursor after delete and I have to move to show the strip so the check might need smth extra, but it doesn't break in any of the cases at least.

damyanpetev: Could also track enter/leave as state, don't believe there's a leave in-between those multiple enters

MayaKirova: @damyanpetev Since you already have the changes locally, might as well commit and make a PR, if that's ok with you. I'll close mine as redundant.

damyanpetev: Eh, what I posted on top was what I have. However, as I said it doesn't _quite_ work since there's a secondary `show()` on the "deleted" row that's ignored, _but_ that row is 'replaced' by the next in context and that's how it works in main right now - the strip self-closes and the row opens it back up immediately, the delete overrides the context and the strip somehow remains open on the same DOM row and works fine. So attempting to avoid the open.. is odd. We might need to discuss further.
TLDR: The behavior I'm trying to avoid is what we apparently rely on to handle normal delete scenarios, which is a bit baffling.

damyanpetev: Yeah, it seems there's no simple way of handling the context reuse scenario (other than perhaps the row `data` setter and some tracking to determine of the action strip was visible on the last state of the row..). So more differently iffy options it seem.

In that case, I suppose the destroy/detach handlers can be our solution for now, though I still dislike the extra show :)

PR #16316, opened 2025-10-15
Potential fix for code scanning alert no. 9: Workflow does not contain permissions

kdinev:
Potential fix for [https://github.com/IgniteUI/igniteui-angular/security/code-scanning/9](https://github.com/IgniteUI/igniteui-angular/security/code-scanning/9)

To fix the problem, we should explicitly declare the required permissions for the GitHub Actions job so that the GITHUB_TOKEN issued to the job has only the privileges necessary for the workflow to operate. The `actions/stale` action only needs to comment on and label issues and pull requests—requiring `issues: write`, `pull-requests: write`, and at least `contents: read` (for basic repository access). The permissions block can be added at either the workflow level (affecting all jobs) or at the job level (affecting only the specific job). Since there’s only one job, placing the block at the job level suffices. The fix involves adding a `permissions:` block directly under the `stale:` job definition, above `runs-on:` (typically at line 10 in the provided code).

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -2,6 +2,8 @@
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
 name: Node.js CI
+permissions:
+  contents: read
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


dkamburov:
we need that fix for 19.2 and 20.1.x

PR #16363, opened 2025-10-23
fix(igxGrid): Fix unmerge when merge groups partially overlap.

### LINKED ISSUE #16358: [Bug]: Cell merging splitting groups splitts more groups than it should
## Description  
When activating a certain cell that's merged in order to split it the feature splits other groups that are supposed to stay merged.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Open the following example https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/cell-merging#angular-cell-merging-example
2. Click on the cell in the last column and in the first row
3. Observe the merged groups

## Result  
See the image cells that are supposed to stay merged were split

## Expected result  
the non active cell/row groups should remain merged

## Attachments  

Image

### LINKED ISSUE #16361: [Bug]: Cell merging custom strategy is adding additional groups when splitting merged groups
## Description  
Scrollbar and records count is increasing when splitting groups.

 * igniteui-angular version: 
 * browser: 

## Steps to reproduce  

1. Go to https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/cell-merging#demo
2. See the video attachment
3. Start clicking on cells from the first and then from the second column

## Result  
Scrollbar is increasing, it seams that in some scenarious there are additional records added(dupicaded)

## Expected result  
Scrollbar and record count should remain the same

## Attachments  

https://github.com/user-attachments/assets/b76538a2-4e72-4d1e-bf42-7fc9386e8d7f

MayaKirova:
Closes #16358  
Closes #16361

Please also re-test: https://github.com/IgniteUI/igniteui-angular/issues/16146
And also do more extensive testing with more complex merge groups and integration scenarios with other features.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/grids/grid/src/grid.pipes.ts
+++ b/projects/igniteui-angular/grids/grid/src/grid.pipes.ts
@@ -1,5 +1,5 @@
 import { Inject, Pipe, PipeTransform } from '@angular/core';
-import { IGridSortingStrategy, IGridGroupingStrategy, cloneArray, DataUtil, FilteringExpressionsTree, FilterUtil, IFilteringExpressionsTree, IFilteringStrategy, IGridMergeStrategy, IGroupByExpandState, IGroupingExpression, ISortingExpression, IGroupByResult, ColumnType } from 'igniteui-angular/core';
+import { IGridSortingStrategy, IGridGroupingStrategy, cloneArray, DataUtil, FilteringExpressionsTree, FilterUtil, IFilteringExpressionsTree, IFilteringStrategy, IGridMergeStrategy, IGroupByExpandState, IGroupingExpression, ISortingExpression, IGroupByResult, ColumnType, IMergeByResult } from 'igniteui-angular/core';
 import { GridCellMergeMode, RowPinningPosition, GridType, IGX_GRID_BASE } from 'igniteui-angular/grids/core';
 
 /**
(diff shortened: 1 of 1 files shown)


projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
```diff
@@ -131,27 +131,17 @@ export class IgxGridUnmergeActivePipe implements PipeTransform {
...
-                    [x.recordRef, ...childRecs],
-                    col.field,
-                    col.mergingComparer,
-                    res,
-                    activeRowIndexes.map(ri => ri - index),
-                    isDate,
-                    isTime,
-                    this.grid);
-
+                data = data.concat([x.recordRef, ...childRecs]);
```
mddragnev: Isnt it redundant to add the uniqueRoot recordRef each time? Can't you just add it once before looping through the columns?

projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
```diff
@@ -131,27 +131,17 @@ export class IgxGridUnmergeActivePipe implements PipeTransform {
         }
         let result = cloneArray(collection) as any;
         uniqueRoots.forEach(x => {
-            const index = result.indexOf(x);
+            const index = collection.indexOf(x);
             const colKeys = [...x.cellMergeMeta.keys()];
             const cols = colsToMerge.filter(col => colKeys.indexOf(col.field) !== -1);
-            let res = [];
+            let data = [];
```
mddragnev: If this is a set from the beginning, I think that you can ommit the concatenation in the for loop and instead of mapping the childRecs on line 141, you can just add them in a forEach loop.

mddragnev:
If you run the performance sample with 1k records with `    [cellMergeMode]="'always'"` on the grid and `[merge]="true"` on the columns, then you get the following:
Screenshot 2025-10-29 at 4 19 54 pm
If you now click on row with ID = 8 and observe the row with ID 10 and 11 you can see that the group layout there is broken:
Screenshot 2025-10-29 at 4 21 11 pm

projects/igniteui-angular/src/lib/grids/grid/grid.pipes.ts
```diff
@@ -129,33 +129,44 @@ export class IgxGridUnmergeActivePipe implements PipeTransform {
...
+                    if (unmergedRec.cellMergeMeta?.get(col.field)) {
+                        // deep clone of object, since we don't want to pollute the original fully merged collection.
+                        const objCopy = {
+                            recordRef: origRecord.recordRef,
+                            ghostRecord: origRecord.ghostRecord,
+                            cellMergeMeta: new Map<string, IMergeByResult>()
+                        };
+                        // deep clone of inner map
+                        for (const [key, value] of origRecord.cellMergeMeta) {
+                            objCopy.cellMergeMeta.set(key, structuredClone(value));
```
mddragnev: I understand why you clone the merged record, but why is the deep cloning at all? Isnt just changing the reference of the object enough? I dont think that using `structuredClone` for a grid with 1M records where a value object might be very nested is a good idea. Having said that, if you open the performance sample with 1M records, sort by name and unmerge the first row, the grid hangs and throws an error.
What if we just copy the object and the override the specific column. Something like:
```
if (unmergedRec.cellMergeMeta?.get(col.field)) {
                        // deep clone of object, since we don't want to pollute the original fully merged collection.
                        const objCopy = {
                            recordRef: origRecord.recordRef,
                            ghostRecord: origRecord.ghostRecord,
                            cellMergeMeta: new Map<string, IMergeByResult>(origRecord.cellMergeMeta.entries())
                        };
                  
                       // update copy with new meta from unmerged data record, but just for this column
                        objCopy.cellMergeMeta?.set(col.field, unmergedRec.cellMergeMeta.get(col.field));
                        result[index + i] = objCopy;
                    } else {
                        // this is the unmerged record, with no merge metadata
                        result[index + i] = unmergedRec;
                    }
```

PR #16506, opened 2025-11-24
fix(*): moving the IgxOverlayOutletDirective and some renamed exports and fixing migration

kdinev:
Closes #  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [x] Refactoring

### Checklist:


projects/igniteui-angular/core/src/services/public_api.ts
```diff
@@ -5,18 +5,15 @@ export * from './animation/animation';
 export * from './csv/csv-exporter';
 export * from './csv/csv-exporter-options';
 export * from './csv/char-separated-value-data';
-export { Direction as ɵDirection, DIR_DOCUMENT as ɵDIR_DOCUMENT, IgxDirectionality as ɵIgxDirectionality } from './direction/directionality';
+export * from './direction/directionality';
```
damyanpetev: Would appreciate if this is not undone - neither the directionality service nor its bits are meant to be public or long-lived for that matter. Those exports were re-done that way for a reason.

projects/igniteui-angular/core/src/services/public_api.ts
```diff
@@ -5,18 +5,15 @@ export * from './animation/animation';
...
 export * from './exporter-common/base-export-service';
 export * from './exporter-common/exporter-options-base';
 export * from './overlay/overlay';
 export * from './overlay/position';
 export * from './overlay/scroll';
-export {
-    AbsolutePosition, ConnectedFit, HorizontalAlignment, OffsetMode, OverlayAnimationEventArgs, OverlayCancelableEventArgs, OverlayClosingEventArgs,
-    OverlayCreateSettings, OverlayEventArgs, OverlaySettings, Point, PositionSettings, RelativePosition, RelativePositionStrategy, Size, VerticalAlignment, Util
-} from './overlay/utilities';
+export * from './overlay/utilities';
```
damyanpetev: And also careful with star exports - at least `OverlayInfo` and `OutOfViewPort` from this file were not public and this exposes those types.

damyanpetev:
@copilot Can you refactor the custom migration logic behind the `shouldMigrateImports` into a separate migration (leaving us with two migrations for version 21.0.0) since the schema is not working/supported on migrations and it should use a separate entry in the collection json with `optional: true` flag to achieve the prompt instead.
See https://github.com/angular/angular-cli/blob/36cf3afb485a01f86c4c90f136b38a3cf338e313/packages/schematics/angular/migrations/migration-collection.json#L8 for reference.

projects/igniteui-angular/core/src/services/public_api.ts
kdinev: @damyanpetev All are used by other entry points and need to be exported. They can be refactored to go into their own entry point, or under a different one, but they need to be exported.

projects/igniteui-angular/core/src/services/public_api.ts
kdinev: I did the star, because something was needed from other entry points, which was not exported. I can revert this one.

kdinev: Reverted it and added the IgxOverlayOutletDirective

PR #16564, opened 2025-12-01
chore(*): update ng@21.0.2

Lipata:
### Checklist:


Lipata:
@copilot can you fix the package-lock.json file so that we don't get the following error:

```console
npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

PR #16708, opened 2026-01-08
fix(perf): Apply base grid scroll performance optimizations.

### LINKED ISSUE #16709: Virtualization performance improvements
Introduce faster and smoother scrolling experience in the virtualization directive. Mainly related to the grid scrolling

Several items to look into. Look for more info into the additional context.

N\A

## Additional context

- [ ] State/data virtualization.
- [ ] Fenwick tree - https://en.wikipedia.org/wiki/Fenwick_tree
- [ ] Smooth scrolling
- [ ] Throttle scroll
- [ ] afterNextRender • Angular - https://angular.dev/api/core/afterNextRender
- [ ] CSS translate instead of DOM units

## Additional ideas after weekly check-in:

- [ ] refactor `initSizesCache` to init array with needed size and then populate it.
- [ ] refactor `isIndexOutsideView` section where embedded views are mapped.

MayaKirova:
Closes #16709

-  throttle scroll
- move diff check out of ngDoCheck
- use `transform` instead of `top` to limit layout reflow.
- remove HostListener that reads `scrollTop` to prevent unnecessary layout reflow.
- remove zone runs where possible.
- use `afterNextRender` when updating position of container in DOM.

### DIFF
--- a/projects/igniteui-angular/grids/core/src/grid-public-row.ts
+++ b/projects/igniteui-angular/grids/core/src/grid-public-row.ts
@@ -181,8 +181,12 @@ abstract class BaseRow implements RowType {
         return false;
     }
 
+    /**
+     * Gets whether the row is disabled.
+     * A disabled row represents a ghost placeholder created by row pinning.
+     */
     public get disabled(): boolean {
-        return this.grid.isGhostRecord(this.data);
+        return this.grid.isGhostRecordAtIndex(this.index);
     }
 
     /**
--- a/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
+++ b/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
@@ -7899,8 +7899,10 @@ export abstract class IgxGridBaseDirective implements GridType,
      * If record is pinned but is not in pinned area then it is a ghost record.
      *
      * @param dataViewIndex The index of that record in the data view.
+     * @hidden
+     * @internal
      */
-    private isGhostRecordAtIndex(dataViewIndex) {
+    public isGhostRecordAtIndex(dataViewIndex) {
         const isPinned = this.isRecordPinned(this.dataView[dataViewIndex]);
         const isInPinnedArea = this.isRecordPinnedByViewIndex(dataViewIndex);
         return isPinned && !isInPinnedArea;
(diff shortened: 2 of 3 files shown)


dkamburov:
The changes look good and functionality works as expected. 

I did notice one minor behavior difference while testing:
- **Issue**: When holding Tab to navigate between cells during editing, the navigation feels slightly slower than before
- **Possible cause**: This might be related to the throttle timing or the `transform` scroll implementation

This is very minor, but wanted to flag it in case it's worth investigating. Otherwise, I'm verifying the PR.

dkamburov:
## Performance Optimization Summary

### Detailed Optimizations

| Optimization | Technique | Before (Baseline) | After (Optimized) | Expected Delta |
|--------------|-----------|-------------------|-------------------|----------------|
| **Array allocation** | push() → pre-allocated new Array(n) | O(n) resizes with copying | O(1) direct assignment | ~20-40% faster for sizesCache building |
| **DOM element lookup** | Repeated .map() → cached getter | Re-computed every scroll | Single traversal, cached | ~15-25% fewer CPU cycles per scroll |
| **Change detection** | ngDoCheck → property setter | Called every CD cycle | Called only on data change | ~30-50% fewer CD calls |
| **CSS positioning** | style.top → transform: translateY() | Triggers layout reflow | GPU-accelerated, no reflow | ~50-70% faster paint |
| **Scroll event handling** | Raw events → 40ms throttle | ~60+ events/sec processed | ~25 events/sec processed | ~60% fewer scroll handlers |
| **Scroll helper listener** | @HostListener removed | Duplicate event processing | Single event path | ~10-15% less overhead |

### Aggregate Impact Estimate (100k rows, 100 scroll operations)

| Metric | Before (est.) | After (est.) | Delta |
|--------|---------------|--------------|-------|
| **Long Tasks Count** | ~80-120 | ~30-50 | **-50% to -60%** |
| **Avg Long Task Duration** | ~80-120ms | ~55-75ms | **-30% to -40%** |
| **Total Blocking Time** | ~8-12 sec | ~3-5 sec | **-50% to -60%** |
| **Frames per second** | ~15-25 fps | ~40-55 fps | **+100% to +150%** |

dkamburov:
> The changes look good and functionality works as expected.
> 
> I did notice one minor behavior difference while testing:
> 
> * **Issue**: When holding Tab to navigate between cells during editing, the navigation feels slightly slower than before
> * **Possible cause**: This might be related to the throttle timing or the `transform` scroll implementation
> 
> This is very minor, but wanted to flag it in case it's worth investigating. Otherwise, I'm verifying the PR.

Please ignore my previous message. Everything seems to be working as expected, likely due to temporary high resource usage.

Issue #16700, opened 2026-01-08
Create Angular Wrapper for GridLite


ChronosSF:
Should follow the best practices for creating ng wrappers from wc as set by IgxChat. 

Should allow native ng templating.

- [ ] https://github.com/IgniteUI/igniteui-angular-samples/issues/3865
- [ ] https://github.com/IgniteUI/igniteui-docfx/issues/6340
- [ ] API Docs

dkamburov:
Api docs are available as part of the igniteui-angular package public apis - https://staging.infragistics.com/api/angular/igniteui-angular/21.2.0/classes/IgxGridLiteComponent/

PR #16777, opened 2026-01-20
feat: Add Grid Lite Angular wrapper

mddragnev:
Related #16700 
 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -231,3 +231,15 @@ In order to test a pull request that is awaiting test, perform the following act
 4. Verify that the expected behavior is observed with the changes in the pull request.
 5. Return the pull request in a not fixed state if you're still reproducing the issue.
 6. Don't forget to make the necessary status updates, as described in the workflow section.
+
+# Running the Dev Demos
+The developer demos for testing purposes are runnable:
+
+```bash
+npm i
+npm start
+```
+
+If you cannot find a suitable demo for a component, when trying to test and reproduce an issue or scenario, then add a new demo for the scenario.
+
+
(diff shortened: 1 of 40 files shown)


damyanpetev:
Just a note:
The current template renders inside the Web Component **Shadow DOM**, so global styles (other than CSS vars) may not react elements inside. That affect templates with other `igx-` components that rely on the global theme in styles.css (`igc-` web components should be fine though). So if such component should be used in the template currently it should also bring the theme styles in as well:
```html
  <igx-grid-lite-column field="active" dataType="boolean" header="Active">
    <ng-template igxGridLiteCell let-value>
      <link rel="stylesheet" href="/styles.css">
      <igx-checkbox [checked]="value"></igx-checkbox>
    </ng-template>
  </igx-grid-lite-column>
```
The example `<link rel="stylesheet" href="/styles.css">` is the same as in the `index.html` and contains the Ignite UI theme; the link is heavily cached/optimized by the browser so it's roughly equivalent to adopting the root styles in the shadow root.

We're likely to address this further, but it will on the web component level to bake in functionality to automatically adopt the root styles for templated cells in a future release.

Edit: will be resolved by integrating https://github.com/IgniteUI/igniteui-grid-lite/pull/46

damyanpetev:
Additional issue, if a class or any other rule targeting the Angular host with some size modifications:
```html
<igx-grid-lite class="grid"
```
```css
.grid {
    width: 50%;
}
```
That obviously doesn't work as the host is `display: inline` with browser defaults. Even if adjusted to say 'block' I somewhat doubt it'll behave like the wrapped component anyway. An option would be setting `:host { display: contents; }` and _transfer_ class to the  wrapped element which works for most cases. Except you can't write rules targeting the host by element selector for much else than CSS variables as other stuff won't apply. And any CSS part items will _always_ need to be applied to the inner `igc-grid-lite` anyway. Might end up a documented quirk of the wrapper.

Edit: Eh, forgot about view encapsulation on the parent component that we can't control :/ So even if the class is transferred, it's not guaranteed to do anything for the component below.

Edit 2: Two options #16896 and #16898 with the first one winning out for simplicity.

PR #16784, opened 2026-01-21
fix(grid-pinning): remove hide call in scrollToRow - master

### LINKED ISSUE #16493: [Elements] Action strip closes when clicking jump to row action
## Description  
Action strip closes when clicking jump to row action. You have to then exit the row to show it agin.

 * browser: any

## Steps to reproduce  

1. Open https://www.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/tree-grid/row-actions#custom-implementation
2. Pin a row
3. Scroll so the row is out of view
4. Click on the `Jump down` button of action strip

## Result  
The action strip closes.

## Expected result  
The action strip shouldn't close

## Attachments  

Image

IMinchev64:
Closes #16493   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
+++ b/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
@@ -7828,18 +7828,19 @@ export abstract class IgxGridBaseDirective implements GridType,
 
     protected hasMenuPinningActions(): boolean {
         const strip = this.actionStrip;
-        if (!strip?.actionButtons?.length) {
+        const actionButtons = strip?.actionButtons;
+
+        if (!actionButtons?.length) {
             return false;
         }
 
-        let hasMenuPinningActions = false;
-        strip.actionButtons.forEach((button) => {
-            if (button instanceof IgxGridPinningActionsComponent && button.asMenuItems) {
-                hasMenuPinningActions = true;
-            }
-        });
-
-        return hasMenuPinningActions;
+        return actionButtons
+            .toArray()
+            .some(
+                (button) =>
+                    button instanceof IgxGridPinningActionsComponent &&
+                    button.asMenuItems
+            );
     }
 
     protected horizontalScrollHandler(event) {


mddragnev:
@skrustev @IMinchev64 The change is fine. However, if the action strip is set as menu this behavior is not perfect IMO.
I think that when having the action strip as menu we should close the menu on every action click
https://github.com/user-attachments/assets/635939be-2a58-4f49-830e-0fa6898cc35a

IMinchev64:
@mddragnev I added a condition to continue hiding the strip when the action strip is a menu.

Note that testing this in elements would be a bit of a hassle currently as:
1. The fix for this issue (#16834) would need to be merged in both master and 21.0.x as the action strip is broken in elements for the respective branches 
2. Even when the action strip works as a whole, using it as a menu is broken in elements as the moment we try to hover over the menu, the action strip gets hidden. This should be logged and fixed separately.

mddragnev:
The issue still reproduces when the record you try to scroll to is out of view

IMinchev64:
> The issue still reproduces when the record you try to scroll to is out of view

I am failing to reproduce this scenario with the same result

IMinchev64:
Replaced the unconditional action-strip hide/reset on vertical scroll with a conditional check: keep it only when the current context row is pinned, still connected to the DOM, and hovered; otherwise keep the previous hide + context reset behavior. This preserves virtualization cleanup while preventing the strip from disappearing during pinned-row jump/scroll scenarios.

dkamburov:
@IMinchev64 please click on **resolve conversation** button if the review comments are ready.

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -7798,7 +7798,7 @@ export abstract class IgxGridBaseDirective implements GridType,
         const keepActionStrip =
             !!context?.pinned &&
             !!contextEl?.isConnected &&
-            !!context?.asMenuItems;
+            !context?.asMenuItems;
```
mddragnev: The context is the row element here. The row does not have asMenuItems property, thus this will evaluate to undefined every time. You should get the info you need from the actionstrip

mddragnev (approved):

PR #16818, opened 2026-01-29
fix(stepper): prevent step indicator from shrinking with long titles - master

### LINKED ISSUE #16816: IgxStepper (vertical + titlePosition="end"): long step title compresses the step indicator into an oval
## Description  
When `IgxStepper` is in vertical orientation and the step title is positioned at the end, providing a very long step title causes the step header layout to shrink the step indicator (the circle with the step number). Instead of staying a fixed circle, the indicator is compressed horizontally and becomes an oval/pill shape.

This looks like a flexbox shrink issue in the step header row: the title is long (and effectively “unshrinkable” due to nowrap/intrinsic width), so the indicator ends up shrinking instead.

 * igniteui-angular version: latest
 * browser: all

## Steps to reproduce  
1. Open the sample: https://stackblitz.com/edit/odv1zhr8?file=src%2Fapp%2Fstepper-label-position-and-orientation-sample%2Fstepper-label-position-and-orientation-sample.component.html
2. Run the preview (the stepper is already configured with):
   - `orientation="vertical"`
   - `titlePosition="end"`
3. Use a very long text as the first step title (already present in the sample via `igxStepTitle`).
4. Observe the first step header.

## Result
The step indicator (the “1” circle) is **squeezed** and rendered as an **oval** instead of a circle when the step title is too long to fit in the available width.

## Expected result
The step indicator should **always preserve its circular shape** (fixed width/height), regardless of how long the title is. If the title is too long, the title should **truncate (ellipsis)** or **wrap** (depending on design), but **the indicator must not shrink**.

## Attachments
- Repro sample: https://stackblitz.com/edit/odv1zhr8?file=src%2Fapp%2Fstepper-label-position-and-orientation-sample%2Fstepper-label-position-and-orientation-sample.component.html
Image

georgianastasov:
Closes #16816   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/core/src/core/styles/components/stepper/_stepper-theme.scss
+++ b/projects/igniteui-angular/core/src/core/styles/components/stepper/_stepper-theme.scss
@@ -203,6 +203,7 @@
         font-size: rem(12px);
         height: $indicator-size;
         width: $indicator-size;
+        min-width: $indicator-size;
         white-space: nowrap;
         border-radius: var-get($theme, 'border-radius-indicator');
         color: var-get($theme, 'indicator-color');
@@ -407,10 +408,6 @@
             flex-direction: row;
             align-items: center;
             //gap: $title-gap--horizontal;
-        }
-
-        %igx-stepper__step-indicator {
-            flex-shrink: 0;
         }
     }
 
@@ -637,10 +634,6 @@
 
         %igx-stepper__step--start,
         %igx-stepper__step--end {
-            %igx-stepper__step-indicator {
-                flex: 1 0 auto;
-            }
-
             %igx-stepper__step-header {
                 @if $variant != 'fluent' {
                     padding: calc(#{$step-header-padding} / 2);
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/core/src/core/styles/components/stepper/_stepper-theme.scss
```diff
@@ -408,6 +408,10 @@
             align-items: center;
             //gap: $title-gap--horizontal;
         }
+
+        %igx-stepper__step-indicator {
+            flex-shrink: 0;
+        }
```
RivaIvanova: So this will affect the indicator not only when the stepper is in vertical orientation, but in horizontal orientation as well. There is a style for the horizontal case [here](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/core/src/core/styles/components/stepper/_stepper-theme.scss#L636) which will override these styles and achieve the same.

Image

Maybe these styles can be removed, and this case can be handled with `min-width` like in [WC](https://github.com/IgniteUI/igniteui-webcomponents/blob/master/src/components/stepper/themes/step/step.base.scss#L226).

Issue #14451, opened 2024-07-01
Stepper: Align content to Title and Subtitle


imincheva:
## Is your feature request related to a problem? Please describe.
Currently, the Content on the vertical Stepper isn't aligned to the Title and Subtitle like it has been designed in the handoff.

## Describe the solution you'd like
Align the Content to the Title and Subtitle. This has already been implemented in Indigo, so it should be applied in Material, Fluent and Bootstrap in both Angular and Web Components

Related PR:
#14380

**Current** (reference from Material)
<img width="413" alt="image" src="https://github.com/IgniteUI/igniteui-angular/assets/116892553/17e179a8-b946-490a-9ca2-5567c9a056b2">

**Expected** (reference from Material)
<img width="420" alt="image" src="https://github.com/IgniteUI/igniteui-angular/assets/116892553/6fb39f65-2ccf-498a-8173-6abae50ba6d4">

PR #16831, opened 2026-01-30
fix(stepper): prevent content shift in vertical stepper - master

### LINKED ISSUE #16822: igx-stepper (vertical + titlePosition="end") step content “jumps” a few pixels left when navigating between steps (regression vs 18.0.0)
## Description  
When switching between steps in **IgxStepper** with:
- `orientation="vertical"`
- `titlePosition="end"`

…the **step content container** (most noticeable on action buttons inside the content) shifts horizontally a few pixels to the **left** during navigation, causing a visible “jitter”/layout shift.

This is reproducible in StackBlitz and also observable in the official Stepper docs sample where orientation/title position can be changed at runtime.

**Regression:** The same sample tested with `igniteui-angular@18.0.0` does **not** show the shift.

 * igniteui-angular version: 
 * ✅ Reproducible: `21.0.x` `20.1.x`
 * ❌ Not reproducible: `18.0.0`
 * browser: all

## Steps to reproduce
1. Open the minimal repro:  
   https://stackblitz.com/edit/odv1zhr8-1qte6l6b?file=src%2Fapp%2Fstepper-label-position-and-orientation-sample%2Fstepper-label-position-and-orientation-sample.component.html
2. Navigate between steps (use next/prev buttons).
3. Observe the left edge of the step content (e.g., the primary action button) shifting a few pixels left/right when the active step changes.

## Result  
The step content horizontally shifts (“jumps”) a few pixels when changing the active step in a vertical stepper with `titlePosition="end"`.

## Expected result  
Step content should remain horizontally aligned and stable (no layout shift/jitter) when navigating between steps.

## Attachments  
Minimal repro (StackBlitz): 
(https://stackblitz.com/edit/odv1zhr8-1qte6l6b?file=src%2Fapp%2Fstepper-label-position-and-orientation-sample%2Fstepper-label-position-and-orientation-sample.component.html)
Image

georgianastasov:
Closes #16822   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,5 +1,6 @@
 name: Npm.js deploy
 permissions:
+  id-token: write
   contents: read
 
 on:
@@ -10,12 +11,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
         with:
           node-version: 22
           cache: 'npm'
           registry-url: 'https://registry.npmjs.org'
+
+      - name: Update NPM
+        run: npm install -g npm@latest
+
       - run: echo "VERSION=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
       - run: echo ${VERSION}
 
@@ -60,11 +65,7 @@ jobs:
       - name: Publish igniteui-angular
         run: npm publish --tag ${NPM_TAG}
         working-directory: dist/igniteui-angular
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
 
       - name: Publish igniteui-angular-i18n
         run: npm publish --tag ${NPM_TAG}
         working-directory: dist/igniteui-angular-i18n
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}


RivaIvanova (commented):
I think all these calculations are unnecessary, or at least I don't see why they are needed. Let me know.
Can't this be achieved by removing the `[aria-selected='true']` rule and setting the `padding-inline-start: $v-line-indent;` on `%igx-stepper__step-content-wrapper`? I saw that these `aria-selected` styles are related to [this issue](https://github.com/IgniteUI/igniteui-angular/issues/14451). The above suggestion seems to cover both cases, but test it across all themes as well.

georgianastasov:
> I think all these calculations are unnecessary, or at least I don't see why they are needed. Let me know. Can't this be achieved by removing the `[aria-selected='true']` rule and setting the `padding-inline-start: $v-line-indent;` on `%igx-stepper__step-content-wrapper`? I saw that these `aria-selected` styles are related to [this issue](https://github.com/IgniteUI/igniteui-angular/issues/14451). The above suggestion seems to cover both cases, but test it across all themes as well.

The jump happens because the content wrapper uses padding: 16px, but when a step becomes active we override only padding-inline-start to $v-line-indent.

$v-line-indent is calculated (step-header-padding + indicator-size/2) and is typically not 16px so during navigation the start padding flips 16px - $v-line-indent, which causes the horizontal jitter.

My fix removes that state-based change by making the content offset constant: effectively the layout previously ended up as margin-start ($v-line-indent) + padding-start ($v-line-indent) when active, so the stable value is 2 * $v-line-indent. That’s why padding-inline-start: calc(#{$v-line-indent} * 2) is used, and the separator is repositioned with calc(#{$v-line-indent} - (#{$separator-size} / 2)) to keep it aligned with the indicator center.

chrome_Xu5ROBfUll

I did try the suggested approach, and actually I started with something similar initially. While it can remove the jitter itself, it introduces a new visual regression: the whole step/content alignment shifts, so the stepper looks moved left/misaligned.

chrome_JYNu5sQgTy

So the current fix is intentionally written this way because it: removes the layout shift, preserves the intended alignment (indicator center + separator + content), avoids introducing a new visual issue in this configuration.

RivaIvanova:
I am not able to reproduce the left shifting/misalignment. I tested with the [sample](https://stackblitz.com/edit/odv1zhr8-1qte6l6b?file=src%2Fapp%2Fstepper-label-position-and-orientation-sample%2Fstepper-label-position-and-orientation-sample.component.html) in the issue.
image
image

georgianastasov:
> I am not able to reproduce the left shifting/misalignment. I tested with the [sample](https://stackblitz.com/edit/odv1zhr8-1qte6l6b?file=src%2Fapp%2Fstepper-label-position-and-orientation-sample%2Fstepper-label-position-and-orientation-sample.component.html) in the issue.

You’re right. I reviewed my earlier attempt and I did make a mistake: I removed margin-inline-start: $v-line-indent and left only padding-inline-start: $v-line-indent. That’s what caused the step/content block to look shifted/misaligned in my screenshot.

I re-tested. I remove the `[aria-selected='true']` rule and set `padding-inline-start: $v-line-indent` directly on `%igx-stepper__step-content-wrapper` while keeping the existing `margin-inline-start` and it fixes the jitter without introducing the alignment regression. 

So yes, we can go with the simpler change. I’ll implement it and push the update.

PR #16835, opened 2026-01-30
feat(overlay) use popover in overlay service

### LINKED ISSUE #16824: Use popover attribute in Overlay service
Overlay service as it is now has two major disadvantages:
- it moves the component shown in the overlay out of the angular application. With this step styling of the component could be corrupted.
- when used in outlet overlay may be shown hidden behind other components with higher stacking context.

To solve these issues instead of using `z-index` overlay could be shown via popover attribute.

wnvko:
Use `popover` global attribute to show the overlay. We are adding the attribute to overlay wrapper element and show the overlay by calling `showPopover`.

Closes #16824 

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:

### DIFF
--- a/projects/igniteui-angular/core/src/services/overlay/position/container-position-strategy.ts
+++ b/projects/igniteui-angular/core/src/services/overlay/position/container-position-strategy.ts
@@ -21,15 +21,9 @@ export class ContainerPositionStrategy extends GlobalPositionStrategy {
         this.io = Util.setupIntersectionObserver(
             outletElement,
             contentElement.ownerDocument,
-            () => this.internalPosition(contentElement)
+            () => this.updatePosition(contentElement)
         );
         this.internalPosition(contentElement);
-    }
-
-    private internalPosition(contentElement: HTMLElement): void {
-        contentElement.classList.add('igx-overlay__content--relative');
-        contentElement.parentElement.classList.add('igx-overlay__wrapper--flex-container');
-        this.updatePosition(contentElement);
     }
 
     /**
@@ -40,6 +34,13 @@ export class ContainerPositionStrategy extends GlobalPositionStrategy {
         this.io = null;
     }
 
+    private internalPosition(contentElement: HTMLElement): void {
+        contentElement.classList.add('igx-overlay__content--relative');
+        contentElement.parentElement.classList.add('igx-overlay__wrapper--flex-container');
+        this.setPosition(contentElement);
+        this.updatePosition(contentElement);
+    }
+
     private updatePosition(contentElement: HTMLElement): void {
         // TODO: consider using new anchor() CSS function when it becomes more widely supported: https://caniuse.com/mdn-css_properties_anchor
         const parentRect = contentElement.parentElement.parentElement.getBoundingClientRect();
@@ -47,6 +48,5 @@ export class ContainerPositionStrategy extends GlobalPositionStrategy {
         contentElement.parentElement.style.height = `${parentRect.height}px`;
         contentElement.parentElement.style.top = `${parentRect.top}px`;
         contentElement.parentElement.style.left = `${parentRect.left}px`;
-        this.setPosition(contentElement);
     }
 }
--- a/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
+++ b/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
@@ -4334,6 +4334,7 @@ export abstract class IgxGridBaseDirective implements GridType,
             }
         });
 
+
         this.zone.runOutsideAngular(() => {
             this.verticalScrollContainer?.getScroll()?.removeEventListener('scroll', this.verticalScrollHandler);
             this.headerContainer?.getScroll()?.removeEventListener('scroll', this.horizontalScrollHandler);


projects/igniteui-angular/core/src/services/overlay/position/container-position-strategy.ts
```diff
@@ -16,6 +16,11 @@ export class ContainerPositionStrategy extends GlobalPositionStrategy {
     public override position(contentElement: HTMLElement): void {
         contentElement.classList.add('igx-overlay__content--relative');
         contentElement.parentElement.classList.add('igx-overlay__wrapper--flex-container');
+        const parentRect = contentElement.parentElement.parentElement.getBoundingClientRect();
+        contentElement.parentElement.style.width = `${parentRect.width}px`;
+        contentElement.parentElement.style.height = `${parentRect.height}px`;
+        contentElement.parentElement.style.top = `${parentRect.top}px`;
+        contentElement.parentElement.style.left = `${parentRect.left}px`;
```
damyanpetev, line 50: Would that handle the container resizing?
In any case this would be preferably a temporary solution until we can stop moving the DOM around and ideally have the wrapper scale to the container with static CSS like 100% or w/e

CHANGELOG.md
```diff
@@ -2,6 +2,23 @@
...
 All notable changes for each version of this project will be documented in this file.
 
+## 21.2.0
+
+### New Features
+
+- `IgxOverlayService`
+    - Integrated HTML Popover API into the overlay service for improved z-index management and layering control.
+    - The overlay service now uses the Popover API to place overlay elements in the top layer, eliminating z-index stacking issues.
+    - Improved positioning accuracy for container-based overlays with fixed container bounds.
```
damyanpetev: Move this under the 21.1 section and remove the merge conflict leftovers :)

simeonoff (approved):

dafo:
1. Run the demos locally
2. Navigate to http://localhost:4200/gridCellEditing
3. Double-click any row in the grid to open the editing dialog
4. While the editing dialog is opened - toggle the nav-drawer (no matter if you will open or close it)

Actual result:
The editing dialog position is broken

screenshot

damyanpetev:
> 1. Run the demos locally
> 2. Navigate to http://localhost:4200/gridCellEditing
> 3. Double-click any row in the grid to open the editing dialog
> 4. While the editing dialog is opened - toggle the nav-drawer (no matter if you will open or close it)
> 
> Actual result: The editing dialog position is broken
> 
> screenshot

I would've thought the resize observer would handle that case, though it is possible to shift the position but that (without the native `anchor` positioning) I'd assume would be a problem in other cases too :/

projects/igniteui-angular/core/src/services/overlay/overlay.spec.ts
```diff
@@ -1,6 +1,12 @@
...
 import { fakeAsync, TestBed, tick, waitForAsync } from '@angular/core/testing';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { PlatformUtil } from 'igniteui-angular';
+import { scaleInVerTop, scaleOutVerTop } from 'igniteui-angular/animations';
+import { IgxAvatarComponent } from 'igniteui-angular/avatar';
+import { IgxCalendarComponent } from 'igniteui-angular/calendar';
+import { IgxCalendarContainerComponent } from 'igniteui-angular/date-picker';
+import { IgxToggleDirective } from 'igniteui-angular/directives';
 import { first } from 'rxjs/operators';
 import { UIInteractions } from '../../../../test-utils/ui-interactions.spec';
```
damyanpetev:
```suggestion
import { UIInteractions, wait } from '../../../../test-utils/ui-interactions.spec';
```

projects/igniteui-angular/core/src/services/overlay/overlay.spec.ts
```diff
@@ -44,6 +45,8 @@ const CLASS_OVERLAY_MAIN = 'igx-overlay';
 const CLASS_SCROLLABLE_DIV = 'scrollableDiv';
 const DEBOUNCE_TIME = 16;
 
+const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
```
damyanpetev:
```suggestion
```

projects/igniteui-angular/core/src/services/overlay/utilities.ts
```diff
@@ -332,3 +337,110 @@ export class Util {
     }
 }
 
+/**
+ * @hidden @internal
+ * Helper class for managing IntersectionObserver-based position updates
+ * Used by position strategies that need to reposition overlays when target elements move
+ */
+export class IntersectionObserverHelper {
```
damyanpetev: Ah, okay, just noticed the public API export;
At this point, I'd much prefer not adding to the pile if possible - and in this case it should be. I think containing this within a `Util` method would be preferable.

projects/igniteui-angular/core/src/services/overlay/utilities.ts
```diff
@@ -332,3 +337,110 @@ export class Util {
...
+
+        // Store initial position
+        this.previousRect = target.getBoundingClientRect();
+
+        // Set up IntersectionObserver to trigger position checks
+        // Use rootMargin to detect when element enters/exits observable area
+        this.intersectionObserver = new intersectionObserver(
+            (_entries) => {
+                // When IntersectionObserver detects visibility change, start continuous polling
+                this.startPositionUpdateLoop(target, onPositionUpdate);
```
damyanpetev: Oh, huge NOPE on this one. The entire `requestAnimationFrame` loop is ridiculous and basically does the detecting _instead_ of the observer using it for just the initial call to toggle itself.

Image
This goes on, the intersection callback doesn't trigger but the setup sure does on every `rAF` that calls the position callback.

projects/igniteui-angular/core/src/services/overlay/utilities.ts
```diff
@@ -332,3 +337,110 @@ export class Util {
...
+            },
+            {
+                root: null,
+                rootMargin: '0px', // Expand detection area to catch layout shifts
```
damyanpetev: Pretty sure that now does nothing for the intersection itself.

The idea was to shrink the root margin to the target so we can observe it move and pretty sure `0px` is the default anyway ([`rootMargin`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#rootmargin)) so "Expand detection area to catch layout shifts" this does.. not 😄
Not that it matters with `root: null` anyway I think 😂

projects/igniteui-angular/grids/core/src/grid.common.ts
```diff
@@ -39,14 +41,28 @@ export class RowEditPositionStrategy extends ConnectedPositioningStrategy {
...
 
         super.position(contentElement, { width: targetElement.clientWidth, height: targetElement.clientHeight },
-                    document, initialCall, targetElement);
+            document, initialCall, targetElement);
+
+        // Use the IntersectionObserverHelper to manage position updates when the target moves
+        this.observerHelper.setupIntersectionObserver(
+            targetElement,
+            document,
+            () => this.position(contentElement, { width: targetElement.clientWidth, height: targetElement.clientHeight }, document, false, targetElement)
```
damyanpetev: If this is calling back `this.position` and the `_size` is not read anyway - don't pass it? IIRC calling `clientWidth` will have a layout cost and it seems wasteful if it won't be [re-]used.

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -4334,7 +4334,7 @@ export abstract class IgxGridBaseDirective implements GridType,
             }
         });
 
-
+        this.rowEditPositioningStrategy.dispose();
```
damyanpetev: This should not needed, the toggle directive should call detach on close/destroy, where the overlay service should do the dispose already.

projects/igniteui-angular/core/src/services/overlay/utilities.ts
```diff
@@ -330,5 +335,41 @@ export class Util {
...
+    ): IntersectionObserver | null {
+        const intersectionObserver = getIntersectionObserver();
+        if (!intersectionObserver) {
+            return null;
+        }
+        if (!element || !doc) {
+            return null;
+        }
+        const rect = element.getBoundingClientRect();
+        const viewPortRect = Util.getViewportRect(document);
```
damyanpetev: Use the passed param?
```suggestion
        const viewPortRect = Util.getViewportRect(doc);
```

gedinakova:
@wnvko When a dialog is open, when resizing smaller the browser, the first time the dialog briefly goes above the nav-drawer before going under. In addition, when the nav-drawer is open, initially the dialog remains in its previous position, it moves only after the resizing starts.

[screen-capture (9).webm](https://github.com/user-attachments/assets/0c912cca-2dc5-4a30-bbd4-6f793d2f5c8c)

damyanpetev:
> @wnvko When a dialog is open, when resizing smaller the browser, the first time the dialog briefly goes above the nav-drawer before going under. In addition, when the nav-drawer is open, initially the dialog remains in its previous position, it moves only after the resizing starts.
> 
>  screen-capture.9.webm

Um, the drawer is initially pinned (not in overlay mode), so pretty sure the dialog going on top is before the drawer unpins, after which the drawer goes on top as expected.

wnvko:
> @wnvko When a dialog is open, when resizing smaller the browser, the first time the dialog briefly goes above the nav-drawer before going under. In addition, when the nav-drawer is open, initially the dialog remains in its previous position, it moves only after the resizing starts.
> 
>  screen-capture.9.webm

This how it is works on 21.0.12 branch
image
I am not even sure where this dialog should be positioned. It is using connected position strategy, but actually is trying to position in the middle of the grid. And then it is draggable so can be moved around. IMO this was not working correctly and before the changes introduced in this PR and in nav-drawer related to popover attribute.

PR #15791, opened 2025-05-12
feat(tooltip): align with WC tooltip

RivaIvanova:
Closes #15726  

### Additional information (check all that apply):
 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [x] All relevant tags have been applied to this PR
 - [x] This PR includes unit tests covering all the new code ([test guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Test-implementation-guidelines-for-Ignite-UI-for-Angular))
 - [ ] This PR includes API docs for newly added methods/properties ([api docs guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines))
 - [x] This PR includes `feature/README.MD` updates for the feature docs
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [x] This PR includes `CHANGELOG.MD` updates for newly added functionality
 - [ ] This PR contains breaking changes
 - [ ] This PR includes `ng update` migrations for the breaking changes ([migrations guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Update-Migrations))
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

PR #16837, opened 2026-01-30
fix(tooltip/snackbar): use overlay service size registry to set initial size - master

### LINKED ISSUE #16458: [Tooltip/Snackbar]: overlay content size is not properly reflected when component is placed inside another element
## Description  
The tooltip does not render properly when placed inside an element.

```html
Placed inside a button
<div class="container end">
  <button igxIconButton="flat" [igxTooltipTarget]="themeTooltipRef">
    <igx-icon>palette</igx-icon>
    <div #themeTooltipRef="tooltip" igxTooltip>Click to toggle between dark/light themes</div>
  </button>
</div>

Placed outside the button
<div class="container end">
  <button igxIconButton="flat" [igxTooltipTarget]="themeTooltipRef2">
    <igx-icon>palette</igx-icon>
  </button>
  <div #themeTooltipRef2="tooltip" igxTooltip>Click to toggle between dark/light themes</div>
</div>
```

Image

The reason for the first behavior is that inside the overlay service's [attach method](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/core/src/services/overlay/overlay.ts#L334), `getBoundingClientRect` returns the tooltip's [min-width](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/core/src/core/styles/components/tooltip/_tooltip-theme.scss#L25) CSS constraint, which is 24px, instead of the actual size.

Image

Apart from the poor formatting of the text, this gives the false impression that the autopositioning strategy has no effect.

Image

Before the major refactoring introduced in 20.1.0 by [this PR](https://github.com/IgniteUI/igniteui-angular/pull/15791), the tooltip behaved differently, giving the impression that it worked in both cases (inside and outside an element), but the overlay content wrapper's width was bigger than the actual tooltip content. This made it seem as if the tooltip is not properly positioned relative to its target.

Image

Image

 * igniteui-angular version: master, 20.1.x, 19.2.x
 * browser: any

## Steps to reproduce  

Play around with the code snippet below in the tooltip dev demos.
```css
.container {
  display: flex;
  align-items: center;
  justify-content: center;
  margin: 64px;
}

.end {
  justify-content: end;
  margin: 24px;
}
```
```html
<div class="container end">
  <button igxIconButton="flat" [igxTooltipTarget]="themeTooltipRef">
    <igx-icon>palette</igx-icon>
    <div #themeTooltipRef="tooltip" igxTooltip><img height="100" src="https://i.ibb.co/6ZdY7cn/Untitled-design-3.png"></div>
    <!-- <div #themeTooltipRef="tooltip" igxTooltip>Click to toggle between dark/light themes</div> -->
  </button>
</div>

<div class="container end">
  <button igxIconButton="flat" [igxTooltipTarget]="themeTooltipRef2">
    <igx-icon>palette</igx-icon>
  </button>
  <div #themeTooltipRef2="tooltip" igxTooltip><img height="100" src="https://i.ibb.co/6ZdY7cn/Untitled-design-3.png"></div>
  <!-- <div #themeTooltipRef2="tooltip" igxTooltip>Click to toggle between dark/light themes</div> -->
</div>
```

## Result  
In the description.

## Expected result  
Tooltip should size properly.

## Attachments  

RivaIvanova:
Closes #16458   

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


wnvko (changes requested):
So we know [this PR](https://github.com/IgniteUI/igniteui-angular/pull/15791) is breaking the tooltip. What is the reason this happening? Can we revert the change that made this happen?
Adding this additional property to the overlay settings and use it in this way in the tooltip looks like a breaking change. IMO it will be better to find out what happened in the above mentioned PR and to revert it.

RivaIvanova:
@wnvko The linked [PR](https://github.com/IgniteUI/igniteui-angular/pull/15791) does not break the tooltip. This PR was merged in version 20.1.0, but the behavior is observed in version 15.0.x too (maybe even before that, but I tested with that version).
tooltip-inside-icon-button

The PR adds a lot of changes, but the one that is related to the current behavior is adding [this CSS](https://github.com/IgniteUI/igniteui-angular/blob/master/projects/igniteui-angular/core/src/core/styles/components/tooltip/_tooltip-theme.scss#L25-L27):
```css
min-width: rem(24px);
max-width: rem(200px);
width: fit-content;
```
The new styles did not break the tooltip, but revealed the already broken behavior.

Overall, it is due to the element's size (width) being measured before being moved to the overlay content. If the element does not have a specific width set, like 100px, its size is affected by the styles that come from the parent container.

For example, consider this template:
```html
<div style="background: red; text-transform: uppercase; letter-spacing: 8px;" [igxTooltipTarget]="themeTooltipRef4">
    Parent Element
    <div #themeTooltipRef4="tooltip" igxTooltip>Click to toggle between dark/light themes</div>
</div>
```

The parent element (red div) has these styles `text-transform: uppercase; letter-spacing: 8px;` and I would expect the `igx-overlay__content` to look somewhat like in the image, with a width of 194.99px (the size of the tooltip).
tooltip-inside-parent-with-styles-expect

However, the result is the following:
tooltip-inside-parent-with-styles-actual

That's because the tooltip/element is measured and then moved to the overlay. While being measured, the tooltip is affected by the styles of the parent, and a bigger size is set as the `initialSize.width`.
tooltip-inside-parent-with-styles-reason

The snackbar behavior is due to [this PR](https://github.com/IgniteUI/igniteui-angular/pull/16440), which was introduced in version [20.1.9](https://github.com/IgniteUI/igniteui-angular/releases/tag/20.1.9). Not exactly breaking it, since before this PR, the snackbar had no size (perhaps because it was not present in the view yet). 
image
And the `igx-overlay__content` had no styles at all, e.g., `style="width: 200px"`.
image

I removed the additional property from the overlay settings and implemented a different approach, but similar to the previous one, it requires moving the element to the overlay and then getting its size.

PR #16871, opened 2026-02-05
refactor(grid): Removed scroll throttle for grid scrolls

rkaraivanov:


ChronosSF:
LGTM

MayaKirova:
@ChronosSF I'm against just removing it. Then we'd have almost no performance gain with large number of rows/cells on scroll.
 
I'd prefer if you give more information on the exact scenario you have tested/measured against and the issue you have with the throttle in them so that we may discuss a throttle variation that works for. Some options we have discussed with @rkaraivanov are:

- Have a smaller throttle.
- Have a throttle that depends on the number of rows/cells that need to render (i.e. 0 for a very small grid, but more for grids with larger sets of rows/cells that need to update on each scroll.)
- Use animationFrameScheduler since it might "feel" smoother.

rkaraivanov:
Closing this in favor of #16879

PR #16902, opened 2026-02-13
Recreate filtering expression condition based on correct condition.Name value [master]

### LINKED ISSUE #16841: Grid State - Extracted filter column state is always set to "contains"
## Description  
When setting e.g. two filters for the same column, the extracted "conditionName" is set to "contains" instead of what has been selected (e.g. "equals"). This sometimes happens to all of the filters and sometimes only to the last one (e.g. the second one in the attached video).

 * igniteui-angular version: 21.0.0
 * browser: Chrome 144.0.7559.110

## Steps to reproduce  

1. Select two column filters with an operation different from "contains"
2. Extract the grid state. The "conditionName" is set to "contains"
3. Apply the grid state. The affected filter is set to "contains" instead of what has been selected

The issue can be seen/reproduced in [this repository](https://github.com/remshams/igx-grid-state-filter-issue)

## Result  
The column filter(s) are set to "contains" instead of what has been selected.

## Expected result  
The column filter(s) should be set to what has been selected.

## Attachments  

https://github.com/user-attachments/assets/d4d2f5e0-d6ee-4e56-a2d0-5d3af953003f

hanastasov:
`conditionName` is used to recreate `expression.condition` when restoring grid state from IgxGridState directive. Without this fix, restoring grid state would change condition from whatever it was to the default condition "contains"

Solution:
1. When restoring, give preference to `condition.name` over `conditionName`
2. Update `conditionName = condition.name`

**What to reproduce and verify:**

1. Use the sample in the issue
2. Add filtering condition different than contains
3. Save grid state
4. Restore grid state - condition is restored to contains, not to what user has set.

Closes #16841 

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


projects/igniteui-angular/core/src/data-operations/expressions-tree-util.ts
```diff
@@ -138,25 +138,22 @@ function recreateOperatorFromDataType(expression: IFilteringExpression, dataType
...
 
     if (!expression.condition && expression.conditionName) {
         throw Error('Wrong `conditionName`, `condition` or `field` provided! It is possible that there is a type mismatch between the condition type and field type.');
     }
 
-    if (!expression.conditionName) {
-        expression.conditionName = expression.condition?.name;
-    }
-
+    expression.conditionName = expression.condition?.name;
```
damyanpetev, line 156: @hanastasov There's something not quite right with the fix;

Inverting the precedence of the props (`condition?.name` before `conditionName`) above or unconditionally overriding the `conditionName` here doesn't sound right.
`conditionName` is intended to be the _actual_ value, `conditon` as indicated by the optional check is.. _optional_ (actually intended to be dropped entirely at some point). Also this line with optional condition can also delete `conditionName` and likely break.

If there's a mismatch between the prop values - something else, elsewhere is wrong. If it's a restore from _cached_ state that was created before introducing the `conditionName` prop, it might be just a case of moving this check, though I can't be sure that's the issue without checking the repro

hanastasov: Thanks for noticing! Will be fixed

hanastasov: @damyanpetev  https://github.com/IgniteUI/igniteui-angular/pull/16922

PR #17030, opened 2026-03-11
fix(grid): Make sure border does not appear when new theme is loaded runtime.

MayaKirova:
Closes https://github.com/IgniteUI/igniteui-angular-samples/issues/3903  
- [x] Understand current theme.css approach
- [x] Create alternate theme SCSS using igniteui-angular dark material preset
- [x] Add alternate theme to angular.json as non-injected style bundle (`inject: false`)
- [x] Remove theme.css asset copy from angular.json assets
- [x] Update grid-cellMerging component to reference bundled stylesheet (`grid-cell-merging-alternate-theme.css`)
- [x] Implement OnDestroy cleanup to remove link element on navigation
- [x] Delete the large theme.css file (15179 lines)
- [x] Fix border reset in `%igx-grid__tr--merged` to use both `border-bottom: 0` and `border-block-end: 0` (removes `!important`)


ChronosSF:
@MayaKirova , is it needed to add the huge theme.css to the samples project? I think you should be able to add the default themes through angular.json to the build and switch them runtime this way?

MayaKirova:
> @MayaKirova , is it needed to add the huge theme.css to the samples project? I think you should be able to add the default themes through angular.json to the build and switch them runtime this way?

@ChronosSF 
I was trying to make it as similar as possible to the original issue, where new theme css is runtime injected via the theming widget over the existing theme. Mainly for testing. 

Not sure if the scenario where the default theme is switched is the same, but sure. @copilot add the default themes through angular.json to the build and switch them runtime this way in the grid-cellMerging sample.

MayaKirova:
@dkamburov resetting verified, since co-pilot changed the fix. Moving back to awaiting test.

dkamburov (approved):

PR #17059, opened 2026-03-18
feat(agents): Add backtracking and feedback loop to bug-fixing agent workflow

ivanvpetrov:
### Description
The bug-fixing orchestrator previously ran as a strict linear pipeline with no mechanism to recover from problems discovered mid-workflow. This PR adds an explicit backtracking loop so the process can self-correct when an agent surfaces an issue with earlier work.


kdinev:
@ivanvpetrov What is the status of this PR?

kdinev (approved):

PR #17183, opened 2026-04-15
Fix migrations & schematics commonjs type for latest package, adjust module output

ChronosSF:
Closes #  

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Fix: `ng update` migrations crash with `"type": "module"`

The latest [ng-packagr@21.2.2 sets `"type": "module"` in the published package.json](https://github.com/ng-packagr/ng-packagr/releases/tag/21.2.2), causing Node.js to treat all `.js` files as ES modules.
```text
[error] ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and 'igniteui-angular-samples\node_modules\igniteui-angular\package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
    at igniteui-angular-samples/node_modules/igniteui-angular/migrations/common/import-helper.js:2:23
    [...]
```
The migrations and schematics are CommonJS so their `exports = ...` syntax throws `exports is not defined in ES module scope`. The fix adds a `{ "type": "commonjs" }` package.json to both the `migrations/` and `schematics/` source directories, overriding the parent ESM setting for those subdirectories. These files are automatically copied to dist by the existing gulp tasks.
Additionally, the specific `import-helper.js` is a pre-compiled file to avoid a TypeScript emit converting dynamic imports to require, which is no longer needed with more modern module output modes so that is removed and updated.


ChronosSF:
@copilot apply changes based on the comments in [this thread](https://github.com/IgniteUI/igniteui-angular/pull/17183#pullrequestreview-4114398203)

projects/igniteui-angular/.npmignore
```diff
@@ -1,2 +1,4 @@
 # Nested package.json's are only needed for development.
-**/package.json
\ No newline at end of file
+**/package.json
+!migrations/package.json
+!schematics/package.json
```
damyanpetev, line 4: Fairly certain this entire file is largely irrelevant - it doesn't get copied to dist, thus should have no effect on the actual publish which is what you use `.npmingnore` for...

damyanpetev (dismissed):

projects/igniteui-angular/.npmignore
damyanpetev: Nvm, it does get copied - I guess ng-packagr does that too

damyanpetev (approved):

PR #17185, opened 2026-04-16
feat(skills): add grid sizing, updating theming, PR template, agents

### LINKED ISSUE #1608: React - Incorrect generation of CategoryChartType.Bar when using the MCP server
### Description

When generating React applications with charts via the `igniteui-cli` MCP server, the generated code sometimes uses an invalid chart type: `CategoryChartType.Bar` (or `"Bar"`), which is not part of the Ignite UI for React API.

---

### Error

Invalid CategoryChartType value: Bar

Image

---

### Actual Generated Code

```tsx
<IgrCategoryChart
  dataSource={data}
  chartType="Bar"
/>
```

---

### Expected

A valid `CategoryChartType` value (e.g. `Column`) instead of `Bar`.

---
### Reproduction

Two separate prompts with which I have produced this behavior are attached below:

[prompt1.txt](https://github.com/user-attachments/files/26467904/prompt1.txt), [prompt2.txt](https://github.com/user-attachments/files/26467903/prompt2.txt)

kdinev:
Fixes https://github.com/IgniteUI/igniteui-cli/issues/1608 for Angular

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


skills/igniteui-angular-grids/references/sizing.md
```diff
@@ -0,0 +1,218 @@
...
+}
+```
+
+```html
+<div class="grid-container">
+  <igx-grid height="100%" ...></igx-grid>
+</div>
+```
+
+## Column Sizing
```
georgianastasov: The section covers per-column width but misses the grid-level `columnWidth`. Suggest adding a note here something like:  

_Use the grid-level `[columnWidth]` input to set a global default width for all auto-sized columns at once:_
```html
<igx-grid [data]="data" [columnWidth]="'200px'" [autoGenerate]="true"></igx-grid>
```
_This is equivalent to setting the same `width` on every column that has no explicit width defined._

skills/igniteui-angular-components/references/charts.md
```diff
@@ -233,7 +233,7 @@ chartComponent.itemsSource = dataArray;
 ```typescript
 // Required
 dataSource: any[];           // Data array (auto-detects numeric fields)
-chartType: ChartType;        // Area, Bar, Column, Line, Waterfall, etc.
+chartType: ChartType;        // Auto, Area, Column, Line, Point, etc.
```
damyanpetev:
```suggestion
chartType: CategoryChartType;        // Auto, Area, Column, Line, Point, etc.
```
https://www.infragistics.com/products/ignite-ui-angular/api/docs/typescript/latest/classes/igniteui_angular_charts.IgxCategoryChartComponent.html#chartType

Also pretty sure the point of the Category Chart is that `Auto` is default, making `chartType` optional (i.e. **not** required). The doc example for the component sets just DS.
Edit: Yup https://codesandbox.io/p/devbox/optimistic-haze-v3hvdw?file=%2Fsrc%2Fapp.component.html%3A15%2C9-15%2C45

damyanpetev: PS: Also, the component does accept those options as string values (case-insensitive IIRC) in markup, which is where you'd normally configure that, so you don't need to reference the enum explicitly. Not sure what this "typescipt" snippet is supposed to be anyway

skills/igniteui-angular-grids/references/sizing.md
```diff
@@ -0,0 +1,218 @@
...
+- [Grid Height](#grid-height)
+- [Column Sizing](#column-sizing)
+- [Grid Cell Spacing Control](#grid-cell-spacing-control)
+- [Key Rules](#key-rules)
+
+## Overview
+
+The grid uses **border-box sizing** — border and padding are included in the width/height calculations. Do **not** override `box-sizing` on the grid element; doing so will break size calculations.
+
+Do **not** set `width` or `height` via CSS styles directly on the grid host element — always use the `width` and `height` **inputs** on the grid component.
```
damyanpetev: Any specific reason?
We absolutely use CSS sizing in various layout scenarios and/or rely on the default behavior on the element to fit inside the container/app layout, which is the same as setting CSS size on the element itself.

I realize there might be some issues with some pre-render virtualization scenarios, but thought we've resolved that and should have provision for such cases, no?

skills/igniteui-angular-grids/references/sizing.md
```diff
@@ -0,0 +1,218 @@
...
+
+> **Docs:** [Grid Sizing — Width](https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/sizing#width)
+
+**Default:** `100%` — the grid fills available width of the parent/window.
+
+### `null` Width
+
+```html
+<igx-grid [width]="null" ...></igx-grid>
+```
```
damyanpetev: Well that's awkward as.. Had to do a double take and check with the docs since it makes _no_ sense for unset width (null) in the DOM to ignore its layout context (say parent flex/grid) and I kind of don't think it could or should.
Honestly, this should be an addendum note for a edge use case where the goal is to disable virtualization, which is splits the grid out, not the first suggestion.

damyanpetev: Yeah, checks out really:
https://stackblitz.com/edit/rzvqtdcn?file=src%2Fapp%2Fgrid%2Fgrid-nested-data-binding-2%2Fgrid-nested-data-bind2.html

Just set width=null and display: flex on the container and you get a broken grid. Sizing is far from limited to width and I don't think we can easily ignore the rest.

skills/igniteui-angular-grids/references/sizing.md
kdinev: @copilot Provide a change based on this comment. The grid does support inheriting size from it's wrapping container, so this section is not correct, as Damyan points out.

skills/igniteui-angular-grids/references/sizing.md
kdinev: @copilot+claude-sonnet-4.6 remove the sections related to `null` Width, because as Damyan points out, they do not provide correct information and `null` width usage should not be encouraged as an option.

skills/igniteui-angular-components/references/charts.md
damyanpetev: @copilot Based on those comments - move the category chart `chartType` out of the required properties sections, clarify `Auto` is the default.
Also move the grid `null` width sub-section actually last in the sections.

damyanpetev (approved):

PR #17200, opened 2026-04-17
fix(select): support igx-select-item inside nested @for/@if control flow blocks

### LINKED ISSUE #17164: [igx-select] items inside if/else condition don't render
## Description  

When putting `<igx-select-item>` inside an @if / @else condition they don't render.

 * igniteui-angular version: 21.1
 * browser: any

## Steps to reproduce  

1. Use a condition like this one:
    ```html
    <igx-select #select>
      <label igxLabel>Even/Odd Select</label>
      @for (item of items; track item; let e = $even) {
        @if ($even) {
          <igx-select-item [value]="item"> even: {{ item }} </igx-select-item>
        } @else {
          <igx-select-item [value]="item"> odd: {{ item }} </igx-select-item>
        }
      }
    </igx-select>
    ```

## Result  
The select dropdown renders empty

## Expected result  
Items are still rendered.

I'm not sure if this is a limitation of the component or of Angular.
The limitation could be avoided by moving the conditions inside the element.

## Attachments  
Stackblitz: https://stackblitz.com/edit/qgbuhzqd?file=src%2Fapp%2Fselect-sample-1%2Fselect-sample-1.component.html

Zneeky:
Closes #17164 

### Issue: 
igx-select fails to render items when igx-select-item is nested inside @for + @if/@else.
The items are detected via ContentChildren, but not projected through <ng-content>, so the dropdown appears empty.

### Changes:
this.children already contains all discovered select items.
For each item, the code checks whether its DOM element is already inside the dropdown list.
If it is already there, nothing happens.
If it is missing, the code appends it to the scroll container so it becomes visible.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:


rkaraivanov (changes requested):
I don't think we should hack around limitations of the Angular framework.
I think there are already logged issue(s) for this in the Angular repository. We can document it as a 3rd party limitation.

PR #17205, opened 2026-04-20
feat(chips-area): add separate theme

### LINKED ISSUE #16769: [scoped-styles] - igx-chip-area styling missing until chip is present in grid drop area
### Description

The `igx-chip-area` component's styles are currently scoped only to the `igx-chip`, so styles for the area are only present when at least one `igx-chip` is on the page. In grids using a drop area (via `igx-chip-area`), the area should be visible even before a chip is created or dropped in.

- igniteui-angular version: scoped-styles branch
- browser: All

### Steps to reproduce

1. Open any grid sample that has column moving functionality enabled
2. Remove all groups if there are any.
3. Observe the chip-area styles.

### Result

The styles for `igx-chip-area` are not available until a chip is present. The drop area lacks styling when empty.
cousing it to shrink to the text size

### Expected result

The styles for the `igx-chip-area` should be present and visible, even if the area is empty (before a column/chip is dropped).

### Attachments

None yet.

**Branch:** `simeonoff/scoped-styles`  
**Component:** `igx-chip-area`

Image

adrianptrv:
Closes #16769


projects/igniteui-angular/chips/src/chips/chips-area/themes/_base.scss
```diff
@@ -0,0 +1,15 @@
...
+
+@include layer(base) {
+    @include b(igx-chip-area) {
+        display: flex;
+        align-items: center;
+        justify-content: flex-start;
+        flex-wrap: wrap;
+        width: 100%;
+
+        &:empty {
```
desig9stein, line 11: Not sure why we need `empty`; the area is never empty, even when no chips are present. Can you test to see what happens if you remove it?

adrianptrv: In the **Grid** context, the empty state is never reached because the **drop-area** element remains inside the **chips-area** even when no chips are added. However, when the **chips-area** is used independently outside of the grid, the `empty` state is applied whenever no chips are present. If we remove the '**empty**' selector; on the `empty` state, the **chips-area** element will still be displayed, but with 0px height:

Screenshot 2026-04-20 at 18 10 13
<br>
Screenshot 2026-04-20 at 18 08 32

desig9stein (approved):

PR #17281, opened 2026-05-18
feat: Added virtual scroll component and sample implementation

### LINKED ISSUE #17607: New Virtual Scroll component
### Component / Area

New Component

### Is your feature request related to a problem?

`igxForOf` requires developers to wire a separate scroll container, helper elements, and container sizing manually.

### Describe the solution you'd like

Expose a new Virtualization component for horizontal and vertical scrolling and deprecate `igxForOf`.

The new component should provide:

* A self-contained viewport
* Signals-first reactive model
* View pooling
* `ChangeDetectionStrategy.OnPush`
* Better DX for remote/infinite scrolling
* No auxiliary helper components in the DOM
* Typed template context

### Proposed API or Usage

```typescript

```

### Describe alternatives you've considered

_No response_

### How important is this feature to you?

Nice to have

### Additional context

_No response_

### LINKED ISSUE #17608: Migrate internal igxForOf usages to the new Virtual Scroll component
Migrate the internal `igxForOf` usages to the new `IgxVirtualScrollComponent`.

### Scope

* Combo
* Simple Combo
* Excel-style filtering search
* Keep Drop Down compatible with both projected `IgxVirtualScrollComponent` and `igxForOf`

The migration should preserve the existing behavior and public APIs, including selection, filtering, grouping, custom templates, keyboard navigation/accessibility, popup sizing, and remote data binding.

### LINKED ISSUE #17610: Replace igxForOf for igxCombo

### LINKED ISSUE #17611: Replace igxForOf for Excel Style Filtering

### LINKED ISSUE #17615: Replace igxForOf for IgxSimpleCombo
### Component / Area

Combo / Simple Combo

### Is your feature request related to a problem?

_No response_

### Describe the solution you'd like

n/a

### Proposed API or Usage

```typescript

```

### Describe alternatives you've considered

_No response_

### How important is this feature to you?

Nice to have

### Additional context

_No response_

### LINKED ISSUE #17616: Test Drop Down for compatibility with  IgxVirtualizationComponent

rkaraivanov:
Closes #17607
Closes https://github.com/IgniteUI/igniteui-angular/issues/17608
Closes https://github.com/IgniteUI/igniteui-angular/issues/17610
Closes https://github.com/IgniteUI/igniteui-angular/issues/17611
Closes https://github.com/IgniteUI/igniteui-angular/issues/17615
Closes https://github.com/IgniteUI/igniteui-angular/issues/17616

## Description
Expose a new virtual scroll component for horizontal/vertical scrolling.

## Motivation / Context
1. Self-contained viewport — The component is the scrollable element. `igxForOf` requires the developer to manually wire a separate scroll container, a `VirtualHelperComponent`/`HVirtualHelperComponent` DOM helper element, and pass `igxForContainerSize` explicitly. That's 3–4 pieces of setup vs. one tag.
2. Signals-first reactive model.
3. View pooling — Active views are kept in a pool and detached/reattached instead of destroyed/recreated on every scroll. `igxForOf` uses `ViewContainerRef.move()` but always keeps a fixed chunk alive, paying create/destroy cost when chunk size changes.
4. Clean `ChangeDetectionStrategy.OnPush` throughout the implementation.
5. Better DX for remote/infinite scrolling.
6. No auxiliary helper components in the DOM.
7. Typed template context.

 - [ ] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

## How Has This Been Tested?

 - [x] Unit tests
 - [x] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **Angular version**: 
 - **Browser(s)**: 
 - **OS**: 

## Checklist:


rkaraivanov:
@simeonoff 
Can someone from your team take a look at styling of the component? I'm asking mainly to check whether it correctly works with the new scoped styles and theming.

simeonoff:
> @simeonoff Can someone from your team take a look at styling of the component? I'm asking mainly to check whether it correctly works with the new scoped styles and theming.

Restructured the CSS and sample. Should be more in line with the other component styles now and the approach we wanna use in the future for authoring component styles.

projects/igniteui-angular/drop-down/src/drop-down/drop-down.component.ts
```diff
@@ -472,16 +494,91 @@ export class IgxDropDownComponent extends IgxDropDownBaseDirective implements ID
...
+            .subscribe(() => this.connectVirtualization());
+
+        // A sliding window reuses its items, so only the adapter reports it; a page
+        // arriving or the list emptying builds or drops them, which reaches this query.
+        // The query settles inside the pass that rendered them, and the element it names
+        // was already checked in it, so the write waits for the render to finish.
+        this.children.changes
+            .pipe(takeUntil(this.destroy$))
+            .subscribe(() => afterNextRender(
+                () => this.refreshActiveDescendant(),
```
ChronosSF: Can we do the same ESF does - it only marksForCheck if the id actually changed.

projects/igniteui-angular/grids/core/src/filtering/excel-style/excel-style-search.component.ts
```diff
@@ -260,17 +266,8 @@ export class IgxExcelStyleSearchComponent implements AfterViewInit, OnDestroy {
...
-            const itemSize = firstItem?.element.getBoundingClientRect().height;
-            if (itemSize) {
-                // Excel filter rows are uniform; use the outer size to keep the scrollbar range stable.
-                this._measuredItemSize = itemSize;
-            }
-            this.virtDir.igxForContainerSize = this.containerSize;
-            this.virtDir.igxForItemSize = this.itemSize;
-            this.virtDir.recalcUpdateSizes();
+        // Only flushes the bindings the surrounding menu changed; the list measures itself.
+        if (this.virtualScroll && !(this.cdr as any).destroyed) {
```
ChronosSF: can you use the component's own destroy flag?

CHANGELOG.md
```diff
@@ -13,8 +13,18 @@ All notable changes for each version of this project will be documented in this
 
 ## 22.2.0
 
+### General
+
+- The Excel style filtering search list, `IgxComboComponent` and `IgxSimpleComboComponent` are now virtualized by `IgxVirtualScrollComponent` instead of the `igxFor` directive. A row is measured in the DOM once it renders and the measured size replaces the estimate it started from; rows that have not rendered keep that estimate.
+    - The list markup changed accordingly: `igx-display-container` and the `igx-vhelper--vertical` scrollbar are replaced by the `igx-virtual-scroll` host and its `igx-vs__item` row wrappers. Applications and tests that reach into those elements directly need updating.
```
ChronosSF: What are these wrappers? Shouldn't this be just `igx-virtual-item` as children of igx-virtual-scroll

projects/igniteui-angular/combo/src/combo/combo.common.ts
```diff
@@ -769,11 +774,8 @@ export abstract class IgxComboBaseDirective implements IgxComboBase, AfterViewCh
     public searchInput: ElementRef<HTMLInputElement> = null!;
 
     /** @hidden @internal */
-    @ViewChild(IgxForOfDirective, { static: true })
-    public virtualScrollContainer!: IgxForOfDirective<any>;
-
-    @ViewChild(IgxForOfDirective, { read: IgxForOfDirective, static: true })
-    protected virtDir!: IgxForOfDirective<any>;
+    @ViewChild('virtualScroll', { static: true })
+    public virtualScrollContainer!: IgxVirtualScrollComponent<any>;
```
ChronosSF, line 778: the type change is a breaking one and should be documented as such and also requires a migration. Same for ESF

projects/igniteui-angular/virtual-scroll/src/virtual-scroll/virtual-scroll.component.ts
```diff
@@ -0,0 +1,1053 @@
...
+
+function onAbort(abort: AbortSignal, cancel: () => void): void {
+  abort.addEventListener("abort", cancel, { once: true });
+}
+
+/**
+ * A virtual scroll component for large lists. Only the items visible in the
+ * viewport (plus a configurable over-scan) are rendered.
+ *
+ * @igxModule IgxVirtualScrollModule
```
ChronosSF: Don't think we have this module right? This component is strictly stand-alone

ChronosSF (changes requested):

ChronosSF:
Some findings during agentic testing:

1. nearest alignment start-aligns out-of-view items. scrollToIndex(i, { block: 'nearest' }) returns the current offset when the item is already visible, but otherwise falls through to start. Native scrollIntoView semantics, which the docs claim to match, end-align an item that is below or past the viewport. Drop-down, combo and the Excel-style filter all navigate with nearest, so ArrowDown past the bottom edge now jumps the row to the top of the list instead of revealing it at the bottom the way igxFor did. Tests: should end-align an item below the viewport for block: nearest and the horizontal inline: nearest mirror both fail with Expected 500 to be 250. The start case for an item above the viewport passes and is included for completeness.

2. Direction is read non-reactively inside _contentTransform. The computed calls _isLTR(), which reads computed style, so it only re-evaluates when the rendered range or the engine version changes. Flipping dir at runtime on a horizontal list while the window stays put leaves the wrapper at translateX(+x) instead of translateX(-x). Tests: should flip the content transform when the direction changes to rtl at runtime and its ltr mirror fail with Expected 'translateX(400px)' to be 'translateX(-400px)' and the reverse. The tests deliberately hold the rendered range constant across the flip, since that is the only state in which the bug is observable. A fix needs the direction itself to be a tracked signal, not a DOM read inside the computed.

Both have tests added that will fail in the latest commit. We need to fix these before merge.

ChronosSF:
Added 3 more tests that fail due to actual issues:

1. Compressed lists cannot reach their last rows
When the total size exceeds the browser's scroll limit, the engine maps DOM scroll to virtual space by scaling the total: virtual start = DOM scroll × totalSize ÷ domSize. At the largest scrollable offset, domSize − viewport, that lands at totalSize − viewport × ratio, and the window only spans one viewport from there. The last viewport × (ratio − 1) virtual pixels are never in the visible range, so with a million 50 px rows Firefox silently drops the final 8 to 33 rows depending on viewport height and Chrome the final 3 to 7. The content-transform clamp does not help, since it only positions rows that were selected for rendering.
**Suggested fix**: compress the scrollable range, not the total. Virtual start = DOM scroll × (totalSize − viewport) ÷ (domSize − viewport), so the maximum DOM offset maps exactly to totalSize − viewport. getVisibleRange, getScrollOffsetForIndex, getAlignedScrollOffset and isIndexInView must all share that mapping, which means the offset conversions need the viewport size, not only the range query. With that mapping all three assertions in the test hold as written.

2. Replacing data with same-size rows leaves the engine on the estimate
Assigning a new array changes every item's identity, so the retain count is zero and every size is rebuilt from `estimatedItemSize`. The recycled wrappers keep their data-index, so `_scheduleItemMeasurement` skips re-observing them, and ResizeObserver stays silent because the element's size did not change. Rendered rows therefore keep the estimate indefinitely, and the track size, scrollToIndex targets and stateChange.totalSize are off by the difference. 
**Fix direction**: whenever a resize discards measurements, re-measure the wrappers currently in the DOM, either by re-registering them with the observer or by reading their border box in the after-render pass.

3. Remote first page lands focus on a group header
When the loaded page starts past index 0, navigateFirst calls navigateItem(0). The scroll resolves before the page for index 0 exists, so skipHeader finds no rendered item and returns. When the page then arrives, index 0 renders as the group header, `aria-activedescendant` names it, and `focusedItem` reports a header. The igxFor path always landed on the first selectable item. The test asserts the loaded page starts past 0 first, so it cannot pass through the local path. 
**Fix direction:** after a page arrives, re-run the header skip for the pending focused index, or defer the skip until the target index is rendered rather than until the scroll settles.

ChronosSF (dismissed):

ChronosSF (approved):

PR #17302, opened 2026-05-29
fix(hierarchical-grid): hide child row editing overlay on scroll - master

### LINKED ISSUE #17275: Row edit overlay flies outside the grid when scrolling
### Description

Go to a hierarchical grid with row editing on child
Start edit in the child
Start scrolling
The row edit overlay goes outside the grid when scrolling

### Framework

Angular

### Angular Version

latest

latest

### Component / Area

Hierarchical Grid

### Browser

Chrome

### Operating System

_No response_

### Command Not Working

```shell

```

### Steps to Reproduce

1. Go to a hierarchical grid with row editing on child
2. Start edit in the child
3. Start scrolling

### Actual Result

The row edit overlay goes outside the grid when scrolling:

### Expected Result

The row edit overlay should be hidden with the current child grid when scrolling down/up.

### Reproduction URL

_No response_

### Attachments
https://github.com/user-attachments/assets/6ee79b6d-e8ad-4977-82c3-dbf779bb6f34

_No response_

georgianastasov:
Closes #17275   

## Description
Fixes an issue where a row editing overlay could remain visible outside the grid boundaries while scrolling, especially in nested Hierarchical Grid row islands.

The fix clips the row editing overlay to the effective visible grid area after it is positioned. Since the overlay is rendered in the top layer, it no longer relies on the grid body overflow to hide it when the edited row scrolls out of view.

The clipping area is calculated from the row edit container and its clipping ancestors, so nested grids respect their parent grid scroll bounds.

## Motivation / Context
When row editing is started inside a child grid and the user scrolls a parent or nested grid, the editing overlay could visually escape the grid body and appear detached from the edited row.

This happened because the overlay is positioned outside the normal grid body clipping context. The fix keeps the overlay visually constrained to the same visible area as the edited row.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [x] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

- Hierarchical Grid
- Row Editing
- Overlay positioning / clipping on scroll

## How Has This Been Tested?
Added unit tests for Hierarchical Grid child row editing overlay behavior during scroll.

 - [x] Unit tests
 - [x] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **Angular version**: 22.0.x
 - **Browser(s)**: all
 - **OS**: all

## Checklist:


RivaIvanova (commented):
The issue is not fixed.

1. Go to http://localhost:4200/hierarchicalGridAddRow
2. Expand the first row `Product: A0` and scroll until you see the second nested island.
```
root hGrid
   -> rowIsland lvl 1 (first nested island)
   -> rowIsland lvl 1 (second nested island)
```
3. In the first nested island, scroll and expand any of the middle records, for example, `Product: A14`.
4. Enter edit mode for any row in the inner island (lvl 2), for example, `Product: A1`.
```
root hGrid
   -> rowIsland lvl 1
      -> rowIsland lvl 2 (inner island)
   -> rowIsland lvl 1
```
5. Scroll up and down on the first nested island lvl 1.

hGridScrollOverlay

projects/igniteui-angular/grids/core/src/grid.common.ts
```diff
@@ -63,6 +63,72 @@ export class RowEditPositionStrategy extends ConnectedPositioningStrategy {
...
+    private getClippingRect(element: HTMLElement): Pick<DOMRect, 'top' | 'right' | 'bottom' | 'left'> {
+        const document = element.ownerDocument;
+        const rect = element.getBoundingClientRect();
+        // Start with the current grid body, then narrow it by every clipping parent.
+        const clippingRect = { top: rect.top, right: rect.right, bottom: rect.bottom, left: rect.left };
+
+        let parent = element.parentElement;
+
+        // Intersect with clipping ancestors so nested grids respect their parent grid scroll bounds.
+        while (parent && parent !== document.body && parent !== document.documentElement) {
```
RivaIvanova: If the clipping rect has already been reduced to zero area, I don't see any benefit in continuing the ancestor walk. I think there should be an early exit like:
```ts
if (clippingRect.top >= clippingRect.bottom || clippingRect.left >= clippingRect.right) {
    break;
}
```

And about the ancestor walk, do we need to go through every parent all the way to `app-root`?

Image

RivaIvanova (commented):
Is this behavior reproducible for the grid and the tree grid, too? In what scenarios? I tested them before the fix and didn't notice. If it is reproducible, then there should be tests for them too. If not, adding the clip logic that is relevant only for the hierarchical grid for all grids does not seem optimal. Maybe extend the strategy and plug the `updateContentClip` logic wherever needed?

georgianastasov:
> Is this behavior reproducible for the grid and the tree grid, too? In what scenarios? I tested them before the fix and didn't notice. If it is reproducible, then there should be tests for them too. If not, adding the clip logic that is relevant only for the hierarchical grid for all grids does not seem optimal. Maybe extend the strategy and plug the `updateContentClip` logic wherever needed?

Good point. I checked flat Grid and Tree Grid and I don’t see the same issue there. In Tree Grid there is a small delayed hide while scrolling near the bottom edge, but it is not the same visual escape that happens in nested Hierarchical Grid row islands. 

The problematic case is specific to child Hierarchical Grids, where the row edit overlay is rendered in the top layer while the child grid itself is clipped by parent grid bodies. So I agree that applying this clipping logic unconditionally through the shared row edit strategy is broader than needed. I’ll scope it to hierarchical child grids only.

projects/igniteui-angular/grids/core/src/grid.common.ts
georgianastasov: Yes, agreed. There is no benefit in continuing once the clipping rect has collapsed, so I will add the early exit.

georgianastasov: For nested Hierarchical Grids the effective visible area is the intersection of the child grid body and its parent grid bodies. Since the row edit overlay is rendered in the top layer, it does not naturally inherit that clipping, while the row island content does. That is the part need to restore narrowly, without walking every parent up to `app-root`.

RivaIvanova (approved):

PR #17325, opened 2026-06-12
fix(for-of): call recalcUpdateSizes via afterNextRender in zoneless mode

viktorkombov:
Closes #17280

## Description
In zoneless Angular (`provideZonelessChangeDetection`), `NoopNgZone.onStable` never emits, so `recalcUpdateSizes()` was never called after scrolling. This left the virtual scroll size cache uncorrected, causing the last visible row in MRL grids to appear partially clipped.

## Motivation / Context
In `for_of.directive.ts`, at all three scroll handler sites, detect zoneless mode and use `afterNextRender({ read: () => recalcUpdateSizes() })` instead of `zone.onStable.pipe(first()).subscribe(recalcUpdateSizes)`. Zone mode keeps the original path unchanged (replacing it unconditionally breaks an existing zone-based scroll test).
In `grid-base.directive.ts`, the resizeNotify handler skips zone.run() in zoneless mode - `notifyChanges()` already calls `markForCheck()` which schedules CD via the zoneless scheduler.

### Type of Change (check all that apply):
 - [x] Bug fix
 - [ ] New functionality
 - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
 - [ ] Refactoring (no functional changes)
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

### Component(s) / Area(s) Affected:
Grid, For-of

## How Has This Been Tested?
Added a zoneless describe block in `for_of.directive.spec.ts` verifying `recalcUpdateSizes` is called after scroll when using `provideZonelessChangeDetection`. 
 - [x] Unit tests
 - [ ] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **Angular version**: 
 - **Browser(s)**: 
 - **OS**: 

## Checklist:
 - [x] All relevant tags have been applied to this PR
 - [x] This PR includes unit tests covering all the new code ([test guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Test-implementation-guidelines-for-Ignite-UI-for-Angular))
 - [ ] This PR includes API docs for newly added methods/properties ([api docs guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines))
 - [ ] This PR includes `feature/README.MD` updates for the feature docs
 - [ ] This PR includes general feature table updates in the root `README.MD`
[...]

PR #17328, opened 2026-06-15
fix(grid): fix zone.onStable patterns broken in zoneless change detection

### LINKED ISSUE #17280: igxGrid - The last row is not fully displayed.
### Description

igxGrid - The last row is not fully displayed.

When a grid is scrolled down to the bottom, the last row is not fully displayed. This behavior has come to be seen after upgrading from 20.1.20. It seems that igx-display-container's top value has an incorrect value.

### Framework

Angular

### Angular Version

21

21.0.8

### Component / Area

Grid

### Browser

Chrome

### Operating System

Windows

### Command Not Working

```shell

```

### Steps to Reproduce

1. npm install and npm start the attached application.
2. Scroll down the grid to the bottom.

### Actual Result

The last row is not fully displayed.

### Expected Result

The last row is fully displayed.

### Reproduction URL

_No response_

### Attachments

Image

[c00246170-app1.zip](https://github.com/user-attachments/files/27945027/c00246170-app1.zip)

### LINKED ISSUE #17318: Zoneless: Grid throws ExpressionChangedAfterItHasBeenCheckedError for aria-colcount
### Description

When the application is running in a zoneless change detection environment, the FinJS sample from the Samples Browser throws an `ExpressionChangedAfterItHasBeenCheckedError` from the Grid.

The error is related to the `aria-colcount` attribute. The Grid initially renders one value for `attr.aria-colcount`, but the value changes after Angular has already checked the expression:

```text
ERROR RuntimeError: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value for 'attr.aria-colcount': '51'. Current value: '48'.
```

### Framework

Angular

### Angular Version

21.2.8

21.2.0

### Component / Area

Grid

### Browser

Chrome

### Operating System

Windows

### Command Not Working

```shell

```

### Steps to Reproduce

1. Open the following sample: https://stackblitz.com/edit/ybtb6xmn-bghatscf?file=package.json
2. Observe the browser console.

### Actual Result

The Grid throws an `ExpressionChangedAfterItHasBeenCheckedError` related to the aria-colcount attribute:

`Previous value for 'attr.aria-colcount': '51'. Current value: '48'.`

### Expected Result

The Grid should render without throwing an `ExpressionChangedAfterItHasBeenCheckedError` in a zoneless change detection environment.

### Reproduction URL

https://stackblitz.com/edit/ybtb6xmn-bghatscf?file=package.json

### Attachments

Image

### LINKED ISSUE #17323: Zoneless: Grid and Tree Grid auto-size columns are misaligned when available width is limited
### Description

When the application is running in a zoneless change detection environment, the auto-size column feature for the Grid and Tree Grid components does not work correctly when the screen is not wide enough to fit all columns.

In this case, the column headers and the columns inside the grid body are rendered with different widths, which makes the Grid / Tree Grid layout appear broken.

This appears to be related to zoneless change detection, as the column widths are not synchronized correctly between the header and body sections when auto-sizing is applied and the available horizontal space is limited.

I have created a sample that reproduces the behavior:
https://stackblitz.com/edit/y6yhtvg7-yfuwtwzm?file=src%2Fmain.ts

### Framework

Angular

### Angular Version

21.2

21.2.0

### Component / Area

Tree Grid

### Browser

Chrome

### Operating System

_No response_

### Command Not Working

```shell

```

### Steps to Reproduce

1. Open the sample: https://stackblitz.com/edit/y6yhtvg7-yfuwtwzm?file=src%2Fmain.ts
2. Observe the alignment between the column headers and the grid body columns.

### Actual Result

The column headers and the body columns are rendered with different widths. The Grid / Tree Grid appears misaligned and visually broken.

### Expected Result

The auto-size column feature should calculate and apply consistent column widths for both the header and body sections, even when the available width is not enough to fit all columns.

### Reproduction URL

https://stackblitz.com/edit/y6yhtvg7-yfuwtwzm?file=src%2Fmain.ts

### Attachments

Image

### LINKED ISSUE #17364: Improve Grid Stability With Angular Zoneless Change Detection
## Component / Area

Grid

## Is your feature request related to a problem?

Angular v21+ uses zoneless change detection by default. Angular v20
applications can also opt in explicitly, and tests can reproduce this runtime
with `provideZonelessChangeDetection()`.

Some Grid paths relied on `NgZone.onStable` as a scheduling point after layout,
scroll, virtualization, or auto-sizing work. Angular's zoneless guide states
that `NgZone.onStable` does not emit in zoneless applications, so these paths
need a real update, render, or DOM boundary instead of depending on zone
lifecycle notifications.

The goal is to keep the existing zone-based behavior stable while making the
confirmed Grid scenarios reliable in zoneless applications.

Resolve the confirmed Grid zoneless issues in this branch with targeted
production changes and regression coverage.

- Use `provideZonelessChangeDetection()` in consumer-style tests.
- Avoid post-action `fixture.detectChanges()` in tests that verify zoneless
  runtime behavior.
- Assert public behavior such as rendered rows, emitted events, measured
  auto-size values, or public Grid state.
- Keep zone-based applications on the existing scheduling path where that path
  already worked.
- Use direct change detection, render callbacks, browser callbacks, or public
  Grid notifications only when they match the work being performed.

## Proposed API or Usage

No public API change is expected.

The branch only changes internal scheduling for confirmed Grid paths that were
blocked by `NgZone.onStable` in zoneless mode.

Example test setup:

```ts
TestBed.configureTestingModule({
    imports: [TestHostComponent],
    providers: [provideZonelessChangeDetection()]
});
```

- Replacing every `NgZone.onStable` usage globally. This is too broad and risks
  changing behavior that is still valid in zone-based applications.
- Triggering change detection everywhere after async work. This can hide
  missing notifications and does not follow Angular's zoneless guidance.
- Keeping all zoneless tests in one bucket. A focused flat-grid zoneless spec is
  useful for cross-cutting runtime coverage, while Pivot Grid behavior belongs
  in the existing Pivot Grid spec because it is component-specific.

## How important is this feature to you?

Important. Zoneless change detection is the default in Angular v21+, so
confirmed Grid scenarios should not depend on zone lifecycle scheduling.

## Additional context

This task is intentionally limited to the issues resolved by the current
branch. Broader Tree Grid, Hierarchical Grid, directive-level, or editing-flow
investigations should be logged separately if they need follow-up work.

Angular zoneless guide:
https://angular.dev/guide/zoneless

### Resolved in this branch

- [x] Flat Grid async sorting renders after `fixture.whenStable()` without an
  additional `fixture.detectChanges()`.
- [x] Flat Grid async filtering and clearing filters render after
  `fixture.whenStable()` without an additional `fixture.detectChanges()`.
- [x] Flat Grid `filteringDone` emits for `filter()` and `clearFilter()` in a
  zoneless test environment.
- [x] Flat Grid horizontal column virtualization emits `parentVirtDir.chunkLoad`
  and observes shifted virtualized column state in zoneless mode.
- [x] Flat Grid fit-content column auto-sizing re-measures after explicit
  recalculation and header data changes.
- [x] Flat Grid recycled virtual row state restore no longer depends on
  `NgZone.onStable` in zoneless mode.
- [x] Pivot Grid row dimension auto-sizing is covered in its existing spec and
  does not require manual change detection after grid sizing.
- [x] Existing zone-based scheduling behavior is preserved for the changed
  paths.

### Test organization

- Flat Grid zoneless regressions are covered in
  `projects/igniteui-angular/grids/grid/src/grid.zoneless.spec.ts` because they
  validate cross-cutting runtime behavior in an explicitly zoneless TestBed.
- Pivot Grid dimension auto-sizing is covered in
  `projects/igniteui-angular/grids/pivot-grid/src/pivot-grid.spec.ts` because
  it is component-specific behavior.

### Suggested validation

- `npm run test:lib:grid`
- `npm run test:lib:pgrid`
- `npm run lint:lib`

### LINKED ISSUE #17385: Tree Grid keyboard navigation does not activate cells outside the virtualized viewport
### Description

When keyboard navigation in the Tree Grid targets a cell outside the currently rendered virtualized viewport, the target cell may not be activated/selected correctly after the grid scrolls.

This can happen with navigation actions such as `Arrow Up` / `Arrow Down`, `Ctrl + Home` / `Ctrl + End`, or `Tab` / `Shift + Tab` when the target row or column is not currently rendered because of virtualization.

In this case, the Tree Grid may scroll, but the active/selected cell remains on the previous cell or moves to an incorrect cell.

The issue is easier to expose in zoneless change detection tests, but the underlying behavior appears to be related to virtualized scroll/navigation timing and is not limited to zoneless applications.

### Framework

Angular

### Angular Version

22.0

22.0.x

### Component / Area

Tree Grid

### Browser

Chrome

### Operating System

Windows 11

### Steps to Reproduce

1. Create a Tree Grid with fixed `height` and enough rows/columns to enable virtualization.
2. Focus or select a cell near the currently visible viewport.
3. Use keyboard navigation to move to a cell outside the current viewport, for example:
   - `ArrowDown` / `ArrowUp` when the target row requires vertical scrolling
   - `Ctrl + End` or `Ctrl + Home`
   - `Tab` or `Shift + Tab` when the next editable cell is outside the visible area
4. Observe the active/selected cell after scrolling completes.

### Actual Result

The Tree Grid scrolls, but the target cell outside the original viewport is not activated/selected correctly.

The active/selected state may remain on the previous cell or move to an incorrect cell.

### Expected Result

Keyboard navigation should scroll the Tree Grid as needed and then activate/select the intended target cell once it is rendered, including when the target cell is outside the current virtualized viewport.

### LINKED ISSUE #17388: [Zoneless]: IgxList with igx-for-of virtualization has wrong placement of the virtualization container
### Description

Having `igx-list` with virtualization has wrong position of the container of the virtualized items

### Framework

Angular

### Angular Version

latest

latest

### Component / Area

List

### Browser

Chrome

### Operating System

_No response_

### Command Not Working

```shell

```

### Steps to Reproduce

1. Open https://stackblitz.com/edit/tyacnprm-3esag12f?file=src%2Fapp%2Fsplitter-horizontal-sample%2Fsplitter-horizontal-sample.component.html
2. Scroll to bottom
3.

### Actual Result

Observe the bottom gap of the list:

Image

### Expected Result

No gap should exist

### Reproduction URL

https://stackblitz.com/edit/tyacnprm-3esag12f?file=src%2Fapp%2Fsplitter-horizontal-sample%2Fsplitter-horizontal-sample.component.html

### Attachments

_No response_

viktorkombov:
Closes #17323
Closes #17364
Closes #17385
Closes #17318
Closes #17280
Closes #17388 

## What

Grid virtualization, autosizing, and keyboard navigation relied on `NgZone.onStable`, which never emits in zoneless apps. As a result, scroll-driven `chunkLoad`, column autosize, filter-row rendering, and post-scroll cell activation silently never ran.

This migrates all of that off `onStable` and onto Angular’s render hooks.

### Core

* Added a new central helper: `runAfterRenderOnce(injector, cb, phase?)` in `core/utils.ts`.

  * This is now the single scheduling point for all deferred-render work.
  * Previously, this logic was scattered across `onStable` / `ɵNoopNgZone` branches.
* Updated every affected call site to go through `runAfterRenderOnce`.
* Removed the private `ɵNoopNgZone`-based zoneless detection.
* Removed the duplicated per-class `runAfterRender` helpers.

### Grids

* `IgxFilteringService.isFilterRowVisible` is now signal-backed, so the `OnPush` header row re-renders without requiring a zone tick.
* Filter cells now observe their own size via `ResizeObserver` to recompute visible chips after resize.
* Column resize now explicitly notifies the grid via `notifyResized`, instead of relying on an empty `zone.run`.
* Keyboard navigation now notifies after any scrolling navigation key, not just `Ctrl`.
* Hierarchical navigation now subscribes to `chunkLoad` before scrolling, avoiding missed synchronous emissions.
* Deferred work in `for_of`, `grid-base`, and `pivot-grid` was migrated to `runAfterRenderOnce`.

### Tests

* Added zoneless duplicates using `provideZonelessChangeDetection` for the affected grid, tree, hierarchical, and pivot suites.
* Centralized async test helpers in `helper-utils.spec.ts`.
* `runAfterRenderOnce` consumers now settle through shared helpers:

  * `waitForGridSettle`
  * `setGridVerticalScrollTop`
  * `dispatchGridScrollEvents`
  * `navigateWithGridScroll`
* Updated the zoned originals to await the new render-boundary timing.

  * Previously, they happened to align with `onStable`.

## Notes

* Behavior for zoned apps is unchanged in practice.

  * Real interaction still triggers a render.
  * Only the timing signal moved from `onStable` to `afterNextRender`.
* No `runOutsideAngular` was removed.
* No forced `detectChanges` was added in zoneless tests.

 - [x] Bug fix
 - [x] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

## How Has This Been Tested?

 - [x] Unit tests
 - [x] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **Angular version**: 
 - **Browser(s)**: 
 - **OS**: 

## Checklist:


mddragnev (commented):
Everywhere you remove the `onStable` subscription you're invoking the callback immediately in the zoneless case. I'm not sure this is equivalent behavior.

`onStable` is typically used because we need to wait until Angular has finished processing and rendering before performing some work. Simply checking whether change detection is zoneless and executing the callback immediately may change the timing and could introduce regressions.

I think we should consider using the new render hooks instead, depending on the intended behavior:

- `afterNextRender()` if the callback should run once after the next render.
- `afterEveryRender()` if it needs to run after every render.

These APIs are designed to work in both zoned and zoneless applications and more closely match the intent of waiting for rendering to complete.

For example, instead of:
```
this.zone.onStable.pipe(first()).subscribe(() => {
    this.cdr.detectChanges();
    this.autoSizeColumnsInView();
});
```
we could do:
```
runInInjectionContext(this.injector, () => {
    afterNextRender(() => {
        this.autoSizeColumnsInView();
// note that we don't need the detectChanges here because the hook is called after the change detection cycle
    });
});
```
the render hooks need to be called within an injection context or we could provide the injector as an option:
```
afterNextRender(() => {
    this.autoSizeColumnsInView();
}, { injector: this.injector });
```
@rkaraivanov What do you think about using the render hooks here instead? I think they'd better preserve the original intent while also supporting zoneless mode.

viktorkombov:
> Everywhere you remove the `onStable` subscription you're invoking the callback immediately in the zoneless case. I'm not sure this is equivalent behavior.

@mddragnev You're right, thanks. I used the same `afterNextRender` approach in #17325, following [Angular's guidance](https://angular.dev/guide/zoneless#remove-ngzoneonmicrotaskempty-ngzoneonunstable-ngzoneisstable-or-ngzoneonstable). The immediate callback here was chosen on the assumption that the work runs directly after `detectChanges()`.

On reviewing the logic again, that is not a safe equivalence. A synchronous `detectChanges()` checks the relevant view and its children, but it does not provide the same boundary as waiting until the next application render has completed. For the DOM and layout work in these paths, `afterNextRender` is therefore the more appropriate zoneless replacement.

The zoneless path now uses `afterNextRender`, while the zone-based path keeps the existing onStable behavior. This avoids changing the timing for existing zoned consumers as part of this PR, since onStable and afterNextRender do not have identical semantics.

viktorkombov:
@mddragnev, @rkaraivanov 

Blocker: there are still grid regressions, but they are not identical between zone-based and zoneless apps.

The main issue reproduces in the dev sample `gridFinJS` in a zone-based app. With many columns and rows, the initial horizontal scroll sizing is not calculated correctly. After horizontal scrolling, the scroll container can leave a large empty space after the last column, which means the virtualized horizontal size/offset state is stale.

Keyboard navigation is affected there as well: `Ctrl + Home`, `Ctrl + End`, `Ctrl + Arrow*`, etc. do not behave reliably once the scroll state gets out of sync.

In zoneless apps I do not see the exact same horizontal empty-space issue, but keyboard navigation is still not fully stable. There is also an initial flicker with shortcuts such as `Ctrl + End`.

This should block the PR until the zone-based horizontal sizing regression is fixed and the zoneless keyboard navigation behavior is verified/fixed separately.

Related to #17323 

zonebased

mddragnev:
@viktorkombov This issue has nothing to do with the zoneless effort. I am going to log it separate and fix it on another PR. #17392

projects/igniteui-angular/grids/core/src/filtering/base/grid-filtering-cell.component.ts
```diff
@@ -94,6 +98,14 @@ export class IgxGridFilteringCellComponent implements AfterViewInit, OnInit, DoC
 
     public ngAfterViewInit(): void {
         this.updateFilterCellArea();
+        // The visible chips calculation measures the rendered cell, so it must rerun when
+        // the cell's size actually changes (column/grid resize). ZoneJS apps got this for
+        // free from zone-triggered ticks; observing the element covers zoneless apps too.
+        this.zone.runOutsideAngular(() => {
```
mddragnev: I cant seem to reproduce the issue this code fixes? It seems to me that without the resize observer the chips work just fine. Could you please give me some steps to reproduce it?

mddragnev: Edit: I think that this fixes only a zoneless test, right? Since the same behavior where the test fails is not reproducible in a sample. If thats correct, I dont think that we should attach a resize observer. Also, I did some digging why that test fails in zoneless: 
I found out that that when chips are added there is logic in the `filtering-row` which "scrolls" them. This scrolling is managed inside a `requestAnimationFrame` which does not automatically trigger a CD cycle after its execution. It looks like this:
```
    private transform(offset: number) {
        requestAnimationFrame(() => {
            this.chipsArea.element.nativeElement.style.transform = `translate(${offset}px)`;
        });
    }
```
If thats the case indeed I think that you can just add a `markForCheck` here

projects/igniteui-angular/grids/core/src/resizing/resizing.service.ts
```diff
@@ -89,7 +102,7 @@ export class IgxColumnResizingService {
         const currentColWidth = this.getColumnHeaderRenderedWidth();
         this.column.width = this.column.getAutoSize();
 
-        this.zone.run(() => { });
+        this.notifyResized();
```
mddragnev: Column width updates already go through the column width setter + notifyChanges decorator chain, which marks grid CD: setting width -> watch-changes -> notifyChanges in the grid.
Because of that, I’m not entirely sure we even need the explicit zone.run call in ZoneJS apps, and I’m also not sure we need the extra notifyChanges scheduling in notifyResized.
Question:
Is there a specific regression or known issue this is fixing in either ZoneJS mode or zoneless mode. If there is a concrete scenario it would help to capture it in a short test note so the intent is explicit and future cleanups don’t accidentally regress it.

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -7076,24 +7075,16 @@ export abstract class IgxGridBaseDirective implements GridType,
             this.cdr.detectChanges();
         }
 
-        if (this.zone.isStable) {
+        runAfterRenderOnce(this.injector, () => {
             this.zone.run(() => {
```
mddragnev, line 7079: There might be a chance that this would trigger more detect changes cycles than the previous version. We should test this thoroughly.
Again here, not entirely sure that we need this `zone.run` because even in zone cd we have implicit trigger of CD. However, you can leave it as is

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -7743,19 +7734,13 @@ export abstract class IgxGridBaseDirective implements GridType,
...
                 this.changeRowEditingOverlayStateOnScroll(this.crudService.rowInEditMode);
             }
         };
-        if (this.isZonelessChangeDetection()) {
-            this.cdr.detectChanges();
-            callback();
-        } else {
-            this.zone.onStable.pipe(first()).subscribe(callback);
-        }
+        runAfterRenderOnce(this.injector, callback, 'read');
```
mddragnev: This should be called in the `mixedReadWrite` phase because `changeRowEditingOverlayStateOnScroll` might trigger reads and writes in/from the DOM

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -7808,12 +7789,12 @@ export abstract class IgxGridBaseDirective implements GridType,
...
 
         this.zone.run(() => {
-            this.zone.onStable.pipe(first()).subscribe(() => {
+            runAfterRenderOnce(this.injector, () => {
                 this.parentVirtDir.chunkLoad.emit(this.headerContainer.state);
                 requestAnimationFrame(() => {
                     this.autoSizeColumnsInView();
                 });
-            });
+            }, 'read');
```
mddragnev: `mixedReadWrite` here as well

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -7848,7 +7828,17 @@ export abstract class IgxGridBaseDirective implements GridType,
 
             return;
         }
+        this.executeNavigationCallback(row, visibleColIndex, cb);
```
mddragnev: You are introducing a behavior change here. This is public facing API that executed the `cb` even if the `row` or `args.target` were nullish.

mddragnev (changes requested):

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
viktorkombov: Yes, I think this is the reason for the bug I reported above. I added it because of an error caused by a missing target while testing column autosizing in the Tree Grid. I’ll check this tomorrow when I’m back at work.

viktorkombov: Correction: I was mistaken in my previous comment. The missing target is related to keyboard navigation, not column autosizing. I have reverted this change and logged a separate issue to investigate and resolve the navigation problem - #17415.

projects/igniteui-angular-elements/src/app/custom-strategy.spec.ts
```diff
@@ -210,6 +212,7 @@ describe('Elements: ', () => {
 
             const thirdGroup = document.querySelector<IgcNgElement>('igc-column-layout[header="Product Stock"]');
             const secondGroup = document.querySelector<IgcNgElement>('igc-column-layout[header="Product Details"]');
+            await waitForChildrenResolved(grid);
```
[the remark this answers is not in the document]
viktorkombov: Good catch. I removed the duplicate await and updated the helper to wait for a single `childrenResolved` boundary where the expected column state has been reached. The subscriptions for subsequent DOM mutations are also created before the mutation to avoid missing the event.

projects/igniteui-angular/directives/src/directives/for-of/for_of.directive.ts
```diff
@@ -962,15 +958,10 @@ export class IgxForOfDirective<T, U extends T[] = T[]> extends IgxForOfToken<T,U
...
-                    this.dc.instance._viewContainer.element.nativeElement.style.transform = `translateY(${-scrollOffset}px)`;
-                }
-              });
-          });
-
-        this._zone.onStable.pipe(first()).subscribe(this.recalcUpdateSizes.bind(this));
+        runAfterRenderOnce(this._injector, () => {
+            this.dc.instance._viewContainer.element.nativeElement.style.transform = `translateY(${-scrollOffset}px)`;
+        }, 'write');
+        runAfterRenderOnce(this._injector, () => this.recalcUpdateSizes(), 'read');
```
mddragnev: `recalcUpdateSizes` may trigger `detectChange` and may update some DOM sizes. So more appropriate would be to be call it in `mixedReadWrite` phase

projects/igniteui-angular/directives/src/directives/for-of/for_of.directive.ts
```diff
@@ -1182,7 +1173,7 @@ export class IgxForOfDirective<T, U extends T[] = T[]> extends IgxForOfToken<T,U
         } else {
             this.dc.instance._viewContainer.element.nativeElement.style.left = -scrollOffset + 'px';
         }
-        this._zone.onStable.pipe(first()).subscribe(this.recalcUpdateSizes.bind(this));
+        runAfterRenderOnce(this._injector, () => this.recalcUpdateSizes(), 'read');
```
mddragnev: Same here

projects/igniteui-angular/directives/src/directives/for-of/for_of.directive.ts
```diff
@@ -1785,14 +1776,10 @@ export class IgxGridForOfDirective<T, U extends T[] = T[]> extends IgxForOfDirec
...
-                write: () => {
-                    this.dc.instance._viewContainer.element.nativeElement.style.transform = `translateY(${-scrollOffset}px)`;
-                    this._zone.onStable.pipe(first()).subscribe(this.recalcUpdateSizes.bind(this, prevState));
-                }
-              });
-          });
+        runAfterRenderOnce(this._injector, () => {
+            this.dc.instance._viewContainer.element.nativeElement.style.transform = `translateY(${-scrollOffset}px)`;
+        }, 'write');
+        runAfterRenderOnce(this._injector, () => this.recalcUpdateSizes(prevState), 'read');
```
mddragnev: same here

projects/igniteui-angular/grids/grid/src/grid-cell-selection.spec.ts
```diff
@@ -1737,6 +1738,44 @@ describe('IgxGrid - Cell selection #grid', () => {
...
+            UIInteractions.simulateClickAndSelectEvent(firstCell);
+            await wait();
+            await fix.whenStable();
+
+            expect(selectionChangeSpy).toHaveBeenCalledTimes(0);
+            GridSelectionFunctions.verifyCellSelected(firstCell);
+            expect(grid.selectedCells.length).toBe(1);
+
+            UIInteractions.triggerKeyDownEvtUponElem('end', firstCell.nativeElement, true, false, true, true);
+            await wait(200);
```
mddragnev: You were using `waitForGridSettle` in the Zone test above instead of `await wait(200)`. Perhaps, you should use it here as well.

projects/igniteui-angular/grids/grid/src/grid-keyBoardNav-headers.spec.ts
```diff
@@ -63,12 +64,13 @@ describe('IgxGrid - Headers Keyboard navigation #grid', () => {
 
         it('should focus first header when the grid is scrolled', async () => {
             grid.navigateTo(7, 5);
-            await wait(250);
-            fix.detectChanges();
+            // navigateTo scrolls both axes; drive the deferred scroll processing so the
+            // columns land in view before focusing the header.
+            await dispatchGridScrollEvents(fix, grid, { waitMs: 250 });
```
mddragnev: Wait, why do you need to dispatch scroll events here? Isn't there a cleaner way?

projects/igniteui-angular/grids/grid/src/grid.component.spec.ts
```diff
@@ -2104,6 +2104,40 @@ describe('IgxGrid Component Tests #grid', () => {
...
+                TestBed.configureTestingModule({
+                    providers: [provideZonelessChangeDetection()]
+                });
+            });
+
+            it('should set correct aria attributes related to total rows/cols count and indexes', async () => {
+                const fix = TestBed.createComponent(IgxGridDefaultRenderingComponent);
+                fix.componentInstance.initColumnsRows(80, 20);
+                fix.detectChanges();
+                fix.detectChanges();
```
mddragnev, line 2122: Why do you need `detectChanges` twice?

mddragnev (changes requested):
We need to talk about all the new methods for testing that you are introducing. There is too much change detection there that I am not a fan of.

projects/igniteui-angular/grids/grid/src/grid-keyBoardNav-headers.spec.ts
```diff
@@ -62,9 +63,11 @@ describe('IgxGrid - Headers Keyboard navigation #grid', () => {
         });
 
         it('should focus first header when the grid is scrolled', async () => {
+            const verticalScroll = waitForGridScroll(fix, grid, 'vertical');
+            const horizontalScroll = waitForGridScroll(fix, grid, 'horizontal');
             grid.navigateTo(7, 5);
-            await wait(250);
-            fix.detectChanges();
+            await verticalScroll;
```
mddragnev: I am kind of worried about these. There might be a case where either horziontal or vertical scroll is not needed. Then wouldn't the test just hang?

projects/igniteui-angular/grids/grid/src/grid-keyBoardNav-headers.spec.ts
```diff
@@ -760,6 +765,72 @@ describe('IgxGrid - Headers Keyboard navigation #grid', () => {
...
+        beforeEach(() => {
+            fix = TestBed.createComponent(SelectionWithScrollsComponent);
+            fix.detectChanges();
+            grid = fix.componentInstance.grid;
+            gridHeader = GridFunctions.getGridHeader(grid);
+        });
+
+        it('should focus first header when the grid is scrolled', async () => {
+            grid.navigateTo(7, 5);
+            await wait(250);
```
mddragnev: This is a copy of a test from the Zone CD testbed, right? Why is there such difference in async waiting?

projects/igniteui-angular/grids/grid/src/grid-mrl-keyboard-nav.spec.ts
```diff
@@ -1991,9 +1980,12 @@ describe('IgxGrid Multi Row Layout - Keyboard navigation #grid', () => {
                 fix.detectChanges();
 
                 // arrow right
+                let verticalChunkLoad = firstValueFrom(grid.verticalScrollContainer.chunkLoad);
```
mddragnev: This looks similar to the use cases above in the file. Could you use `navigateWithGridScroll` here as well?

projects/igniteui-angular/grids/grid/src/grid-mrl-keyboard-nav.spec.ts
```diff
@@ -2010,9 +2002,12 @@ describe('IgxGrid Multi Row Layout - Keyboard navigation #grid', () => {
                 fix.detectChanges();
 
                 // arrow left
+                verticalChunkLoad = firstValueFrom(grid.verticalScrollContainer.chunkLoad);
```
mddragnev: Same question here

projects/igniteui-angular/grids/grid/src/grid-mrl-keyboard-nav.spec.ts
```diff
@@ -2123,7 +2116,7 @@ describe('IgxGrid Multi Row Layout - Keyboard navigation #grid', () => {
 
                 // first cell should be active and is fully in view
                 firstCell = grid.gridAPI.get_cell_by_index(0, 'ID');
-                expect(firstCell.active).toBe(true);
+                expect(fix.componentInstance.selectedCell.column.field).toMatch('ID');
```
mddragnev: Any specific reason you removed the active check?

projects/igniteui-angular/grids/grid/src/grid-mrl-keyboard-nav.spec.ts
```diff
@@ -2665,6 +2655,195 @@ describe('IgxGrid Multi Row Layout - Keyboard navigation #grid', () => {
...
+                provideZonelessChangeDetection()
+            ]
+        }).compileComponents();
+    }));
+
+    beforeEach(() => {
+        fix = TestBed.createComponent(ColumnLayoutTestComponent);
+    });
+
+    it(`should navigate to the last cell from the layout by pressing Home/End and Ctrl key
```
mddragnev: This test differs a lot from the Zone counterpart above.

projects/igniteui-angular/grids/grid/src/grid.master-detail.spec.ts
```diff
@@ -667,15 +668,17 @@ describe('IgxGrid Master Detail #grid', () => {
...
             fix.detectChanges();
 
             const targetCellElement = grid.gridAPI.get_cell_by_index(52, 'CompanyName');
             UIInteractions.simulateClickAndSelectEvent(targetCellElement);
-            fix.detectChanges();
 
-            UIInteractions.triggerEventHandlerKeyDown('ArrowUp', gridContent, false, false, true);
-            await waitForActiveNodeChange(grid);
-            fix.detectChanges();
+            await waitForGridNavigation(
```
mddragnev: Isn't his the same as calling `navigateWithGridScroll` ?

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.navigation.spec.ts
```diff
@@ -170,23 +221,24 @@ describe('IgxHierarchicalGrid Navigation', () => {
...
 
-            const childLastCell =  childGrid.dataRowList.toArray()[9].cells.toArray()[3];
+            const childLastCell = childGrid.gridAPI.get_cell_by_index(9, 'childData2');
             GridFunctions.focusCell(fixture, childLastCell);
 
             const childGridContent =  fixture.debugElement.queryAll(By.css(GRID_CONTENT_CLASS))[1];
-            UIInteractions.triggerEventHandlerKeyDown('home', childGridContent, false, false, true);
-            await wait(DEBOUNCE_TIME * 3);
-            fixture.detectChanges();
+            const parentVerticalScroll = waitForGridScroll(fixture, hierarchicalGrid, 'vertical');
```
mddragnev: Aren't `waitForGridScroll` and `navigateAndWaitForScroll` doing the same job? Why do you need them both?

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.navigation.spec.ts
```diff
@@ -943,17 +982,26 @@ describe('IgxHierarchicalGrid Navigation', () => {
             const firstChildGrid = childGrids[0].query(By.directive(IgxHierarchicalGridComponent)).componentInstance;
             const secondChildGrid = childGrids[1].query(By.directive(IgxHierarchicalGridComponent)).componentInstance;
 
-            firstChildGrid.verticalScrollContainer.scrollTo(9);
-            fixture.detectChanges();
-            await wait(DEBOUNCE_TIME);
-
-            const firstChildCell =  firstChildGrid.gridAPI.get_cell_by_index(9, 'Col1');
+            let firstChildCell = firstChildGrid.gridAPI.get_cell_by_index(9, 'Col1');
```
mddragnev: Any reason you choose to manually scroll instead of using the API?

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.navigation.spec.ts
```diff
@@ -943,17 +982,26 @@ describe('IgxHierarchicalGrid Navigation', () => {
...
+                await firstChildVerticalScroll;
+                firstChildCell = firstChildGrid.gridAPI.get_cell_by_index(9, 'Col1');
+            }
             GridFunctions.focusCell(fixture, firstChildCell);
 
             const childGridContent =  fixture.debugElement.queryAll(By.css(GRID_CONTENT_CLASS))[1];
-            UIInteractions.triggerEventHandlerKeyDown('arrowdown', childGridContent, false, false, false);
-            fixture.detectChanges();
-            await wait(DEBOUNCE_TIME);
+            await runGridAction(fixture, [firstChildGrid, secondChildGrid, hierarchicalGrid], async () => {
```
mddragnev: Is all of this necessary?

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.navigation.spec.ts
```diff
@@ -989,16 +1037,13 @@ describe('IgxHierarchicalGrid Navigation', () => {
                 rowID: 10
             };
             await wait(16);
-            hierarchicalGrid.navigation.navigateToChildGrid([path]);
-            await wait(DEBOUNCE_TIME);
+            await runGridAction(fixture, [hierarchicalGrid], () =>
```
mddragnev: Is this necessary as well? You may use another already introduced scrolling helpers.

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.navigation.spec.ts
```diff
@@ -1018,18 +1063,480 @@ describe('IgxHierarchicalGrid Navigation', () => {
                 rowID: 5
             };
 
-            hierarchicalGrid.navigation.navigateToChildGrid([targetRoot, targetNested]);
+            const rootRowIsland = getRowIslandByKey(hierarchicalGrid.childLayoutList, targetRoot.rowIslandKey);
```
mddragnev: This looks so heavy. Waiting for initializations and etc. Are you sure that we are not missing anything regarding the change detection?

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.navigation.spec.ts
```diff
@@ -1018,18 +1063,480 @@ describe('IgxHierarchicalGrid Navigation', () => {
...
+            baseHGridContent = GridFunctions.getGridContent(fixture);
+            GridFunctions.focusFirstCell(fixture, hierarchicalGrid);
+        }));
+
+        afterEach(() => {
+            clearGridSubs();
+        });
+
+        it('should allow navigating to end in child grid when child grid target row moves outside the parent view port.', async () => {
+            const childGrid = hierarchicalGrid.gridAPI.getChildGrids(false)[0];
```
mddragnev, line 1065: Wait, almost all of the zoneless tests look so much better around the navigation part.  How is there such difference between zone and zoneless tests?

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.virtualization.spec.ts
```diff
@@ -548,3 +542,126 @@ export class IgxHierarchicalGridNoScrollTestComponent extends IgxHierarchicalGri
...
+
+        (hierarchicalGrid.dataRowList.toArray()[2].nativeElement.children[0] as HTMLElement).click();
+        await wait();
+        await fixture.whenStable();
+
+        (hierarchicalGrid.dataRowList.toArray()[1].nativeElement.children[0] as HTMLElement).click();
+        await wait();
+        await fixture.whenStable();
+
+        let chunkLoad = firstValueFrom(hierarchicalGrid.verticalScrollContainer.chunkLoad);
```
mddragnev: This is different from how you scroll in the zone test

projects/igniteui-angular/grids/hierarchical-grid/src/hierarchical-grid.virtualization.spec.ts
```diff
@@ -548,3 +542,126 @@ export class IgxHierarchicalGridNoScrollTestComponent extends IgxHierarchicalGri
...
+        (firstRow.nativeElement.children[0] as HTMLElement).click();
+        await wait();
+        await fixture.whenStable();
+        expect(firstRow.expanded).toBeTruthy();
+
+        const verticalScroll = hierarchicalGrid.verticalScrollContainer;
+        const elem = verticalScroll['scrollComponent'].elementRef.nativeElement;
+
+        // scroll down so the expanded row is recycled out of view
+        let chunkLoad = firstValueFrom(hierarchicalGrid.verticalScrollContainer.chunkLoad);
```
mddragnev: This is different in the zone test as well

projects/igniteui-angular/grids/tree-grid/src/tree-grid-keyBoardNav.spec.ts
```diff
@@ -846,4 +851,455 @@ describe('IgxTreeGrid - Key Board Navigation #tGrid', () => {
...
+            const firstCell: CellType = treeGrid.gridAPI.get_cell_by_index(5, 'ID');
+            UIInteractions.simulateClickAndSelectEvent(firstCell);
+            await fix.whenStable();
+
+            TreeGridFunctions.verifyTreeGridCellSelected(treeGrid, treeGrid.gridAPI.get_cell_by_index(5, 'ID'));
+            expect(treeGrid.selected.emit).toHaveBeenCalledTimes(1);
+
+            for (let i = 5; i < 9; i++) {
+                let cell = treeGrid.gridAPI.get_cell_by_index(i, 'ID');
+                const chunkLoad = firstValueFrom(treeGrid.verticalScrollContainer.chunkLoad);
```
mddragnev: This is different as well from the zone CD test

mddragnev (changes requested):

viktorkombov:
> We need to talk about all the new methods for testing that you are introducing. There is too much change detection there that I am not a fan of.

@mddragnev You are right to question the additional change-detection handling in the tests. I reverted the per-test synchronization changes and kept the zone-based tests in their original master form.

[The remaining centralized helper](https://github.com/IgniteUI/igniteui-angular/pull/17328/changes#diff-7fd5bdf5361c9dad2e02bca9773c0d880751e6eb69603f4ee964dc4bb4faca58R28) is needed because these are zone-based `ComponentFixture` tests with manual change detection. Unlike a bootstrapped application view, the fixture view does not participate in regular application-wide render ticks unless `autoDetectChanges()` is enabled.

This became relevant after replacing `NgZone.onStable` with `afterNextRender`. `onStable` invoked the callback directly when the zone became stable, whereas a view-bound `afterNextRender` callback requires Angular to traverse the fixture view. Without a fixture render, that callback can remain pending in these tests even though the bootstrapped view would be traversed in a real application.

The existing setup helper now listens to the real `gridScroll` event and schedules one deferred, coalesced `fixture.detectChanges()` pass. It does not emit `chunkLoad`, activate cells, or complete navigation. It only provides the render boundary missing from the manual fixture, after the scroll handler and its synchronous size corrections have completed.

I validated this approach against the affected zone suites: it reduced the failures from roughly 25 to 8, and the remaining failures were investigated individually and were unrelated to this helper. As a negative control, disabling the production `chunkLoad.emit` caused the scroll-dependent tests to fail, confirming that they still depend on the real scroll, virtualization, and navigation pipeline.

`fixture.autoDetectChanges()` is the broader Angular alternative, but enabling it would change the change-detection behavior of the existing suites. The centralized helper keeps the adjustment limited to the scroll flows affected by the migration.

I have removed the additional zoneless test variants for now because their synchronization had diverged too much from the original scenarios. I will add them separately after reworking their setup so they exercise the zoneless scheduler without test-specific timing workarounds.

projects/igniteui-angular/grids/grid/src/grid-mrl-keyboard-nav.spec.ts
```diff
@@ -1991,9 +1993,12 @@ describe('IgxGrid Multi Row Layout - Keyboard navigation #grid', () => {
                 fix.detectChanges();
 
                 // arrow right
+                let verticalChunkLoad = firstValueFrom(grid.verticalScrollContainer.chunkLoad);
```
mddragnev: Doesnt `fix.autoDetectChanges();` help with navigation here as well?

projects/igniteui-angular/grids/grid/src/grid.master-detail.spec.ts
```diff
@@ -1276,6 +1278,158 @@ describe('IgxGrid Master Detail #grid', () => {
...
+            providers: [
+                provideZonelessChangeDetection(),
+                IgxGridMRLNavigationService,
+                { provide: SCROLL_THROTTLE_TIME_MULTIPLIER, useValue: 0 }
+            ]
+        }).compileComponents();
+    }));
+
+    afterEach(() => {
+        clearGridSubs();
```
mddragnev: Is this a leftover?

projects/igniteui-angular/grids/grid/src/grid.search.spec.ts
```diff
@@ -1082,14 +1083,21 @@ describe('IgxGrid - search API #grid', () => {
...
+            await fix.whenStable();
 
+            const chunkLoad = firstValueFrom(grid.verticalScrollContainer.chunkLoad);
             (grid as any).scrollTo(9, 0);
-            await firstValueFrom(grid.verticalScrollContainer.chunkLoad);
             fix.detectChanges();
+            await wait();
+            await fix.whenStable();
+            fix.detectChanges();
+            await chunkLoad;
```
mddragnev: Doesn't `fix.autoDetectChanges();` help here as well?

mddragnev (changes requested):

projects/igniteui-angular/grids/grid/src/grid.search.spec.ts
viktorkombov: Yes, `autoDetectChanges()` is sufficient for the render scheduling introduced by `afterNextRender`, so I restored the rest of the test to its `master` form. I kept the existing `chunkLoad` wait from `master`, since it represents completion of the virtualized scroll. Removing it still makes the test fail because auto-detection alone does not wait for the browser scroll to complete

projects/igniteui-angular/grids/grid/src/grid.master-detail.spec.ts
viktorkombov: Yes, this was a leftover. The zoneless tests do not create these subscriptions, so I removed the `clearGridSubs()` call.

projects/igniteui-angular/grids/grid/src/grid-mrl-keyboard-nav.spec.ts
viktorkombov: Yes, `autoDetectChanges()` is sufficient here. I restored the rest of the test to its `master` form and removed the newly added `verticalChunkLoad` handling. I verified that the original test fails without auto-detection and passes with only `fix.autoDetectChanges()` added.

mddragnev (approved):

rkaraivanov (approved):

PR #17389, opened 2026-07-07
chore(elements): add csSuppress to ISummaryExpression - master

IMinchev64:
Related to https://github.com/Infragistics-Developer-Tools/dev-tools/pull/3754

 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

## How Has This Been Tested?

 - [ ] Unit tests
 - [ ] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **Angular version**: 
 - **Browser(s)**: 
 - **OS**: 

## Checklist:

### DIFF
--- a/projects/igniteui-angular/core/src/data-operations/grid-types.ts
+++ b/projects/igniteui-angular/core/src/data-operations/grid-types.ts
@@ -430,6 +430,7 @@ export interface IPathSegment {
     rowIslandKey: string;
 }
 
+/* csSuppress */
 /* tsPlainInterface */
 /* marshalByValue */
 export interface ISummaryExpression {


projects/igniteui-angular/core/src/data-operations/grid-types.ts
```diff
@@ -430,6 +430,7 @@ export interface IPathSegment {
     rowIslandKey: string;
 }
 
+/* csSuppress */
 /* tsPlainInterface */
 /* marshalByValue */
```
damyanpetev, line 435: In that case might want to drop the other meta tags, since they are not going to be needed? And it seems this is actually included as necessary type in the analyzer for some reason, so remove from there and there's a good likelihood even the `csSuppress` might not be needed.

damyanpetev: Eh, wait. That's actually an internal type?! That's also exported from the public_api barrel and generates API docs https://www.infragistics.com/api/angular/igniteui-angular/22.0.0/interfaces/ISummaryExpression/?

~Um, actually, please put `@hidden @internal` instead on that type 😶~

Plot thickens.. Is that supposed to be the argument type of the _**public**_ grid method `enableSummaries()`?! In that case, it should generate alright across the board and we've just have broken API signatures for some reason that should be fixed instead.

IMinchev64: Well I added `csSuppress` here for Blazor specifically as to avoid any duplicates as `SummaryExpression.cs` seems to already be [manually added](https://github.com/Infragistics-Developer-Tools/dev-tools/blob/iminchev/sync-generated-grid-input-description-output/XPlatform/Main/Source/WebGrids/Shared/Grids/SummaryExpression.cs) to `WebGrids`.

damyanpetev: It seems the generated file is more/less identical, so we might as well have it; usually prefer over manual.

Also add `disableSummaries()` to the list and possibly more; Good thing the spec https://github.com/IgniteUI/igniteui-angular/wiki/Summaries-Specification has some API listed... not 🫠

PR #15125, opened 2024-12-02
refactor(*): bundle styles with components

simeonoff:
Closes #13770, #16794

### Additional information (check all that apply):
 - [ ] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD

### Checklist:
 - [ ] All relevant tags have been applied to this PR
 - [ ] This PR includes unit tests covering all the new code ([test guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Test-implementation-guidelines-for-Ignite-UI-for-Angular))
 - [ ] This PR includes API docs for newly added methods/properties ([api docs guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Documentation-Guidelines))
 - [ ] This PR includes `feature/README.MD` updates for the feature docs
 - [ ] This PR includes general feature table updates in the root `README.MD`
 - [ ] This PR includes `CHANGELOG.MD` updates for newly added functionality
 - [ ] This PR contains breaking changes
 - [ ] This PR includes `ng update` migrations for the breaking changes ([migrations guidelines](https://github.com/IgniteUI/igniteui-angular/wiki/Update-Migrations))
 - [ ] This PR includes behavioral changes and the feature specification has been updated with them

PR #17440, opened 2026-07-21
fix(grid): eliminate empty space after last column and scrollbar not reaching the end

### LINKED ISSUE #17428: Grid with horizontal scrollbar leaves empty space after the last column
### Description

Having grid with horizontal scrollbar leaves empty space after the last column and the scrollbar doesn't move to the end.

### Framework

Angular

### Angular Version

latest

22.0.x, 22.1.x

### Component / Area

Grid

### Browser

Chrome

### Operating System

_No response_

### Command Not Working

```shell

```

### Steps to Reproduce

1. Open almost any sample with horizontal scrollbar, i.e grid filtering dev sample
2. Scroll to the right

### Actual Result

You can see that there is additional space after the last column and the horizontal scrollbar is not at the end.

Image

### Expected Result

There should be no space left

### Reproduction URL

_No response_

### Attachments

_No response_

Zneeky:
Closes #17428  

## Description
Fixes the Grid's horizontal scrollbar leaving empty, unscrollable space after the last
column, with the scrollbar thumb never reaching the end of its track.

Root cause: `igx-horizontal-virtual-helper` (and its vertical counterpart) used a shared
`overflow: auto` on both axes. The horizontal helper's height is pinned to exactly the
native scrollbar thickness, so once the real horizontal scrollbar claims that space,
`clientHeight` becomes `0`. Since `overflow-y` was also `auto`, the helper's 1px-tall
placeholder content then always "overflows" that 0px client height, causing the browser
to spawn a phantom vertical scrollbar. That phantom scrollbar eats ~16-17px of width from
the helper, shrinking its effective scrollable range below the real content width — so
the grid can never scroll far enough to bring the last column fully into view.

Fix: constrain each virtual helper to scroll only along its intended axis —
`overflow-y: hidden` on the horizontal helper and `overflow-x: hidden` on the vertical
helper (defensive, symmetrical fix for the same latent failure mode) in
`_igx-vhelper.scss`.

## Motivation / Context
Reported bug: opening almost any grid sample with a horizontal scrollbar (e.g. the Grid
Filtering dev sample) and scrolling all the way to the right leaves a gap after the last
column, and the horizontal scrollbar thumb stops short of the track's end.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

Grid (and other grid-based components sharing `igxFor` virtualization: Tree Grid,
Hierarchical Grid, Pivot Grid) — horizontal/vertical virtual scrollbar helpers.

## How Has This Been Tested?
 - [ ] Unit tests
 - [x] Manual testing
 - [ ] Automated e2e tests

Manually reproduced and verified via the Grid Filtering dev sample: measured
`clientWidth`/`scrollWidth` of `igx-horizontal-virtual-helper` before and after the fix,
and confirmed `scrollLeft` now reaches exactly `scrollWidth - clientWidth`, with the last
column ("Index") flush against the grid's right edge.

### Test Configuration:
 - **Angular version**: 22.1.x (also verified applies cleanly to 22.0.x and 21.2.x)
 - **Browser(s)**: Chrome (Windows)
 - **OS**: Windows

## Checklist:
 - [x] Accessibility (ARIA, keyboard navigation, focus management) has been verified (no ARIA/keyboard impact — style-only fix)

### DIFF
--- a/projects/igniteui-angular/core/src/core/styles/components/_common/_igx-vhelper.scss
+++ b/projects/igniteui-angular/core/src/core/styles/components/_common/_igx-vhelper.scss
@@ -13,10 +13,12 @@
     top: 0;
     inset-inline-end: 0;
     width: var(--vhelper-scrollbar-size);
+    overflow-x: hidden;
 }
 
 %vhelper--horizontal {
     width: 100%;
+    overflow-y: hidden;
 }
 
 %vhelper-content--vertical {
--- a/projects/igniteui-angular/grids/grid/src/cell.spec.ts
+++ b/projects/igniteui-angular/grids/grid/src/cell.spec.ts
@@ -201,13 +201,10 @@ describe('IgxGrid - Cell component #grid', () => {
             const lastCell = cells[cells.length - 1];
             expect(GridFunctions.getValueFromCellElement(lastCell)).toEqual('990');
 
-            const scrollbarBorderWidth = parseFloat(getComputedStyle(
-                grid.nativeElement.querySelector('.igx-grid__tbody-scrollbar')
-            ).getPropertyValue('border-inline-start-width')) || 0;
-            // Calculate where the end of the cell is. Relative left position should equal the grid calculated width.
+            // Calculate where the end of the cell is. Relative left position should equal the grid calculated
+            // width, which is already reduced by the vertical scrollbar's width.
             expect(lastCell.nativeElement.getBoundingClientRect().left +
-                lastCell.nativeElement.offsetWidth +
-                grid.scrollSize).toEqual(grid.calcWidth + scrollbarBorderWidth);
+                lastCell.nativeElement.offsetWidth).toEqual(grid.calcWidth);
         }));
 
         it('should not reduce the width of last pinned cell when there is vertical scroll.', () => {


ChronosSF:
@Zneeky , please resolve the conflict

mddragnev (changes requested):
I don't think this is the right fix. It looks like we're patching a symptom of a regression rather than the cause, and the regression itself isn't present in `21.2.x`. I did some digging: this works correctly in `21.2.x`, where the scroll strip `%grid-scroll-main`has no border. In `22.0.x`, PR #17333  introduced a new 1px border on the scroll component:
```
$grid-scrollbar-borders: rem(1px) solid var(--ig-grid-summary-border-color, var(--row-border-color));

%grid-scroll-main {
    border-block-start: $grid-scrollbar-borders;
}
```

That border is never accounted for in the scrollbar size calculations. calculateScrollNativeSize() returns scrollWidth + 1, and the layout relies on that 1px cushion for the placeholder content. The new `border-block-start` consumes it out of the content box, so the 1px content now overflows vertically.
So the root cause is the untracked 1px border from #17333, not something that needs a workaround here. I'd suggest fixing it at the sizing calculations.

mddragnev:
So, i`ve looked into this more carefully. As I already said this is not reproducible in 21.2.x. Additionally, after the newly introduced scoped styles in master (https://github.com/IgniteUI/igniteui-angular/pull/15125) the issue is not reproducible there as well. So I am closing this now because once we release 22.1 we will drop support for 22.0 which is the only version the issue is now reproducible.

PR #17457, opened 2026-07-23
fix(grid): prevent NaN column width when hidden grid has all columns sized

### LINKED ISSUE #17418: Grid throws "Maximum call stack size exceeded" and becomes unresponsive when binding an empty data source while hidden (display:none)
### Description

When an IgbGrid is placed inside a container with `display:none`, binding an empty data source (0 records) can cause the following browser error:
```
Uncaught RangeError: Maximum call stack size exceeded
```
After the error occurs, the application also becomes noticeably sluggish. UI interactions may take several seconds to respond.

A sample project and a video demonstrating the behavior are attached.

### Hosting Model

Blazor WebAssembly

### .NET Version

10.0

### Ignite UI for Blazor Version

26.1.51

### Component / Area

Other

### Browser

Edge

### Operating System

Windows

### Steps to Reproduce

1. Run the attached sample application.
2. Verify that the page initially displays an IgbGrid populated with 50 records.
3. Click "Hide the Grid" to hide the grid's parent container (`display:none`).
4. While the grid remains hidden, click "Bind Empty Data" to bind an empty collection (0 records) to the grid.
5. Open the browser Developer Tools console.

### Actual Result

The browser console reports:
```
Uncaught RangeError: Maximum call stack size exceededShow
```
In addition, the application becomes significantly less responsive after the error occurs. For example, repeatedly clicking the Show/Hide the Grid button may result in delayed UI updates and poor overall responsiveness.

### Expected Result

No JavaScript error should occur when an empty data source is bound while the grid is hidden.
The grid should update normally and application responsiveness should remain unaffected.

### Reproduction URL

_No response_

### Attachments

[📦C_00246521_BlazorWasmApp1.zip](https://github.com/user-attachments/files/29995993/C_00246521_BlazorWasmApp1.zip)

https://github.com/user-attachments/assets/da1bd869-559e-43af-a22d-f93736b4069c

### Notes
The issue does not always reproduce under all environments and appears to be timing and performance dependent.
For me, the issue reproduces almost 100% of the time when:
- The browser window width is relatively small.
- Windows Energy Saver mode is enabled.
- The sample application is run with the attached configuration.

The issue also seems less likely to reproduce when:

- Running on a high-performance machine.
- Using a larger screen resolution.
- The grid contains fewer columns.

The attached video demonstrates both:

- The error being triggered by the reproduction steps.
- The application becoming sluggish after the error occurs.

#### Affected Versions
Reproduced with:

- IgniteUI.Blazor 25.2.77 or higher, including 26.1.51 (latest at the time of writing)

#### Last Known Good Version
The issue does not reproduce with:

- IgniteUI.Blazor 25.2.38 or earlyer

This suggests the issue was introduced sometime between 25.2.38 and 25.2.77.

#### Application Types
Reproduced in both:
- Blazor WebAssembly
- Blazor Server

#### Browsers
Reproduced in the latest versions (as of July 14, 2026) of:
- Microsoft Edge
- Google Chrome
- Mozilla Firefox

### Additional Investigation
I tested a similar sample using the Ignite UI for Web Components grid and could not reproduce the issue with either of the following versions:

- Ignite UI for Web Components 7.2.1
- Ignite UI for Web Components 6.0.4

Therefore, this appears to be specific to the Blazor wrapper and/or its interaction with the underlying grid implementation.

viktorkombov:
Closes #17418 

## Description
`getPossibleColumnWidth()` divided the available width by `columnsToSize` when every visible column already had an explicit or constrained width.

When the grid is hidden through a wrapper and has no measurable width, `calcWidth` falls back to the sum of the column widths. In this case, `computedWidth` equals `sumExistingWidths` and `columnsToSize` is `0`, resulting in `0 / 0 = NaN`. This then poisons the cached column widths and can cause an infinite sizing loop.

When the grid is using the column-width sum fallback and there are no columns left to auto-size, return the `"0px"` sentinel so the existing valid column widths are preserved.

 - [x] Bug fix
 - [ ] New functionality
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [x] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

## How Has This Been Tested?

 - [x] Unit tests
 - [x] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **Angular version**: 
 - **Browser(s)**: 
 - **OS**: 

## Checklist:

### DIFF
--- a/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
+++ b/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
@@ -5636,13 +5636,13 @@ export abstract class IgxGridBaseDirective implements GridType,
             return '0px';
         }
 
-        // When every visible column already has an explicit or constrained width there is
-        // nothing left to auto-size. Dividing the remaining space by `columnsToSize` (0) would
-        // produce NaN/±Infinity, which then poisons the cached column widths - most notably when
-        // the grid is hidden through a wrapper and falls back to summing the column widths, so
-        // `computedWidth` equals `sumExistingWidths`. Fall back to the default minimum width.
-        if (columnsToSize <= 0) {
-            return this.minColumnWidth + 'px';
+        // When there is nothing left to auto-size (columnsToSize === 0) and the grid is sizing to
+        // the sum of its column widths - which happens when it has no width of its own and is
+        // hidden through a wrapper - computedWidth equals sumExistingWidths, so the division below
+        // would be 0 / 0 = NaN and poison the cached column widths. Return the same "0px" sentinel
+        // used above so callers preserve the existing (valid) column widths instead.
+        if (columnsToSize <= 0 && this.isColumnWidthSum) {
+            return '0px';
         }
 
         computedWidth -= this.featureColumnsWidth();
--- a/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
+++ b/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
@@ -5636,11 +5636,12 @@ export abstract class IgxGridBaseDirective implements GridType,
             return '0px';
         }
 
-        // When there is nothing left to auto-size (columnsToSize === 0) and the grid is sizing to
-        // the sum of its column widths - which happens when it has no width of its own and is
-        // hidden through a wrapper - computedWidth equals sumExistingWidths, so the division below
-        // would be 0 / 0 = NaN and poison the cached column widths. Return the same "0px" sentinel
-        // used above so callers preserve the existing (valid) column widths instead.
+        // When the grid has no measurable width, calculateGridWidth() falls back to the
+        // sum of its column widths and sets isColumnWidthSum. If all visible columns
+        // already have explicit or constrained widths, columnsToSize is 0 and
+        // computedWidth equals sumExistingWidths, resulting in 0 / 0 = NaN.
+        // Return the "0px" sentinel so _derivePossibleWidth() preserves the existing
+        // valid column widths.
         if (columnsToSize <= 0 && this.isColumnWidthSum) {
             return '0px';
         }
--- a/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
+++ b/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
@@ -5636,6 +5636,15 @@ export abstract class IgxGridBaseDirective implements GridType,
             return '0px';
         }
 
+        // When every visible column already has an explicit or constrained width there is
+        // nothing left to auto-size. Dividing the remaining space by `columnsToSize` (0) would
+        // produce NaN/±Infinity, which then poisons the cached column widths - most notably when
+        // the grid is hidden through a wrapper and falls back to summing the column widths, so
+        // `computedWidth` equals `sumExistingWidths`. Fall back to the default minimum width.
+        if (columnsToSize <= 0) {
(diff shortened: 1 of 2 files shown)


projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -5636,6 +5636,15 @@ export abstract class IgxGridBaseDirective implements GridType,
...
         }
 
+        // When there is nothing left to auto-size (columnsToSize === 0) and the grid is sizing to
+        // the sum of its column widths - which happens when it has no width of its own and is
+        // hidden through a wrapper - computedWidth equals sumExistingWidths, so the division below
+        // would be 0 / 0 = NaN and poison the cached column widths. Return the same "0px" sentinel
+        // used above so callers preserve the existing (valid) column widths instead.
+        if (columnsToSize <= 0 && this.isColumnWidthSum) {
+            return '0px';
+        }
```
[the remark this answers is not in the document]
viktorkombov: columnsToSize === 0 also occurs during normal MRL sizing. We tested the suggested unconditional "0px" return, and it regresses “should initialize correctly when widths are set in %” with 64/128px layout differences. The "0px" sentinel is only appropriate when isColumnWidthSum indicates that the grid has no measurable width and is using the column-width sum fallback. Therefore, the guard is intentionally scoped to that state.

ChronosSF (approved):

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -5636,6 +5636,16 @@ export abstract class IgxGridBaseDirective implements GridType,
             return '0px';
         }
 
+        // When the grid has no measurable width, calculateGridWidth() falls back to the
+        // sum of its column widths and sets isColumnWidthSum. If all visible columns
+        // already have explicit or constrained widths, columnsToSize is 0 and
+        // computedWidth equals sumExistingWidths, resulting in 0 / 0 = NaN.
+        // Return the "0px" sentinel so _derivePossibleWidth() preserves the existing
+        // valid column widths.
+        if (columnsToSize <= 0 && this.isColumnWidthSum) {
```
damyanpetev, line 5645: Bit late & likely not an that much of an issue since the measures probably run again (possibly even more than once 🥲) when the grid is shown, but I remember there was some min (136px?) assigned for columns by default.
In case that'll spare parts of next ops, might consider returning that min instead of `0`, with the hope that in a the future where we've reduced the calculations, the column is more likely to settle on the min instead. I see `0` is already used for another case as well, so this fix is fine, just food for thought :)

viktorkombov: @damyanpetev Thanks for the suggestion! I considered returning the default minimum as well. The reason I kept `"0px"` is that `_derivePossibleWidth()` already treats it as a sentinel:
https://github.com/IgniteUI/igniteui-angular/blob/0c95c1b0fc331918a252e1e455153ead14a8fc68/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts#L6761-L6772
The `possibleWidth === "0px"` branch returns before `_columnWidth` is reassigned, leaving its previously calculated value unchanged. If that value is valid, `_updateColumnDefaultWidths()` reapplies it to the columns. Returning `minColumnWidth` would instead go through the normal path and replace `_columnWidth` while the grid has no measurable width.

Issue #14196, opened 2024-05-08
IgxTooltip and IgxDropdown is dependend on order in the html


jmulw:
## Description  
If you order the IgxDropdown and IgxTooltips incorrectly in the attributes of a component the drop down behaviour breaks completely.

 * igniteui-angular version: "^17.0.0"
 * browser: Any

## Steps to reproduce  
https://stackblitz.com/edit/7wzpjd?file=package.json
- Hover over Options button to open the dropdown
- Try to hover over any option for a few seconds

## Result  
What is the actual result after following the steps to reproduce?
The dropdown dissapears while you're still hovering over it

## Expected result  
What is the expected result after following the steps to reproduce?
You'd expect to be able to hover over the dropdown for as long as you want

## Notes
If you change the order of the igxTooltipTarget and the igxToggleAction it works as expected

PR #17483, opened 2026-08-05
refactor(igniteui-angular): enabling strict true for the library

### LINKED ISSUE #17403: Review & update Grid summaries API surface
_Sparked off from an discussion in https://github.com/IgniteUI/igniteui-angular/pull/17389#discussion_r3538148400_

### Thus far discovered:
We have `ISummaryExpression` that's public, but not documented. It's exported and listed in the API docs https://www.infragistics.com/api/angular/igniteui-angular/22.0.0/interfaces/ISummaryExpression/, yet it's not used _**anywhere.**_ - I don't see it in docs and/or samples https://www.infragistics.com/products/ignite-ui-angular/angular/components/grid/summaries

There's exactly 1 use of `ISummaryExpression` and that's in a protected method:
https://github.com/IgniteUI/igniteui-angular/blob/9ad405c40d9df3e0edc4f666beab6ff1b3bfe811/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts#L7290
However, that method is in:
https://github.com/IgniteUI/igniteui-angular/blob/9ad405c40d9df3e0edc4f666beab6ff1b3bfe811/projects/igniteui-angular/grids/grid/src/grid-base.directive.ts#L5135-L5153
Seems to me `rest` is actually `ISummaryExpression[]`, technically `ISummaryExpression[] | [fieldName: string, customSummary?: any] (destructed ISummaryExpression)`?! looking at the way the description/logic is, but that should really be in **overloads** instead. Also the spread might be absolute bull, considering the logic uses `[0]` a lot... -.-
Pretty sure the same thing applies to the equivalent `public disableSummaries()` on first glance.

Then there's the `customSummary` prop:
https://github.com/IgniteUI/igniteui-angular/blob/9ad405c40d9df3e0edc4f666beab6ff1b3bfe811/projects/igniteui-angular/core/src/data-operations/grid-types.ts#L435-L439
Whatever that is kind of hard to track, but it appears that's actually be `IgxSummaryOperand` but that's tracked from it being assigned to the respective column `summaries`, which opens another can of worms:
- `IgxSummaryOperand` is a class, with seemingly  abstract `static count()`, of course none of that is disclosed on the type
- Samples usage just ignores the type entirely on what's passed to the column [`class MySummary { ... }`](https://github.com/IgniteUI/igniteui-angular-samples/blob/aa7991c9033d4f9167dce09027385b233418043c/src/app/grid/grid-sample-3/grid-sample-3.component.ts#L11)
- And that's because the column `summaries` is untyped as well 🫠:
  https://github.com/IgniteUI/igniteui-angular/blob/9ad405c40d9df3e0edc4f666beab6ff1b3bfe811/projects/igniteui-angular/grids/core/src/columns/column.component.ts#L1126

I've confirmed that's how the API entered initially (see #845) and there's nothing of note described in https://github.com/IgniteUI/igniteui-angular/wiki/Summaries-Specification, so usage mostly needs to be inferred from tests and samples I guess.

### Bottomline:
- Clean up the Grid summaries API & update the spec
- Update docs & samples guidance
- Enable `noImplicitAny`/`strict` and consider lint `no-explicit-any` errors in general. Beyond user-provided data and derived values, `any` is rarely an acceptable type on public API.

kdinev:
Closes #17403

 - [ ] Bug fix
 - [ ] New functionality
 - [x] Refactoring (no functional changes)
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [ ] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

## How Has This Been Tested?

 - [x] Unit tests
 - [ ] Manual testing
 - [ ] Automated e2e tests

### Test Configuration:
 - **Angular version**: 
 - **Browser(s)**: 
 - **OS**: 

## Checklist:


projects/igniteui-angular/directives/src/directives/tooltip/tooltip-target.directive.ts
```diff
@@ -276,7 +276,8 @@ export class IgxTooltipTargetDirective extends IgxToggleActionDirective implemen
      * @hidden
      */
     @Input('igxTooltipTarget')
-    public override set target(target: any) {
+    public override set target(target: IgxTooltipDirective) {
```
viktorkombov: Could we revisit this type narrowing?

The plain-text tooltip syntax still appears to be a supported use case:

```html
<igx-icon igxTooltipTarget [tooltip]="'Text'">
```

This syntax is documented in both the changelog and the directive JSDoc, and I could not find a migration or deprecation notice for it.

With the new `IgxTooltipDirective` type, `npm start` reports `TS2322` for this syntax. The guard for #14196 also indicates that non-tooltip values can legitimately reach this setter from a sibling igxToggleAction.

Could the input type be adjusted to represent these existing input paths while retaining the current runtime narrowing? This would preserve the documented plain-text syntax without falling back to any.

Image

rkaraivanov: Sure, you can union it with a string

projects/igniteui-angular/directives/src/public_api.ts
```diff
@@ -28,7 +28,6 @@ export * from './directives/mask/mask.directive';
 // export { IgxRadioGroupDirective } from 'igniteui-angular/radio';
 export * from './directives/ripple/ripple.directive';
 export * from './directives/scroll-inertia/scroll_inertia.directive';
-export * from './directives/size/ig-size.directive';
```
viktorkombov, line 31: It looks like `IgSizeDirective` was intentionally moved from the library into the demo application, but `AppComponent` still imports it from `igniteui-angular/directives`. This causes npm start to fail because the corresponding public export was removed. Could we update the demo to import the new local directive instead?

Since this also removes an existing public export, could we confirm whether any changelog or migration follow-up is needed?

Image

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
```diff
@@ -3074,7 +3075,7 @@ export abstract class IgxGridBaseDirective implements GridType,
     /**
      * @hidden
      */
-    protected _pagingState;
+    protected _pagingState!: IPagingState;
```
viktorkombov, line 3078: This typing change exposes a pre-existing issue in `src/app/grid/grid.sample.html`: it accesses `grid2.pagingState?.paging?.index`, but `IPagingState` exposes `index` directly. Could we update it to `grid2.pagingState?.index` as part of this PR? This currently causes `npm start` to fail.

Image

projects/igniteui-angular/migrations/update-22_1_0_import-migration/index.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+describe(`Update to ${version}`, () => {
+    let appTree: UnitTestTree;
+    const schematicRunner = new SchematicTestRunner('ig-migrate', path.join(__dirname, '../migration-collection.json'));
+
+    beforeEach(() => {
+        appTree = setupTestTree();
+    });
+
+    const migrationName = 'migration-58';
```
viktorkombov: It looks like this spec runs `migration-58`, while the new import migration is registered as `migration-59`. Could we update `migrationName` accordingly? The current value runs the existing `update-22_1_0` migration instead and causes these tests to fail.
```suggestion
    const migrationName = 'migration-59';
```

projects/igniteui-angular/migrations/update-22_1_0_import-migration/index.ts
```diff
@@ -0,0 +1,203 @@
...
+        return null;
+    }
+
+    // Generate new import statements
+    const newImports: string[] = [];
+
+    // Add remaining imports for the original entry point first
+    if (remaining.length > 0) {
+        const sortedImports = remaining.sort();
+        newImports.push(`import { ${sortedImports.join(', ')} } from '${importPath}';`);
```
viktorkombov: Could we also preserve type-only imports during the rewrite? For example, `import type { IGroupingDoneEventArgs } ...` is currently regenerated as a regular `import { IGroupingDoneEventArgs } ...`, which may affect projects using `verbatimModuleSyntax`.

projects/igniteui-angular/migrations/update-22_1_0_import-migration/index.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    // Add moved imports
+    for (const [entryPoint, imports] of entryPointGroups) {
+        const sortedImports = imports.sort();
+        newImports.push(`import { ${sortedImports.join(', ')} } from '${source.basePackage}/${entryPoint}';`);
```
viktorkombov: Could we merge migrated symbols into an existing import from the target entry point? For example, when `IgxSummaryResult` is already imported from `igniteui-angular/core`, migrating the summary operands currently creates a second import from the same module.

Image

projects/igniteui-angular/directives/src/public_api.ts
mddragnev: I think that this was never meant to be public and I dont think anyone has ever used it. So we should be safe without changelog entry and migrations

projects/igniteui-angular/action-strip/src/action-strip/action-strip.component.ts
```diff
@@ -187,6 +187,7 @@ export class IgxActionStripComponent implements IgxActionStripToken, AfterViewIn
                 return false;
             }
         }
+        return undefined!;
```
mddragnev: There are a lot of instances where we `return undefined!;` in order to suppress the error from typescript.
In this case the code is not reachable since the menu would always have either 0 or more items. However, to suppress the warning I think that returning `false` is better.

projects/igniteui-angular/calendar/src/calendar/days-view/days-view.component.ts
```diff
@@ -656,6 +658,6 @@ export class IgxDaysViewComponent extends IgxCalendarBaseDirective implements Af
     }
 
     private setPreviewRangeDate(value?: Date) {
-        this.previewRangeDate = value;
+        this.previewRangeDate = value!;
```
mddragnev: The parameter is optional, so `value` may be `undefined`, but the non-null assertion suggests it is guaranteed to exist. Could we remove `!` and make the property type explicitly support undefined, or validate/provide a fallback here?

projects/igniteui-angular/calendar/src/calendar/month-picker/month-picker.component.ts
```diff
@@ -512,7 +512,7 @@ export class IgxMonthPickerComponent extends IgxCalendarBaseDirective implements
 
         const formatObject = Array.isArray(value)
             ? value.map((date, index) => construct(date, index))
-            : construct(value, i);
+            : construct(value, i!);
```
mddragnev: The `generateContext` method of the `month-picker` is never used without providing the index. We could probably change the parameter to be required here and remove the need of non null assertion.

projects/igniteui-angular/carousel/src/carousel/carousel.component.ts
```diff
@@ -1004,7 +1004,7 @@ export class IgxCarouselComponent extends IgxCarouselComponentBase implements On
         slide.nativeElement.style.opacity = '';
     }
 
-    private pan(event) {
+    private pan(event: any) {
```
mddragnev: This could be typed with `IgxGestureEvent`
```suggestion
    private pan(event: IgxGestureEvent) {
```

projects/igniteui-angular/directives/src/public_api.ts
kdinev: @mddragnev This directive has been developed entirely for demos-dev purposes and I have no clue why it has been put in the library in the first place and then respectively exported.

projects/igniteui-angular/grids/grid/src/grid-base.directive.ts
kdinev: @viktorkombov I wonder whether the sample ever worked properly, especially since the `igx-paginator` component was made standalone. I'm pretty sure I updated the internal API access to go directly through the paging state, but this demo has probably not been updated.

mddragnev (approved):

PR #17509, opened 2026-08-18
fix(esf): let templated filter operations fill the filter menu

ChronosSF:


ChronosSF:
@simeonoff , it seems this code existed but was either intentionally removed with the scoped style change or was removed during a wrong conflict resolution either by you or @viktorkombov . Can you please confirm if it should be there or not?

(this is the PR that fixes the broken test in master we discussed today)

simeonoff (approved):

PR #17566, opened 2026-09-02
fix(forms): Support Signal Forms in form controls

### LINKED ISSUE #17556: Signal Forms interop NgControl causes Ignite UI Angular form control initialization failure
### Description

Ignite UI's form controls still read legacy observables from `NgControl`. They subscribe to `statusChanges` (and, in the input group, `valueChanges`) from `ngAfterViewInit` to drive invalid styling. Signal Forms provides an interop `NgControl` backed by signals instead, which carries neither observable. The subscriptions fail with `"Cannot read properties of undefined (reading 'pipe')"`/`Cannot read properties of undefined (reading 'subscribe')`, and the control never finishes wiring itself up. Interacting with the controls results in: `this.ngControl.control.markAsTouched is not a function`.The issue occurs when using `[formField]` on an Ignite UI control.

```html
<igx-input-group>
    <input igxInput type="text" [formField]="userForm.firstName" />
    <label igxLabel>First Name</label>
</igx-input-group>
```

```ts
interface User {
    firstName: string;
}

public readonly userSignal = signal<User>({
    firstName: '',
});
public readonly userForm = form(this.userSignal, (path) => {
    required(path.firstName);
});
```

### Framework

Angular

### Angular Version

22.1

22.1.1

### Component / Area

Other

### Browser

Chrome

### Operating System

_No response_

### Command Not Working

```shell

```

### Steps to Reproduce

1. Use `[formField]` on an Ignite UI control (or use the attached sample).
2. Open the DevTools Console and observe the errors.
3. Interact with the control.

### Actual Result

Image

### Expected Result

No errors on initialization and when interacting with the controls.

### Reproduction URL

_No response_

### Attachments

[SignalForm.zip](https://github.com/user-attachments/files/31647144/SignalForm.zip)

rkaraivanov:
## Description

The `[formField]` interop `NgControl` exposes signal-backed getters only. It has no `statusChanges`, `valueChanges`, `validator`, `markAsTouched` or `setValue`, so igxInput, checkbox, switch, radio group, select, combo, simple combo and the date, time and date range pickers threw on init.

Add `NgControlAdapter` in core as the single access path to the bound `NgControl`. It detects the backend and derives the missing observables from a root effect over the signal getters, keeping change detection order identical to the observable case. Controls no longer read `NgControl` internals directly.

Closes #17556 #17617 #17618

 - [x] Bug fix

## How Has This Been Tested?
 - [x] Unit tests

## Checklist:

### DIFF
--- a/projects/igniteui-angular/core/src/core/ng-control-adapter.ts
+++ b/projects/igniteui-angular/core/src/core/ng-control-adapter.ts
@@ -1,0 +1,142 @@
+import { effect, EnvironmentInjector, Injector, untracked } from '@angular/core';
+import { AbstractControl, NgControl, TouchedChangeEvent, Validators } from '@angular/forms';
+import { filter, Observable } from 'rxjs';
+
+/** Source of the state behind an `NgControl`. */
+export type NgControlBackend = 'observable' | 'signal';
+
+/** Whether a control took a value written through `setValue`. */
+export type ValueWriteResult = 'accepted' | 'ignored';
+
+/**
+ * Uniform access to the `NgControl` bound to a form control.
+ *
+ * Reactive and template-driven forms provide an `NgControl` backed by an
+ * `AbstractControl` with observables. Signal Forms (`[formField]`) provide
+ * an interop `NgControl` backed by signals, without `statusChanges`,
+ * `valueChanges`, `validator` or `markAsTouched`.
+ *
+ * @hidden @internal
+ */
+export class NgControlAdapter {
+    public readonly backend: NgControlBackend;
+
+    private readonly envInjector: EnvironmentInjector;
+
+    /** Wraps `ngControl`, or returns `null` when there is none. */
+    public static from(ngControl: NgControl | null, injector: Injector): NgControlAdapter | null {
+        return ngControl ? new NgControlAdapter(ngControl, injector) : null;
+    }
+
+    constructor(private readonly ngControl: NgControl, injector: Injector) {
+        this.envInjector = injector.get(EnvironmentInjector);
+        this.backend = 'statusChanges' in ngControl ? 'observable' : 'signal';
+    }
+
+    public get disabled(): boolean {
+        return !!this.ngControl.disabled;
+    }
+
+    public get valid(): boolean {
+        return !!this.ngControl.valid;
+    }
+
+    public get invalid(): boolean {
+        return !!this.ngControl.invalid;
+    }
+
+    public get touchedOrDirty(): boolean {
+        const control = this.ngControl.control;
+        return !!(control?.touched || control?.dirty);
+    }
+
+    /** Signal Forms expose no validator list, only `required` and the current errors. */
+    public get hasValidators(): boolean {
+        if (this.backend === 'signal') {
+            return this.required || this.invalid;
+        }
+
+        const control = this.ngControl.control;
+        return !!(control?.validator || control?.asyncValidator);
+    }
+
+    public get required(): boolean {
+        if (this.backend === 'signal') {
+            return !!this.ngControl.control?.hasValidator(Validators.required);
+        }
+
+        const validator = this.ngControl.control?.validator;
+        if (!validator) {
+            return false;
+        }
+
+        // Probe with an empty control so `required` is detected regardless of the current value.
+        return !!validator({} as AbstractControl)?.required;
+    }
+
+    /**
+     * Emits when validity, disabled, dirty or pending state changes.
+     * Signal Forms `submit()` only marks fields touched, so touched changes count too
+     * or the errors would never surface.
+     */
+    public get statusChanges(): Observable<unknown> {
+        if (this.backend === 'signal') {
+            return this.watch(() => [
+                this.ngControl.valid, this.ngControl.invalid, this.ngControl.pending,
+                this.ngControl.disabled, this.ngControl.dirty, this.ngControl.touched
+            ]);
+        }
+
+        return this.ngControl.statusChanges!;
+    }
+
+    public get touchedChanges(): Observable<unknown> {
+        if (this.backend === 'signal') {
+            return this.watch(() => [this.ngControl.touched]);
+        }
+
+        return this.ngControl.control!.events.pipe(filter(e => e instanceof TouchedChangeEvent));
+    }
+
+    public get valueChanges(): Observable<unknown> {
+        if (this.backend === 'signal') {
+            return this.watch(() => [this.ngControl.value]);
+        }
+
+        return this.ngControl.valueChanges!;
+    }
+
+    /** No-op for Signal Forms: `[formField]` tracks touch through blur and `registerOnTouched`. */
+    public markAsTouched(): void {
+        this.ngControl.control?.markAsTouched?.();
+    }
+
+    /**
+     * Signal Forms ignore the write: they read the value from the view
+     * (DOM or `ControlValueAccessor`), so the caller must update that instead.
+     */
+    public setValue(value: unknown): ValueWriteResult {
+        const control = this.ngControl.control;
+        if (!control?.setValue) {
+            return 'ignored';
+        }
+
+        control.setValue(value);
+        return 'accepted';
+    }
+
+    // Signal-backed getters are reactive, so an effect over them replaces the missing observables.
+    // A root effect runs before change detection, like an observable would; a view effect would
+    // run after the host bindings were checked. `untracked` allows subscribing from within another
+    // effect. `toObservable` is not used: it replays and lives until the environment is destroyed.
+    private watch(read: () => unknown[]): Observable<void> {
+        return new Observable<void>(subscriber => {
+            const ref = untracked(() => effect(() => {
+                read();
+                untracked(() => subscriber.next());
+            }, { injector: this.envInjector }));
+
+            return () => ref.destroy();
+        });
+    }
+}
--- a/projects/igniteui-angular/date-picker/src/date-range-picker/date-range-picker.component.ts
+++ b/projects/igniteui-angular/date-picker/src/date-range-picker/date-range-picker.component.ts
@@ -62,6 +62,7 @@ import {
     DateTimeUtil,
     IgxPickerActionsDirective,
     isDateInRanges,
+    NgControlAdapter,
     PickerCalendarOrientation,
     THEME_TOKEN,
     ThemeToken
@@ -603,12 +604,7 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
     }
 
     private get required(): boolean {
-        if (this._ngControl && this._ngControl.control && this._ngControl.control.validator) {
-            const error = this._ngControl.control.validator({} as AbstractControl);
-            return (error && error.required) ? true : false;
-        }
-
-        return false;
+        return this._control?.required ?? false;
     }
 
     private get calendar(): IgxCalendarComponent {
@@ -640,6 +636,7 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
     private _originalValue!: DateRange | null;
     private _overlayId: string = '';
     private _ngControl!: NgControl;
+    private _control: NgControlAdapter | null = null;
     private _statusChanges$!: Subscription;
     private _calendar!: IgxCalendarComponent;
     private _calendarContainer?: HTMLElement;
@@ -861,6 +858,7 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
     /** @hidden */
     public ngOnInit(): void {
         this._ngControl = this._injector.get<NgControl>(NgControl, null);
+        this._control = NgControlAdapter.from(this._ngControl, this._injector);
     }
 
     /** @hidden */
@@ -875,8 +873,8 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
 
         this.setRequiredToInputs();
 
-        if (this._ngControl) {
-            this._statusChanges$ = this._ngControl.statusChanges!.subscribe(this.onStatusChanged.bind(this));
+        if (this._control) {
+            this._statusChanges$ = this._control.statusChanges.subscribe(this.onStatusChanged.bind(this));
         }
 
         // delay invocations until the current change detection cycle has completed
@@ -939,23 +937,15 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
     };
 
     private setValidityState(inputDirective: IgxInputDirective, isFocused: boolean) {
-        if (this._ngControl && !this._ngControl.disabled && this.isTouchedOrDirty) {
-            if (this.hasValidators && isFocused) {
-                inputDirective.valid = this._ngControl.valid ? IgxInputState.VALID : IgxInputState.INVALID;
+        if (this._control && !this._control.disabled && this._control.touchedOrDirty) {
+            if (this._control.hasValidators && isFocused) {
+                inputDirective.valid = this._control.valid ? IgxInputState.VALID : IgxInputState.INVALID;
             } else {
-                inputDirective.valid = this._ngControl.valid ? IgxInputState.INITIAL : IgxInputState.INVALID;
+                inputDirective.valid = this._control.valid ? IgxInputState.INITIAL : IgxInputState.INVALID;
             }
         } else {
             inputDirective.valid = IgxInputState.INITIAL;
         }
-    }
-
-    private get isTouchedOrDirty(): boolean {
-        return (this._ngControl.control!.touched || this._ngControl.control!.dirty);
-    }
-
-    private get hasValidators(): boolean {
-        return (!!this._ngControl.control!.validator || !!this._ngControl.control!.asyncValidator);
     }
 
     private handleSelection(selectionData: Date[]): void {
@@ -1307,6 +1297,11 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
             const _value = this.value ? this.toRangeOfDates(this.value) : null;
             start.updateInputValue(_value?.start || null!);
             end.updateInputValue(_value?.end || null!);
+
+            // `validate()` reads the editors, which Signal Forms validate before writing to; re-run it.
+            if (this._control?.backend === 'signal') {
+                this.onValidatorChange();
+            }
         }
     }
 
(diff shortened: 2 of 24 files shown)


projects/igniteui-angular/directives/src/directives/checkbox/checkbox-base.directive.ts
```diff
@@ -415,14 +411,10 @@ export class CheckboxBaseDirective implements AfterViewInit {
...
-                (this.ngControl.control!.touched || this.ngControl.control!.dirty)
-            ) {
+        if (this.control) {
+            if (!this.disabled && !this.readonly && this.control.touchedOrDirty) {
                 // the control is not disabled and is touched or dirty
-                this.invalid = this.ngControl.invalid!;
+                this.invalid = this.control.invalid;
             } else {
                 //  if the control is untouched, pristine, or disabled, its state is initial. This is when the user did not interact
                 //  with the checkbox or when the form/control is reset
```
viktorkombov: `required` detection here now goes through `NgControlAdapter.required`, which probes the validator with a bare `{}`:

```ts
return !!validator({} as AbstractControl)?.required;
```

That hands every custom validator a control with **`value === undefined`**, which they were never written to handle. Anything that dereferences the value throws before it can return:

```ts
const v = (c: AbstractControl): ValidationErrors | null =>
    (c.value as string[]).length === 0 ? { empty: true } : null;

form = new FormGroup({ accepted: new FormControl<unknown>([], v) });
```

```html
<form [formGroup]="form">
    <igx-checkbox formControlName="accepted">Accept</igx-checkbox>
</form>
```

`TypeError: Cannot read properties of undefined (reading 'length')` out of `ngAfterViewInit`, and **change detection stops at that control**. **Master returns `false` for the same setup**, which I checked by reverting this file on its own.


`radio-group.directive.ts:520` has the **same change and the same failure**. Reactive Forms in both cases, **no Signal Forms involved**.

projects/igniteui-angular/radio/src/radio/radio-group/radio-group.directive.ts
viktorkombov:
```ts
if (this.ngControl!.disabled) {
    button.disabled = this.ngControl!.disabled;
}
```

This **only ever writes `true`**, so the group never comes back:

```ts
model = signal({ season: 'Summer' });
isDisabled = signal(false);
f = form(this.model, p => disabled(p.season, { when: () => this.isDisabled() }));
```

```html
<igx-radio-group [formField]="f.season">
    <igx-radio value="Summer">Summer</igx-radio>
    <igx-radio value="Winter">Winter</igx-radio>
</igx-radio-group>
```

Set `isDisabled` true, then false. **`IgxRadioComponent.disabled` stays `true` on every button.**

This **isn't introduced by the PR**. `IgxRadioGroupDirective` doesn't implement `setDisabledState`, so with Reactive Forms `control.disable()` never reaches the buttons at all. `initialize()` sits in an `effect()` that doesn't re-run when a non-signal `ngControl.disabled` changes. Signal Forms make that effect reactive, which is why the latch fires and the group ends up stuck.

Separate symptom going into disabled: `IgxRadioComponent.disabled` becomes `true` but the native `<input>` **stays enabled and clickable**, even though the template binds `[disabled]="disabled"`. The group writes the property without marking anything dirty, so an OnPush host never gets checked. **Reproduces under zone.js too.**

projects/igniteui-angular/core/src/core/ng-control-adapter.ts
```diff
@@ -0,0 +1,142 @@
...
+        // Probe with an empty control so `required` is detected regardless of the current value.
+        return !!validator({} as AbstractControl)?.required;
+    }
+
+    /**
+     * Emits when validity, disabled, dirty or pending state changes.
+     * Signal Forms `submit()` only marks fields touched, so touched changes count too
+     * or the errors would never surface.
+     */
+    public get statusChanges(): Observable<unknown> {
```
viktorkombov, line 82: `required` isn't in the watched tuple, so a rule that flips while the field stays valid produces no emission and `onStatusChanged` doesn't run.

```
required on              igx-input-group--required = false   (stale)
first interaction        igx-input-group--required = true    (catches up)
required off             igx-input-group--required = true    (stale again)
any further interaction  igx-input-group--required = true    (stuck)
```

It corrects on the next transition of `valid`, `invalid`, `pending`, `disabled`, `dirty` or `touched`. In practice that's the user's first interaction, which flips `touched` and `dirty`.

Both of those are one-way, so nothing refreshes it after that. A programmatic value change doesn't help either.

Same on combo, date picker and time picker. `igxInput` only goes stale on `aria-required`, since its asterisk comes through the directive's own `@Input`.

Specific to the signal backend added here — the Reactive Forms path still emits through `updateValueAndValidity()` and is untouched.

projects/igniteui-angular/core/src/core/ng-control-adapter.ts
```diff
@@ -0,0 +1,142 @@
...
+        return !!this.ngControl.invalid;
+    }
+
+    public get touchedOrDirty(): boolean {
+        const control = this.ngControl.control;
+        return !!(control?.touched || control?.dirty);
+    }
+
+    /** Signal Forms expose no validator list, only `required` and the current errors. */
+    public get hasValidators(): boolean {
```
viktorkombov, line 54: `required || invalid` — a valid, non-required field with a custom rule is neither. So it reports no validators, and `updateValidityState` drops to `INITIAL` where Reactive Forms give `VALID`.

To see it: use `validate(p.role, ...)` instead of `required()`, then satisfy it while the control has focus. The success state never appears. Same for a resolved `validateAsync`.

The observable branch below keeps the old `validator || asyncValidator` check, so this is specific to the signal backend added here.

Note the check sits inside the `touchedOrDirty` branch, which makes the result order-dependent.

projects/igniteui-angular/directives/src/directives/checkbox/checkbox-base.directive.ts
```diff
@@ -262,18 +263,13 @@ export class CheckboxBaseDirective implements AfterViewInit {
...
 
-            if (
-                this.ngControl.control!.validator ||
-                this.ngControl.control!.asyncValidator
-            ) {
-                this._required = this.ngControl.control!.hasValidator(
-                    Validators.required
-                );
+            if (this.control.hasValidators) {
+                this._required = this.control.required;
```
viktorkombov, line 272: `NgControlAdapter.required` probes the validator with a bare `{}`, so every custom validator gets a control with `value === undefined`. Anything that dereferences the value throws:

```ts
const v = (c: AbstractControl): ValidationErrors | null =>
    (c.value as string[]).length === 0 ? { empty: true } : null;

form = new FormGroup({ accepted: new FormControl<unknown>([], v) });
```

```html
<igx-checkbox formControlName="accepted">Accept</igx-checkbox>
```

`TypeError: Cannot read properties of undefined (reading 'length')` out of `ngAfterViewInit`, and change detection stops at that control. Master returns `false` for the same setup. `radio-group.directive.ts:520` has the same change and the same failure.

projects/igniteui-angular/date-picker/src/date-range-picker/date-range-picker.component.ts
```diff
@@ -877,8 +875,8 @@ export class IgxDateRangePickerComponent extends PickerBaseDirective
 
         this.setRequiredToInputs();
 
-        if (this._ngControl) {
-            this._statusChanges$ = this._ngControl.statusChanges!.subscribe(this.onStatusChanged.bind(this));
+        if (this._control) {
+            this._statusChanges$ = this._control.statusChanges.subscribe(this.onStatusChanged.bind(this));
```
viktorkombov, line 879: The asterisk is CSS on `.igx-input-group--required`, and that class is a host binding over `inputGroup.isRequired`, so the value has to be written and then painted. Two things go wrong.

Cause one: this subscription doesn't emit for a required-only change. `onStatusChanged` never runs, `setRequiredToInputs()` (line 938) isn't called, and nothing is written.

Cause two: when something else does make it run, it writes inside a `Promise.resolve().then()` (line 1078) with no `markForCheck()`. The write sits unpainted until a later pass — by which point the rule may have flipped back:

```
                    isRequired written      --required painted
required on         false, false, false     false, false, false
fields touched      true,  true,  true      false, false, false
required off        true,  true,  true      true,  true,  true
```

On where these come from. The first is new, since it's the signal backend of the adapter this PR adds. The second is not — it reproduces on the plain Reactive Forms path with no Signal Forms involved, which is worth knowing on its own:

```ts
ctrl.addValidators(Validators.required);
ctrl.updateValueAndValidity();
// isRequired written = true, --required painted = false, and it never paints
```

So fixing the adapter won't clear this one. With the adapter watching `required`, the marker still didn't render until a `markForCheck()` went in.

Two reasons the new spec stays green:

- its setup is `detectChanges(); tick(); detectChanges();`, which manually does the job of the missing `markForCheck()`
- the test component uses `required(path.range)` with no `when`, so the rule is on from the start and the initial emission covers it

---

projects/igniteui-angular/directives/src/directives/checkbox/checkbox-base.directive.ts
viktorkombov, line 272: Side effect of the same change: checkbox, switch and radio now report `required = true` and `aria-required="true"` for `Validators.requiredTrue`, where master reported `false`. `igxInput` already reported `true` on master, so this aligns them. Flagging in case it should be in the CHANGELOG.

viktorkombov (changes requested):
Tested on a zoneless demo page (`provideZonelessChangeDetection()`) with `[formField]` on each of the ten controls. All six also reproduce under zone.js, so none of them are zoneless-specific.

Two blocking:

- **Checkbox / switch / radio throw** on any validator that reads `control.value`, because `required` detection now probes with `{}`. Reactive Forms path, no Signal Forms involved. Confirmed as a regression by reverting the two files to master individually.
- **Radio group gets stuck disabled** under Signal Forms — it disables but never re-enables. Looks pre-existing rather than introduced here, but this is what makes it reachable.

Four more, not blocking:

- **The conditional `required` marker goes stale** on select, combo and both pickers. It corrects on the user's first interaction with the field, then sticks for good.
- **A satisfied custom rule never reaches `VALID`**, where Reactive Forms do.
- **The date range picker ends up drawing the asterisk for a rule that is switched off.** Half of that one is pre-existing and hits Reactive Forms too.
- **`Validators.requiredTrue` now counts as required** on checkbox/switch/radio where master said no.

Also checked, and clean, so nobody re-treads them:

- the adapter's effects tear down on `@if` destruction and repeated mount/unmount, with no leaked emissions
- switching `[formField]` between fields at runtime
- `submit()` marking untouched fields touched
- `clear()` writing back to the model
- the date range picker's projected two-input write fallback, on both backends
- rapid model updates coalescing into one transition

viktorkombov:
`radio-group.directive.ts:524`, inside `initialize()`:

```ts
if (this.ngControl!.disabled) {
    button.disabled = this.ngControl!.disabled;
}
```

The write is guarded by the condition it's writing, so it only ever sets `true`:

```ts
isDisabled = signal(false);
f = form(this.model, p => disabled(p.season, { when: () => this.isDisabled() }));
```

Set `isDisabled` true, then false. `IgxRadioComponent.disabled` stays `true` on every button.

Not introduced here — `IgxRadioGroupDirective` has no `setDisabledState`, so with Reactive Forms `control.disable()` never reaches the buttons at all, and `initialize()` sits in an `effect()` that doesn't re-run for a non-signal `ngControl.disabled`. Signal Forms make that effect reactive, which is what surfaces the latch.

Also, going *into* disabled the native `<input>` stays clickable even though the component flag is `true`. The group writes the property without marking anything dirty, so an OnPush host never gets checked.

viktorkombov:
- `input.directive.ts:294` — master had `if (this.ngControl.control)` around this. It's unconditional now, and `touchedChanges` does `this.ngControl.control!.events`. Was the guard safe to drop, or can `control` still be null at `ngAfterViewInit`?
- `radio-group.directive.ts:516` — `subscribe(() => this.invalid = false)`. The signal backend emits an initial value where the observable one emits nothing, so `invalid` gets wiped at a point master didn't wipe it.
- A field in `PENDING` renders as `INVALID` on both backends. Pre-existing and unchanged here, but Signal Forms inherit it.

projects/igniteui-angular/input-group/src/input-group/directives-input/input.directive.ts
```diff
@@ -292,22 +288,10 @@ export class IgxInputDirective implements AfterViewInit, OnDestroy {
...
-                    .pipe(filter(e => e instanceof TouchedChangeEvent))
-                    .subscribe(
-                        this.updateValidityState.bind(this)
-                    );
-            }
+        if (this.control) {
+            this._statusChanges$ = this.control.statusChanges.subscribe(this.onStatusChanged.bind(this));
+            this._valueChanges$ = this.control.valueChanges.subscribe(this.onValueChanged.bind(this));
+            this._touchedChanges$ = this.control.touchedChanges.subscribe(this.updateValidityState.bind(this));
         }
```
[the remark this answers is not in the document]
rkaraivanov: Correct for the signal backend. `updateValidityState` is idempotent; the second run costs two attribute writes. The overlap exists because the other controls subscribe to `statusChanges` only and need `submit()` to surface errors. Dropping touched from `statusChanges` and subscribing them to `touchedChanges` changes reactive `markAllAsTouched()` behavior. Leaving as is.

projects/igniteui-angular/core/src/core/ng-control-adapter.ts
rkaraivanov: Fixed. required is in the watched tuple.

projects/igniteui-angular/core/src/core/ng-control-adapter.ts
rkaraivanov: Fixed as far as the interop control allows. It exposes no rule list, so the adapter remembers that the field was ever invalid or pending and reports rules from then on. A rule satisfied from the start and never violated still reads as no validators. Stated in the comment.

projects/igniteui-angular/directives/src/directives/checkbox/checkbox-base.directive.ts
rkaraivanov: Fixed. The probe is guarded; a validator that throws on the empty control reads as not required, which is what master returned for checkbox and radio. igxInput threw on master for the same validator and no longer does.

projects/igniteui-angular/date-picker/src/date-range-picker/date-range-picker.component.ts
rkaraivanov: Both fixed. Cause one: required is watched now. Cause two: setRequiredToInputs calls markForCheck after the deferred write. Your repro paints on 22.x because updateValueAndValidity flips the signal-backed status, which the ng-* host classes read. With a valid value the status is unchanged and nothing repaints. The new zoneless spec covers that case.

projects/igniteui-angular/directives/src/directives/checkbox/checkbox-base.directive.ts
rkaraivanov: Added to the changelog.

rkaraivanov:
> `radio-group.directive.ts:524`, inside `initialize()`:
> 
> ```ts
> if (this.ngControl!.disabled) {
>     button.disabled = this.ngControl!.disabled;
> }
> ```
> 
> The write is guarded by the condition it's writing, so it only ever sets `true`:
> 
> ```ts
> isDisabled = signal(false);
> f = form(this.model, p => disabled(p.season, { when: () => this.isDisabled() }));
> ```
> 
> Set `isDisabled` true, then false. `IgxRadioComponent.disabled` stays `true` on every button.
> 
> Not introduced here — `IgxRadioGroupDirective` has no `setDisabledState`, so with Reactive Forms `control.disable()` never reaches the buttons at all, and `initialize()` sits in an `effect()` that doesn't re-run for a non-signal `ngControl.disabled`. Signal Forms make that effect reactive, which is what surfaces the latch.
> 
> Also, going _into_ disabled the native `<input>` stays clickable even though the component flag is `true`. The group writes the property without marking anything dirty, so an OnPush host never gets checked.

Fixed. IgxRadioGroupDirective implements setDisabledState: stores the state, writes every button, marks for check. initialize() applies the stored state to late buttons and no longer reads ngControl.disabled. Covers control.disable() / enable() and the Signal Forms rule. Specs for both.

viktorkombov:
> `radio-group.directive.ts:524`, inside `initialize()`:
> 
....
> Also, going _into_ disabled the native `<input>` stays clickable even though the component flag is `true`. The group writes the property without marking anything dirty, so an OnPush host never gets checked.

Correction on this comment: the last paragraph ("Also, going into disabled the native `<input>` stays clickable...") was wrong. The native `<input>` does follow the component's `disabled` flag. My check was reading the wrong element, so it always came back `false`. On the code before your fix, disabling the group disabled the inputs as well. The part about the buttons never re-enabling was right. Nothing to change on your side, I just didn't want it left in the thread as a real bug.

viktorkombov:
I've logged the remaining findings as separate issues so they can be handled outside this PR:

#17617 [Checkbox] required is lost when combined with a validator that reads the value
#17618 [Radio Group] control.enable() enables options disabled in the template
#17619 [Input Group] Valid state shown for a field with no rules after switching [formField]
#17620 [Input Group] aria-required stays true after a conditional required rule turns off

They reproduce on `rkaraivanov/fix-17556 branch`. The first two might be worth a look before merging, since neither happens on master and both affect existing Reactive Forms setups.

viktorkombov (dismissed):

rkaraivanov:
@viktorkombov 
Both #17617 and #17618 should be fixed with the latest commit.

viktorkombov:
> @viktorkombov Both #17617 and #17618 should be fixed with the latest commit.

@rkaraivanov, thanks, I checked the latest commit and both #17617 and #17618 are fixed. I tested them in an OnPush component with zoneless change detection, with Reactive Forms and Signal Forms. I also tried a control that starts disabled and a radio button added while the control is disabled.

viktorkombov (approved):

PR #17628, opened 2026-09-15
 fix(elements): attach nested elements to their parent's view

### LINKED ISSUE #17598: [Elements] Grid column actions indicators are not updated when column state changes
## Description

In the Ignite UI Web Components examples, some Grid column actions indicators are not updated when the underlying column state changes.

This can be reproduced with both **Advanced Filtering** and **Column Pinning**.

For **Advanced Filtering**, after applying a filter through the Advanced Filtering dialog, the toolbar button does not indicate that a filter is active. The indicator is updated only after opening the Advanced Filtering dialog again.

A similar issue occurs with **Column Pinning**. If a column is initially hidden, then pinned through the Column Pinning action and later made visible through Column Hiding, the number of pinned columns displayed by the Column Pinning toolbar button is not updated. It is refreshed only after opening the Column Pinning dropdown again.

The same scenarios can also be reproduced in the Elements demos in `igniteui-angular`. In that case, the UI state changes also trigger `ExpressionChangedAfterItHasBeenCheckedError` errors.

### Framework

Angular Elements

### Angular Version

22.1.x

22.2.x

### Component / Area

Grid

### Browser

Chrome

### Operating System

Windows

### Command Not Working

```shell id="a9sy97"
```

### Steps to Reproduce

The issue can be reproduced either in the `igniteui-wc-examples` repository or in the Angular Elements demos from `igniteui-angular`.

#### Option 1: `igniteui-wc-examples`

Run the Grid toolbar samples and use the first Grid containing the relevant column actions.

**Advanced Filtering**

1. Open **Advanced Filtering** from the Grid toolbar.
2. Apply a filter and close the dialog.
3. Observe the Advanced Filtering toolbar button.
4. Open the Advanced Filtering dialog again.

**Column Pinning**

1. Hide a column using **Column Hiding**.
2. Open **Column Pinning** and pin the hidden column.
3. Close the Column Pinning dropdown.
4. Open **Column Hiding** and make the pinned column visible again.
5. Observe the pinned-column count in the Column Pinning toolbar button.
6. Open the Column Pinning dropdown again.

#### Option 2: `igniteui-angular` Elements demo

1. Run the Elements demos:

```shell
npm run start:elements
```

2. Open the Grid toolbar demo and use the first Grid.
3. Repeat the same Advanced Filtering and Column Pinning scenarios described above.
4. Keep the browser DevTools Console open while performing the actions.

### Actual Result

In the Web Components examples:

* The Advanced Filtering button does not immediately indicate that a filter has been applied.
* The Column Pinning button does not immediately update the number of pinned columns after a hidden pinned column is made visible.
* In both cases, the indicator is refreshed only after reopening the corresponding action UI.

In the Elements demos from `igniteui-angular`, the same interactions also result in `ExpressionChangedAfterItHasBeenCheckedError` errors in the corresponding Angular components, including:

```text id="b9tj3a"
ExpressionChangedAfterItHasBeenCheckedError
Expression location: _IgxColumnActionsComponent component.

ExpressionChangedAfterItHasBeenCheckedError
Expression location: _IgxGridToolbarPinningComponent component.

ExpressionChangedAfterItHasBeenCheckedError
Expression location: _IgxGridToolbarAdvancedFilteringComponent component.
```

### Expected Result

Grid toolbar column action indicators should react immediately when the corresponding grid state changes.

* Applying or removing an advanced filter should immediately update the Advanced Filtering indicator.
* Pinning, unpinning, hiding, or showing columns should immediately update the Column Pinning count.
* The Elements demos should not produce `ExpressionChangedAfterItHasBeenCheckedError` errors for these interactions.

### Reproduction URL

_No response_

### Attachments

Image

Image

Image

mddragnev:
Closes #17598 

## Description
Nested custom elements are now attached into their parent element's view instead of being attached to the `ApplicationRef` as standalone roots.

`IgxCustomNgElementStrategy` resolves a `ViewContainerRef` from the parent component's injector (a container anchored at the parent's host element) and inserts the child's host view there. 

Because the anchor is no longer needed, the `@igxElementsAnchor` plumbing is removed: the `anchor` property on `IgxGridBaseDirective` and the `<div #sink>` markup in the grid, tree grid, hierarchical grid, pivot grid and row island templates.

The DOM position restore that follows the insert now derives its reference node from the last of the host view's root nodes rather than from `element.nextSibling`. `ViewContainerRef.insert()` relocates *every* root node of the view, and some components (`igc-action-strip`, `igc-grid-state`) have a trailing anchor comment in addition to the element - using it as the reference node makes `insertBefore` throw `NotFoundError` once it has been moved.

## Motivation / Context
Elements runs with zoneless change detection, which only walks views reachable from whatever was marked dirty. A detached root is unreachable from its parent, so a grid's own `notifyChanges()`/`markForCheck()` never refreshed nested elements and they kept rendering stale state until something ticked their own root.

This surfaced in the grid toolbar actions, which bind directly to live grid state rather than to inputs:
- `igc-grid-toolbar-pinning` kept showing the old `pinnedColumnsCount` after pinning/unpinning a column
- `igc-grid-toolbar-advanced-filtering` kept showing the old column count after applying a filter

The same scenarios work under zone.js, where every tick re-checks all attached views regardless of the view hierarchy, and in a plain Angular app, where these components are genuine descendants of the grid's view.

 - [x] Bug fix
 - [ ] New functionality
 - [x] Refactoring (no functional changes)
 - [ ] Documentation
 - [ ] Demos
 - [ ] CI/CD
 - [x] Tests
 - [ ] Changelog
 - [ ] Skills/Agents

Grid, Tree Grid, Hierarchical Grid, Pivot Grid

## How Has This Been Tested?
 - [x] Unit tests
 - [ ] Manual testing
 - [ ] Automated e2e tests
 
Three tests added to `custom-strategy.spec.ts`, all verified to fail when the behaviour is reverted:
- **View hierarchy** - only the grid's host view is among the `ApplicationRef` root views; toolbar, title, actions, hiding, pinning, column and paginator are not.
- **DOM position** - nothing is stranded next to the grid after the insert, parent/child relationships and sibling order are kept as authored, and the projected paginator keeps its position deeper inside the grid.
- **Change detection** - pinning/unpinning a column through the grid API alone refreshes `pinnedColumnsCount` in the nested toolbar action, with no input or event reaching that element.

### Test Configuration:
 - **Angular version**: 
 - **Browser(s)**: 
 - **OS**: 

## Checklist:


projects/igniteui-angular-elements/src/app/custom-strategy.ts
```diff
@@ -132,17 +132,12 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
...
-                    || element.tagName.toLocaleLowerCase() === 'igc-paginator') {
-                    // NOPE: viewcontainerRef will re-render this node again, no option for rootNode :S
-                    // this.componentRef = parentAnchor.createComponent(this.componentFactory.componentType, { projectableNodes, injector: childInjector });
-                    parentComponentRef = await parent?.ngElementStrategy[ComponentRefKey];
-                    parentAnchor = parentComponentRef?.instance.anchor;
-                }
+                // Container anchored at the parent's host element, so this becomes a child in the parent's
+                // view tree instead of a standalone root. Zoneless CD only walks views reachable from what
+                // was marked dirty, so the parent's `markForCheck()` has to be able to reach it.
+                parentAnchor = parentInjector.get(ViewContainerRef);
```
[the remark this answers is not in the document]
mddragnev: This is not related. The initialization of the custom element is done before the angular hooks are called.

damyanpetev (commented):
Minor nits on the tests and comments, overall LGTM;

projects/igniteui-angular-elements/src/app/custom-strategy.spec.ts
```diff
@@ -371,6 +383,111 @@ describe('Elements: ', () => {
...
+                (await gridEl.querySelector<IgcNgElement>(selector).ngElementStrategy[ComponentRefKey]).hostView;
+            // views the ApplicationRef ticks directly; anything else is reached through its parent
+            const rootViews = (injector.get(ApplicationRef) as any)._views as unknown[];
+
+            // no element parent to attach to, so the grid stays a root
+            expect(rootViews.includes((await gridEl.ngElementStrategy[ComponentRefKey]).hostView)).toBeTrue();
+
+            for (const selector of ['igc-grid-toolbar', 'igc-grid-toolbar-title', 'igc-grid-toolbar-actions',
+                'igc-grid-toolbar-hiding', 'igc-grid-toolbar-pinning', 'igc-column', 'igc-paginator']) {
+                expect(rootViews.includes(await hostViewOf(selector)))
```
damyanpetev: Assuming there was no (even more brittle) way to check the actual hierarchy here? Though haven't interestingly what `hostView` returns for each

projects/igniteui-angular-elements/src/app/custom-strategy.spec.ts
```diff
@@ -371,6 +383,111 @@ describe('Elements: ', () => {
...
+                .toEqual(['IGC-GRID-TOOLBAR-HIDING', 'IGC-GRID-TOOLBAR-PINNING']);
+
+            // the paginator is projected deeper (into the footer) - that spot survives the attach too
+            expect(gridEl.contains(paginatorEl)).toBeTrue();
+            expect(paginatorEl.parentElement).not.toBe(gridEl);
+        });
+
+        it('should refresh a nested toolbar action when only the parent grid is marked for check', async () => {
+            // nothing reaches the toolbar action here - no input, no event. It re-renders only because the
+            // grid's own `notifyChanges()` reaches it through the view hierarchy.
```
damyanpetev: The view hierarchy kind-of guarantees that (if it can be validated) and this test can drop out then.
That statement may be (hopefully soon) inaccurate as well, assuming the actions are not reliant on the grid to run changes;

projects/igniteui-angular-elements/src/app/custom-strategy.spec.ts
```diff
@@ -371,6 +383,111 @@ describe('Elements: ', () => {
...
+
+            // nothing stranded next to the grid, where the insert temporarily moves elements
+            expect(Array.from(testContainer.children).map(x => x.tagName)).toEqual(['IGC-GRID']);
+
+            const toolbarEl = gridEl.querySelector<HTMLElement>('igc-grid-toolbar');
+            const actionsEl = gridEl.querySelector<HTMLElement>('igc-grid-toolbar-actions');
+            const paginatorEl = gridEl.querySelector<HTMLElement>('igc-paginator');
+
+            expect(toolbarEl.parentElement).toBe(gridEl);
+            expect(gridEl.querySelector<HTMLElement>('igc-grid-toolbar-title').parentElement).toBe(toolbarEl);
```
damyanpetev: toolbarEl.querySelector? Also just in case the queries might want to be `?.` since those can miss too

projects/igniteui-angular-elements/src/app/custom-strategy.ts
```diff
@@ -132,17 +132,12 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
...
-                if (element.tagName.toLocaleLowerCase() === 'igc-grid-toolbar'
-                    || element.tagName.toLocaleLowerCase() === 'igc-paginator') {
-                    // NOPE: viewcontainerRef will re-render this node again, no option for rootNode :S
-                    // this.componentRef = parentAnchor.createComponent(this.componentFactory.componentType, { projectableNodes, injector: childInjector });
-                    parentComponentRef = await parent?.ngElementStrategy[ComponentRefKey];
-                    parentAnchor = parentComponentRef?.instance.anchor;
-                }
+                // Container anchored at the parent's host element, so this becomes a child in the parent's
+                // view tree instead of a standalone root. Zoneless CD only walks views reachable from what
+                // was marked dirty, so the parent's `markForCheck()` has to be able to reach it.
```
damyanpetev: so much for short comments :D

```suggestion
                // Use anchor to attach to the parent's view tree instead of a standalone root.
                // Zoneless CD only walks views reachable from what  was marked dirty.
```
And the second line is optional TBH, we shouldn't have to explain that as well, just producing valid view hierarchy is enough.

projects/igniteui-angular-elements/src/app/custom-strategy.ts
```diff
@@ -197,9 +192,13 @@ class IgxCustomNgElementStrategy extends ComponentNgElementStrategy {
             // const parentViewRef = parentInjector.get<ViewContainerRef>(ViewContainerRef);
             // preserve original position in DOM (in case of projection, e.g. grid pager):
             const domParent = element.parentElement;
-            const nextSibling = element.nextSibling;
-            parentAnchor.insert((this as any).componentRef.hostView); //bad, moves in DOM, AND need to be in inner anchor :S
-            //restore original DOM position
+            // `insert` moves all root nodes, and some components have more than the element itself
+            // (igc-action-strip/igc-grid-state also have a trailing anchor comment). Skipping past them
+            // keeps the reference node outside the view, so it stays put and `insertBefore` won't throw.
```
damyanpetev:
```suggestion
            // `insert` moves all root nodes & some components have more than one,
            // so a potential `nextSibling` is always the one after the _last_ root node.
```

projects/igniteui-angular-elements/src/app/custom-strategy.spec.ts
mddragnev: Changed the test to check the actual view hierarchy

projects/igniteui-angular-elements/src/app/custom-strategy.spec.ts
mddragnev: This is true for the current setup with `CDS.Eager`. So I left the test just in case but removed the comment saying that it depends on grid's `notifyChanges` because as you mentioned we hope that it wouldnt be needed soon when we move on `onPush` and signals.

damyanpetev (approved):
LGTM!

Note: This doesn't dismiss the need to resolve the toolbar actions rendering dependency on the grid running full change detection (prob move to signals), it just makes Elements consistent to Angular's runtime.

PR #17643, opened 2026-09-18
fix(migrations): keep brackets balanced when updating theme properties

### LINKED ISSUE #17642: 22.2.0 migration produces invalid SCSS for nested scrollbar-theme() calls
### Description

The `update-22_2_0` migration (which removes the deprecated `scrollbar-theme` parameters `$sb-size`, `$sb-thumb-min-height`, etc.) produces **syntactically invalid SCSS** when the theme function is used as a nested argument inside another call, e.g. the documented `@include scrollbar(scrollbar-theme(...))` pattern.

The migration drops the closing parenthesis of the **enclosing** call, so every affected file fails to compile.

Found while migrating the samples app to `22.2.0-rc.0`: IgniteUI/igniteui-angular-samples#4031

Examples from that PR (`projects/app-lob/src/app/grid-dynamic-chart-data/data-analysis-dock-manager/data-analysis-dock-manager.component.scss`):

```diff
-    @include scrollbar(scrollbar-theme($sb-size: 6px));
+    @include scrollbar(scrollbar-theme();
```

```diff
-            @include scrollbar(scrollbar-theme($sb-size: 16px));
+            @include scrollbar(scrollbar-theme();
```

### Root cause

In `UpdateChanges.updateThemeProps()` (`projects/igniteui-angular/migrations/common/UpdateChanges.ts`):

* L402 — the owner call is matched with a non-greedy `${change.owner}\([\s\S]+?\);`, which is not paren-aware. For `@include scrollbar(scrollbar-theme($sb-size: 6px));` the match is `scrollbar-theme($sb-size: 6px));` — it swallows the `)` that belongs to the outer `scrollbar(...)` mixin call.
* L414–415 — `closing` is taken as the trailing `);`, so `body` becomes `$sb-size: 6px)`. The stray `)` ends up riding on the **last** argument.
* L419–431 — when that last argument is one of the removed ones, it is dropped together with the stray `)`.
* L433 — the rebuilt string is `scrollbar-theme(` + remaining params + `);`, one `)` short of the original.

Reproduction of the exact logic:

| Input | Output |
| --- | --- |
| `@include scrollbar(scrollbar-theme($sb-size: 6px));` | `@include scrollbar(scrollbar-theme();` ❌ |
| `@include scrollbar(scrollbar-theme($thumb-bg: red, $sb-size: 6px));` | `@include scrollbar(scrollbar-theme($thumb-bg: red);` ❌ |
| `@include scrollbar(scrollbar-theme($sb-size: 6px, $thumb-bg: red));` | `@include scrollbar(scrollbar-theme($thumb-bg: red));` ✔ |
| `$t: scrollbar-theme($sb-size: 6px);` | `$t: scrollbar-theme();` ✔ |

So the corruption happens whenever the theme call is **nested inside another call** *and* the **last** argument (or all arguments) get removed. Because `update-22_2_0` removes the entire `$sb-*` set, practically every `@include scrollbar(scrollbar-theme($sb-...))` usage is broken.

This is generic `UpdateChanges` behavior, not specific to `scrollbar-theme` — any future `type: "property"` / `remove: true` theme change hits it. It also affects the `replaceWith` path, where the renamed last argument would keep the stray `)` and then get a second one appended.

### Framework
Angular

### Angular Version
22.0.0

22.2.0-rc.0 (migration shipped in `update-22_2_0`)

### Component / Area
Schematics / Migrations (Theming / Styles)

### Browser
N/A — build-time SCSS compilation failure

### Operating System
Windows, macOS, Linux

### Steps to Reproduce

1. In an app on `igniteui-angular` 22.1.0, add a stylesheet with the documented nested usage:
   ```scss
   @use 'igniteui-angular/theming' as *;

   .selection-area {
       @include scrollbar(scrollbar-theme($sb-size: 6px));
   }
   ```
2. Run `ng update igniteui-angular@22.2.0`.
3. Inspect the stylesheet / run `ng build`.

### Actual Result

The file is rewritten to invalid SCSS and the build fails:

```scss
.selection-area {
    @include scrollbar(scrollbar-theme();
}
```

### Expected Result

The removed parameters are stripped while the surrounding call stays balanced:

```scss
.selection-area {
    @include scrollbar(scrollbar-theme());
}
```

### Suggested fix

Make `updateThemeProps` paren-aware: locate `change.owner(` and scan forward counting brackets (and skipping strings/comments) to find that call's own matching `)`, instead of relying on the `[\s\S]+?\);` regex and the `\s*\);$` closing match. Worth adding specs for:

* theme function nested in a mixin include (`@include scrollbar(scrollbar-theme(...))`),
* removal of the *last* argument,
* removal of *all* arguments,
* a nested theme call not terminated by `;` on the same statement.

### Attachments

Diff of the affected files in the samples repo: https://github.com/IgniteUI/igniteui-angular-samples/pull/4031/files

ChronosSF:
Closes #17642

### What

`UpdateChanges.updateThemeProps()` matched the owner call with a non paren-aware regex:

```ts
const searchPattern = String.raw`${change.owner}\([\s\S]+?\);`;
```

For a theme function nested in another call — the documented `@include scrollbar(scrollbar-theme($sb-size: 6px));` pattern — the match ran to the first `);` and therefore swallowed the closing bracket of the **surrounding** call. That stray `)` ended up attached to the last argument, and when that argument was one of the removed ones it was dropped along with it, producing invalid SCSS:

```scss
@include scrollbar(scrollbar-theme();
```

Since `update-22_2_0` removes the whole `$sb-*` set from `scrollbar-theme`, practically every such usage broke. Real fallout in the samples app: IgniteUI/igniteui-angular-samples#4031.

The bug is in the shared `UpdateChanges` code, not in the 22.2.0 changes file, so any future `type: "property"` / `remove: true` theme change would have hit it too.

### How

* `findFunctionCalls` / `findClosingBracket` locate each `owner(` and scan forward counting brackets to find that call's *own* `)`. Brackets inside strings and comments are ignored, and `isLineCommentStart` keeps `url(https://…)` from being mistaken for a `//` comment.
* The rewritten argument list is spliced back in by index instead of via `String.replace(match, …)`, and the calls are rewritten back to front so the collected indices stay valid.
* The whitespace in front of the closing bracket is preserved separately, so multi-line theme calls keep their formatting exactly as before.
* `splitFunctionProps` now skips strings as well, so a `(` or `,` inside a quoted value no longer throws off the argument split.

Dropping the `\);` anchor also means a theme call that is not terminated by `;` on the same statement is now migrated instead of silently skipped.

### Behavior

| Input | Before | After |
| --- | --- | --- |
| `@include scrollbar(scrollbar-theme($sb-size: 6px));` | `@include scrollbar(scrollbar-theme();` ❌ | `@include scrollbar(scrollbar-theme());` |
| `@include scrollbar(scrollbar-theme($thumb-bg: red, $sb-size: 6px));` | `@include scrollbar(scrollbar-theme($thumb-bg: red);` ❌ | `@include scrollbar(scrollbar-theme($thumb-bg: red));` |
| `@include scrollbar(scrollbar-theme($sb-size: 6px, $thumb-bg: red));` | correct | unchanged |
| `$t: scrollbar-theme($sb-size: 6px);` | correct | unchanged |

Formatting of the non-nested, single-theme-per-statement case — by far the common one — is byte-for-byte what it was; the existing specs cover that and pass untouched.

### Tests

* `UpdateChanges.spec.ts` — new spec over the generic path: nested call with the only / last / first argument removed, a rename inside a nested call, and brackets and commas hiding in a string, a `url()` and a trailing comment.
* `update-22_2_0/index.spec.ts` — the two `@include scrollbar(scrollbar-theme(…))` shapes from the samples repo, plus a theme call with no terminating `;`.

Both new specs fail on `master` and pass with the fix. `npm run test:schematics`: 468 specs, 0 failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -397,51 +397,134 @@ export class UpdateChanges {
...
+    /**
+     * Returns the argument list boundaries of every top-level `owner(...)` call in the content.
+     * The brackets are tracked, so a call nested in another one -
+     * `@include scrollbar(scrollbar-theme($sb-size: 6px))` - reports its own closing bracket
+     * rather than the one of the call surrounding it.
+     */
+    private findFunctionCalls(content: string, owner: string): { bodyStart: number; bodyEnd: number }[] {
+        const calls: { bodyStart: number; bodyEnd: number }[] = [];
+        const opening = `${owner}(`;
+        let index = content.indexOf(opening);
```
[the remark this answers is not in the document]
ChronosSF: Good catch, this was real. Fixed in 7c4cd7b0fb.

`findFunctionCalls` now scans for the opening the same way the bracket matching already did, stepping over strings and comments through a shared `skipNonCode` helper that both it and `findClosingBracket` use.

The failure mode was worse than just a spurious rewrite. Given:

```scss
// igx-theme-func($remove-me: 1px) is gone, use the grid's own borders
$var: igx-theme-func($remove-me: 4px, $prop1: red);
```

the old scan rewrote the *comment* to `// igx-theme-func()` and, having consumed that as a call, never reached the real one on the last line - so the actual usage went unmigrated. Added as a spec in `UpdateChanges.spec.ts` covering a line comment, an unterminated `(` in a block comment, and a quoted value; it fails on the previous commit and passes now.

One adjacent thing I deliberately left alone: the opening match has no identifier boundary, so an owner like `grid-theme` also matches inside `igx-grid-theme(`. That predates this PR and several migrations declare overlapping owners (`theme` / `igx-theme`, `button-theme` / `flat-button-theme`), so tightening it is a behavior change across old migrations that belongs in its own PR rather than riding along here.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -879,6 +993,10 @@ export class UpdateChanges {
 
         for (let i = 0; i < body.length; i++) {
             const char = body[i];
+            if (char === '\'' || char === '"') {
```
viktorkombov: This input comes back with `$sb-thumb-bg-color` removed:

```scss
$my-scrollbar: scrollbar-theme($sb-size: 6px /* user's size */, $sb-thumb-bg-color: red);
```

```scss
// master
$my-scrollbar: scrollbar-theme( $sb-thumb-bg-color: red);
// this branch
$my-scrollbar: scrollbar-theme();
```

`$sb-thumb-bg-color` is one of the two properties 22.2.0 keeps working, and this deletes a value the user wrote.

Tracing it: `splitFunctionProps` now skips strings but not comments, so it disagrees with `skipNonCode`, which handles both. The apostrophe in the comment opens a string that never closes, the body stops splitting on commas, and the whole argument list collapses into a single parameter. That single parameter matches `^\s*\$sb-size:`, so the retained color is dropped with it.

Ordering decides the failure mode. Removed property first deletes the kept ones, as above. Removed property last silently skips the removal instead, leaving obsolete arguments behind. A plain `it's` or `don't` is enough, and a stray double quote does the same.

Recommendation: route this through `skipNonCode` instead of `skipString` before merge. That covers both orderings, and also clears the pre-existing case where a `(`, `)` or `,` inside a comment breaks the split — unless there's a reason `splitFunctionProps` was kept on `skipString` that points at a narrower fix.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
```diff
@@ -397,51 +397,165 @@ export class UpdateChanges {
...
+    }
+
+    /**
+     * Tells apart a `//` line comment from the `//` of a protocol - `url(https://...)` -
+     * by looking at what precedes it.
+     */
+    private isLineCommentStart(content: string, index: number): boolean {
+        const previous = content[index - 1];
+
+        return previous === undefined || /[\s,(;{]/.test(previous);
```
viktorkombov: **Observed: valid theme calls are skipped in both directions of the whitelist.**

The `//` of an unquoted protocol-relative URL reads as a comment, because `(` is accepted as a preceding character, and the rest of the line is swallowed. Where the URL sits decides how the call is lost. Inside the argument list, `findClosingBracket` returns -1 and the `break` at line 470 abandons this call and every later one in the file. Before the call on the same line, the top-level scan steps straight over the call and never sees it.

The PR's own `url()` fixture from `UpdateChanges.spec.ts` shows the first shape most directly. Only the protocol differs:

```scss
// as written in the PR's spec, migrates correctly
$var2: igx-theme-func($image: url(https://example.com/a.png), $remove-me: 6px);
// protocol-relative, silently not migrated
$var2: igx-theme-func($image: url(//example.com/a.png), $remove-me: 6px);
```

`scrollbar-theme` itself takes no image argument, so for 22.2.0 the trigger is the second shape, a URL earlier on the same line. This is valid SCSS and comes back unchanged:

```scss
.a { background: url(//cdn.example.com/bg.png); $my: scrollbar-theme($sb-size: 16px, $sb-thumb-bg-color: red); }
```

In the other direction, `}`, `)`, an identifier and a closing quote are rejected, so a comment glued to one of them is scanned as code. An apostrophe in it then hides every theme call below:

```scss
.selection-area { color: red; }// don't set $sb-size here

$my-scrollbar: scrollbar-theme($sb-size: 16px, $sb-thumb-bg-color: blue);
```

Also unchanged.

Recommendation: treat an unquoted `url(` as its own token in `skipNonCode` and skip to its matching bracket, which lets `//` be a comment unconditionally and covers both directions. I'd want a decision on this one, though it's separable from R0.

viktorkombov (changes requested):
I ran the PR through the real `migration-61` schematic, then re-ran the same inputs with master's `UpdateChanges.ts` swapped back in, to separate what this branch changes from what was already there.

The nested `@include scrollbar(scrollbar-theme(...))` shapes migrate properly now, and I confirmed separately that the new specs fail on master.

Alongside that, five inputs migrate correctly on master and not on this branch. One of them (R0) deletes a property the migration is documented to keep. They trace back to three places in the new scanner:

- `splitFunctionProps` — R0, R1
- `isLineCommentStart` — R2, R3
- `findFunctionCalls` — R4

With the regression specs added to the repo:

| Revision | Result | Failing |
|---|---|---|
| This branch | `474 specs, 5 failures` | the five regression specs |
| Master | `474 specs, 4 failures` | the four specs this PR adds |

The five regression specs pass on master, so the two failure sets don't overlap.

R0 deletes retained configuration. R1–R3 leave migrations incomplete. R4 removes a declaration parameter and causes the demonstrated mixin invocation to fail compilation.

Recommendation: resolve R0 before merge — it's the one that loses user data. I'd want a decision on the other four, but they're separable from it. Details inline.

Each defect below carries its own reproduction, so nothing needs to be downloaded to follow them. Attached for convenience: the five as Jasmine specs to drop into `projects/igniteui-angular/migrations/update-22_2_0/`, and a script that runs the same inputs through the schematic and prints each one next to master's output.

[repro-17642.zip](https://github.com/user-attachments/files/32480812/repro-17642.zip)

viktorkombov:
## CHANGELOG.md

22.2.0 shipped with this migration corrupting `@include scrollbar(scrollbar-theme(...))` into invalid SCSS, which is user-visible, and AGENTS.md asks for notable user-visible fixes where they fit the existing section structure. They do here.

I argued the opposite earlier on the grounds that past `fix(migrations)` commits skip the changelog. That precedent covers migration infrastructure, not a shipped migration that damaged real projects, and it does not override the written rule. Withdrawing that objection.

Recommendation: a line at the end of the list under Unreleased, Bug Fixes, in the two-level shape the neighboring entries use.

```markdown
- **Theming**
    - The 22.2.0 `ng update` migration no longer drops the closing bracket of a surrounding call when a theme function is nested in one, as in `@include scrollbar(scrollbar-theme($sb-size: 6px));`. The migration produced invalid SCSS whenever the last argument was one of the removed properties.
```

viktorkombov:
**Not blocking:**

Two things I noticed but would not hold the PR for. The owner name is matched without an identifier boundary, so a project's own `custom-scrollbar-theme($sb-size: ...)` loses its argument; master does this too, so it predates this branch. And `eslint.config.mjs` line 10 excludes `migrations/`, so a passing `npm run lint:lib` says nothing about these files either way.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
ChronosSF: Confirmed and fixed in b14d59c2b9. You were right on the cause and on the recommendation.

`splitFunctionProps` now routes through `skipNonCode` rather than `skipString`, so it treats comments exactly the way the rest of the scanner does. That covers both orderings and, as you noted, also clears the pre-existing case where a `(`, `)` or `,` inside a comment broke the split.

There was no reason it was on `skipString` other than my oversight: I added string skipping to it in the same commit that introduced `skipNonCode` and never reconciled the two.

I reproduced R0 before touching anything, and it is as bad as you describe - `scrollbar-theme()`, with the user's `red` gone:

```
4) should not delete a retained property because of a comment containing an apostrophe
   Expected '$my-scrollbar: scrollbar-theme();'
      to equal '$my-scrollbar: scrollbar-theme( $sb-thumb-bg-color: red);'
```

Both orderings are now specs in `update-22_2_0/index.spec.ts`.

projects/igniteui-angular/migrations/common/UpdateChanges.ts
ChronosSF: Confirmed and fixed in b14d59c2b9, taking the direction you suggested.

`skipNonCode` now treats an unquoted `url(` as its own token and skips to its matching bracket, which lets `//` start a comment unconditionally. `isLineCommentStart` is gone. As you said, the previous-character whitelist was the wrong shape for the problem - simultaneously too strict for R2 and too loose for R3 - and matching Sass's own special-casing of `url()` resolves both directions at once.

Two related things came out of it:

**R4** is fixed in the same commit. `findFunctionCalls` now skips a name that directly follows `@mixin` or `@function`, so a user declaration is left alone. I checked the output with `npx sass` against your `sass-check/` files: the declaration compiles to `.demo { width: 16px; }` again instead of `Error: Undefined variable`.

**The `break`** you flagged inside R3's first shape is now a `continue`. Abandoning every later call in the file was the wrong failure mode even for genuinely unbalanced input, so it now drops just the call it cannot parse.

Verification, in your harness and in the suite:

| | Result |
|---|---|
| `repro.js` | `All 5 cases behave correctly` |
| `npm run test:schematics` | `474 specs, 0 failures` |

I folded your five specs into `update-22_2_0/index.spec.ts` rather than keeping a separate file, per the repo's "reuse the existing spec structure". They still fail on the previous commit and pass on this one - I re-ran them both ways to be sure.

Thank you for the repro bundle. Running the real schematic and diffing against master is what made these unambiguous, and the warning about stale `dist/` was warranted: I hit exactly that when I removed a spec file and the count did not drop until I cleared the compiled leftovers.

ChronosSF:
On the owner-boundary issue from your summary and the not-blocking comment - I dug into it and I'm deliberately leaving it out of this PR, with a reason beyond "it's pre-existing".

I confirmed it still reproduces exactly as you wrote, and that my `@mixin`/`@function` fix does **not** incidentally cover it:

```scss
@function custom-scrollbar-theme($sb-size: 6px, $thumb: red) { @return ($size: $sb-size); }
// still, on this branch and on master:
@function custom-scrollbar-theme( $thumb: red) { @return ($size: $sb-size); }
```

`isDeclaredName` walks back from the owner's own start, hits `custom-`, and stops short of the `@function`. I could widen it to walk the full identifier first, and it would catch this - but I think that would be a mistake. Today the mangled declaration leaves `$sb-size` undefined in the body, so Sass fails loudly. Protecting only the declaration while calls keep losing the argument would turn that into a silent wrong value. Half-fixing it is worse than not fixing it.

The correct fix is the identifier boundary you propose, which fixes declarations and calls together. What makes it too big for this PR is that the loose matching may be load-bearing for older migrations:

| Migration | Owner | Would stop matching |
|---|---|---|
| `update-15_0_0` | `theme` | every `*-theme(` call, incl. `igx-light-theme(` |
| `update-15_0_0` | `typography` | `igx-typography(` |
| `update-19_1_4`, `update-20_1_0` | `grid-theme` | `tree-grid-theme(`, `igx-grid-theme(` |

`update-19_1_4` removes `$root-summaries-background` from `grid-theme`, and `tree-grid-theme` is not declared as an owner anywhere - so today a tree-grid theme is migrated only through the suffix match. Some of these look enumerated deliberately (`typography` and `igx-typography` are both listed), others look like they rely on the looseness. Telling those apart needs someone with the theming history, and getting it wrong silently stops an old migration from doing its job.

Happy to open a separate issue for it with this table if that's useful.

On `eslint.config.mjs` excluding `migrations/` - you're right that `npm run lint:lib` proves nothing here. I ran `npx eslint --no-ignore` on the changed files and compared against master: the same three findings appear on both (two unused-disable warnings and one unused `err`), all pre-existing and outside this diff, so the change adds nothing new. The exclusion looks worth a separate look too.

simeonoff (approved):

viktorkombov (approved):

